PR #211, opened 2014-05-11
$dates array support

### LINKED ISSUE #109: $date (date-mutators) does not work with dot notation
For example if you try to let it handle `feeds.date`, where `feeds` is an array.

There might not be easy workaround for this one... just reporting for the sake of known limits.

calumks:
Allows you to work with Carbon/DateTime objects instead of MongoDate objects within arrays.

If you specify a full array in $dates, any dates within will be converted into carbon objects. You can also use dot notation to specify dates within arrays.

Works with toArray() and any date that passes through asDateTime().

Resolves #109

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,7 +298,7 @@ All basic insert, update, delete and select methods should be implemented.
 
 ### Dates
 
-Eloquent allows you to work with Carbon/DateTime objects instead of MongoDate objects. Internally, these dates will be converted to MongoDate objects when saved to the database. If you wish to use this functionality on non-default date fields you will need to manually specify them as described here: http://laravel.com/docs/eloquent#date-mutators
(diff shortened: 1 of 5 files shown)


spawn-guy:
this /should/ be supported via embedOne/Many and then no dot-notation would be needed.

and please be careful with `array_set`. 
i created same code recently, and now i have to admit, that `array_set` f*cks up the returned array. 
"last" array's key value is re-written with the last processed (array set) date value :/ (even if it is an array)

i had to rewrite to array_walk_recursive, as a more stable version(i have deadlines over here)

erkanarslan:
This feature is needed +1

drewlag:
News about this?

daywong1119:
I need this function too, hope this can be merge soon

Smolevich:
@jenssegers I think that this old issue should be closed

Giacomo92:
It can be closed @Smolevich @divine . There is also tests [here](https://github.com/jenssegers/laravel-mongodb/blob/master/tests/ModelTest.php#L384)

alagaccia:
So how can you cast to MongoDates something like this?
For example I want to cast date1, date2 from "last_dates" array.

`array('dates' => array(first_dates => array(date1, date2), last_dates => array(date1, date2)))`

PR #238, opened 2014-06-20
Add options in increment and decrement method

potsky:
Why ? To increment and decrement in a shot with upsert option for example


CammoKing:
Is this not easily added on the Eloquent/Builder because it returns the parent EloquentBuilder which does not have the placeholder for the $options array? Thanks much!

PR #383, opened 2014-12-14
Prevent getAttribute() from calling static methods

NSmithUK:
I hit an issue after upgrading to v2 that I could no longer access attributes that had the same name as static method declared on Jenssegers\Mongodb\Model (or its parents).

For example I had attributes called 'updated' and 'deleted'. When trying to access $model->updated, PHP threw a fatal error because `$relations = $this->$camelKey();` tried calling `\Illuminate\Database\Eloquent\Model::updated($callback)`.

This pull request helps mitigate the issue by ensuring only non-static methods are assumed to represent a embedded model.

### DIFF
--- a/src/Jenssegers/Mongodb/Model.php
+++ b/src/Jenssegers/Mongodb/Model.php
@@ -13,6 +13,7 @@ use Carbon\Carbon;
 use DateTime;
 use MongoId;
 use MongoDate;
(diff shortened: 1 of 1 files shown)


jenssegers:
I had to manually merge your changes, could not get the github merge to work :(

PR #405, opened 2015-01-30
Remove unused imports from Model

duxet:
It also fixes problem with:
> Cannot use Illuminate\Database\Eloquent\Collection as Collection because the name is already in use


jenssegers:
Thanks man!

PR #428, opened 2015-02-18
Update aggregation builder Laravel 4.2

giboow:
Old request is here,
The new is compatible with laravel 4.2


jenssegers:
Would you mind adding a unit test as well?

jenssegers:
Are you sure this is the only way to do pagination?

jenssegers:
I added this feature, but in a slightly different way. I would be cool if I could get some feedback from you.

PR #442, opened 2015-03-07
Convert overloaded Eloquent model to Traits

hackel:
This PR is only meant as a starting point for discussion, not at all final.  The change I'm proposing here is potentially significant.  I have converted Jenssegers/Eloquent/Model and Jenssegers/Mongodb/Model to traits.  I did this while attempting to get BootstrapCMS/Credentials and Sentry to work with laravel-mongodb.  Now my models can extend Illuminate\Database\Eloquent\Model like normal, but still get all the extra MongoDB functionality from this package.  This eliminates the need for  laravel-mongodb-sentry as well, which hadn't yet been updated to L5.

My class definitions now look like this:
```php
use Jenssegers\Mongodb\Mongo;
use Jenssegers\Mongodb\Contracts\Mongo as MongoContract;

class User extends Model implements MongoContract
{
	use Mongo;
```

The interface in Mongodb\Contracts\Mongo is currently empty, but method signatures could certainly be added if it's deemed useful.  It's simply there so that the is_subclass_of() checks will pass for model relationships.  Instead, I think we could just check whether the model's $connection == 'mongodb' and remove the interface all together.

I moved the implementation from Jenssegers/Mongodb/Model to the Mongo trait, but kept the class in place in an effort to maintain backward-compatibility.  I haven't tested this at all, however.

I still need to add/fix tests for this, but wanted to see what people thought first.  This could potentially make testing easier as well.  (At least for me using PHPSpec, which doesn't allow to mock parent methods).

Feedback is appreciated!


jenssegers:
For the v3 version on the develop branch, I have converted the class for the hybrid sql models, instead of replacing the class for those models, they will have to use the `HybridRelations` trait.

I don't know about converting the whole Mongo class to a trait though. I will think about it.

hackel:
Moved to https://github.com/hackel/laravel-mongodb/tree/use-traits if anyone's interested in pursuing this further.  Unless I find a better solution, I will be using this for one of my own projects so will try to keep the fork in sync and minimize any changes.

PR #471, opened 2015-04-16
Mongodb support for non-existent field

foxbit19:
Mongodb has no fixed structure for collections. So a field may be non-existent in mongodb.
In this case, the property that represents the foreignKey cannot be retrieved and an ErrorException will be thrown.
I've added a try-catch construct to avoid the case in which the field doesn't exists.

### DIFF
--- a/src/Jenssegers/Mongodb/Relations/BelongsTo.php
+++ b/src/Jenssegers/Mongodb/Relations/BelongsTo.php
@@ -21,7 +21,8 @@ class BelongsTo extends \Illuminate\Database\Eloquent\Relations\BelongsTo {
             }
             catch (ErrorException $ex)
             {
-            	// The property doesn't exists. Set it to null
+            	// The property doesn't exists.
+            	// Set it to null
                 $foreign_key = null;
             }
 
--- a/src/Jenssegers/Mongodb/Relations/BelongsTo.php
+++ b/src/Jenssegers/Mongodb/Relations/BelongsTo.php
@@ -11,10 +11,24 @@ class BelongsTo extends \Illuminate\Database\Eloquent\Relations\BelongsTo {
 	{
 		if (static::$constraints)
 		{
+			// The foreign key field may be non-existent in mongodb.
+			// In this case, the property $this->parent->{$this->foreignKey} cannot be retrieved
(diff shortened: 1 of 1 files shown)


jenssegers:
In what case would you get this error?

foxbit19:
I've no foreign keys fields for some documents in a collection. The foreign keys fields are inserted progessively by a laravel application. I think this problem comes up because I've first inserted the documents inside some collections, then I've defined the relations between them.

jenssegers:
Could you show me a code example when this would happen? I will be able to add it to the tests then and fix it more easily :)

foxbit19:
Just define your model without relations:

    /*
    * Keyboard's key :)
    */
    class Key extends Moloquent 
    {
        protected $collection = 'keys';
    }

Then use the model in your code to populate the collection.

You want to add another collection for the `keyboard`. But your `key` collection is populated and you don't want to delete all the documents inside it. You just want to add the relation between the collection without touching any data.

So you add the `Keyboard` model with the correct relation.

    class Keyboard extends Moloquent 
    {
        protected $collection = 'keyboards';

        public function getKeys() 
        {
            return $this->hasMany('App\Models\Key','keys');
        }
    }

Add now add an inverse relation to the model `Key`:

    class Key extends Moloquent 
    {
        protected $collection = 'keys';

        /*
        * Since this is a one to many relation, I want a 'keyboard' field
        * inside the collection using the inverse relation
        */
        public function getKeyboard() 
        {
            return $this->belongsTo('App\Models\Keyboard', 'keyboard');
        }
    }

You expected to work with the field `keyboard` of the collection `key` but when you use this code

    $keyboard = $key->getKeyboard;

you get an `ErrorException`.

The above code fails because `belongsTo` search for a field named `keyboard` inside the collection `key`. This field cannot exist at this moment.

Consider the case in which I want to add some one-to-many relations to my collections and my data cannot be touched because they are used daily by my clients (through my production systems).

*This is the context in which this error happen and I think the same behaviour can be extended to other inverse relations.*

foxbit19:
Any news of this?

jenssegers:
Doing the try/catch feels like a work-arround. Would it not be better to just check if the parent instance is set?

PR #483, opened 2015-05-04
Update README.md

lukepolo:
Before anyone pulls their hair out, In mongo 3 if you have auth you must supply the auth database ! 

http://docs.mongodb.org/manual/reference/program/mongotop/#cmdoption--authenticationDatabase


jenssegers:
Thanks!

PR #487, opened 2015-05-07
If an attribute is a date; and used as a dot-notation key

spawn-guy:
fix it after parent attributeToArray.

as i've noticed fromDateTime behavior on setAttribute. i still don't get it, why it's not in the getAttributesToArray


jenssegers:
I had to merge you changes manually, they are now in dev-master, thanks!

PR #510, opened 2015-06-04
Retries failed connection to fix #508.

### LINKED ISSUE #508: Retry closed connections.
We're encountering problems with object rocket and it's replica sets. I hope someone else here has had this issue too.

We think the issue can be solved by retrying the connection (see links below). I'm going to test that retrying the connection does solve the issue for us. If it does, would you be happy for me to make a pull request to wrap the connection attempt in a try-catch and retry? I know it's an edge case, but we need this fixed asap and I don't want to point composer to a forked version of this repo if possible.

http://stackoverflow.com/questions/22399795/mongo-db-connection-problems
https://www.compose.io/articles/achieving-high-availability-and-the-mongodb-php-driver/
https://github.com/jenssegers/laravel-mongodb/blob/master/src/Jenssegers/Mongodb/Connection.php#L153

ryansmith94:
#508

### DIFF
--- a/src/Jenssegers/Mongodb/Connection.php
+++ b/src/Jenssegers/Mongodb/Connection.php
@@ -126,9 +126,10 @@ class Connection extends \Illuminate\Database\Connection {
      * @param  string  $dsn
      * @param  array   $config
      * @param  array   $options
+     * @param  int     $remaining_retries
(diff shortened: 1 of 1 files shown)


ryansmith94:
Closing because of inactivity.

PR #522, opened 2015-06-21
Update installation instructions for L5 and Lumen

### LINKED ISSUE #470: Lumen support?
Hi,

I loved using this with Laravel and would like to continue using this on Lumen. I tried the following:

1. Registered the service provider on bookstrap/app.php with `$app->register('Jenssegers\Mongodb\MongodbServiceProvider');`

2. Since Lumen does not have config/database, I had to edit vendor\laravel\lumen-framework\config\database.php to add MongoDB configuration.

But when I actually try to do something, I get:

`InvalidArgumentException: Unsupported driver [mongodb]`

So I guess even though I register the service provider, it does not create the necessary database connector for MongoDB? I know this Lumen is not officially supported, but is there any chance to get this working?

kieranajp:
Fixes #470

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,16 +28,30 @@ For Laravel 5, install the latest stable version using composer:
 composer require jenssegers/mongodb
 ```
 
+And add the service provider in `config/app.php`:
+
+```php
+Jenssegers\Mongodb\MongodbServiceProvider::class,
+```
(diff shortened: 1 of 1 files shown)


jenssegers:
Thanks!

PR #619, opened 2015-10-15
Update QueryBuilder get method to run getFresh method

### LINKED ISSUE #618: Laravel 5.1.20 Issue - mySQL TO Mongo
Looks like the new update has broken the ability to relate things from mySQL to Mongo.

5.1.19 works fine but once update we get the error :

"Call to a member function compileSelect() on null"

Anyone else seeing this?

loonytoons:
Following the latest update of the laravel framework (5.1.20), the BaseQuery builder that used to just run the getFresh method (overriden by the mongodb child builder) in the get function, now calls the runSelect method that relies on the mysql Grammar object being present that this builder does not need to run.
This results in the error `` Call to a member function compileSelect() on null `` when trying to run a query.

This update simply removes the parent get() call and directly calls it's own getFresh method, thus return the functionality to what it was before the laravel update.

The particular change in laravel that causes this issue can be seen here:
https://github.com/laravel/framework/compare/v5.1.19...v5.1.20#diff-cc3d43e01028390141596ae0da939930L1373

Fixes #618

### DIFF
--- a/src/Jenssegers/Mongodb/Query/Builder.php
+++ b/src/Jenssegers/Mongodb/Query/Builder.php
@@ -147,7 +147,7 @@ class Builder extends BaseBuilder {
      */
     public function get($columns = [])
     {
-        return parent::get($columns);
+        return $this->getFresh($columns);
     }
 
     /**


jenssegers:
Thanks!

fuzonmedia:
Same issue facing .. "Fatal error: Call to a member function compileSelect() on null"
Using 
"laravel/framework": "5.2.*",
"jenssegers/mongodb": "^3.0"

Can you help how to solve ?

fuzonmedia:
I fixed it by changing the extends class from "Model" to "Eloquent" in my DB model

ahmedash95:
@fuzonmedia i faced this problem now but i can't understand which file you had change model to eloquent ?

fuzonmedia:
@ahmedash95  you have to change it in your model file 

class User extends **Eloquent** {} 

- User is the laravel model

ahmedash95:
@fuzonmedia okay i got it , but i change it in the Authenticate class in the laravel core itself and it works fine for me

tjntun:
I face the same issue when using Laravel 5.2
`"laravel/framework": "5.2.*",
"jenssegers/mongodb": "^3.0",
"laravel/cashier": "~6.0"`

I also tried @fuzonmedia method but it doesn't work. Any idea everyone?

Ravaelles:
@ahmedash95 Because overriding core classes is always wrong as they will be overriden in the next Laravel upgrade via composer, I think it's a good idea to introduce new App\MongoUser class over which you have full control.
`class MongoUser extends \Jenssegers\Mongodb\Eloquent\Model implements
AuthenticatableContract, AuthorizableContract, CanResetPasswordContract
{
    use Authenticatable,
        Authorizable,
        CanResetPassword;
}`

And then change your User model to:
`class User extends MongoUser`

That works for me :+1:

ahmedash95:
@Ravaelles thanks ^_^ 👍

deanmcpherson:
For reference, I was getting the above error when using Cashier (User and Subscriptions are MySQL backed). The issue was because the default connection is set to mongo, the Cashier Models where looking in the wrong place.

I forked Cashier to fix, setting the connection direction on the subscription model.

chintan13:
thanks to @Ravaelles

for some one new with same problem her is full Class:

```
<?php

namespace App;

use Illuminate\Auth\Authenticatable;
use Illuminate\Auth\Passwords\CanResetPassword;
use Illuminate\Foundation\Auth\Access\Authorizable;
use Illuminate\Contracts\Auth\Authenticatable as AuthenticatableContract;
use Illuminate\Contracts\Auth\Access\Authorizable as AuthorizableContract;
use Illuminate\Contracts\Auth\CanResetPassword as CanResetPasswordContract;

class MongoUser extends Eloquent implements AuthenticatableContract, 
			AuthorizableContract, CanResetPasswordContract 
{ 
	use Authenticatable, Authorizable, CanResetPassword; 
}
```

and then 

`class User extends MongoUser`

Works like a charm

I have used this with Laravel 5.3

PR #620, opened 2015-10-15
Fix #607 exists() bug and assure compatibility with Laravel 5.1.19

### LINKED ISSUE #607: Laravel 5.1.19 breaks exists()
This commit https://github.com/laravel/framework/commit/cf668a67bd4959d9265b8e5ea46d0fce24b871f3 broke previously working exists() method with `Call to a member function compileExists() on null` exception.

alexandre-butynski:
Laravel 5.1.19 version breaks the `exists()` method (see #607) because of a complete rewriting of the Laravel base method.

I think that the method should be rewrite in this package. This implementation is close to the best in term of performance but a small improvement can be done by using `db.collection.find().limit(1)->count(1)` but I don't know how to use the count method like that (if there is no argument, `count()` don't care about `limit()`).

If this pull request is accepted, a new release would be appreciated :)

### DIFF
--- a/src/Jenssegers/Mongodb/Query/Builder.php
+++ b/src/Jenssegers/Mongodb/Query/Builder.php
@@ -350,6 +350,16 @@ class Builder extends BaseBuilder {
     }
 
     /**
+     * Determine if any rows exist for the current query.
+     *
+     * @return bool
+     */
+    public function exists()
+    {
+        return ! is_null($this->first());
+    }
+
+    /**
      * Force the query to only return distinct results.
      *
      * @return Builder
--- a/tests/QueryTest.php
+++ b/tests/QueryTest.php
@@ -206,6 +206,12 @@ class QueryTest extends TestCase {
         $this->assertEquals(6, $count);
     }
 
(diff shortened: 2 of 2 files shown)


jenssegers:
Can't we use the same logic that Laravel had before instead of calling `first()`? Probably more efficient to force `limit(1)` on the exists query.

alexandre-butynski:
`$this->first()` is faster than `$this->limit(1)->count()` because `count()` method ignores the effect of `limit()` method. A quick test on a 2 000 documents dataset shows a request 6 times faster.

See this doc page for more informations : http://docs.mongodb.org/manual/reference/method/cursor.count. The best performance would be reached by setting `applySkipLimit` to `true` to the cursor `count()` method but a simple `find()` is at least better than the previous solution.

jenssegers:
Ok I will merge this until we find actual complications :)

alexandre-butynski:
Thanks.

Maybe did I have made this PR in the wrong branch but this should be integrated in the 2.2 version of the package too !

PR #623, opened 2015-10-15
Add job queue support for mongodb.

jeremy-brenner:
This adds job queue support by making a few small changes to the Database job queue libraries.  It has a separate service provider making it's usage completely optional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,17 +105,7 @@ You can connect to multiple servers or replica sets with the following configura
 ),
 ```
 
-For job queue support add a mongodb connection in config/queue.php:
-```
-        'mongodb' => [
-            'driver' => 'mongodb',
-            'table' => 'jobs',
-            'queue' => 'default',
-            'expire' => 60,
-        ],
-```
-
-You also need to set your QUEUE_DRIVER to 'mongodb' in your environment or .env
+You also need to set your QUEUE_DRIVER to 'database' in your environment or .env
 
 Eloquent
 --------
@@ -934,4 +924,3 @@ DB::connection()->disableQueryLog();
 ```
 
 *From: http://laravel.com/docs/database#query-logging*
-
--- a/src/Jenssegers/Mongodb/MongodbQueueServiceProvider.php
+++ b/src/Jenssegers/Mongodb/MongodbQueueServiceProvider.php
@@ -1,0 +1,29 @@
+<?php namespace Jenssegers\Mongodb;
+
+use Illuminate\Support\ServiceProvider;
+use Jenssegers\Mongodb\Queue\MongodbConnector;
+
+class MongodbQueueServiceProvider extends ServiceProvider {
+
+    /**
+     * Bootstrap the application events.
+     */
+    public function boot()
+    {
+
+    }
+
+    /**
+     * Register the service provider.
+     */
+    public function register()
+    {
+        $this->app->resolving('queue', function($queue)
+        {
+          $queue->extend('mongodb', function() {
+            return new MongodbConnector($this->app['db']);
+          });
+        });
+    }
+
+}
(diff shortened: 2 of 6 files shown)


CurosMJ:
Instead of having a `table` config parameter, we should have a `collection` parameter, since this is Mongo.

CurosMJ:
And, why not have a seperate `mongodb` queue driver, instead of extending the database driver?

jeremy-brenner:
It isn't a driver per se.  The 'database' setting uses whatever database driver Laravel is already using.  If you set your queue driver to database while the laravel-mongodb driver is in use it will try to use your mongo database and fail.  This really just patches it so that it can work as intended.  I have not really looked at how the other queues are implemented but since mongodb isn't really a queue then I doubt anybody would want a standalone mongodb queue while using some other database as default laravel db driver.  That is my reasoning anyway. :)

johngieselmann:
Any progress?

jenssegers:
@jeremy-brenner What did you change in the `MongoQueue` class? It looks almost identical as the `DatabaseQueue` class.

leakingminds:
Any news on this? Does it work?

jenssegers:
Currently waiting one @jeremy-brenner's response.

leakingminds:
hmm... ok well, since this post is pretty old, i guess there wont be any response... 
has anyone tried it out yet? i havent really worked with qeues, so i cant confirm this!

jeremy-brenner:
The difference is that Mongo can't "select for update" so it was replaced with a simple loop.  As for the naming, the "database" and "table" nomenclature was kept because really the "database" driver has nothing to do with the db and really means "use the ORM" which in this case happens to be Moloquent rather than Eloquent.  If I switch then name then I have to do a little bit of redirecting instead of just using the ORM directly like it already does.  It's a trivial difference so I left it.  It works fine, I use it personally.  I have not done a recent merge and test but I use it in a couple projects.

leakingminds:
Sweet! Pretty excited to try it out :)

One more question, which i didnt really understand from your post: Do you use Eloquent or Moloquent for this?

jeremy-brenner:
This is a simple connector to make job queues work with Moloquent.  It just adds the necessary files and makes no changes to Moloquent directly.  You just need to load them as outlined in the docs to make it work.

jeremy-brenner:
Is there any more I can do to help this go more smoothly?  Any other outstanding questions that I have not covered?

jenssegers:
I prefer extending original Laravel classes rather than copy pasting all their code. So I need to figure out what was changed before I can create an extended class. Or if you want to refactor your code that's good as well :)

jeremy-brenner:
The issue is pretty straightforward.  
In releaseJobsThatHaveBeenReservedTooLong in Illuminate/Queue/DatabaseQueue.php the query chains an update.  This doesn't work with mongo. 

    protected function releaseJobsThatHaveBeenReservedTooLong($queue)
    {
        $expired = Carbon::now()->subSeconds($this->expire)->getTimestamp();

        $this->database->table($this->table)
                    ->where('queue', $this->getQueue($queue))
                    ->where('reserved', 1)
                    ->where('reserved_at', '<=', $expired)
                    ->update([
                        'reserved' => 0,
                        'reserved_at' => null,
                        'attempts' => new Expression('attempts + 1'),
                    ]);
    }

Thats really it.  I tried extending the Illuminate classes but had little luck.  Another approach might be to simply emulate the ->update() behavior in the db driver leaving the database queue driver alone entirely.

jenssegers:
Thanks @jeremy-brenner, I have added your code using an extending class in my latest commits.

leakingminds:
Nice, that was fast :)

How about using and setting everything else up? Is it basically the same as:

https://laravel.com/docs/5.1/queues

Nothing else to keep in mind?

Thanks guys!

jenssegers:
@leakingminds I updated the readme as well.

PR #630, opened 2015-10-24
Fix for Builder compileWheres relation id

scrummitch:
With belongsTo/hasOne/hasMany relations, there seems to be a problem detecting the relation foreign key because its searching for an ending of "._id" instead of just "_id".

Example:
User has many Photos using the normal eloquent method, you insert  $user->photos()->save($photo) and a "user_id" of type MongoId is attached to the photo but when you go to search all the users photos, it doesnt have the "._id" at the end of the key so it cant find any children.

### DIFF
--- a/src/Jenssegers/Mongodb/Query/Builder.php
+++ b/src/Jenssegers/Mongodb/Query/Builder.php
@@ -861,7 +861,7 @@ class Builder extends BaseBuilder {
             }
 
             // Convert id's.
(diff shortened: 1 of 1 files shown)


scrummitch:
Is there a way of saving a one a to many relationship using the traditional eloquent method of $object->children->save($child) instead of Child::create([object_id = $object->_id]) ? I would be willing to help make this change

pi0:

Thanks :)

PR #660, opened 2015-12-02
Update to support non-MongoDate instances in DatabaseTokenRepository.php 

wuori:
Ref: https://github.com/jenssegers/laravel-mongodb/issues/493#issuecomment-161142179

Sorry, not fully adjusted to making pull requests on public projects, nor do I know for certain this is a solid fix. But it works well in my Homestead environment and, as an elseif with two conditionals, doesn't seem to pose much risk.

Thanks for all the work you put into this!


jenssegers:
I can't really merge this PR with all the other commits attached to it :p

wuori:
Alright, I'll fix this up. To confirm, should I be working of the "develop" branch or master? Develop seems a bit old.

wuori:
Conferred with some buddies at work, they suggested creating fix branch off of master then pull requesting just my one commit, so that should fix the problem.

PR #666, opened 2015-12-04
Add sparse unique function to Blueprint

pengkong:
this will allow us to create a sparse unique index in migrations.


jenssegers:
You should be able to do `->unique('xx', ['sparse' => true])`

PR #669, opened 2015-12-06
Import the Moloquent class before extending it

djug:


jenssegers:
You are referencing the new mongodb driver which is not compatible with this library.

PR #691, opened 2016-01-01
Compatible with pecl-mongodb

i0:
Compatible with pecl-mongodb for PHP 7 and earlier versions,
pecl-mongo package has deprecated and isn't available for PHP 7

### DIFF
--- a/src/Jenssegers/Mongodb/Model.php
+++ b/src/Jenssegers/Mongodb/Model.php
@@ -10,10 +10,8 @@ use Jenssegers\Mongodb\Query\Builder as QueryBuilder;
 use Jenssegers\Mongodb\Relations\EmbedsMany;
 use Jenssegers\Mongodb\Relations\EmbedsOne;
 use Jenssegers\Mongodb\Relations\EmbedsOneOrMany;
-use MongoDate;
-use MongoId;
 use ReflectionMethod;
-
+use MongoDB;
 abstract class Model extends BaseModel {
 
     use HybridRelations;
@@ -54,8 +52,8 @@ abstract class Model extends BaseModel {
             $value = $this->attributes['_id'];
         }
 
-        // Convert MongoId's to string.
-        if ($value instanceof MongoId)
+        // Convert MongoDB\BSON\ObjectID's to string.
+        if ($value instanceof MongoDB\BSON\ObjectID)
         {
             return (string) $value;
         }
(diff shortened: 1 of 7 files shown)


src/Jenssegers/Mongodb/Relations/BelongsToMany.php
```diff
@@ -126,9 +126,7 @@ public function sync($ids, $detaching = true)
         {
             $this->detach($detach);
 
-             $changes['detached'] = (array) array_map(function ($v) {
-                return is_numeric($v) ? (int) $v : (string) $v;
-            }, $detach);
+            $changes['detached'] = (array) array_map(function ($v) { return (int) $v; }, $detach);
```
jenssegers: Why was this changed?

i0: I didn't change it,
Probably my laravel-mongodb package was older.

src/Jenssegers/Mongodb/Model.php
```diff
@@ -202,11 +201,11 @@ protected function getDateFormat()
...
      * Get a fresh timestamp for the model.
      *
-     * @return MongoDate
+     * @return MongoDB\BSON\UTCDateTime
      */
     public function freshTimestamp()
-    {
-        return new MongoDate;
+    {	
+	    return round(microtime(true) * 1000);
```
jenssegers: Shouldn't this return a UTCDateTime instance?

i0: In this case, it gives an exception because that the object values are not same between MongoDate and UTCDateTime, but the script will change generated microtime to an UTCDateTime instance later.

jenssegers:
Do you know if there is a way of telling mongodb to return arrays instead of objects?

i0:
There is a method named "toArray" for MongoDB\Driver\Cursor
see: http://php.net/manual/en/mongodb-driver-cursor.toarray.php

Symfomany:
You must use MongoDB\BSON\UTCDateTime for handle datetime

tobias-kuendig:
Make sure to rename the other `getName` and `ensureIndex` calls since they don't exist on https://github.com/mongodb/mongo-php-library anymore:

https://github.com/OFFLINE-GmbH/laravel-mongodb/compare/2dc68b03be0e9f9878937832029c16ca7d8fcf9d...OFFLINE-GmbH:ec2d90b26e6ec3fb3c2e24eea09ee873ef6666e8

austinh:
Any update on this?

andijames:
Yeah i'd be interested to know if there's any update too. If any help is required i'm more than happy to lend a hand where i can.

qutwala:
+1, i'm too interested and waiting.

jenssegers:
These changes are on the pecl-mongodb branch; but not all tests are green yet.

PR #710, opened 2016-01-26
Use the driver's native count() method where possible 

ht2:
We were running into real problems relating to using the aggregation pipeline to perform complex counts. As a result, this should attempt to use the MongoCollection's `count()` (
http://php.net/manual/en/mongocollection.count.php) method where the query is deemed simple enough to warrant it.

I've also included phpunit as a dependency as it was not being included by composer? I notice it is also included on your master branch, so I used the same as that.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
         "illuminate/events": "4.2.*"
     },
     "require-dev": {
+        "phpunit/phpunit": "^4.0|^5.0",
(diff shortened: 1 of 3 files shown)


ht2:
Err, some interesting Style failures! Not sure of the best way to rectify them

pi0:

Thanks :)

PR #723, opened 2016-02-04
Should be global MongoDB Class

nayzawoo:


jenssegers:
This is actually not returning MongoDB instances anymore.

PR #731, opened 2016-02-09
Added connection type check back

### LINKED ISSUE #673: `Jenssegers\Mongodb\Model` do not work with other databases
Hi,

Since version 2.1.8..2.1.9 was the change that prevents use of models inheriting `Jenssegers\Mongodb\Model` with other db drivers. For example it is impossible to use sqlite db during tests with this models. 

Exception:
`Argument 1 passed to Jenssegers\Mongodb\Query\Builder::__construct() must be an instance of  Jenssegers\Mongodb\Connection, instance of lluminate\Database\SQLiteConnection given, called in 
/[...]/vendor/jenssegers/mongodb/src/Jenssegers/Mongodb/Model.php on line 519`

tobias-kuendig:
This fixes #673 and enables the use of alternative drivers such as sqlite for unit tests.

### DIFF
--- a/src/Jenssegers/Mongodb/Eloquent/Model.php
+++ b/src/Jenssegers/Mongodb/Eloquent/Model.php
@@ -522,7 +522,12 @@ abstract class Model extends BaseModel
     {
         $connection = $this->getConnection();
 
-        return new QueryBuilder($connection, $connection->getPostProcessor());
+        // Check the connection type
+        if ($connection instanceof \Jenssegers\Mongodb\Connection) {
+            return new QueryBuilder($connection, $connection->getPostProcessor());
+        }
+
+        return parent::newBaseQueryBuilder();
     }
 
     /**


jenssegers:
Can you give me an example use case why this would be useful?

tobias-kuendig:
When running unit tests I set sqlite as a global database driver to run tests in memory. This means that `newBaseQueryBuilder` gets called with an sqlite instance. Since `Jenssegers\Mongodb\Query\Builder` only accepts a `Jenssegers\Mongodb\Connection` all tests fail with the following exception:

```
Argument 1 passed to Jenssegers\Mongodb\Query\Builder::__construct() must be an instance of Jenssegers\Mongodb\Connection, instance of lluminate\Database\SQLiteConnection given, called in /[...]/vendor/jenssegers/mongodb/src/Jenssegers/Mongodb/Model.php on line 519
```

With this check, the code falls back to the default `Illuminate\Database\Query\Builder` and you are able to run the tests with the sqlite driver.

Obviously not the most beautiful solution, but without the check is is pretty hard to use in memory tests.

You know of a better solution to this problem?

jenssegers:
I'm not sure if the model would even work if you would give it an SQLiteConnection.

PR #735, opened 2016-02-12
Fix timestamp errors

zhaff:
1. To generate valid token's timestamp using Carbon

2. To fix error: MongoDB\Driver\Exception\InvalidArgumentException with message 'MongoDB\BSON\UTCDateTime::__construct() expects parameter 1 to be integer, float given'

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,17 +249,11 @@ This service provider will slightly modify the internal DatabaseReminderReposito
 
 ### Queues
 
-If you want to use MongoDB as your database backend, change the default queue driver in `config/queue.php`:
-
-```php
-'default' => 'mongodb',
-```
-
-And add the following connection:
+If you want to use MongoDB as your database backend, change the the driver in `config/queue.php`:
 
 ```php
 'connections' => [
-    'mongodb' => [
+    'database' => [
         'driver' => 'mongodb',
         'table'  => 'jobs',
         'queue'  => 'default',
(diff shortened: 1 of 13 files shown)


jenssegers:
Why is `UTCDateTime` not working?

zhaff:
Yup, its not working, its returning: Uncaught MongoDB\Driver\Exception\InvalidArgumentException: MongoDB\BSON\UTCDateTime::__construct() expects parameter 1 to be integer

Tried these, also not working. I think the easiest way is to use Carbon.

```
$utcdatetime = new MongoDB\BSON\UTCDateTime(round(microtime(true)*1000));

$datetime = $utcdatetime->toDateTime();

var_dump($datetime);

// Return: Uncaught MongoDB\Driver\Exception\InvalidArgumentException: MongoDB\BSON\UTCDateTime::__construct() expects parameter 1 to be integer

$utcdatetime = new MongoDB\BSON\UTCDateTime((int) round(microtime(true)*1000));

$datetime = $utcdatetime->toDateTime();

var_dump($datetime);

// return: 1969-12-24 11:48:21.-164000000"


$utcdatetime = new MongoDB\BSON\UTCDateTime(time());

$datetime = $utcdatetime->toDateTime();

var_dump($datetime);

// return: 1970-01-17 20:15:45.34000000"
```

zhaff:
I just realized the problem is caused by my PHP_INT_MAX is 2147483647. Fyi, I am on windows x86_64. Just hold this PR at the moment.

zhaff:
Tested on homestead, all run just fine. I think only affected on Window.

zhaff:
Cancelled.

Issue #736, opened 2016-02-12
EmbedsOne relationship always have a model associated


Rezouce:
I upgraded Moloquent from 2.2.4 to 2.3.2.
Before, when no object was associated to an embedOne relation, I had a null value when trying to retrieve it, now I have an empty object.

Is this an intended new behaviour? If so, how should we check if the object is a prepopulated one or one we associated ourselves?

PR #741, opened 2016-02-15
EmbedsOne relationship returns null when no model is associated

Rezouce:
See issue #736 which has been closed yet was not resolved.


jenssegers:
Could you also add a unit test so that I does not happen again? :)

Rezouce:
Done.

PR #748, opened 2016-02-21
urlencode mongo password

timemachine3030:


jenssegers:
@timemachine3030 Do you know if there is an alternate way of passing credentials?

rjvandoesburg:
@jenssegers Hey, so by a coincidence I was having credential problems and I stumbled upon this ticket. I had a @ in my password that caused a problem (see password parameter here: http://uk3.php.net/manual/en/mongoclient.construct.php) I know this is deprecated but it seems the problem persists in the newer driver.

The new manager seems to still accept the same parameters (or at least the credentials)

So to answer the question you asked @timemachine3030 yes, there is an alternative way to pass the credentials, even other parameters.

I've have made some changes locally and am willing to put up a pull request but we might want to expand the parameters. What are your thoughts on this?

See the diff of the change I've made so far: [diff.txt](https://github.com/jenssegers/laravel-mongodb/files/146459/oei2BFBe.txt)

jenssegers:
If that works, please send the PR!

rjvandoesburg:
Actually, silly old me (only actually started using your project today) this is already possible.
See: https://github.com/rjvandoesburg/laravel-mongodb/commit/eaf2b8f1fdd297f2cd612c437cc2d18724746306

By adding the credentials to options in the config they are automatically passed as extra options in the MongoDB constructor :)

However this change will cause problems when people update because the config needs to be updated.

PR #753, opened 2016-02-25
Pass credentials as parameters

rjvandoesburg:
These changes will pass the credentials as parameters at the creation of a new MongoDB connection instead of passing them on the DSN.

reference: https://github.com/jenssegers/laravel-mongodb/pull/748

**Username and password now need to be in the options section of the DB config**

### DIFF
--- a/src/Jenssegers/Mongodb/Connection.php
+++ b/src/Jenssegers/Mongodb/Connection.php
@@ -137,11 +137,11 @@ class Connection extends \Illuminate\Database\Connection
         }
 
         // Check if the credentials are not already set in the options
-        if (!isset($options['username'])) {
-            $options['username'] = isset($config['username']) ? $config['username'] : '';
+        if (!isset($options['username']) && isset($config['username'])) {
+            $options['username'] = $config['username'];
         }
-        if (!isset($options['password'])) {
-            $options['password'] = isset($config['password']) ? $config['password'] : '';
+        if (!isset($options['password']) && isset($config['password'])) {
+            $options['password'] = $config['password'];
         }
 
         return new Client($dsn, $options, $driverOptions);
--- a/src/Jenssegers/Mongodb/Connection.php
+++ b/src/Jenssegers/Mongodb/Connection.php
@@ -136,6 +136,13 @@ class Connection extends \Illuminate\Database\Connection
             $driverOptions = $config['driver_options'];
         }
 
+        if (!isset($options['username'])){
+            $options['username'] = isset($config['username']) ? $config['username'] : '';
+        }
+        if (!isset($options['password'])){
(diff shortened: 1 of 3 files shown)


jenssegers:
I would prefer having the username and password outside of the `options` in the Laravel configuration (like it used to be). And pass them to the `options` in the `Connection` class.

rjvandoesburg:
Ok, I will change that back and add the credentials to the constructor in the `Connection` class

rjvandoesburg:
Ok, so the changes I've made are as followed.
In the `Connection` class I've added a couple of checks to see if a person has already provided credentials in the options, if not the credentials will be used from the config (the old way)

I've reverted the readme as no changes are there and I've changed the access level of a method because it was different from the access level of the method it was overriding.

But it seems I've managed to mess up the tests... I'll be fixing those next :)

src/Jenssegers/Mongodb/Connection.php
```diff
@@ -136,6 +136,14 @@ protected function createConnection($dsn, array $config, array $options)
             $driverOptions = $config['driver_options'];
         }
 
+        // Check if the credentials are not already set in the options
+        if (!isset($options['username'])) {
```
jenssegers: Can you turn this into:

```
if (! isset($options['username']) && isset($config['username'])) {
$options['username'] = $config['username'];
```

rjvandoesburg: Yes, I've seen that that is indeed where the problem was.

I've got the test suite running locally now as well so I can test before pushing bad code again.
I want to test the entire suite before pushing so I just need to get MySQL up and running, after that I will submit the new changes again

rjvandoesburg:
Tests are working again. Applied the change you found as well.

jenssegers:
Can you squash your commits?

rjvandoesburg:
Ah sorry missed the notification.
Cheers for accepting the PR

oneso:
You are not checking if the credentials are empty. This resulty in an "Authentication failed." exception.
Two options:

- add empty() check
or
- edit readme:
remove 'DB_USERNAME/DB_PASSWORD' from .env if empty and change database.php from:
```
env($test . 'DB_USERNAME', '')
env($test . 'DB_PASSWORD', '')
```
to:
```
env($test . 'DB_USERNAME')
env($test . 'DB_PASSWORD')
```

Or is it just me?

jenssegers:
I added it back again.

rjvandoesburg:
Hmm, I see what you mean.
However I do think there is a difference in having empty fields or no fields at all.
I'd suggest changing the readme so it states to remove the username and password if you wish to use anonymous login.

By having empty username and password a user might be alerted for misconfiguration in the .env file.

PR #769, opened 2016-03-04
update MongoQueue.php

giolvani:
Some fixes to delete successfully processed jobs on queues.


giolvani:
I replaced some methods because the base methos doesn't work with mongodb, perhaps because the primary key is ```_id``` and not ```id```

I overwrite the methods ```markJobAsReserved()``` and ```deleteReserved()``` and copied ```releaseJob()``` from @Bodom78 implementation.

davidchchang:
Thanks @giolvani!
@jenssegers: can this fix be merged to 2.3 and the other branches as well?

giolvani:
Your welcome guys!

Issue #752, opened 2016-02-25
 cannot delete reserved queue


nutjaa:
[https://github.com/laravel/framework/blob/5.2/src/Illuminate/Queue/DatabaseQueue.php#L242](url)
```

 /**
     * Delete a reserved job from the queue.
     *
     * @param  string  $queue
     * @param  string  $id
     * @return void
     */
    public function deleteReserved($queue, $id)
    {
        $this->database->table($this->table)->where('id', $id)->delete();
    }
```
No id field in jobs collection 
Did i miss some configuration ?

giolvani:
+1

davidchchang:
@nutjaa: you are right, it should be checking for _id instead of id. Pull request (#770) has been submitted and awaiting review.

davidchchang:
I just noticed that @jenssegers has merged this to 2.3 already. I believe this issue can be closed now.

PR #770, opened 2016-03-04
#752: Override DatabaseQueue's deleteReserved function to use Mongo _id.

davidchchang:
The default deleteReserved query checks the jobs table for records by id instead of _id. We can address this issue by overriding this method in MongoQueue.php.

This addresses the issue mentioned in #752.


jenssegers:
This was merged: https://github.com/jenssegers/laravel-mongodb/pull/769/files

PR #778, opened 2016-03-10
Allow Query builder pluck to work with _id

rjvandoesburg:
- As of 5.2 lists is deprecated so I moved the ObjectID to string code to pluck as well
  I kept pluck the same for backwards compatibility

If only Laravel 5.2 support is enough the lists method can be removed from the `Builder` class because the parent only calls pluck.

### DIFF
--- a/src/Jenssegers/Mongodb/Query/Builder.php
+++ b/src/Jenssegers/Mongodb/Query/Builder.php
@@ -566,6 +566,19 @@ class Builder extends BaseBuilder
      */
     public function pluck($column, $key = null)
     {
+        if ($key == '_id') {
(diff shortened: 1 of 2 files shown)


rjvandoesburg:
Ah, I've just noticed there are tests for the query builder so allow me to write some tests as well.

andycaramba:
Please, merge this PR. I have an error after updating to Laravel 5.2 with `_id` as `pluck` method `key` parameter. In Laravel 5.2 `lists` method just an alias to `pluck`.

pi0:
`lists` fully removed on last release :) so maybe better closing this PR?

PR #786, opened 2016-03-18
Fixes #743 for version 3.x

### LINKED ISSUE #743: Warning: Possible queue duplication with multiple processes
@jenssegers I previously implemented queues for a project using an older version of this library and looking at the MongoQueue.php in the master branch I noticed that it may suffer from the same issue as I did.

Running the queue as a daemon with multiple processes will cause a race condition where more then 1 process can select the same job before another has marked it as reserved. This in turn will produce randomly duplicated jobs.

The fix is to to get the next job and mark it as reserved using Mongo's findAndReplace to lock the document.

I created a "getNextAvailableJobAndReserve" method and updated the "pop" method to use it.

For reference here is what my MongobdQueue class method looks like, Take note of the modified **pop** and how it uses the **getNextAvailableJobAndReserve** method.

Hope this can be of some help.

```
<?php

namespace App\Mongodb\Queue;

use Carbon\Carbon;
use Illuminate\Queue\DatabaseQueue;
use DB;

class MongodbQueue extends DatabaseQueue
{
    /**
     * Pop the next job off of the queue.
     *
     * @param  string $queue
     *
     * @return \Illuminate\Contracts\Queue\Job|null
     */
    public function pop($queue = null)
    {
        $queue = $this->getQueue($queue);

        if (!is_null($this->expire))
        {
            $this->releaseJobsThatHaveBeenReservedTooLong($queue);
        }

        if ($job = $this->getNextAvailableJobAndReserve($queue))
        {
            return new MongodbJob(
                $this->container, $this, $job, $queue
            );
        }
    }

    /**
     * Release the jobs that have been reserved for too long.
     *
     * @param  string $queue
     *
     * @return void
     */
    protected function releaseJobsThatHaveBeenReservedTooLong($queue)
    {
        $expired  = Carbon::now()->subSeconds($this->expire)->getTimestamp();
        $reserved = $this->database->collection($this->table)
                                   ->where('queue', $this->getQueue($queue))
                                   ->where('reserved', 1)
                                   ->where('reserved_at', '<=', $expired)->get();

        foreach ($reserved as $job)
        {
            $attempts = $job['attempts'] + 1;
            $this->releaseJob($job['_id'], $attempts);
        }
    }

    /**
     * Get the next available job for the queue and mark it as reserved.
     *
     * @param  string|null $queue
     *
     * @return \StdClass|null
     */
    protected function getNextAvailableJobAndReserve($queue)
    {
        $job = DB::getCollection($this->table)->findAndModify(
            [
                'queue'        => $this->getQueue($queue),
                'reserved'     => 0,
                'available_at' => ['$lte' => $this->getTime()],

            ],
            [
                '$set' => [
                    'reserved'    => 1,
                    'reserved_at' => $this->getTime(),
                ],
            ],
            null,
            [
                'new' => true,
            ]
        );

        return $job ? (object)$job : null;
    }

    /**
     * Release the given job ID from reservation.
     *
     * @param  string $id
     *
     * @return void
     */
    protected function releaseJob($id, $attempts)
    {
        $this->database->table($this->table)->where('_id', $id)->update([
            'reserved'    => 0,
            'reserved_at' => null,
            'attempts'    => $attempts,
        ]);
    }

    /**
     * Delete a reserved job from the queue.
     *
     * @param  string $queue
     * @param  string $id
     *
     * @return void
     */
    public function deleteReserved($queue, $id)
    {
        $this->database->table($this->table)->where('_id', $id)->delete();
    }
}
```

Bodom78:
Resolves issue where possible daemons could process the same record by using the  findOneAndUpdate to lock a job record while flagging it as reserved at the same time.

### DIFF
--- a/src/Jenssegers/Mongodb/Queue/MongoQueue.php
+++ b/src/Jenssegers/Mongodb/Queue/MongoQueue.php
@@ -2,31 +2,78 @@
 
 use Carbon\Carbon;
 use Illuminate\Queue\DatabaseQueue;
+use Illuminate\Queue\Jobs\DatabaseJob;
+use MongoDB\Operation\FindOneAndUpdate;
+use DB;
 
 class MongoQueue extends DatabaseQueue
 {
     /**
-     * Get the next available job for the queue.
+     * Pop the next job off of the queue.
      *
-     * @param  string|null  $queue
+     * @param  string $queue
+     *
+     * @return \Illuminate\Contracts\Queue\Job|null
+     */
+    public function pop($queue = null)
+    {
+        $queue = $this->getQueue($queue);
+
+        if (!is_null($this->expire))
+        {
+            $this->releaseJobsThatHaveBeenReservedTooLong($queue);
+        }
+
+        if ($job = $this->getNextAvailableJobAndReserve($queue))
+        {
+            return new DatabaseJob(
+                $this->container, $this, $job, $queue
+            );
+        }
+    }
+
+    /**
+     * Get the next available job for the queue and mark it as reserved.
+     *
+     * When using multiple daemon queue listeners to process jobs there
+     * is a possibility that multiple processes can end up reading the
+     * same record before one has flagged it as reserved.
+     *
+     * This race condition can result in random jobs being run more then
+     * once. To solve this we use findOneAndUpdate to lock the next jobs
+     * record while flagging it as reserved at the same time.
+     *
+     * @param  string|null $queue
+     *
      * @return \StdClass|null
      */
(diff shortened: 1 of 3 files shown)


jenssegers:
Any chances you could write unit tests for this?

Bodom78:
@jenssegers I'm not too familiar with unit testing but I'll give it a go. Is this why the coverage is decreased?

Bodom78:
@jenssegers I've added some tests that cover necessary methods. 

Testing also showed that the getNextAvailableJobAndReserve method was not returning the updated document so that has been resolved too.

All checks now pass.

PR #792, opened 2016-03-23
Fixes #789

### LINKED ISSUE #789: Artisan queue [failed] [retry] do not work as expected
**artisan queue:failed** 
Does not work because the failed_at column is not stored as a timestamp.
Exception thrown: strstr() expects parameter 1 to be string, array given 

**artisan queue:retry id** 
Does re queue the job but does not delete the record in the failed_jobs collection.
Exception thrown: Undefined property: stdClass::$id

**artisan queue:retry all** 
Does not find any failed jobs even when there are some present.
Error: No failed job matches the given ID [].

Bodom78:
Docs updated with required information to use MongoDB for failed jobs.

Artisan queue failed, retry with id and retry all now work as expected.


Bodom78:
I don't understand why StyleCI analysis is failing when it just reports a close brace is to be replaced with a close brace?

pi0:
@Bodom78 StyleCI fails because it should be an empty line at the end of file :)

pi0:
@Bodom78 StyleCI fails because it should be an empty line at the end of file :)

Bodom78:
@pi0 Thanks for that.

pi0:

Thanks :)

PR #795, opened 2016-03-29
importing DateTime and DateTimeZone from the global namespace

dschreck:
Looks like someone fixed one of the \DateTimeZone calls but missed the other global namespace classes. Placed use statements to include them, class no longer throws runtime exception.


dschreck:
Seems related to https://github.com/jenssegers/laravel-mongodb/pull/751

PR #798, opened 2016-03-30
Fixed logQuery to display ObjectID in parameters

moura137:
Also Collection class tests was created


gnrcampos:
+1, debugbar giving me a hard time

PR #825, opened 2016-04-28
adds troubleshooting section

vivex:
Troubleshooting section was missing so adding that.


jenssegers:
This is for the old mongo pecl extension, the new one is `mongodb`.

PR #827, opened 2016-04-29
update composer.json

triadev:
use the laravel version 5.2


jenssegers:
This is not necessary at the moment. I prefer to keep it compatible as long as possible.

PR #835, opened 2016-05-11
Fix #556 Adding sparce & unique indices

### LINKED ISSUE #556: Unique & Sparse indices
Hey,

In the scenario we're working with, sign in is optional, but we still create a User model for each person which stays with them even as guests. This means we will have multiple models with a `null` email address field. However, for those persons who have elected to sign in, and therefore we have their email address, the email field needs to be unique to ensure no duplicate accounts.

If we simply apply the Unique index to the email field, as soon as a second guest enters the app, we get the following error message:

    MongoDuplicateKeyException in Collection.php line 42:
    127.0.0.1:27017: E11000 duplicate key error index: genii.Auth_EndUsers.$email_1 dup key: { : null }

No biggie! MongoDB already thinks of this, and provides us a way of only checking entries where the field is not null (see [Sparse indices](http://docs.mongodb.org/manual/core/index-sparse/)).

It's nice to see that this library supports sparse indices, however, when I try to apply the index (using `$collection->sparse('email');`) to the already-unique index I get the following error message:

    ben$ php artisan migrate:refresh
    Nothing to rollback.

    [MongoResultException]                                                                      
    127.0.0.1:27017: exception: Index with name: email_1 already exists with different options  

Apologies if any of this is unclear. Has anyone else come across this, or found the solution?

benargo:
Provides a handy shortcut for allowing both sparse and unique indices on columns. MongoDB provides a method of only checking unique values where the field is not null (or 'sparse'). At present, using the `unique` and `sparse` functions separately throws an `MongoDuplicateKeyException` (see issue #556). 

You can get around this by using the `index` function, but my understanding is this should be avoided (please correct me if I'm wrong).

### DIFF
--- a/src/Jenssegers/Mongodb/Schema/Blueprint.php
+++ b/src/Jenssegers/Mongodb/Schema/Blueprint.php
@@ -225,14 +225,14 @@ class Blueprint extends \Illuminate\Database\Schema\Blueprint
      */
     public function sparse_and_unique($columns = null, $options = [])
     {
-      $columns = $this->fluent($columns);
+        $columns = $this->fluent($columns);
 
-      $options['sparse'] = true;
-      $options['unique'] = true;
+        $options['sparse'] = true;
+        $options['unique'] = true;
 
-      $this->index($columns, $options);
+        $this->index($columns, $options);
 
-      return $this;
+        return $this;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


pi0:
@jenssegers why not merging this ?

PR #842, opened 2016-05-16
Little timesaving fix at read.me

Denys-Bushulyak:
Wrong param at options array.


jenssegers:
Could you link me to a documentations page for this? Yesterday I got a PR for this exact change. Which one is the correct one?

Denys-Bushulyak:
Sorry, it was mistake. When I put 'db' key instead 'database' key it became workable. But it's because it became to use default 'admin' database.

PR #845, opened 2016-05-19
Add transactions to MongoDB

korobkadima:


jenssegers:
Could you make a commit with only the actual changes?

PR #851, opened 2016-05-26
Add auto-increment _id support

Gtskk:
Add auto-incrementing _id support.

With the value of `incrementing` property, you can easily enable or disable the auto-incrementing  _id support.

**Note:** By default, the function is enabled. You can set `incrementing=false` to disable the function.

PS: Forgiving my poor English.

### DIFF
--- a/src/Jenssegers/Mongodb/Eloquent/Model.php
+++ b/src/Jenssegers/Mongodb/Eloquent/Model.php
@@ -38,6 +38,51 @@ abstract class Model extends BaseModel
      */
     protected $parentRelation;
 
+
+    /**
+     * Get next id value.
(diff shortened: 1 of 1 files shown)


RTLer:
adding auto-incrementing id to mongo like this, is not good idea considering race condition
it can cause duplicate key error.

Gtskk:
@RTLer  And a better solution?

RTLer:
i use redis to to generate true incremental id.

amiri27:
@Gtskk auto_increment using mongodb is not safe with race conditions. You can use redis cache to auto_increment indexes which is safe regarding race conditions.

PR #871, opened 2016-06-19
Overwrite transaction functions to avoid errors

### LINKED ISSUE #854: laravel ^5.2.34 with Queue transaction error
Hello,
In laravel ^5.2.34 have  added a transaction in DatabaseQueue.php:pop() https://github.com/laravel/framework/commit/2715509a2842ee0402fd8ceebc21aab92c6d1f91
From that version and above transaction are not working for me because this package does not support it. Is there a workaround or some fix we can do?

nunomazer:
MongoDB do not have transactions, so when Laravel tries to begin one an exception is throwed. Fix #854.

To avoid it this PR overwrites trnasactions functions, removing the lines that call PDO transaction at original code. This only preserves flag propertie increments and decrements, also the events fired.

### DIFF
--- a/src/Jenssegers/Mongodb/Connection.php
+++ b/src/Jenssegers/Mongodb/Connection.php
@@ -208,6 +208,43 @@ class Connection extends \Illuminate\Database\Connection
     }
 
     /**
+     * Start a new database transaction.
+     *
+     * @return void
+     * @throws Exception
+     */
+    public function beginTransaction()
+    {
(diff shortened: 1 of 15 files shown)


nunomazer:
I don't know how to set the tests to work, since none of the failures has something to do with my chenges ... can someone help with it?

nunomazer:
Seens Queues were implemented, thanks .. closing this

Issue #528, opened 2015-06-27
INFO: Laravel Auth and Unique


dereckmartin:
This is not really an issue.  I am hoping maybe to help others out there with some auth while using mongodb, so do what you want with this ticket.  I'm using the latest laravel, and I had to extend several of the functions from the auth stuff to get user authentication working.  

I don't know if anyone has tested this, but checks for uniqueness is case sensitive as well as attempting auth. 

I had to extend the user create after registration to strtolower() the email.  Other wise someuser@domain.tld was different than SomeUser@domain.tld.

When authenticating a user, I had to extend the postLogin in auth/AuthController to transform email addresses into a uniform format, so that I could auth someone that couldn't remember if they registered SomeUser@domain.tld vs someuser@domain.tld.

``` php
    /**
     * Handle a login request to the application.
     *
     * @param  \Illuminate\Http\Request  $request
     * @return \Illuminate\Http\Response
     */
    public function postLogin(Request $request)
    {

        /* Added code that differs from the laravel's postLogin */
        $rd = $request->all();
        $rd['email'] = strtolower($request->input('email'));
        $request->replace($rd);
        /* end of added code */

        $this->validate($request, [
            'email' => 'required|email|exists:users', 
            'password' => 'required',
        ]);

        $credentials = $this->getCredentials($request);

        var_dump($credentials);

        if (Auth::attempt($credentials, $request->has('remember'))) {
            return redirect()->intended($this->redirectPath());
        }

        return redirect($this->loginPath())
            ->withInput($request->only('email', 'remember'))
            ->withErrors([
                'email' => $this->getFailedLoginMessage(),
            ]);
    }
```

shehi:
Closing this issue would be nice. Search engines will pick up on it anyway.

PR #900, opened 2016-07-25
[Closed][Stale Code] Fix working of unique and exists

CurosMJ:
Before this, unique rule was case sensitive, that means it would pass strings with same chars but different case. Example, if curosmj@gmail.com exists in database, applying unique rule on curosMj@gmail.com SHOULD have failed, but it does not, since mongo is case sensitive, and case sensitiveness cannot be configured. See https://github.com/laravel/framework/issues/9430. Also seen in #528 

I've implemented a DatabasePresenceVerifier for mongodb, which overrides the default presence verifier, ONLY IF MongodbValidationServiceProvider is used. Also added test cases to make sure 'unique' and 'exists' validation rules function in a case insensitive manner.

This is an issue that affects almost all applications that would be using this plugin,  and it is pretty serious.

Ping @jenssegers.


pi0:

Thanks :)

CurosMJ:
See #1196

PR #909, opened 2016-08-06
Fixed an error in readme.md

rhbvkleef:
Example with syntax error:

Examples/Incrementing or decrementing a value of a column

`User->inc....`

changed to

`User::inc...`


pi0:
@jenssegers should it be fixed ? :D

jenssegers:
I don't think this is a typo.

PR #912, opened 2016-08-08
add sub document aggregation functionality (sum, count, ...)

fieteboerner:
this commit allows you to  count, sum, ... on a sub document attributes like:

Model::sum('subarray.price');
this method is much simpler as to use as complex raw aggregations

to get this behaviour to work a $unwind directive will be pushed
before the $group directive


fieteboerner:
I have noticed my pr is in conflict with simple subdocument aggragation functions.
A update that fixes this issue will coming soon.

PR #918, opened 2016-08-13
Aggregate subdocument arrays

fieteboerner:
With this code it is possible to use aggregation functions (sum, count, max, min, avg, ...) on subdocument arrays (embedsMany). To use this functionality you have to pass `array.*.field` to the corresponding aggregation function.

Check out the code example in `testSubdocumentArrayAggregate` in `QueryBuilderTest.php`.

As you can see with this solution it is very simple to apply the aggregation functions instead of writing complex raw aggregations. In my scenario, the code is much simpler and cleaner.

I would be glad to get feedback if you are interested to this functionality, or if you have suggestions for improvements or alternatives.

### DIFF
--- a/src/Jenssegers/Mongodb/Query/Builder.php
+++ b/src/Jenssegers/Mongodb/Query/Builder.php
@@ -251,7 +251,7 @@ class Builder extends BaseBuilder
             }
 
             // apply unwinds for subdocument array aggregation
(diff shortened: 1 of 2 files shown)


pi0:

Thanks :)

fieteboerner:
ok, i've just updated it :)

PR #930, opened 2016-08-29
5.3 UnitTests

pi0:
I'm working on Unit Tests to be fully passed.

Update: All tests passed!

### DIFF
--- a/src/Jenssegers/Mongodb/Eloquent/Builder.php
+++ b/src/Jenssegers/Mongodb/Eloquent/Builder.php
@@ -2,6 +2,7 @@
 
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
 use Illuminate\Database\Eloquent\Relations\Relation;
+use Illuminate\Support\Collection;
 use MongoDB\Driver\Cursor;
 use MongoDB\Model\BSONDocument;
 
@@ -167,9 +168,10 @@ class Builder extends EloquentBuilder
         $query = $hasQuery->getQuery();
 
         // Get the number of related objects for each possible parent.
+        $relations = $query->pluck($relation->getHasCompareKey());
         $relationCount = array_count_values(array_map(function ($id) {
             return (string) $id; // Convert Back ObjectIds to Strings
-        }, $query->pluck($relation->getHasCompareKey())));
+        }, is_array($relations) ? $relations : $relations->toArray()));
 
         // Remove unwanted related objects based on the operator and count.
         $relationCount = array_filter($relationCount, function ($counted) use ($count, $operator) {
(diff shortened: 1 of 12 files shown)


pi0:
@jenssegers Finally all tests fully passed :) Would you please Review and Approve this PR ?

src/Jenssegers/Mongodb/Eloquent/Builder.php
```diff
@@ -167,9 +168,10 @@ protected function addHasWhere(EloquentBuilder $hasQuery, Relation $relation, $o
         $query = $hasQuery->getQuery();
 
         // Get the number of related objects for each possible parent.
+        $relations = $query->pluck($relation->getHasCompareKey());
         $relationCount = array_count_values(array_map(function ($id) {
             return (string) $id; // Convert Back ObjectIds to Strings
-        }, $query->pluck($relation->getHasCompareKey())));
+        }, is_array($relations) ? $relations : $relations->toArray()));
```
jenssegers: This seems like a big change?

jenssegers: Why was this needed?

jenssegers:
Looking good!

src/Jenssegers/Mongodb/Eloquent/Model.php
```diff
@@ -233,11 +234,11 @@ public function getAttribute($key)
 
                 // This attribute matches an embedsOne or embedsMany relation so we need
                 // to return the relation results instead of the interal attributes.
-                if ($relations instanceof EmbedsOneOrMany) {
+                if ($relations instanceof EmbedsOneOrMany or $relations instanceof MorphTo) {
```
jenssegers: This seems a bit odd, this is very specific to embedded relations, why was `MorpTo` added here?

src/Jenssegers/Mongodb/Eloquent/Builder.php
pi0: This tests fail without this check:
- RelationsTest.php:384
- RelationsTest.php:433

Error: `array_map(): Argument #2 should be an array`

Laravel `has` returns arrays instead of Collections!

src/Jenssegers/Mongodb/Eloquent/Model.php
pi0: RelationsTest.php:367 : `Failed asserting that null is an instance of class "User".`
As a workaround to working MorphTo Relations will be directly returned and resolved.
Is it necessary to separate this logical part?

jenssegers: Might be better to split the logic indeed, I'll have to check what exactly is going wrong because this is weird.

jenssegers:
Let me know when your changes are done!

pi0:
@jenssegers last commit was due to accidental fast-forward :) i you think something is missing just let me know :)

jenssegers:
Ill release this as an RC and go over everything soon.

PR #948, opened 2016-09-10
Remove dependency on Laravel Application

morrislaptop:


morrislaptop:
@acidjazz @pi0 i've switched to your repo, i'll take one less namespace anyday!

pi0:
@jenssegers isn't this a safe solution ? and would you consider merging this ?

PR #953, opened 2016-09-14
Convert whereBetween values to UTCDateTime

ricardofontanelli:
Added a protected method (dateTimeConvertion) to convert DateTime objects to MongoDB\BSON\UTCDateTime and added a verification for DateTime objects in $where['values'] array to ensure that all DateTime values will be properly converted to MongoDB\BSON\UTCDateTime.

### DIFF
--- a/src/Jenssegers/Mongodb/Query/Builder.php
+++ b/src/Jenssegers/Mongodb/Query/Builder.php
@@ -866,7 +866,16 @@ class Builder extends BaseBuilder
 
             // Convert DateTime values to UTCDateTime.
             if (isset($where['value']) and $where['value'] instanceof DateTime) {
-                $where['value'] = new UTCDateTime($where['value']->getTimestamp() * 1000);
+                $where['value'] = $this->dateTimeConvertion($where['value']);
+            }
+
+            // Convert DateTime values to UTCDateTime in $where['values'] key.
+            if (isset($where['values'])) {
+                foreach ($where['values'] as $key => $value) {
+                    if ($value instanceof DateTime) {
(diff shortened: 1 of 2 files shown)


src/Jenssegers/Mongodb/Query/Builder.php
```diff
@@ -1018,6 +1029,20 @@ protected function compileWhereRaw($where)
...
+    /**
+     * Convert to MongoDB\BSON\UTCDateTime if $value is a DateTime object.
+     *
+     * @param  mixed   $value
+     * @return mixed
+     */
+    protected function dateTimeConvertion($value)
+    {
+        if ($value instanceof DateTime) {
+            return new UTCDateTime($value->getTimestamp() * 1000);
```
GreenLamantin: Why UTCDateTime instead UTCDatetime?

ricardofontanelli: According PHP [documentation](http://php.net/manual/pt_BR/class.mongodb-bson-utcdatetime.php), the class name is UTCDate**T**ime

aditya-cherukuri: Isn't the if conditional check redundant? You are already checking if the value is an instance of DateTime before calling the dateTimeConversion function.

Opsse:
This issue isn't fixed yet ?

ricardofontanelli:
It was merged at moloquent/moloquent, I don't know why nobody has merged this PR here!

D3nnisH:
already fixed by #1328

divine:
> already fixed by #1328
Then this PR should be closed.

PR #1012, opened 2016-10-31
Set the correct auth database in connection string

philipbrown:
Hey,

This is related to https://github.com/jenssegers/laravel-mongodb/pull/971

I've just realised that the connection string should be using the auth database from the `options` array, rather than the main database.

This PR corrects that problem. Sorry, only just noticed after you merged my previous PR 😞


philipbrown:
Thanks Jens ❤️

jenssegers:
No problem! Will tag new version after some more fixes/tweaks.

PR #1055, opened 2016-12-07
Don't set updated_at if using operator on collection

hackel:
When using an operator (key starts with $), we should not automatically
set the updated_at field as this generates an error.

Additional logic could be added to check whether a $set operator is present
and add the updated_at field to that, or create it, but for now I've simply
omitted it, as the developer probably knows what he or she is doing when using
a custom operator.

### DIFF
--- a/src/Jenssegers/Mongodb/Eloquent/Builder.php
+++ b/src/Jenssegers/Mongodb/Eloquent/Builder.php
@@ -35,7 +35,11 @@ class Builder extends EloquentBuilder
             return 1;
         }
 
-        return $this->query->update($this->addUpdatedAtColumn($values), $options);
+        if (! starts_with(key($values), '$')) {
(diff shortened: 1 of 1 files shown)


hackel:
Ugh, really, Composer?  How does this make any sense?
```
- satooshi/php-coveralls v0.6.0 requires psr/log 1.0.0 -> satisfiable by psr/log[1.0.0].
- satooshi/php-coveralls v0.6.1 requires psr/log 1.0.0 -> satisfiable by psr/log[1.0.0].
- Conclusion: don't install psr/log 1.0.0
- Installation request for satooshi/php-coveralls ^0.6 -> satisfiable by satooshi/php-coveralls[v0.6.0, v0.6.1].
```

Anyway, this is solved by updating the `satooshi/php-coveralls` constraint to "^1.0" instead of "^0.6", but that is a different issue.

PR #1063, opened 2016-12-23
MongoDB document key validation

jsertx:
This pull request add the validation rule to check if a field is a correct ObjectID value.
The ObjectID value is a hexadecimal value with 24 digit.


jsertx:
I don't understand how to solve the style check, can anyone explain to me?

PR #1069, opened 2017-01-03
Fix attribute read for values nested on multiple levels

bstanescu:
There is a bug in the retrieval of intermediary nested values
Let's assume the following structure of data:
`$foo['a.b.c'] = 1;`

```php
>>> $foo['a'];
=> [
     "b" => [
       "c" => 1,
     ],
   ]

>>> $foo['a.b'];
=> null // <-- this is wrong as it should return this:

>>> $foo['a']['b'];
=> [
     "c" => 1,
   ]

>>> $foo['a.b.c'];  // this works fine
=> 1
```

### DIFF
--- a/src/Jenssegers/Mongodb/Eloquent/Model.php
+++ b/src/Jenssegers/Mongodb/Eloquent/Model.php
@@ -237,16 +237,7 @@ abstract class Model extends BaseModel
      */
     protected function getAttributeFromArray($key)
     {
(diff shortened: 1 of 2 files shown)


jenssegers:
Could you still use the if that checks if the key contains a dot? So that we still use the parent logic in any other case.

bstanescu:
Good point (pun intended 😄 ). Done.

PR #1089, opened 2017-01-25
Undefined property: Jenssegers\Mongodb\Relations\BelongsTo::$otherKey

vincentri:
I upgrade to laravel 5.4 and i got an error when using relation belongs to.

Here is the file after I correct it in belongs.php

<?php namespace Jenssegers\Mongodb\Relations;

class BelongsTo extends \Illuminate\Database\Eloquent\Relations\BelongsTo
{
    /**
     * Set the base constraints on the relation query.
     */
    public function addConstraints()
    {
        if (static::$constraints) {
            // For belongs to relationships, which are essentially the inverse of has one
            // or has many relationships, we need to actually query on the primary key
            // of the related models matching on the foreign key that's on a parent.
            $this->query->where($this->ownerKey, '=', $this->parent->{$this->foreignKey});
        }
    }

    /**
     * Set the constraints for an eager load of the relation.
     *
     * @param  array  $models
     */
    public function addEagerConstraints(array $models)
    {
        // We'll grab the primary key name of the related models since it could be set to
        // a non-standard name and not "id". We will then construct the constraint for
        // our eagerly loading query so it returns the proper models from execution.
        $key = $this->otherKey;

        $this->query->whereIn($key, $this->getEagerModelKeys($models));
    }
}


jenssegers:
There seems to be something wrong with your PR?

vincentri:
Sorry already solved

 Thank you

Sent from my iPad

On Feb 14, 2017, at 21:15, Jens Segers <notifications@github.com<mailto:notifications@github.com>> wrote:


There seems to be something wrong with your PR?

PR #1095, opened 2017-01-31
Updates PasswordBroker to work with Laravel 5.4

wuori:
### Bad code alert: 

This code is a shim until new dot release is available. As reset-password isn't used as often as other functions, the version check's added process should be negligible, but it still isn't the best way, obviously. 

*TODO:* Remove check for next dot release (non-compatible with Laravel < 5.4).

### PR Info:

Laravel 5.4 [implements Hasher into `DatabaseTokenRepository`](https://github.com/laravel/framework/commit/c454c87f75c179043a956071a8572af538c926a9#diff-f682f161553e5df2ef0864ef427eaa40) and therefore requires this parameter to be passed to it from `PasswordBrokerManager.php` in laravel-mongodb.

This PR adds a check to see if we're on 5.4+ and passes hasher if so, allowing it to remain compatible with prior versions of Laravel.

### DIFF
--- a/src/Jenssegers/Mongodb/Auth/PasswordBrokerManager.php
+++ b/src/Jenssegers/Mongodb/Auth/PasswordBrokerManager.php
@@ -12,11 +12,26 @@ class PasswordBrokerManager extends BasePasswordBrokerManager
      */
     protected function createTokenRepository(array $config)
     {
-        return new DatabaseTokenRepository(
-            $this->app['db']->connection(),
-            $config['table'],
-            $this->app['config']['app.key'],
-            $config['expire']
-        );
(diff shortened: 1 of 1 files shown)


jenssegers:
Can you use this to compare the version? http://php.net/manual/en/function.version-compare.php

wuori:
Sure I'll try to get that done soon, thanks!

wuori:
@jenssegers Doesn't seem like this is needed any longer?

divine:
> @jenssegers Doesn't seem like this is needed any longer?

Should be closed, merged in https://github.com/jenssegers/laravel-mongodb/pull/1162

Thank you for your contribution.

cc @Smolevich

PR #1100, opened 2017-02-02
Authenticatable User Class

artdarek:


jenssegers:
Any chance you could only include the user class and not your second commit?

artdarek:
Hi, I reverted second commit... will that work for you?

PR #1106, opened 2017-02-05
Relations: Add support for columns name that ends with '_id'

estebanprimost:
If the mongo's collection have a field of type **`ObjectID`** that references another collection and it ends with `_id`, the query (i.e.: `hasMany` relation setup) does not match any results.

### DIFF
--- a/src/Jenssegers/Mongodb/Query/Builder.php
+++ b/src/Jenssegers/Mongodb/Query/Builder.php
@@ -895,7 +895,7 @@ class Builder extends BaseBuilder
             }
 
             // Convert id's.
-            if (isset($where['column']) and ($where['column'] == '_id' or ends_with($where['column'], '._id'))) {
+            if (isset($where['column']) and ($where['column'] == '_id' or ends_with($where['column'], '_id'))) {
                 // Multiple values.
                 if (isset($where['values'])) {
                     foreach ($where['values'] as &$value) {


jenssegers:
Can you provide some more information about what is going wrong?

estebanprimost:
Hi @jenssegers, of course. My collection `Response` has an attribute named like `words_id`, that holds an array of Mongo's [ObjectId](https://docs.mongodb.com/manual/reference/bson-types/#objectid) that represents the `_id` of a `word` document. 

Collections as JSON:
```
// Response collection document
{
  _id: '58a86d72aef73451484710d8',
  words_id: ['542c2b97bac0595474108b48', '507f191e810c19729de860ea'],
  ...
}

// Word collection document
{
   _id: '542c2b97bac0595474108b48'
   name: 'Blahblah'
   ...
}
```

In the `Word` Model I have a `belongsToMany` relationship defined as:
```
public function responses() {
    return $this->belongsToMany('App\Models\Response', '_id', 'words_id');
}
```

Those relationship not working with current configuration. So I [made a little change](https://github.com/jenssegers/laravel-mongodb/pull/1106/commits/a6915f70ceb46dee9df37ec5ca25d55f4aea88f9) in the Builder.php class to allow me work with this schema. Test are not passing, but I think it's because other problem with tests itself. I want to see if this change broke something else.

Thanks and sorry for my English.

PR #1115, opened 2017-02-14
Fixed "Error parsing [float] as 64-bit integer" in Model::freshTimestamp()

macdabby:
Fixes an error when `round(microtime(true) * 1000)` works out to be a float. Running in HHVM 3.15.4

https://github.com/jenssegers/laravel-mongodb/issues/712

### DIFF
--- a/src/Jenssegers/Mongodb/Eloquent/Model.php
+++ b/src/Jenssegers/Mongodb/Eloquent/Model.php
@@ -191,7 +191,7 @@ abstract class Model extends BaseModel
      */
     public function freshTimestamp()
     {
-        return new UTCDateTime(round(microtime(true) * 1000));
+        return new UTCDateTime((int) round(microtime(true) * 1000));
     }
 
     /**


jenssegers:
Can you change the casting to `(int)`?

jenssegers:
Can you change the casting to `(int)`?

macdabby:
Absolutely - done!

elghobaty:
@jenssegers @macdabby Overflow?!
```
microtime(true) = 1487826998.2899
microtime(true) * 1000 = 1487826998289.9
round(microtime(true) * 1000) = 1487826998290
round(microtime(true) * 1000) = 1487826998290
(int) round(microtime(true) * 1000) = 1768313874
```
`new UTCDateTime((int) round(microtime(true) * 1000))` will evaluate to a date in 1970-ish.

macdabby:
Interesting - php version?

macdabby:
This function should also accept null to use the current time. @jenssegers should we use that instead?

jenssegers:
@macdabby Wow, why didn't we see that. Are you up for making a PR?

PR #1116, opened 2017-02-15
Get the Other/Owner Key name based on different version of Illuminate/Database

netcode:
Recently Taylor changed the names in relations otherKey to ownerKey
see commit [commit](https://github.com/illuminate/database/commit/6a35698d72e276f435324b7e29b3cd37ef7d5d9c)


jenssegers:
Any chance you could use `getOwnerKey` as the function name?

netcode:
Yea , it will be more readable :+1:

PR #1121, opened 2017-02-17
Model class referencing according to Laravel 5.4

JanisOzolins:
As of Laravel 5.4, the way model class names can be references has been changed. Now you must user a namespace path in front of the model class in order for it to work (e.g., 'App\User' instead of just 'User'). README.md examples and "Relations" section updated accordingly.


divine:
Hello!

PR #1127, opened 2017-02-23
Fixed 32-bit integer overflow by allowing UTCDateTime() to get it's own current time

macdabby:
@jenssegers This fixes the issue over on https://github.com/jenssegers/laravel-mongodb/pull/1115

### DIFF
--- a/src/Jenssegers/Mongodb/Auth/DatabaseTokenRepository.php
+++ b/src/Jenssegers/Mongodb/Auth/DatabaseTokenRepository.php
@@ -12,7 +12,7 @@ class DatabaseTokenRepository extends BaseDatabaseTokenRepository
      */
     protected function getPayload($email, $token)
     {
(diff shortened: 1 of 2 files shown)


macdabby:
I ran tests with this code in the freshTimestamp() method:

        print new UTCDateTime().get;
        print "\n";
        print new UTCDateTime((int)round(microtime(true) * 1000));
        print "\n";

What's interesting is that the output was as follows:

1487777270195
1487777270195
1487777270197
1487777270197
1487777270197
1487777270197
1487777270197
1487777270198
1487777270197
1487777270198
1487777270198
1487777270198
1487777270198
1487777270198
1487777270198
1487777270199
1487777270198
1487777270199

In most cases, you'll see the matching pair, however in a few cases they were off by one second. I suppose that comes from rounding.

jenssegers:
Can you change `src/Jenssegers/Mongodb/Auth/DatabaseTokenRepository.php` as well?

macdabby:
Yeah, I also added an explicit null as a parameter to prevent notices. Tested to make sure I got the same output as above.

PR #1162, opened 2017-03-24
Added missing dependency

dees040:
The DatabaseTokenRepository in PasswordBrokerManager missed the Hasher parameter.


jenssegers:
This might break things for older Laravel versions. Probably we will need to check the Laravel version first.

dees040:
I made a smell check to verify the Laravel version.

vaxilicaihouxian:
Why does this commit not merge to the master branch?

PR #1170, opened 2017-04-04
Update Builder.php

mikield:
Do not use app function because this function is absent in other frameworks (as silex for example)
We can check if interface exists
casue from laravel code:
```<?php

namespace Illuminate\Database\Eloquent;

/**
 * @deprecated since version 5.2. Use Illuminate\Database\Eloquent\Scope.
 */
interface ScopeInterface extends Scope
{
}


mikield:
@jenssegers does my PR seems not useful?

divine:
Hello,

Sorry for late reply, this will be in upcoming version 4.

Thank you.

PR #1186, opened 2017-04-26
Aggregation shouldn't mutate querybuilder

lintaba:
Aggregations (like $query->count() ) modifies the internal state of the queryBuilder object, therefore the $query looses eg. the projection.

Eloquent solves this in the following way:
https://github.com/laravel/framework/blob/5.3/src/Illuminate/Database/Query/Builder.php#L2036


lintaba:
related: 
https://github.com/moloquent/moloquent/pull/64
https://github.com/moloquent/moloquent/issues/63

PR #1196, opened 2017-05-21
[Bug Fix][Validation][Security] Fix working of unique and exists

CurosMJ:
**NOTE : This is the same issue as #900**

Before this, unique rule was case sensitive, that means it would pass strings with same chars but different case. Example, if curosmj@gmail.com exists in database, applying unique rule on curosMj@gmail.com SHOULD have failed, but it does not, since mongo is case sensitive, and case sensitiveness cannot be configured. See laravel/framework#9430. Also seen in #528

I've implemented a DatabasePresenceVerifier for mongodb, which overrides the default presence verifier, ONLY IF ValidationServiceProvider is used. Also added test cases to make sure 'unique' and 'exists' validation rules function in a case insensitive manner.

This is an issue that affects almost all applications that would be using this plugin, and it is pretty serious.


prashant-pokhriyal:
@CurosMJ, after this merge, the unique validation rule is getting failed for sub-text.

        User::create(['name' => 'John Doe']);
        $validator = Validator::make(
            ['name' => 'John'],
            ['name' => 'required|unique:users']
        );
        $this->assertFalse($validator->fails()); // it fails

Problem is in regex

https://github.com/jenssegers/laravel-mongodb/blob/bc78c7575b879351eedc4a57b5ac0410ee765b51/src/Jenssegers/Mongodb/Validation/DatabasePresenceVerifier.php#L20

Instead, it should be

        $query = $this->table($collection)->where($column, 'regex', '/^'.$value.'$/i'); 

One issue is already opened for the same #1616

PR #1197, opened 2017-05-22
PHP7 Support for Laravel 4.2

Ardakilic:
This pull request aims to bring PHP7 support for Laravel 4.2 packages, which officially support PHP7.

All credits goes to @glowdan from his comment here: https://github.com/alcaeus/mongo-php-adapter/issues/107#issuecomment-219393254 

Also, another discussion regarding this issue here: https://github.com/LearningLocker/learninglocker/issues/893

If this specific version is selected as requirement, this may break compatibility with PHP versions lower than 5.6, and since there hasn't been an update to 2.0 branch for a while, I believe, if this would be accepted, a simple note would suffice for users.

Thanks in advance,

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ composer.lock
 *.sublime-project
 *.sublime-workspace
 *.project
+/.idea
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
         "illuminate/support": "4.2.*",
         "illuminate/container": "4.2.*",
         "illuminate/database": "4.2.*",
-        "illuminate/events": "4.2.*"
+        "illuminate/events": "4.2.*",
+        "alcaeus/mongo-php-adapter": "^1.1.0"
     },
     "require-dev": {
         "orchestra/testbench": "2.2.*",
--- a/src/Jenssegers/Mongodb/Collection.php
+++ b/src/Jenssegers/Mongodb/Collection.php
@@ -56,7 +56,22 @@ class Collection {
 
         $start = microtime(true);
 
-        $result = call_user_func_array(array($this->collection, $method), $parameters);
+        //$result = call_user_func_array(array($this->collection, $method), $parameters);
+
+        // Fix for PHP7
+        // https://github.com/LearningLocker/learninglocker/issues/893#issue-203456708
+        // based on https://github.com/alcaeus/mongo-php-adapter/issues/107#issuecomment-219393254
+        if (PHP_VERSION_ID >= 70000 && in_array($method, array('insert', 'batchInsert', 'save'))) {
+            $saveData = array_shift($parameters);
+            $saveParams = array_shift($parameters);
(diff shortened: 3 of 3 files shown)


jenssegers:
Thanks for your contribution @Ardakilic!

Ardakilic:
@jenssegers Awesome! Thank you for reviewing and merging this! 🎉

opb:
Hi @Ardakilic I'm having problems with this change in our L4 app. I seem to be having the same issue as https://github.com/LearningLocker/learninglocker/issues/893#issuecomment-287790662, though we're not using LearningLocker. Turning error reporting off allows the insert to work, if not, the document is still inserted, but we don't have an _id to return.

Any suggestions would be much appreciated!

Ardakilic:
@opb Hey there, I'm not using LearlngLocker neither. Just found the fix there. Can you give me some example code please to reproduce the issue ?

adiyodishreyas:
@Ardakilic  I am having issues with _id not getting returned after my insert query
For eg:
$post = array('x' => 1)
Model::insert($post);
After the insert query was executed the $post model used to get updated with the _id 
i.e array('_id' => BSON, 'x' => 1);

Am I missing something?
I am using laravel 4.2.20, jenssegers/mongodb 2.0.x-dev, php 7
Thanks

adiyodishreyas:
@Ardakilic even $model->save(); seems to be failing. 
I get the error "Undefined index _id"

Thanks.

Ardakilic:
@adiyodishreyas I guess your issue is same with @opb 's , this one: https://github.com/LearningLocker/learninglocker/issues/893#issuecomment-287790662

Sorry, I just brought the commit as a pull request. I've tried to fix it further but stuck on same issue. Have you tried moving your issue to here? https://github.com/alcaeus/mongo-php-adapter/issues/107#issuecomment-219393254 Sorry can't help you further.

Ardakilic:
Late bump.

If you're still in need of this in PHP7 With Laravel 4. I made a commit to prevent the undefined index: _id exception issue for the cases where the row's just been created and on the same request that field's being called. This'll work, but only if you don't need the _id field beginning at all.

Check this commit:

https://github.com/Ardakilic/laravel-mongodb/commit/99ddd15e05da498a4e1dc728fe6eb6e82eb56550

This prevents the  ` "Undefined index _id" ` issue, but please consider my warn on the first paragraph. I didn't need the _id field upon creation, so this works for me, also everything else seem to work nicely.

To use my fork, simply alter your `composer.json` like this:

```json
{
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/Ardakilic/laravel-mongodb"
        }
    ],
    "require": {
        "jenssegers/mongodb": "dev-2.0-php7"
    }
}
```

`minimum-stability` can still be still set as `stable`. No worries there.

opb:
Thanks @Ardakilic this works!

PR #1280, opened 2017-08-30
Update for Laravel 5.5

SimplyCorey:
This updates the `belongsToMany` signature for Laravel 5.5

See: https://laravel.com/docs/5.5/upgrade

### DIFF
--- a/src/Jenssegers/Mongodb/Eloquent/HybridRelations.php
+++ b/src/Jenssegers/Mongodb/Eloquent/HybridRelations.php
@@ -207,14 +207,18 @@ trait HybridRelations
     /**
      * Define a many-to-many relationship.
      *
-     * @param  string $related
-     * @param  string $collection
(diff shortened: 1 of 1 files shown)


SimplyCorey:
Laravel support is going to need to be bumped to 5.5 in the composer.json file.  Not sure how you want to handle that.

RemiCollin:
Nope, checks the composer.json, version constraint is `^5.1`, so it will not prevent installing it from a 5.5 version. 

What we might need is to tag a new package version (3.3 ?) to reflect the breaking change.

akaroot:
see PR #1282

SimplyCorey:
3.3 would also have to drop support for php5.6 in the test suite.

SimplyCorey:
Closing in favor of: https://github.com/jenssegers/laravel-mongodb/pull/1300

PR #1312, opened 2017-09-22
Fix for This database driver does not support dropping all tables Exception and couple of support methods

thilanga:
- addHasWhere
- dropIfExists
- dropAllTables
- getAllTables

dropAllTables and getAllTables are there to fix `This database driver does not support dropping all tables`


thilanga:
StyleCI issue has been fixed

jenssegers:
Is it possible to make a PR without the docblock changes? This makes it more difficult to review.

thilanga:
Will do it tomorrow 👍

jackmiras:
@thilanga just for curiosity, why you are adding functions with table sufix? Mongo is a data base that is document and collections '"based". Wouldn't be more idiomaticly if you call the function dropAllTables as dropAllCollections and the function getAllTables as getAllCollections?

PR #1317, opened 2017-10-02
Fix issue with chunkById.

DFurnes:
The `chunkById` method doesn't work with MongoDB, because the framework [assumes that the ID will be numeric](https://github.com/laravel/framework/blob/997e67cc2001006a43b73eaedf17fa1a2555fd66/src/Illuminate/Database/Query/Builder.php#L1836) & doesn't provide an easy way for the user to override that at runtime. I was able to fix this by overriding the [`forPageAfterId`](https://github.com/laravel/framework/blob/997e67cc2001006a43b73eaedf17fa1a2555fd66/src/Illuminate/Database/Query/Builder.php#L1541-L1556) method so that it would convert Laravel's assumed "zero" ID into an equivalent "zero" ObjectId.

I've also added a test to demonstrate the bug & fix.

### DIFF
--- a/src/Jenssegers/Mongodb/Eloquent/Builder.php
+++ b/src/Jenssegers/Mongodb/Eloquent/Builder.php
@@ -146,6 +146,14 @@ class Builder extends EloquentBuilder
     /**
      * @inheritdoc
      */
+    public function chunkById($count, callable $callback, $column = '_id', $alias = null)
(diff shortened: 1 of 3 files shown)


DFurnes:
@jenssegers Would you mind reviewing this when you have a chance? Thanks!

PR #1320, opened 2017-10-05
Added support for "whereAll" queries

lindelius:


lindelius:
Can we get this reviewed and merged, please? @jenssegers

lindelius:
Thanks!

GromNaN:
Why did you need this method @lindelius?

Both
```php
->where('tags', 'all', ['ssl', 'security'])
```
and
```php
->whereAll('tags', ['ssl', 'security'])
```

Produced the same MongoDB query
```ts
{ tags: { $all: ["ssl", "security"] } }
```

lindelius:
Not really, I haven't used the library for years, but it was problaby because there was either no other way to do it back then or that I just wanted a nicer way to do it than to explicitly have to type the "all" parameter. I really can't remember, though. Sorry.

lindelius:
FWIW, the projects I worked on back then are no longer using the method

GromNaN:
Thanks for checking.

PR #1326, opened 2017-10-11
fix for returning incorrect builder in hybrid relations

pimlie:
@jenssegers @Tucker-Eric 

Commit 266305caeed70cf8ee787bcd0140c9fca2f0bcd6 introduced a bug when using HybridRelations in combination with e.g. raw methods. The `newEloquentBuilder` method always returned a `Illuminate\Database\Eloquent\Builder` but for `Jenssegers\Mongodb\Eloquent\Model` models it should return a `Jenssegers\Mongodb\Eloquent\Builder`.

### DIFF
--- a/tests/models/User.php
+++ b/tests/models/User.php
@@ -9,8 +9,7 @@ use Illuminate\Contracts\Auth\CanResetPassword as CanResetPasswordContract;
 
 class User extends Eloquent implements AuthenticatableContract, CanResetPasswordContract
 {
-    use Authenticatable, CanResetPassword;
-    use HybridRelations;
+    use Authenticatable, CanResetPassword, HybridRelations;
 
     protected $connection = 'mongodb';
     protected $dates = ['birthday', 'entry.date'];
--- a/tests/models/User.php
+++ b/tests/models/User.php
@@ -1,6 +1,7 @@
 <?php
 
 use Jenssegers\Mongodb\Eloquent\Model as Eloquent;
+use Jenssegers\Mongodb\Eloquent\HybridRelations;
 use Illuminate\Auth\Authenticatable;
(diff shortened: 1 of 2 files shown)


pimlie:
@jenssegers Sorry for the failed test, I have fixed the CI and now it passes. I would be obliged if you could merge this :)

tests/models/User.php
```diff
@@ -1,14 +1,15 @@
...
 use Jenssegers\Mongodb\Eloquent\Model as Eloquent;
+use Jenssegers\Mongodb\Eloquent\HybridRelations;
 use Illuminate\Auth\Authenticatable;
 use Illuminate\Auth\Passwords\CanResetPassword;
 use Illuminate\Contracts\Auth\Authenticatable as AuthenticatableContract;
 use Illuminate\Contracts\Auth\CanResetPassword as CanResetPasswordContract;
 
 class User extends Eloquent implements AuthenticatableContract, CanResetPasswordContract
 {
-    use Authenticatable, CanResetPassword;
```
jenssegers: This should not be needed right? Since `Eloquent` had this trait as well?

pimlie: Not sure what you mean, but I added the `HybridRelations` trait for the User model as using that trait immediately gives an error without this fix. In the first commit I used 2 `use` statements which I replaced to be 1 statement, therefor this line shows as deleted.

I dont think the Eloquent model class (either mongodb's or laravel's) implements one of these traits already

jenssegers: `Jenssegers\Mongodb\Eloquent\Model` already has the `HybridRelations` trait, which the model extends from.

pimlie: Ah ok. Thanks for merging.

Issue #1276, opened 2017-08-29
Issues with date whereBetween not work


im-zhangxi:
my code:
$startDate = new \DateTime($startDate);
$endDate = new \DateTime($endDate);
User::whereBetween('created_at', [$startDate, $endDate ]);

Execute SQL:
User.aggregate([{"$match":{{"created_at":{"$gte":{"date":"2016-09-01 00:00:00.000000","timezone_type":3,"timezone":"Asia\/Shanghai"},"$lte":{"date":"2017-09-01 00:00:00.000000","timezone_type":3,"timezone":"Asia\/Shanghai"}}},{"deleted_at":null}]}},{"$group":{"aggregate":{"$sum":1},"_id":null}}],{"typeMap":{"root":"array","document":"array"}})

Please help

hartono-sulaiman-kaskus:
I also had this problem

Wu4m4n:
I have the same issue.

im-zhangxi:
@Wu4m4n me too

mustyzod:
> I have the same issue.
> To fix this I use
> 
> `where($dateField,'>=',$begin)->where($dateField,'<=',$end)`
> 
> and works great!

i have tried all this earlier, it doesnt work with time, only the date is considered in the quey, i need it to filter date and time

PR #1328, opened 2017-10-11
Fix for issue #1276

lindelius:
This fixes the `DateTime` issue (#1276) that occurs when using methods like `whereBetween()` that sets the `values` field instead of the `value` field when compiling "wheres."


lindelius:
Can we get a review and merge on this one, too? @jenssegers

PR #1343, opened 2017-10-23
Fix Unable to set a custom collection for models

robbytaylor:
Laravel allows models to define a custom collection class. However, at the moment the Collection class name is hard-coded breaking this functionality.

This PR calls the related model's newCollection method to return a collection.

See https://laravel.com/docs/5.5/eloquent-collections#custom-collections


robbytaylor:
Thanks 👍

PR #1350, opened 2017-10-30
fix queue

jgcl:
fix queue
https://github.com/jenssegers/laravel-mongodb/issues/1349


jgcl:
fix MaxAttemptsExceededException inserted on all jobs

PR #1352, opened 2017-11-03
Fixed Error "Cannot use object of type MongoDB\BSON\UTCDateTime as array"

nolanle:
When I reset password with token but have error like "Cannot use object of type MongoDB\BSON\UTCDateTime as array"
And this is solution for fix it.
Thanks alot!


arthurkirkosa:
@jenssegers this does solve the reset password problem because of laravel latest changes.

arunmisra2273:
I get error Parse error: syntax error, unexpected 'if' (T_IF) when I use this fix

arunmisra2273:
works. Thank you! @nolanle

mikekamornikov:
@jenssegers There are at least 3 issues/prs here with "reset password" fix. Could you merge this (or any other) one?

koAndy:
@jenssegers Tested and worked! Please merge this!

PR #1375, opened 2017-11-23
Convert ALL DateTime values to UTCDateTime in Query\Builder::compileWheres

D3nnisH:
```
$builder->where('timestamp', '>=', $carbonDateObject1)->where('timestamp', '<=', $carbonDateObject2)
```
was working, but
```
$builder->whereBetween('timestamp', [$carbonDateObject1, $carbonDateObject2])
```
wasn't.

This was because in the `compileWheres` function only converted `$where['value']` to `UTCDateTime`, but not `$where['values']`, where the former was filled in easy queries and the latter one in between queries. MongoDB then couldn't handle the given [Carbon](http://carbon.nesbot.com/) objects (possibly autoconverted to strings)

### DIFF
--- a/src/Jenssegers/Mongodb/Query/Builder.php
+++ b/src/Jenssegers/Mongodb/Query/Builder.php
@@ -920,6 +920,13 @@ class Builder extends BaseBuilder
             }
 
             // Convert DateTime values to UTCDateTime.
+            if (isset($where['values'])) {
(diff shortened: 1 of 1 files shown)


D3nnisH:
oh, I see it's a duplicate of https://github.com/jenssegers/laravel-mongodb/pull/953/files , but that one has merge conflicts

D3nnisH:
#1328 contains the exact same diff and has been merged

PR #1377, opened 2017-11-27
fix: gets key name from model rather than assuming _id

willtj:
As merged in Moloquent repo under https://github.com/moloquent/moloquent/pull/73


willtj:
We GitHub amateurs have been inadvertently pushing into this PR - closing this one and I'll open another for its intended purpose.

PR #1378, opened 2017-11-30
Change Carbon include to fix SerializeUsing & jsonSerialize incosistency

Christophvh:
Changed the Carbon include to use the extended version of Laravel.

`use Carbon\Carbon  => use Illuminate\Support\Carbon`

This way we can use the `jsonSerialize()` and  `serializeUsing()`. methods

For example in the new ApiResources in Laravel 5.5 we can specify a global serialization for Carbon dates: 
```
   
Carbon::serializeUsing(function ($carbon) {
            return $carbon->format('U');
        });
```

https://laravel.com/docs/5.5/eloquent-serialization#date-serialization

### DIFF
--- a/src/Jenssegers/Mongodb/Eloquent/Model.php
+++ b/src/Jenssegers/Mongodb/Eloquent/Model.php
@@ -2,7 +2,7 @@
 
 namespace Jenssegers\Mongodb\Eloquent;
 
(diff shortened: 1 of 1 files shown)


divine:
Hello!

Giacomo92:
It can be closed @jenssegers

PR #1385, opened 2017-12-08
Refactoring tests

carusogabriel:
I've refactored several tests, using:
- `assertCount` instead of `count` function;
- `assertInternalType` instead of `is_*` functions;
- `assertArrayHasKey` and `assertArrayNotHasKey` instead of `array_key_exists` function;
- `assertTrue` instead of comparison with `true` keyword;
- `assertFalse` instead of comparison with `false` keyword;
- `assertNull` instead of comparison with `null` keyword;
- `assertContains` and `assertNotContains` instead of `in_array` function;
- `assertObjectNotHasAttribute` instead of `isset` function;
- `assertLessThan` for mathematical comparisons.

### DIFF
--- a/tests/ConnectionTest.php
+++ b/tests/ConnectionTest.php
@@ -44,7 +44,7 @@ class ConnectionTest extends TestCase
     // public function testDynamic()
     // {
     //     $dbs = DB::connection('mongodb')->listCollections();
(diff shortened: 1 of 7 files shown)


jenssegers:
Thanks!

PR #1386, opened 2017-12-08
Test against PHP 7.2

carusogabriel:
With the release of [`PHP 7.2`](http://php.net/archive/2017.php#id2017-11-30-1), would be interesting to test it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: php
 php:
   - 7
   - 7.1
+  - 7.2
 
 matrix:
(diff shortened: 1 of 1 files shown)


jenssegers:
Looks like there are some issues with installing mongodb

carusogabriel:
@jenssegers By [Travis error log](https://travis-ci.org/jenssegers/laravel-mongodb/jobs/327205384#L489), should we skip the installation in `PHP 7.2`?

PR #1392, opened 2017-12-18
fix: find() now uses getKeyName() for primary key

### LINKED ISSUE #1354: Overriding $primaryKey does not work with find($id)
I get the error:
Undefined index: myId in /Jenssegers/Mongodb/Query/Builder.php:575

It looks like the Builder class is hard-coded to expect the key as _id:

`return $sequence == '_id' ? $result->getInsertedId() : $values[$sequence];`

This causes problems when doing a find($id) when you've modified the primaryKey to be something else in the model class.

`public $primaryKey = 'myId';`

I'm using the latest master version using PHP7 and Mongo 3.4

DMNSteve:
fixes #1354

### DIFF
--- a/src/Jenssegers/Mongodb/Query/Builder.php
+++ b/src/Jenssegers/Mongodb/Query/Builder.php
@@ -198,7 +198,7 @@ class Builder extends BaseBuilder
      */
     public function find($id, $columns = [])
     {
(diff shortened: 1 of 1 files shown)


danjohnson95:
🎉

PR #1411, opened 2018-01-24
Update mongodb/mongodb to support all options and added hint

pedrolopes10:
Updating to mongodb/mongodb 1.2.0 uses this fix:
https://jira.mongodb.org/browse/PHPLIB-264

Also added hint to options.

Warning: requires driver ext-mongodb ^1.3.0


divine:
Should be closed in favor of https://github.com/jenssegers/laravel-mongodb/pull/1556 which has a test.

PR #1451, opened 2018-02-27
Passport support

dmitov:
Passport package has embedded models, which are tightly coupled to the Illuminate's Eloquent. One way of making Passport work is taking ownership of all the models, but this way you are limited if Passport package updates those models.

A service provider could take responsibility for mapping Illuminate models to Jenssegers' models.

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,7 +298,7 @@ This service provider will slightly modify the internal DatabaseReminderReposito
 
 ### Passport
 
-Passport models are strictly `Illuminate\Database\Eloquent\Model` class. One way is to take ownership of the models, or simply include this service provider:
+Laravel Passport models are strictly using the `Illuminate\Database\Eloquent\Model` class. One way to make the package work is to take ownership of the models, or simply include this service provider:
 
 ```php
 'Jenssegers\Mongodb\Auth\PassportServiceProvider',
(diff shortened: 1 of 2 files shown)


RemiCollin:
Facing the same issue right now. I'm afraid this solution won't work as Passport models are not resolved by the IoC container, but instantiated using `new`.

dmitov:
I am using the solution and it works. The models are indeed instantiated using `new` but their base model is in the `use` case, therefore the IoC container will take over the class resolving. 

Maybe we could use some tests here?

RemiCollin:
No the IoC container won't take over unless called explicitely using `make()` . Simple to verify : 

```php 

    app()->when(\Laravel\Passport\Token::class)->needs(\Illuminate\Database\Eloquent\Model::class)->give(\Jenssegers\Mongodb\Eloquent\Model::class);

    $model = new \Laravel\Passport\Token;
    dd($model instanceof \Jenssegers\Mongodb\Eloquent\Model); // false

```

StevePorter92:
Please feel free to use https://github.com/designmynight/laravel-mongodb-passport for passport support with https://github.com/jenssegers/laravel-mongodb

sadnub:
I just created a console command to fix the passport vendor files. Just run `php artisan fix:passport`.
It is available here: https://github.com/sadnub/laravel-mongodb-passport-fix

Issue #1456, opened 2018-03-08
No suitable servers found (`serverSelectionTryOnce` set): [Failed to resolve 'localhost:27017']


halaei:
This config used to work in version 3.3 (laravel 5.5) it is not working now in 3.4 (laravel 5.6)
```php
        'mongodb' => [
            'driver'   => 'mongodb',
            'dsn' => 'mongodb://localhost:27017',
            'database' => env('MONGODB_DATABASE'),
        ],
```
## Error:
```
In Collection.php line 515:
                                                                                                   
  No suitable servers found (`serverSelectionTryOnce` set): [Failed to resolve 'localhost:27017'] 
```

halaei:
#1397 is the braking pull request.
In `Connection` class `$this->getDsn()` returns the invalid dsn `"mongodb://localhost%3A27017"`

ping @StevePorter92

Issue #1449, opened 2018-02-26
Connection string broken after upgrading to Laravel 5.6


Bowens20832:
I am attempting to upgrade my install from 5.5 to 5.6. I am getting  No suitable servers found (`serverSelectionTryOnce` set) on my production and sandbox clusters. 

Any ideas whats going on? Everything works fine in 5.5. 

My database.php looks like:

'mongodb' => [
            'driver'   => 'mongodb',
            'dsn' =>env('MONGO_PATH', ''),
            'database' => ' databasename ',
        ],

MordiSacks:
Had the same problem, I think something with the dsn parsing is broken
this worked for me

```
        'mongodb' => [
            'driver'   => 'mongodb',
            'host'     => [
                'xxx-00-00-mfddc.mongodb.net',
                'xxx-00-01-mfddc.mongodb.net',
                'xxx-00-02-mfddc.mongodb.net',
            ],
            'port'     => 27017,
            'database' => env('MONGODB_DATABASE'),
            'username' => env('MONGODB_USERNAME'),
            'password' => env('MONGODB_PASSWORD'),
            'options'  => [
                'ssl'        => true,
                'replicaSet' => 'replicaSetName',
            ],
        ],
```

Bowens20832:
@MordiSacks lol I literally took your approach a couple hours ago which fixed the issue. Thank you for your suggestion.

divine:
Closing, merged in https://github.com/jenssegers/laravel-mongodb/pull/1457

PR #1457, opened 2018-03-08
fix dsn config when not unix socket

halaei:
It is assumed in #1397 that dsn is only for unix sockets. But the assumption is wrong, causing issues #1456 and #1449 

IMO, a better approach would be having a separate `unix_socket` config to be encoded via `rawurlencode()` in this package or just let the users call `rawurlencode()` themselves when providing such URIs in 'dsn'.

### DIFF
--- a/src/Jenssegers/Mongodb/Connection.php
+++ b/src/Jenssegers/Mongodb/Connection.php
@@ -4,7 +4,6 @@ namespace Jenssegers\Mongodb;
 
 use Illuminate\Database\Connection as BaseConnection;
 use Illuminate\Support\Arr;
-use Illuminate\Support\Str;
 use MongoDB\Client;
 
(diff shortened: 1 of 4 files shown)


wppd:
`rawurlencode($dsn_string)` causes problem if not using unix socket......

jgcl:
merge this pull request, please!!!!!!

hlorofos:
@halaei good job on this, I have to fork the project and apply your fix to use the driver.
Its ridiculous this PR still haven't merged yet 8-)
I also added an ability to parse out database name from DSN string here https://github.com/jenssegers/laravel-mongodb/pull/1491

pendexgabo:
I would like to offer my help in any way in order to push this live. #1397 broke many things at once

jenssegers:
Sorry for the delay!

PR #1459, opened 2018-03-09
New way for drop index

### LINKED ISSUE #1036: Drop Index in Migration Can't Find Index with Name
I get this output when running the migration rollback command. But the up method works fine, and creates the index as expected. It's rolling back(dropping the index that's the issue).

screen shot 2016-11-17 at 2 28 23 pm

This is the migration file content:

```
<?php

use Illuminate\Database\Schema\Blueprint;
use Illuminate\Database\Migrations\Migration;

class CreateProductFacetsIndex extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        Schema::connection('productsdb')->table('products', function (Blueprint $collection) {
            $collection->index(
                [
                    "facets.name" => 1,
                    "facets.value" => 1,
                ],
                'facets_index',
                null,
                ['name' => 'facets_index']
            );
        });
    }

    /**
     * Reverse the migrations.
     *
     * @return void
     */
    public function down()
    {
        Schema::connection('productsdb')->table('products', function (Blueprint $collection) {
            $collection->dropIndex(['facets_index']);
        });
    }
}

```

I think this is a bug. Looking at the code, I see that '_1' is always appended to the index name when the dropIndex method is called.

Dmitry-Kucher:
Сhanged `dropIndex` method to add support for correct creation/dropping index by index name.
Added test case with index creation using `$name` argument, which failed with current implementation of `dropIndex`

Closes #1036

### DIFF
--- a/src/Jenssegers/Mongodb/Schema/Blueprint.php
+++ b/src/Jenssegers/Mongodb/Schema/Blueprint.php
@@ -80,19 +80,24 @@ class Blueprint extends \Illuminate\Database\Schema\Blueprint
     {
         $columns = $this->fluent($columns);
 
-        // Columns are passed as a default array.
-        if (is_array($columns) && is_int(key($columns))) {
-            // Transform the columns to the required array format.
-            $transform = [];
-
-            foreach ($columns as $column) {
-                $transform[$column] = $column . '_1';
-            }
-
-            $columns = $transform;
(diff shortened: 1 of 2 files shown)


divine:
Should be closed, https://github.com/jenssegers/laravel-mongodb/pull/1479 already merged.

PR #1468, opened 2018-03-17
fix tests: assertObjectNotHasAttribute cannot test magic attributes

halaei:
assertObjectNotHasAttribute() actually tests nothing. Go ahead and change note1 to note2 in tests and you see nothing fails.


divine:
Hello!

Issue #1321, opened 2017-10-06
Using outside laravel with packages 5.5


lpj145:
When i used this package outside laravel, i received this error:
`Call to a member function all() on array`

On this line:
![image](https://user-images.githubusercontent.com/11980109/31279579-e8eeb2ee-aa7f-11e7-9c6d-4a42760901a5.png)

Capsule instance code:
```
$config = $container->get('config')['datasource']['mongodb'] ?? false;

        if (!$config) {
            throw new \InvalidArgumentException('Check \'datasource\' key, have \'mongodb\' config ?');
        }

        $capsule = new Capsule();
        $capsule->addConnection($config);
        $capsule->getDatabaseManager()->extend('mongodb', function($config){
            return new Connection($config);
        });
        $capsule->bootEloquent();
        return $capsule;
```
My code:
```
return new JsonResponse(AuthRepository::all())
```

lpj145:
lol #978

lpj145:
#1322

Issue #978, opened 2016-09-28
Call to a member function all() on array


gp187:
Model:

``` php
use Jenssegers\Mongodb\Eloquent\Model as Moloquent;
class Product extends Moloquent
{
    protected $connection="mongodb";
    protected $dates = [
        'created_at', 'last_modified'
    ];
    protected $collection="products";
}
```

Controller:

``` php
class ProductCTRL extends Controller
{
    public function index(Request $request, $id)
    {
        $products = Product::where('cid', '=', $id)->get();
        return parent::formatResponse($products, true, 200);
    }
}
```

After running a `composer update`, gives out this error:

```
FatalThrowableError in Builder.php line 613:
Call to a member function all() on array
1. in Builder.php line 613
2. at Builder->getModels(array('*')) in Builder.php line 318
3. at Builder->get() in /srv/API/app/Http/Controllers/ProductCTRL.php line 44


```

_line `44` is where the `->get()` is_

If I do it like this, it works:

``` php
class ProductCTRL extends Controller
{
    public function index(Request $request, $id)
    {
        $db = DB::connection('mongodb')->collection("products");
        $products = $db->where('cid', '=', $id)->get();

        return parent::formatResponse($products, true, 200);
    }
}
```
- I have the latest version
- I reinstalled the plugin
- Still nothing

pi0:
Would you provide your `composer.json` ?

gp187:
Here it is
```json
{
    "name": "laravel/laravel",
    "description": "The Laravel Framework.",
    "keywords": ["framework", "laravel"],
    "license": "MIT",
    "type": "project",
    "require": {
        "php": ">=5.6.4",
        "laravel/framework": "5.3.*",
        "jenssegers/mongodb": "^3.0",
        "guzzlehttp/guzzle": "^6.2",
        "predis/predis": "^1.1",
        "symfony/finder": "^3.1",
        "symfony/event-dispatcher": "2.8.9",
        "rocket-code/shopify": "^2.0"
    },
    "require-dev": {
        "fzaninotto/faker": "~1.4",
        "mockery/mockery": "0.9.*",
        "phpunit/phpunit": "~5.0",
        "symfony/css-selector": "3.1.*",
        "symfony/dom-crawler": "3.1.*",
        "laravel/homestead": "^3.0",
        "squizlabs/php_codesniffer": "^2.6"
    },
    "autoload": {
        "classmap": [
            "database"
        ],
        "psr-4": {
[...]

PR #1473, opened 2018-03-22
Fixed error for usage outside Laravel.

dinosf:
An error `Call to a member function all() on array` will encouter when calling `Model::all()` outside Laravel. #978 #1321 

I found that the problem is because the function `shouldUseCollections()` in `Jenssegers\Mongodb\Query\Builder` only checks the version of Laravel, and does not handle the usage outside Laravel.

I added a newline so that the variable `useCollections`is now `true` by default.

### DIFF
--- a/src/Jenssegers/Mongodb/Query/Builder.php
+++ b/src/Jenssegers/Mongodb/Query/Builder.php
@@ -123,7 +123,7 @@ class Builder extends BaseBuilder
     ];
 
     /**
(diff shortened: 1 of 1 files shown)


taha:
Hi, is this fix going to be released soon? :)

xkovacikm:
Yeah, that would be awesome. I can't use this pacakge outside of Laravel now. (I actually can, but with very ugly workaround only)

PR #1477, opened 2018-03-27
bug fix: formatting exception so it can be saved to the database.

ruslanstarikov:
I have encountered the problem with saving failed jobs to mongo, - the exception field is always blank. 
The problem was reproduced using amqp and mysql as queue drivers, and mongo db version 3.6.2 on ubuntu 14.04. 
I have formatted the exception field into the array with the message and stack trace as keys. 
This way exception can be saved to the mongo db database.


ruslanstarikov:
I have messed up the formatting with my tabbage. I'll close and re-create the pull request. Thank you.

PR #1480, opened 2018-03-29
bug fix: formatting exception so it can be saved to the database.

ruslanstarikov:
I have encountered the problem with saving failed jobs to mongo, - the exception field is always blank.
The problem was reproduced using amqp and mysql as queue drivers, and mongo db version 3.6.2 on ubuntu 14.04.
I have formatted the exception field into the array with the message and stack trace as keys.
This way exception can be saved to the mongo db database.


divine:
Hello!

ruslanstarikov:
Done. Please review.

Smolevich:
@ruslanstarikov 
Can you explain profit of this changes?
If you add tests, it will be very good

ruslanstarikov:
I'll add some tests this weekend. I did it almost two years ago. 
The purpose of this patch is to make sure that the exception is persisted into the failed_jobs record, same way it does into the SQL databases. Without the patch, the exception field will always be blank.

divine:
@ruslanstarikov any update?

Thank you!

divine:
Closing in favor of https://github.com/jenssegers/laravel-mongodb/pull/1961

PR #1491, opened 2018-04-19
Gets database name from DSN string

hlorofos:
This contains PR https://github.com/jenssegers/laravel-mongodb/pull/1457
and in addition I've added an ability to use database name from DSN string, so it a developer passed DSN string its not necessary to pass database name.


divine:
Hello!

PR #1509, opened 2018-05-04
When comparing new value to original, take type into account

katiecrane:
This function is called from `getDirty()` to determine if a field on a model has been updated. When comparing dates, the types still need to be checked. You may be trying to replace the string “1994-04-01 00:00:00” with a date value of 1994-04-01 00:00:00, and for these two values, comparison with `==` is true while `===` is false. If we use `==`, the value in the database will remain a string instead of updating to a date as we are trying to do.


halaei:
Hi. In case you are still interested, please check #1664 which uses fromPHP() function for getDirty().

divine:
Should be closed in favor of PR https://github.com/jenssegers/laravel-mongodb/pull/1664

cc @Smolevich

PR #1535, opened 2018-06-08
Fixed queue attempts bug

gatisdukurs:
- Fixed queue attempts bug, now attempts are increased only when job gets pulled from the queue.
- Removed unnecessary reserved flag for job since reserved_at already does the job.
- Fixed Docker file error (travis builds where failing).


gatisdukurs:
@jenssegers please, take a look at my pull request works with laravel 5.6 and supervisor in parallel. Currently queue is broken in master and Docker file is broken for builds. 

Thanks

sbourdette:
up

divine:
Looks like it was fixed and merged in https://github.com/jenssegers/laravel-mongodb/pull/1855

Should be closed then.

cc @Smolevich

PR #1543, opened 2018-06-17
fix chunkById for types other than ObjectId and laravel >= 5.6.25

### LINKED ISSUE #1538: chunkById() does not work when _id is integer
In order to reproduce, make a collection with documents having integer ids. Then run the following:

```php
\DB::connection('mongodb')->enableQueryLog();
User::query()->chunkById(100, function ($collection) {
    dd($collection->count());
});
dd(\DB::connection('mongodb')->getQueryLog());
```

The expected behavior would be dumping the count of first chunk.
What actually happens is this query getting dumped:

```
users.find({"_id":{"$gt":"000000000000000000000000"}},{"sort":{"_id":1},"limit":100,"typeMap":{"root":"array","document":"array"}})
```
Apparently, no integer is less than '000000000000000000000000' in MongoDb.

A Related PR: https://github.com/jenssegers/laravel-mongodb/pull/1317
ping @DFurnes 
Any idea on how to fix this?

halaei:
Fixes #1538 
This PR should be merged after laravel 5.7 release.
The removed lines used to fix an issue that is currently being handled in a better way inside laravel/framework. The lines can actually break the code if _id is not of type ObjectId.

### DIFF
--- a/src/Jenssegers/Mongodb/Query/Builder.php
+++ b/src/Jenssegers/Mongodb/Query/Builder.php
@@ -631,12 +631,6 @@ class Builder extends BaseBuilder
      */
     public function forPageAfterId($perPage = 15, $lastId = 0, $column = '_id')
     {
-        // When using ObjectIDs to paginate, we need to use a hex string as the
-        // "minimum" ID rather than the integer zero so the '$lt' query works.
-        if ($column === '_id' && $lastId === 0) {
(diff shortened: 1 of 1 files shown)


halaei:
@jenssegers Can you merge this. These lines are now always redundant and sometimes just buggy.

jenssegers:
Merged 👌

PR #1545, opened 2018-06-17
Fix inserted _id on association of child to parent in EmbedsMany

### LINKED ISSUE #1310: Creating a model without a _id?
Is it possible to create a model without a _id automatically created?

I'm looking to create an embedded model in another, and I don't want the subdocuments to have a _id key, as they already have other primary keys. As we unfortunately can't cast properties in embedded documents (i have a date in the subdocument), I want to use the embeds many functionality, but don't want to end up with a bunch of useless ids attached.

I looked at primaryKey, but that doesn't seem to do anything.

timsweb:
When an object is associated to parent in `associateNew`, a Mongo ID is created and added to the model regardless of the key set on the model. This commit changes the behaviour to only add the _id in instances where the primary key is set to be `_id`. This _may_ be a breaking change if code is depending on this behaviour; but the existing behaviour is undocumented.

Fixes #1310

### DIFF
--- a/src/Jenssegers/Mongodb/Relations/EmbedsMany.php
+++ b/src/Jenssegers/Mongodb/Relations/EmbedsMany.php
@@ -235,8 +235,8 @@ class EmbedsMany extends EmbedsOneOrMany
      */
     protected function associateNew($model)
     {
-        // Create a new key if needed.
-        if (!$model->getAttribute('_id')) {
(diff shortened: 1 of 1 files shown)


divine:
Should be closed, fixed and merged in https://github.com/jenssegers/laravel-mongodb/pull/1740

cc @Smolevich 

PR #1553, opened 2018-06-26
Changes in travis build

Smolevich:
On my opinion we don't need execute sed command in `.travis.yml` because Docker engine has arguments in Dockerfile and we can use it 
- Add arg PHP_VERSION in `Dockerfile`
- Set arg PHP_VERSION in `.travis.yml` from env variable `TRAVIS_PHP_VERSION`
- Package `zlib1g-dev` replaced from `pecl install` to `apt-get install`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,6 +10,8 @@ services:
         volumes:
             - .:/code
         working_dir: /code
+        environment:
+          PHP_VERSION: ${TRAVIS_PHP_VERSION}
         command: bash -c "composer install --prefer-source --no-interaction && php ./vendor/bin/phpunit"
         depends_on:
           - mysql
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,7 +11,7 @@ services:
         working_dir: /code
         environment:
           PHP_VERSION: ${PHP_VERSION}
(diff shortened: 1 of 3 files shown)


Smolevich:
@jenssegers, can you review,  please?

docker-compose.yml
```diff
@@ -10,6 +9,8 @@ services:
         volumes:
             - .:/code
         working_dir: /code
+        environment:
+          PHP_VERSION: ${PHP_VERSION}
```
jenssegers: This can be a fixed value I guess, or at least provide a default value. Could you change this?

Great job on these changes btw!

Smolevich: Yes, i can set default value, but env variable `PHP_VERSION` set in travis script build

Smolevich: Also in Docker file i popose install of composer to replace on this construction
`
FROM composer as composer-build #official docker image with composer (https://github.com/composer/docker/blob/master/1.6/Dockerfile)
COPY --from=composer-build /usr/bin/composer /usr/local/bin/composer
`

Smolevich:
@jenssegers, can you review changes?

jenssegers:
Great work!

PR #1611, opened 2018-09-21
Support for _id of binary type (to use UUIDs as _id)

mfauveau:
While using the ObjectID type is the default implementation in MongoDB, it's technically possible to use anything for the primary key as long as you use a unique value. Some implementations (including mine) use a UUID as the primary key. UUIDs are stored in MongoDB as Binary types. The provider in its current state doesn't work in that situation.

These PR are the changes I've made to make it work. The only thing not included is a trait that automatically generates the UUID on the "creating", for example:
```
<?php

namespace App;

use MongoDB\BSON\Binary;
use Ramsey\Uuid\Uuid;

trait MongodbBinaryUuid
{
    /**
     * Boot the trait.
     *
     * @return void
     */
    protected static function bootMongodbBinaryUuid()
    {
        static::creating(function (Model $model) {
            if ($model->{$model->getKeyName()}) {
                return;
            }

            $model->{$model->getKeyName()} = new Binary(static::encodeUuid(Uuid::uuid1()), Binary::TYPE_UUID);
        });
    }
}
```
Note that my actual trait is actually more complete, with scopes and such.

I'm not sure this is the best way but it works for my use case and I'm curious to hear feedback. Surely I can be the only one with a need to use UUIDs as _id right? ;)

### DIFF
--- a/src/Jenssegers/Mongodb/Eloquent/Model.php
+++ b/src/Jenssegers/Mongodb/Eloquent/Model.php
@@ -9,6 +9,7 @@ use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use Jenssegers\Mongodb\Query\Builder as QueryBuilder;
+use MongoDB\BSON\Binary;
 use MongoDB\BSON\ObjectID;
 use MongoDB\BSON\UTCDateTime;
 use Illuminate\Contracts\Queue\QueueableEntity;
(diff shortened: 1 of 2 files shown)


mfauveau:
@jenssegers any chance of merging this?

halaei:
I am wondering how hard is it to pass an object of type Binary instead of string. I don't like the auto casting behavior even for ObjectId, now we have it for Binary and for properties that ends with `_id`.

I hope for a branch that stop doing so: https://github.com/jenssegers/laravel-mongodb/issues/1761

Smolevich:
@mfauveau, do you run tests on this PR locally? Build fails in travis.

Smolevich:
@jenssegers, this [commit](https://github.com/jenssegers/laravel-mongodb/commit/83d0b8b4a5159d22836c11d7ee52fea641679f83) has breaking changes, tests failed, because this [change](https://github.com/jenssegers/laravel-mongodb/pull/1611/files#diff-41ad12b428acec5b124beded4cc97ef2R171) don't casting ObjectId to string

PR #1638, opened 2018-10-29
add lists method to eloquent builder.

shawn-lin:
[Illuminate\Database\Eloquent\Builder](https://github.com/laravel/framework/blob/5.2/src/Illuminate/Database/Eloquent/Builder.php#L467) has lists method

so it's will never call [Jenssegers\Mongodb\Query\Builder](https://github.com/jenssegers/laravel-mongodb/blob/2.3/src/Jenssegers/Mongodb/Query/Builder.php#L622)'s lists method

### DIFF
--- a/src/Jenssegers/Mongodb/Eloquent/Builder.php
+++ b/src/Jenssegers/Mongodb/Eloquent/Builder.php
@@ -232,4 +232,22 @@ class Builder extends EloquentBuilder
 
         return $results;
     }
+
+    /**
+     * Override lists method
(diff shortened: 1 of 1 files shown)


divine:
Should be closed since it's deprecated and removed.

> The lists method on the Collection, query builder and Eloquent query builder objects has been renamed to pluck. The method signature remains the same.

See https://laravel.com/docs/5.2/upgrade#upgrade-5.2.0 

cc @Smolevich

PR #1664, opened 2018-11-29
use mongodb driver to check for dirtiness

halaei:
To check if a field is dirty we can simply use `fromPHP()` function of the mongodb driver. Besides, we shouldn't use the SQL-specific logic of `Illuminate\Eloquent\Model::originalIsEquivalent()` for mongodb. In sql, fields keep their schema data types and changing integer 1 to string '1' actually does nothing. However, in mongodb fields accept their types as they are inserted/updated. Therefore, if we change value of a field from `(int)1` to `(string)'1'` we actually must regard it as dirty.

**This change is backward incompatible**. Don't know if anyone likes the current behaviour more.


divine:
Hello!

divine:
Hi there,

This PR contains a breaking change and will be released within next version.

Closing for now.

Thanks.

halaei:
@divine I am glad this project became active. I'll send a PR within a day.

Issue #397, opened 2015-01-20
EmbedsOne doesn't allow eager Loading


juaniiton1:
I have two models, one embed in the other likes this:

```
class Media extends Moloquent {

    public function user()
    {
        return $this->belongsTo('User');
    }

    public function comments()
    {
        return $this->embedsMany('Comment');
    }

    public function video()
    {
        return $this->embedsOne('Video');
    }

}
```

```
class Video extends Moloquent {

    ...

}
```

And I want to get a Media with the Video embedded with some getters mutators.

```
$media = Media::with(['comments', 'user', 'video'])->find($id);
```

The load of `comments` and `user` is working great, but when I add the `video` loading it crashes showing the next error:

```
[2015-01-20 09:50:42] alpha.ERROR: exception 'ErrorException' with message 'Argument 2 passed to Jenssegers\Mongodb\Relations\EmbedsOneOrMany::match() must be an instance of Illuminate\Database\Eloquent\Collection, null given, called in /home/vagrant/sites/media/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php on line 478 and defined' in /home/vagrant/sites/media/vendor/jenssegers/mongodb/src/Jenssegers/Mongodb/Relations/EmbedsOneOrMany.php:112
Stack trace:
#0 /home/vagrant/sites/media/vendor/jenssegers/mongodb/src/Jenssegers/Mongodb/Relations/EmbedsOneOrMany.php(112): Illuminate\Exception\Handler->handleError(4096, 'Argument 2 pass...', '/home/vagrant/s...', 112, Array)
#1 /home/vagrant/sites/media/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php(478): Jenssegers\Mongodb\Relations\EmbedsOneOrMany->match(Array, NULL, 'video')
#2 /home/vagrant/sites/media/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php(445): Illuminate\Database\Eloquent\Builder->loadRelation(Array, 'video', Object(Closure))
#3 /home/vagrant/sites/media/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php(158): Illuminate\Database\Eloquent\Builder->eagerLoadRelations(Array)
#4 /home/vagrant/sites/media/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php(125): Illuminate\Database\Eloquent\Builder->get(Array)
[...]

PR #1711, opened 2019-02-25
Fix embedded relationship on broadcast serialization

gazben:


gazben:
Related to #397.
If you want to serialize an embedded collection the second parameter will be null, not an empty collection. We need to allow this.

gazben:
Don't merge. The embeds serialization is broken. Help needed

spaceemotion:
@gazben 	We just had this today as well, currently thinking about opening a PR, but the fix is a bit more involved.

The `Collection` parameter is required as it's typehinted by the laravel framework itself. Nothing we can do here.

The real culprit is the `get()` Method on the Relation. The SQL implementations of those always return a collection. This package simply calls the `getResult()` Method which is incorrect for this use case: For 1-1 relationships this returns a single object instead of a collection as all the others do (even the native BelongsTo does this).

It would be more accurate to create a new collection out of it (using the `newCollection` method on the query builder).

Will try to post an initial PR for this tomorrow.

spaceemotion:
Small update: We're having serious issues using this library, as even updates no longer work properly.

1. Create a new embedded instance
2. Set a property on the parent
3. dump() shows everything OK
4. call save()
5. Find the new properly on the embedded instance instead, leaving the original parent's empty.

Looks like this either needs more investigation in our temp. solution to this or the library in general.

divine:
Hello!

Also, @spaceemotion any update?

divine:
Closing due to inactivity, free free to create new PR or provide an update here.

spaceemotion:
@divine sorry for not replying earlier. We ended up switching to SQL again, as we had other problems with a NoSQL based architecture, so I never actually ended up working further on that prototype (we did get more issues further down the line with it though, so it was kinda unstable....). Sorry for getting any hopes up!

PR #1732, opened 2019-03-17
Update travis.yml

Smolevich:
- Update travis ci


Smolevich:
@jenssegers, i move changes from this [PR](https://github.com/jenssegers/laravel-mongodb/pull/1729)

jenssegers:
I pushed a new travis config the master today. I would rather stick with the docker setup because then I can better predict what the travis result will be when I run tests locally in docker.

PR #1738, opened 2019-03-21
use env values from phpunit.xml, update database.php and queue.php

Smolevich:
- Add values for env variables to phpunit.xml
- Use env values instead hardcode values

It is can usefull if one of variables (`MONGO_HOST` for example) is different from default hardcode value

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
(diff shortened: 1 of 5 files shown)


Smolevich:
@jenssegers, can you review please?

jenssegers:
Looks good!

PR #1799, opened 2019-08-01
Get base query before update so that scopes are applied

### LINKED ISSUE #1798: Global scopes are ignored on Query::update
Given the following model:

```php
<?php

use Jenssegers\Mongodb\Eloquent\Builder;

class Scoped extends Eloquent
{
    protected $collection = 'scoped';
    protected $fillable = ['name', 'favorite'];

    protected static function boot()
    {
        parent::boot();

        static::addGlobalScope('favorite', function (Builder $builder) {
            $builder->where('favorite', true);
        });
    }
}
```

If I run the following:

```php
Scoped::query()->update(['name' => 'Johnny']);
```

It should only update models where `favorite` is `true`, because of the `favorite` scope. However, it updates all models.

goodevilgenius:
If we do `toBase()` on update (as `Illuminate\Database\Eloquent\Builder` does), then it applies global scopes to the query so that it updates only those models that it should.

Fixes #1798

### DIFF
--- a/src/Jenssegers/Mongodb/Eloquent/Builder.php
+++ b/src/Jenssegers/Mongodb/Eloquent/Builder.php
@@ -44,7 +44,7 @@ class Builder extends EloquentBuilder
             return 1;
         }
 
-        return $this->query->update($this->addUpdatedAtColumn($values), $options);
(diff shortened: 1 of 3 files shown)


jenssegers:
Looks good! Thanks!

PR #1804, opened 2019-08-10
change position project in pipeline

kratosvn:


kratosvn:
I got two problems with your vendor:
+ It doesn't work with projection because core builder don't remove it when calculating paginator. You can see here: `laravel/framework/src/Illuminate/Database/Query/Builder.php` line 2203.
+ Query so slow I found the solution here: https://jira.mongodb.org/browse/SERVER-7568. We should push "project" before "sort" to improve query performance.

Smolevich:
Let's go slack. Link [here](https://github.com/jenssegers/laravel-mongodb/issues/1791).

src/Jenssegers/Mongodb/Query/Builder.php
```diff
@@ -325,9 +327,7 @@ public function getFresh($columns = [])
             if ($this->limit) {
                 $pipeline[] = ['$limit' => $this->limit];
             }
-            if ($this->projections) {
-                $pipeline[] = ['$project' => $this->projections];
-            }
+
```
divine: Hello,

Could you please remove this line so styleci check passes?

Also, please kindly update with the base branch again, so many updates coming to this library finally!

divine:
This needs to be closed. cc @Smolevich @jenssegers 

1. It has been fixed upstream 
> This issue is fixed as of commit [8d048a3bb2](https://github.com/mongodb/mongo/commit/8d048a3bb2f0f2f81cf99ce76ff21112bf3963d6). Users should no longer need to work around the problem with hints or index filters, or by rewriting queries. The fix will first become available in development release 4.3.1.

2. There is no guarantee that will work correctly. 
> Can you confirm by implementing the $project workaround we will lose the benefits of indexed sorts?
As per the docs: 
$sort operator can take advantage of an index when placed at the beginning of the pipeline or placed before the $project, $unwind, and $group aggregation operators. If $project, $unwind, or $group occur prior to the $sort operation, $sort cannot use any indexes.
https://docs.mongodb.com/manual/reference/operator/aggregation/sort/#sort-operator-and-performance

Source of the comment [here](https://jira.mongodb.org/browse/SERVER-7568?focusedCommentId=2538093&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2538093)

PR #1806, opened 2019-08-13
fix for non string id

halaei:
#1611 breaks the following query:
```php
$query->where('_id', 'mod', [2, 1]);
```
With the exception:
```strlen() expects parameter 1 to be string, array given```.

To fix it we have to check if $id is string before call to strlen().


odycceus:
Please merge this PR

lapiy:
I went here wanting to suggest exactly the same PR
Looking forward this to be merged ASAP...

Issue #1821, opened 2019-09-03
Fails to install on Laravel 6.0


ReeceM:
Hi @jenssegers, 
# Question

Is this going to be updated or are there problems if this package were to be updated to require the latest versions? i.e. breaking changes?

I have just tried to run a `composer update` against my application in order to upgrade to Laravel 6.0. 

But I get hit with the following error, which seems to be caused by the requirements of the `composer.json` for the Illuminate packages

https://github.com/jenssegers/laravel-mongodb/blob/01ac069cf88a2d1b2749cf3a2df45496b753d11b/composer.json#L13-L19

Errror from `composer update || composer require jenssegers/mongodb`

```
- Conclusion: don't install jenssegers/mongodb v3.5.3
    - Conclusion: don't install jenssegers/mongodb v3.5.2
    - Conclusion: remove jenssegers/mongodb v3.5.1
    - Conclusion: don't install jenssegers/mongodb v3.5.1
    - Conclusion: don't install jenssegers/mongodb v3.5.0
    - Conclusion: don't install laravel/framework v6.0.0
    - Installation request for jenssegers/mongodb ^3.5 -> satisfiable by jenssegers/mongodb[v3.5.0, v3.5.1, v3.5.2, v3.5.3].
    - Conclusion: don't install laravel/framework 6.x-dev
    - jenssegers/mongodb v3.5.1 requires illuminate/support ^5.8 -> satisfiable by laravel/framework[v5.8.34, 5.8.x-dev], illuminate/support[5.8.x-dev, v5.8.0, v5.8.11, v5.8.12, v5.8.14, v5.8.15, v5.8.17, v5.8.18, v5.8.19, v5.8.2, v5.8.20, v5.8.22, v5.8.24, v5.8.27, v5.8.28, v5.8.29, v5.8.3, v5.8.30, v5.8.31, v5.8.32, v5.8.33, v5.8.34, v5.8.35, v5.8.4, v5.8.8, v5.8.9].
```

I gather this is because the package is not pulling the related `illuminate/* ^6.0` for its requirements.

Smolevich:
[...]

PR #1824, opened 2019-09-04
Feat/support laravel 6

ReeceM:
This PR would allow a version of laravel-mongodb support installs on laravel 6.0.

It could be a different release tag or branch as it seems to have been done before.

Or possibly left until laravel 6.0 is more stable and the majority of user base has adopted and migrated to an updated codebase?

for #1821

(apologies for the first brief pr message..., I pressed enter to return the line.)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,15 +11,15 @@
     ],
     "license" : "MIT",
     "require": {
-        "illuminate/support": "^5.8",
-        "illuminate/container": "^5.8",
-        "illuminate/database": "^5.8",
-        "illuminate/events": "^5.8",
+        "illuminate/support": "^6.0",
+        "illuminate/container": "^6.0",
+        "illuminate/database": "^6.0",
+        "illuminate/events": "^6.0",
         "mongodb/mongodb": "^1.4"
     },
     "require-dev": {
-        "phpunit/phpunit": "^6.0|^7.0",
-        "orchestra/testbench": "^3.1",
+        "phpunit/phpunit": "^8.0",
+        "orchestra/testbench": "^4.0",
         "mockery/mockery": "^1.0",
         "satooshi/php-coveralls": "^2.0",
         "doctrine/dbal": "^2.5"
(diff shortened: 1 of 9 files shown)


composer.json
```diff
@@ -11,15 +11,15 @@
     ],
     "license" : "MIT",
     "require": {
-        "illuminate/support": "^5.8",
-        "illuminate/container": "^5.8",
-        "illuminate/database": "^5.8",
-        "illuminate/events": "^5.8",
+        "illuminate/support": "^6.0",
```
jenssegers: Are there any super breaking changes that would make the package only support the latest version? Can we keep compatibility with 5.8?

composer.json
```diff
@@ -11,15 +11,15 @@
...
+        "illuminate/container": "^6.0",
+        "illuminate/database": "^6.0",
+        "illuminate/events": "^6.0",
         "mongodb/mongodb": "^1.4"
     },
     "require-dev": {
-        "phpunit/phpunit": "^6.0|^7.0",
-        "orchestra/testbench": "^3.1",
+        "phpunit/phpunit": "^8.0",
```
jenssegers: This will probably break travis tests on older PHP versions.

composer.json
ReeceM: Could this fix be put with a new 3.x release, as it seems if it requires the the ^5.8 branches it seems the composer install fails as it doesn't have the same versions?

jenssegers: Why would it fail?

jenssegers: Try `"^5.8|^6.0"`

composer.json
ReeceM: Okay, I see that it has done that already on this PR...

What would be the less breaking way of doing that?

jenssegers: `"^6.0|^7.0|^8.0",` should work I think.

Smolevich:
@jenssegers, @ReeceM, maybe delete changes  for tests in this PR and create in another PR?

Smolevich:
https://github.com/jenssegers/laravel-mongodb/pull/1823 this PR with changes only composer.json

jenssegers:
Supported added in 3.6.0

PR #1838, opened 2019-10-02
Fix job queue attempts

djaross97:
At the moment, if I have more than 3 job queues in the jobs collection and the $tries property is set to 3 for example, and run the queue worker it will fail on the 4th job. The reason is, when I run the worker it will look for all reserved jobs in the queue however the lookup query is wrong as it's trying to find the ones where reserver_at is null and that will return all 4 jobs not just the one that's currently being processed. So that will then update all 4 job's attempts field by adding 1, that means when the third job runs, the 4th will already have attempts = 3 so it will fail when the worker tries to process that row as it reached our $tries property value.

To fix this I have changed the whereNull to whereNotNull, so it will only return and update the attempts value to the job that is currently being processed.


composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "jenssegers/mongodb",
+    "name": "ridown/mongodb",
```
divine: Any explanation why are you trying to change this?

divine:
Closing due to inactivity.

Issue #1373, opened 2017-11-23
"php artisan queue:retry" command is not working with MongodbQueueServiceProvider


hamisor:
Hi,

It appears that retrying failed jobs, using mongodb, does not work. I configured my environment as described in (https://github.com/jenssegers/laravel-mongodb#queues), project versions below:
```
"laravel/framework":     v5.4.36
"jenssegers/mongodb":    v3.2.3
"fhteam/laravel-amqp":   dev-master
mongodb v3.4
```

The error I receive is:

```
[ErrorException]                      
  Trying to get property of non-object 
```

The Laravel `RetryCommand->retryJob($job)` function is expecting the `$job` param to be a **`stdClass`** (instead of an **`array`**). 

When using mongo, under the hood the `MongoFailedJobProvider` class (from `MongodbQueueServiceProvider`) calls the `find()` function to retrieve a single failed job, which returns an **`array`** (instead of an **`stdClass`**).

This type mismatch appears to be the problem.

A similar problem exists at (https://github.com/jenssegers/laravel-mongodb/issues/1059) - namely,  the `laravel-mongodb` query builder returns an **`array`** rather than an **`stdClass`** when performing  db operations such as find(), get() first(), etc.

A quick fix I would suggest would be to make `MongodbQueueServiceProvider` work with the Laravel queue command, by casting inside the `find()` function in the `MongoFailedJobProvider` class, like so:

```
   /**
     * Get a single failed job.
     *
     * @param  mixed $id
     * @return stdClass
     */
    public function find($id)
    {
        $job = $this->getTable()->find($id);

        $job['id'] = (string) $job['_id'];

        // CAST HERE
        return (object)$job;
    }
```

@jenssegers, I would appreciate any suggestions on the proposed fix, I am happy to make a pull request for it.

Thanks

PR #1851, opened 2019-10-17
fix queued mongodb usage check

denissonleal:
for cases that use more than one connection with mongodb


denissonleal:
solves problem reported in #1373

geidsonc:
Great job!

jenssegers:
Thanks!

PR #1854, opened 2019-10-17
Added aggregate-sample function to generate random result from query

adlanarifzr:
The ```sample()``` function is used to generate random result based on the given size. It can be used together with eloquent functions like ```where()`` and such.

### DIFF
--- a/src/Jenssegers/Mongodb/Query/Builder.php
+++ b/src/Jenssegers/Mongodb/Query/Builder.php
@@ -210,7 +210,9 @@ class Builder extends BaseBuilder
     }
 
     /**
-     * @inheritdoc
+     * Get a randomize result
+     * @param int $size
+     * @return $this
      */
     public function sample($size = 1)
     {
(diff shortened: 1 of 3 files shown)


jenssegers:
Is this a new Laravel method?

Smolevich:
Laravel doesn't have this function in query builder

src/Jenssegers/Mongodb/Query/Builder.php
```diff
@@ -209,6 +209,15 @@ public function get($columns = [])
         return $this->getFresh($columns);
     }
 
+    /**
+     * @inheritdoc
```
jenssegers: @inheritdoc should be replaced with it’s own docblock since it’s not part of Laravel.

usamanoman:
Very interested in this PR. Let's hope it's available soon.

src/Jenssegers/Mongodb/Query/Builder.php
adlanarifzr: @jenssegers I updated. Kindly recheck.

divine:
Hello!

divine:
Closed in favor of https://github.com/jenssegers/laravel-mongodb/pull/1967

PR #1855, opened 2019-10-17
hotfix/increment-attempts

denissonleal:
- increment attempts only when taken from list.

solves problems of "has been attempted too many times or run too long. The job may have previously timed out." even when there are no job errors.

many issues report the problem.


jenssegers:
Thanks!

PR #1870, opened 2019-11-09
Use Carbon::now() for fresh timestamps so that Carbon's time mocking works

CurosMJ:


passionInfinite:
@CurosMJ This will break the default behaviour of `Carbon::now` . As it will return you the string not the instance which breaks the further usage of `toDateTimeString()`.

src/Jenssegers/Mongodb/Eloquent/Model.php
```diff
@@ -118,7 +118,7 @@ public function getDateFormat()
      */
     public function freshTimestamp()
     {
-        return new UTCDateTime(time() * 1000);
+        return new UTCDateTime(Carbon::now());
```
jenssegers: Shouldn't this be:

```suggestion
        return new UTCDateTime(Carbon::now()->getTimestamp() * 1000);
```

jenssegers: It looks like `UTCDateTime` also supports `DateTime` interfaces 👌

PR #1872, opened 2019-11-13
add pipeline method for raw mongodb queries.

erblinbobaj:
Add pipeline method for raw mongodb queries.


rennokki:
This repo seems like it gets slow updates. Would you mind switching for https://github.com/rennokki/laravel-mongodb and submitting this PR down there?

jenssegers:
Would it be possible to add a test and possibly an example in the documentation?

rennokki:
I think it's the same like the `whereRaw` method.

divine:
Hello,

Sorry for late reply, this will be in upcoming version 4.

Thank you.

PR #1886, opened 2019-12-13
Update README.md

ryan7n:
Add other regex demo code 

Because this code is not working in MongoDB extension version => 1.5.5
User::where('name', 'regexp', '/.*doe/i')->get();


Smolevich:
@ryan7n https://github.com/jenssegers/laravel-mongodb/pull/1827 i try to research README on old or deprecated things, but @jenssegers don't finish review

jenssegers:
If the first example isn't supported anymore I would update it to what's currently working. Otherwise more people might run into the same issue :)

PR #1915, opened 2020-01-20
[feature] StyleCI

rennokki:
StyleCI is also used by Laravel official packages, so i think it's a good idea to stick with the standards in the same way.


Smolevich:
@rennokki please add description

rennokki:
> @rennokki please add description

Fixed.

PR #1916, opened 2020-01-20
[feature] Added ->toSql() for raw wheres

rennokki:
See this open PR i closed a while: https://github.com/jenssegers/laravel-mongodb/pull/1887

It's good to make it work not only because of debugging (either it's not called SQL), but also you can cache query-by-query easier using packages like this: https://github.com/rennokki/laravel-eloquent-query-cache

~Needs testing.~

Actually it is safe, it detects arrays and automatically serializes them to string, which is exactly what's needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ Table of contents
 * [Eloquent](#eloquent)
 * [Optional: Alias](#optional-alias)
 * [Query Builder](#query-builder)
+* [`toSql()`](#to-sql)
 * [Schema](#schema)
 * [Extensions](#extensions)
 * [Examples](#examples)
@@ -239,6 +240,20 @@ $user = DB::connection('mongodb')->collection('users')->get();
 ```
 
 Read more about the query builder on http://laravel.com/docs/queries
+
+->toSql()
+-------------
+Hence we are in a NoSQL driver, it is counter-intuitive to use `toSql()` to output the string that will be executed.
+
+However, if you still need the string that looks like the sql for some inspection, you can use the `toSql()` method:
+```php
+User::where('some_field', '>', 300)
+    ->whereRaw(['age' => ['$gt' => 30, '$lt' => 40]])
+    ->toSql();
+
+// output
+"select * from "users" where "some_field" > ? and {"age":{"$gt":30,"$lt":40}}"
+```
 
 Schema
 ------
--- a/README.md
+++ b/README.md
@@ -896,6 +896,7 @@ Query Builder
 -------------
 
 ### Basic Usage
+
 The database driver plugs right into the original query builder.
 
 When using MongoDB connections, you will be able to build fluent queries to perform database operations.
--- a/README.md
+++ b/README.md
@@ -7,33 +7,48 @@ Laravel MongoDB
 [![codecov](https://codecov.io/gh/jenssegers/laravel-mongodb/branch/master/graph/badge.svg)](https://codecov.io/gh/jenssegers/laravel-mongodb/branch/master)
 [![Donate](https://img.shields.io/badge/donate-paypal-blue.svg)](https://www.paypal.me/jenssegers)
 
-An Eloquent model and Query builder with support for MongoDB, using the original Laravel API. *This library extends the original Laravel classes, so it uses exactly the same methods.*
+This package adds functionalities to the Eloquent model and Query builder for MongoDB, using the original Laravel API. *This library extends the original Laravel classes, so it uses exactly the same methods.*
 
-Table of contents
------------------
-* [Installation](#installation)
-* [Upgrading](#upgrading)
-* [Configuration](#configuration)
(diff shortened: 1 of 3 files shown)


rennokki:
Tests are failing due to Coveralls. This might be needed to PR before re-checking the tests: https://github.com/jenssegers/laravel-mongodb/pull/1922

tests/ModelTest.php
```diff
@@ -572,4 +572,15 @@ public function testChunkById(): void
...
     }
+
+    public function toSqlTest(): void
+    {
+        $this->assertTrue(
+            is_string(User::where('age', 30)->toSql())
+        );
+
+        $this->assertTrue(
+            is_string(User::whereRaw(['age' => ['$gt' => 30, '$lt' => 40]])->toSql())
```
Smolevich: Add additional test case with assert that result equal to generated string (for example `select * from user where age ...`)

rennokki: Fixed in https://github.com/jenssegers/laravel-mongodb/pull/1916/commits/1ad945e303fb9a527db5eb1d7aa4a0606473c3f2

README.md
```diff
@@ -17,6 +17,7 @@ Table of contents
 * [Eloquent](#eloquent)
 * [Optional: Alias](#optional-alias)
 * [Query Builder](#query-builder)
+* [`toSql()`](#to-sql)
```
jenssegers: Maybe we can put this under a category called "debugging" or "inspecting queries".

Smolevich: Add else info about DB::getQueryLog()

Smolevich: It old feature but getQueryLog() return original query struncture to mongodb

rennokki: @jenssegers maybe fix this after #1917

rennokki: @Smolevich i will look into it as soon as i get home

rennokki: Found a solution but some of the code from `getFresh()` needs refactoring.

rennokki: I suggest leaving this PR as-is, in the SQL form, because the `getFresh()` needs a lot of refactoring.

Giacomo92: What `getFresh()` is used for? @rennokki

rennokki: It is used to compile the wheres, options and get data for the `get()` method.

There are some edge cases, like distinct or aggregations, both checks are in `getFresh()`. I assume that the best way is to compile the query in the builder and pass it to the `Collection` class where the call is actually made to the `MongoCollection`'s methods that execute the needed query. 🤷‍♂

Smolevich: @rennokki what about to add info about using `DB::getQueryLog()`
For example 

```
DB::enableQueryLog();
$user = factory(User::class)->create();
$query = DB::connection()->table(config('auth.passwords.users.table'));
dd(DB::getQueryLog());
```
Example of element of array that function `getQueryLog` returned 

```
1 => array:3 [
    "query" => "users.find({"email":"oziemann@example.org"},{"limit":1,"typeMap":{"root":"array","document":"array"}})"
    "bindings" => []
    "time" => 4.49
  ]
```

rennokki: The scope of this PR was to be able to get the query string before the actual query was executed, so i can be able to add cache query-by-query.

I can add the query log thing in the readme, because that's useful.

rennokki: So far, the single issue here is that the basic `get()` and the aggregate and groups are treated differently. This means that the final query is split into two methods rather than a single `get()` one.

I repeat - the purpose of this PR was to get the string for each query, so I can be able to query the cache query-by-query.

rennokki: @Smolevich @Giacomo92 

So far, the single issue here is that the basic `get()` and the aggregate and groups are treated differently. This means that the final query is split into two methods rather than a single `get()` one.

I repeat - the purpose of this PR was to get the string for each query, so I can be able to query the cache query-by-query.

Smolevich: @rennokki, do you see this method https://github.com/jenssegers/laravel-mongodb/blob/a8918acadb79f4f90d42eaddf0e996054ddd8184/src/Jenssegers/Mongodb/Query/Builder.php#L413?

src/Jenssegers/Mongodb/Query/Grammar.php
```diff
@@ -6,4 +6,19 @@
 
 class Grammar extends BaseGrammar
 {
+    /**
+     * {@inheritdoc}
```
Smolevich: Add param doc for argument `$query` please

README.md
rennokki: @Smolevich I think `$options` and `$pipeline` has to be attributes in the class and be passed to the `generateCacheKey` method.

Smolevich:
@rennokki  please resolve conflcts

divine:
@rennokki this has been open for almost a month, will you take care?

Looks like no as you've deleted forked repo while at the same trying to reinvent a wheel.

Good luck, closing.

Issue #1918, opened 2020-01-20
Mongo multi-version testing


rennokki:
Currently, the docker image used for test is latest, which is 4.2.2. I am pretty sure it's not the only version used by the users and services like AWS DocumentDb are locked to 3.6 for more than a year, for example.

What would be a good solution? LTS branches or matrix to test 3.6, 4.0 and 4.2?

Smolevich:
I think that we update matrix in ci

PR #1920, opened 2020-01-21
Add matrix for mongodb

Smolevich:
- Create matrix mongodb with versions 3.6, 4.0, 4.2

"Closes: #1918"

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -61,3 +61,4 @@ jobs:
       run: vendor/bin/php-coveralls -v
       env:
         COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


jenssegers:
That coveralls dependency is really being annoying. Might be better to just disable it for now.

Smolevich:
> That coveralls dependency is really being annoying. Might be better to just disable it for now.

We can use case as webonyx (https://github.com/webonyx/graphql-php/commit/599b03b59bf990e289337b1b79dc6aea31ea882e) or use https://github.com/cedx/coveralls.php before accepting this PR https://github.com/php-coveralls/php-coveralls/pull/275

Smolevich:
@jenssegers i add flag `continue-on-error: true` for step with send coveralls

jenssegers:
Awesome!

PR #1924, opened 2020-01-24
[hotfix] The Mongo version in CI for 4.0 uses actually the 4

rennokki:
It seems like the CI's Mongo version `4.0` is treated by Github Actions like `4` rather than actual `4.0`.

mongo

According to DockerHub, the `4` tag is associated to `4.2.2` and `4.2` rather than `4.0`:

tag

When the matrix elements are changed to string values, the right mongo version is used.

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -11,9 +11,9 @@ jobs:
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
-        php: [7.1, 7.2, 7.3, 7.4]
+        php: ['7.1', '7.2', '7.3', '7.4']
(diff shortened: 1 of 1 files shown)


Smolevich:
I try to understand that four checks in ci are `Waiting for status to be reported`
Снимок экрана 2020-01-24 в 10 41 50

Smolevich:
> I try to understand that four checks in ci are `Waiting for status to be reported`
> Снимок экрана 2020-01-24 в 10 41 50

@jenssegers  https://github.community/t5/GitHub-Actions/GitHub-Actions-Pull-requests-always-include-some-outdated-checks/td-p/30030

PR #1933, opened 2020-01-30
[Updated PR#1746] Add MustVerifyEmail

divine:
This PR was recreated from https://github.com/jenssegers/laravel-mongodb/pull/1746 with updated branch or resolved conflicting files.

--

Added support for Laravel authentication when users need to check their emails to connect

### DIFF
--- a/src/Jenssegers/Mongodb/Auth/User.php
+++ b/src/Jenssegers/Mongodb/Auth/User.php
@@ -2,8 +2,8 @@
 
 namespace Jenssegers\Mongodb\Auth;
 
-use Illuminate\Auth\MustVerifyEmail;
 use Illuminate\Auth\Authenticatable;
+use Illuminate\Auth\MustVerifyEmail;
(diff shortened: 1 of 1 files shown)


src/Jenssegers/Mongodb/Auth/User.php
```diff
@@ -15,5 +16,5 @@ class User extends Model implements
     AuthorizableContract,
     CanResetPasswordContract
 {
-    use Authenticatable, Authorizable, CanResetPassword;
+    use Authenticatable, Authorizable, CanResetPassword, MustVerifyEmail;
```
Smolevich: Check that we extends https://github.com/laravel/framework/blob/6.x/src/Illuminate/Foundation/Auth/User.php and remove interfaces, please

Smolevich: It is hard because we should extends inheritor from https://github.com/jenssegers/laravel-mongodb/blob/master/src/Jenssegers/Mongodb/Eloquent/Model.php

PR #1745, opened 2019-04-04
Bugfix create collection with options

deviouspk:
Adding a fix for #1734 .

I might come up with a pull request for json schema's:
https://docs.mongodb.com/manual/reference/operator/query/jsonSchema/#op._S_jsonSchema

@jenssegers would you like to have this feature?

PR #1734, opened 2019-03-18
Add arg options for creating collections with options

Smolevich:
- Add argument options for creating collections with options

PR #1953, opened 2020-02-09
[Updated PR#1745] Bugfix create collection with options

divine:
This PR was recreated from #1745 with updated branch and added tests.

--

Create collection with options
Adding a fix for #1734 .


src/Jenssegers/Mongodb/Schema/Blueprint.php
```diff
@@ -236,14 +236,14 @@ public function expire($columns, $seconds)
     /**
      * @inheritdoc
```
Smolevich: Maybe replace doc block on actual?

divine: I've added correct doc block, thank you! 👍

PR #1954, opened 2020-02-09
[Updated PR#1491] Gets database name from DSN string

divine:
This PR was recreated from #1491 with updated branch and resolved conflicting file.

--

This contains PR #1457
and in addition I've added an ability to use database name from DSN string, so it a developer passed DSN string its not necessary to pass database name.

### DIFF
--- a/src/Jenssegers/Mongodb/Connection.php
+++ b/src/Jenssegers/Mongodb/Connection.php
@@ -38,7 +38,7 @@ class Connection extends BaseConnection
         $this->connection = $this->createConnection($dsn, $config, $options);
 
         // Select database
-        $this->db = $this->connection->selectDatabase($config['database']);
+        $this->db = $this->connection->selectDatabase($this->getDatabaseDsn($dsn, $config['database']));
 
         $this->useDefaultPostProcessor();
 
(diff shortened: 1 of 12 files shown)


Giacomo92:
Schermata 2020-02-09 alle 11 50 02

Smolevich:
@Giacomo92 i see this moment and case when `$dsn` be syntax incorrect will be happend
For example,
`$dsn = 'mongodb://';`
function return empty string

src/Jenssegers/Mongodb/Connection.php
```diff
@@ -38,7 +38,7 @@ public function __construct(array $config)
         $this->connection = $this->createConnection($dsn, $config, $options);
 
         // Select database
-        $this->db = $this->connection->selectDatabase($config['database']);
+        $this->db = $this->connection->selectDatabase($this->getDatabaseDsn($dsn, $config['database']));
```
derekrprice: When I don't set `$config['database']`, I get this error:
`Undefined index: database {"userId":7,"exception":"[object] (ErrorException(code: 0): Undefined index: database at /var/www/device-service/vendor/jenssegers/mongodb/src/Jenssegers/Mongodb/Connection.php:41)`

I think you should pass `@$config['database']` here.

src/Jenssegers/Mongodb/Connection.php
```diff
@@ -188,10 +188,21 @@ protected function getHostDsn(array $config)
...
+    /**
+     * Get database name from DSN string, if there is no database in DSN path - returns back $database argument.
+     * @param string $dsn
+     * @param $database
+     * @return string
+     */
+    protected function getDatabaseDsn($dsn, $database)
+    {
+        $dsnDatabase = trim(parse_url($dsn, PHP_URL_PATH), '/');
+        return trim($dsnDatabase) ? $dsnDatabase : $database;
```
derekrprice: I think this should be ` return $database ?: trim($dsnDatabase);`.  $database should not be a required configuration option but should override the URL DB name when set.

PR #1145, opened 2017-03-06
items instance Eloquent Collection in paginate

SamsamBabadi:
in embeds many relation in paginate method items must be instance of Eloquent Collection same as get method

PR #1959, opened 2020-02-11
[Updated PR#1145] Fix paginate in EmbedsMany

Giacomo92:
This PR was recreated from #1145 
--

As describe [here](https://laravel.com/api/5.0/Illuminate/Pagination/LengthAwarePaginator.html), `$items` (a parameter of `LengthAwarePaginator`) need to be an instance of Collection. I also converted the `slice()` operation with the collection's method and I have updated the docs.

Co-Authored-By: @SamsamBabadi

### DIFF
--- a/src/Jenssegers/Mongodb/Relations/EmbedsMany.php
+++ b/src/Jenssegers/Mongodb/Relations/EmbedsMany.php
@@ -273,7 +273,7 @@ class EmbedsMany extends EmbedsOneOrMany
      */
     public function paginate($perPage = null, $columns = ['*'], $pageName = 'page', $page = null)
     {
-        $page = Paginator::resolveCurrentPage();
+        $page = $page ?: Paginator::resolveCurrentPage($pageName);
         $perPage = $perPage ?: $this->related->getPerPage();
 
         $results = $this->getEmbedded();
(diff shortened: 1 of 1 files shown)


src/Jenssegers/Mongodb/Relations/EmbedsMany.php
```diff
@@ -265,25 +265,36 @@ protected function associateExisting($model)
...
-     * @param int $perPage
-     * @return \Illuminate\Pagination\AbstractPaginator
+     * @param null $perPage
+     * @param array $columns
+     * @param string $pageName
+     * @param null $page
+     * @return \Illuminate\Contracts\Pagination\LengthAwarePaginator
      */
-    public function paginate($perPage = null)
+    public function paginate($perPage = null, $columns = ['*'], $pageName = 'page', $page = null)
```
Smolevich: Maybe to add additional test case for this [model]( https://github.com/jenssegers/laravel-mongodb/blob/2d76d22a5ea4dccc6f5a352b0003c1483f72e0f0/tests/models/Address.php#L14)

Giacomo92: The result will not change.  In the method the only parameter used is `$perPage`. I add the other parameters only to respect the original paginate method.

src/Jenssegers/Mongodb/Relations/EmbedsMany.php
```diff
@@ -265,25 +265,36 @@ protected function associateExisting($model)
...
     {
         $page = Paginator::resolveCurrentPage();
```
divine:
```suggestion
        $page = $page ?: Paginator::resolveCurrentPage($pageName);
```

https://github.com/laravel/framework/blob/6.x/src/Illuminate/Database/Eloquent/Builder.php#L716

PR #1961, opened 2020-02-14
[Fix] Format exception to string in failed jobs

divine:
Format exception to string in failed job so it can be saved correctly.

See: https://github.com/laravel/framework/blob/6.x/src/Illuminate/Queue/Failed/DatabaseFailedJobProvider.php#L59

Closes: https://github.com/jenssegers/laravel-mongodb/issues/1800, https://github.com/jenssegers/laravel-mongodb/pull/1524, https://github.com/jenssegers/laravel-mongodb/pull/1480

### DIFF
--- a/src/Jenssegers/Mongodb/Queue/Failed/MongoFailedJobProvider.php
+++ b/src/Jenssegers/Mongodb/Queue/Failed/MongoFailedJobProvider.php
@@ -12,11 +12,14 @@ class MongoFailedJobProvider extends DatabaseFailedJobProvider
      * @param string $connection
      * @param string $queue
      * @param string $payload
+     * @param  \Exception  $exception
      * @return void
      */
     public function log($connection, $queue, $payload, $exception)
     {
         $failed_at = Carbon::now()->getTimestamp();
+
+        $exception = (string) $exception;
 
         $this->getTable()->insert(compact('connection', 'queue', 'payload', 'failed_at', 'exception'));
     }


divine:
We might actually do something like this:
```php
    /**
     * Log a failed job into storage.
     * @param string $connection
     * @param string $queue
     * @param string $payload
     * @param  \Exception  $exception
     * @return \MongoDB\BSON\ObjectId|null
     */
    public function log($connection, $queue, $payload, $exception)
    {
        $failed_at = Carbon::now()->getTimestamp();

        $exception = (string) $exception;

        return $this->getTable()->insertGetId(compact(
            'connection', 'queue', 'payload', 'exception', 'failed_at'
        ));
    }
```

However `insertGetId` returns `@inheritdoc` which is `int ` but in our case it's `MongoDB\BSON\ObjectId`

See:
https://github.com/jenssegers/laravel-mongodb/blob/master/src/Jenssegers/Mongodb/Eloquent/Builder.php#L66
and
https://github.com/jenssegers/laravel-mongodb/blob/master/src/Jenssegers/Mongodb/Query/Builder.php#L572

Giacomo92:
Add tests, see [here](https://github.com/spatie/laravel-failed-job-monitor/blob/8127afac1fe52c8a3c6d0ebf2a88476bd084831c/tests/FailedJobMonitorTest.php#L68)

divine:
> Add tests, see [here](https://github.com/spatie/laravel-failed-job-monitor/blob/8127afac1fe52c8a3c6d0ebf2a88476bd084831c/tests/FailedJobMonitorTest.php#L68)

There is no tests at all for jobs so I'm not sure if we should do that now?

Also function `log` doesn't return insert id which should do that at first, then we could work out on adding good tests.

I think better example for tests would be https://github.com/laravel/framework/tree/6.x/tests/Queue

Giacomo92:
Yes, Laravel tests are more complete. It would be awesome if you add tests for all job features.

divine:
> Yes, Laravel tests are more complete. It would be awesome if you add tests for all job features.

Even making a breaking changes is fine? 😀

Issue #1493, opened 2018-04-19
get by random order


saeedvz:
Laravel has inRandomOrder() method on collections.

How to get documents by random order here?

Is it possible?

xxCryptoxx:
`$randomUser = User::all()->inRandomOrder()->first();`
this might work if not
`$randomUser = DB::table('users') ->inRandomOrder() ->first();`

thureos:
Thanks @xxCryptoxx , but non of this worked.

I got this with both eloquent or query builder.

` [MongoDB\Driver\Exception\ServerException] bad sort specification`

leelijin:
same question

Didstriker:
Got the same problem and just found out that MongoDB has no built-in way to return random documents.

There is something similar though, from version 3.2 (https://docs.mongodb.com/manual/reference/operator/aggregation/sample/) but as the docs say, it might return more than once the same record on a request. 

Used it on my code like; 
`$result = myModel::raw(function($collection){
return $collection->aggregate([
                ['$sample' => ['size' => 3]]
            ]);
})`

Jaffer2x:
> 
> 
> Got the same problem and just found out that MongoDB has no built-in way to return random documents.
> 
> There is something similar though, from version 3.2 (https://docs.mongodb.com/manual/reference/operator/aggregation/sample/) but as the docs say, it might return more than once the same record on a request.
> 
> Used it on my code like;
> `$result = myModel::raw(function($collection){ return $collection->aggregate([ ['$sample' => ['size' => 3]] ]); })`

Thank u!! worked like a charm. I just need to figure out how to add a condition and i done.

farad4y:
@Didstriker Thank you 👶🏼

rajatmasih-web4realty:
You should try this.
`$total_records = $your_model->count();`
`if ($total_records > 0){`
`	$randIndex = rand(0, $total_records-1);`
`	$obj = $your_model->skip($randIndex)->get();`
`}`

GromNaN:
Thanks for the provided solution https://github.com/mongodb-php/laravel-mongodb/issues/1493#issuecomment-452664988

Giacomo92:
@GromNaN Do you think a new method sample() or inRandomOrder() can be added?

See #1967

GromNaN:
We are going to work on an aggregation builder in the coming month, which will provide a lot more features with an unified interface. I don't think an opininiated method is necessary at this point.

[...]

Issue #107, opened 2014-01-10
Return random record


gmarineau:
The following code return a error

 $articles = Article::orderBy(DB::raw('RAND()'))->get();

Illegal offset type in /Mongodb/Query/Builder.php Line 284

mdhama:
if number of articles equals to random number than it will gives `null`. should use

`$article = Article::take(1)->skip(rand(0,$noOfArticles-1))->first();`

ragazzid:
db.collection.aggregate([{$sample: {size: 5} }])

laratutodotcom:
https://laratuto.com/how-to-get-random-records-in-laravel/ see this guys............

PR #1967, opened 2020-02-18
[Updated PR#1854] Added aggregate-sample function to generate random result from query

Giacomo92:
This PR was recreated from #1854 
--

The sample() function is used to generate random result based on the given size. It can be used together with eloquent functions like ```where()`` and such.

Co-Authored-By: @adlanarifzr

### DIFF
--- a/README.md
+++ b/README.md
@@ -405,6 +405,7 @@ $price = Product::max('price');
 $price = Product::min('price');
 $price = Product::avg('price');
 $total = Product::sum('price');
+$random = Product::sample(5);
 ```
 
 Aggregations can be combined with **where**:
--- a/src/Jenssegers/Mongodb/Query/Builder.php
+++ b/src/Jenssegers/Mongodb/Query/Builder.php
@@ -55,6 +55,12 @@ class Builder extends BaseBuilder
     public $paginating = false;
 
     /**
+     * Indicate the number of random row requested
+     * @var int
+     */
+    public $sample;
+
+    /**
      * All of the available clause operators.
      * @var array
      */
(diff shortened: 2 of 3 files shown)


divine:
Do we really need to add this feature?

jenssegers:
While it does look like a neat feature, I also think it's really important to add it to the library. My idea of this library is to make it super easy for Laravel developers to switch between SQL/Mongo database implementations without having to change too much of their code. This sample method feels very Mongo oriented. But then again, there are already quite some Mongo specific features in the library right now.

Giacomo92:
There are some of issues asking for randomness.
https://github.com/jenssegers/laravel-mongodb/issues/107
https://github.com/jenssegers/laravel-mongodb/issues/1493

For now the only way to get random data is to use collection. In Laravel there is a method to called `inRandomOrder`  see [here](https://laravel.com/docs/5.7/queries#ordering-grouping-limit-and-offset) to achieve that without using the collection. In my opinion If this package has to be "super easy for Laravel developers", it is not possible to run a raw query to get random data from the database.

divine:
> There are some of issues asking for randomness.
> #107
> #1493
> 
> For now the only way to get random data is to use collection. In Laravel there is a method to called `inRandomOrder` see [here](https://laravel.com/docs/5.7/queries#ordering-grouping-limit-and-offset) to achieve that without using the collection. In my opinion If this package has to be "super easy for Laravel developers", it is not possible to run a raw query to get random data from the database.

My suggestion would be to add this function in the next version when we refactor a little bit of pipeline, not this maintenance release which is almost fixing everything.

Regarding #107, #1493 we have important issues such not working relations, enchantments could be worked on when everythings works as expected at least.

What do you think about this @Smolevich ?

divine:
After internal discussion we have decided this is not the right time to add this new feature in the current release.

mean-cj:
I thank this feature we are need. 
better than sample() to inRandomOrder()  can support both SQL/Mongo it better.

PR #1971, opened 2020-02-26
[fix] default database detection from dsn

divine:
Refactored default database function which detects based on db config or tries to detect it from dsn;

Closes: https://github.com/jenssegers/laravel-mongodb/issues/1861

### DIFF
--- a/tests/ConnectionTest.php
+++ b/tests/ConnectionTest.php
@@ -32,6 +32,15 @@ class ConnectionTest extends TestCase
         $this->assertInstanceOf(\MongoDB\Client::class, $connection->getMongoClient());
     }
 
+    public function testDsnDb()
+    {
+        $connection = DB::connection('dsn_mongodb_db');
+        $this->assertInstanceOf(\MongoDB\Database::class, $connection->getMongoDB());
+
+        $connection = DB::connection('dsn_mongodb_db');
+        $this->assertInstanceOf(\MongoDB\Client::class, $connection->getMongoClient());
+    }
+
     public function testCollection()
     {
         $collection = DB::connection('mongodb')->getCollection('unittest');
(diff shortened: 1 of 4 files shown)


tests/ConnectionTest.php
```diff
@@ -32,6 +32,15 @@ public function testDb()
         $this->assertInstanceOf(\MongoDB\Client::class, $connection->getMongoClient());
     }
 
+    public function testDsnDb()
+    {
+        $connection = DB::connection('dsn_mongodb_db');
+        $this->assertInstanceOf(\MongoDB\Database::class, $connection->getMongoDB());
+
+        $connection = DB::connection('dsn_mongodb_db');
```
Smolevich: What is the profit to create new instance connection?

tests/ConnectionTest.php
```diff
@@ -32,6 +32,15 @@ public function testDb()
         $this->assertInstanceOf(\MongoDB\Client::class, $connection->getMongoClient());
     }
 
+    public function testDsnDb()
+    {
+        $connection = DB::connection('dsn_mongodb_db');
+        $this->assertInstanceOf(\MongoDB\Database::class, $connection->getMongoDB());
```
Smolevich: Add `\MongoDB\Database` to section use

tests/ConnectionTest.php
```diff
@@ -32,6 +32,15 @@ public function testDb()
...
+
+        $connection = DB::connection('dsn_mongodb_db');
+        $this->assertInstanceOf(\MongoDB\Client::class, $connection->getMongoClient());
```
Smolevich: Leave only `Client::class` please

divine: I've just copied from upper function, see https://github.com/jenssegers/laravel-mongodb/blob/master/tests/ConnectionTest.php#L26

Should we refactor all tests? It's not like me who did that 😉

Smolevich: OK, I suggest refactor tests in another PR

derekrprice:
That worked perfectly, thanks!

PR #1988, opened 2020-03-05
[4.x] Laravel 7

divine:
Added Laravel 7 compatibility.


src/Jenssegers/Mongodb/Eloquent/Model.php
```diff
@@ -223,36 +223,37 @@ public function getCasts()
     /**
      * {@inheritdoc}
```
Smolevich: I suggest to remove this dock block and same in this file

tests/DatabaseEloquentTimestampsTest.php
```diff
@@ -0,0 +1,107 @@
...
+
+use Illuminate\Support\Carbon;
+use Jenssegers\Mongodb\Eloquent\Model as Eloquent;
+
+class DatabaseEloquentTimestampsTest extends TestCase
+{
+    /**
+     * Tear down the database schema.
+     *
+     * @return void
```
Smolevich: Let's to remove dock block if return type hints are exists

tests/DatabaseEloquentTimestampsTest.php
```diff
@@ -0,0 +1,107 @@
...
+     */
+    protected function tearDown(): void
+    {
+        $this->schema()->drop('users');
+        $this->schema()->drop('users_created_at');
+        $this->schema()->drop('users_updated_at');
+    }
+
+    /**
+     * Tests...
```
Smolevich: I suggest to remove this doc blocks . They looks like unnecessary

tests/DatabaseEloquentTimestampsTest.php
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    public function testUserWithCreatedAt()
+    {
+        $now = Carbon::now();
+        $user = UserWithCreated::create([
+            'email' => 'test@test.com',
+        ]);
+
+        $this->assertEquals($now->toDateTimeString(), $user->created_at->toDateTimeString());
```
Smolevich: Maybe better to add test case assertNull on `$user->updated_at`

tests/DatabaseEloquentTimestampsTest.php
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    public function testUserWithUpdatedAt()
+    {
+        $now = Carbon::now();
+        $user = UserWithUpdated::create([
+            'email' => 'test@test.com',
+        ]);
+
+        $this->assertEquals($now->toDateTimeString(), $user->updated_at->toDateTimeString());
```
Smolevich: Maybe better to add test case assertNull on $user->created_at

PR #1667, opened 2018-12-01
Unset field

halaei:
This Changes the way fields are unset, so that by a single update query, changes fields get updated and unset fields get removed from the document.

PR #1990, opened 2020-03-06
[4.x] use mongodb driver to check for dirtiness

halaei:
Reopening of #1664 

> To check if a field is dirty we can simply use fromPHP() function of the mongodb driver. Besides, we shouldn't use the SQL-specific logic of Illuminate\Eloquent\Model::originalIsEquivalent()  for mongodb. In sql, fields keep their schema data types and changing  integer 1 to string '1' actually does nothing. However, in mongodb  fields accept their types as they are inserted/updated. Therefore, if we  change value of a field from (int)1 to (string)'1' we actually must regard it as dirty. This change is backward incompatible. Don't know if anyone likes the current behaviour more.
--

### DIFF
--- a/src/Jenssegers/Mongodb/Eloquent/Model.php
+++ b/src/Jenssegers/Mongodb/Eloquent/Model.php
@@ -14,6 +14,8 @@ use Jenssegers\Mongodb\Query\Builder as QueryBuilder;
 use MongoDB\BSON\Binary;
 use MongoDB\BSON\ObjectID;
 use MongoDB\BSON\UTCDateTime;
+use function MongoDB\BSON\fromPHP;
+use MongoDB\Driver\Exception\UnexpectedValueException;
 
 abstract class Model extends BaseModel
 {
@@ -239,20 +241,11 @@ abstract class Model extends BaseModel
             return false;
         }
 
-        if ($this->isDateAttribute($key)) {
-            $current = $current instanceof UTCDateTime ? $this->asDateTime($current) : $current;
-            $original = $original instanceof UTCDateTime ? $this->asDateTime($original) : $original;
-
-            return $current == $original;
+        try {
+            return fromPHP([$current]) === fromPHP([$original]);
+        } catch (UnexpectedValueException $e) {
+            return false;
         }
-
-        if ($this->hasCast($key)) {
-            return $this->castAttribute($key, $current) ===
-                $this->castAttribute($key, $original);
-        }
-
-        return is_numeric($current) && is_numeric($original)
-            && strcmp((string) $current, (string) $original) === 0;
     }
 
     /**
--- a/tests/ModelTest.php
+++ b/tests/ModelTest.php
@@ -555,11 +555,50 @@ class ModelTest extends TestCase
     public function testGetDirtyDates(): void
     {
         $user = new User();
-        $user->setRawAttributes(['name' => 'John Doe', 'birthday' => new DateTime('19 august 1989')], true);
+        $user->name = 'John Doe';
+        $user->birthday = new DateTime('19 august 1989');
+        $user->syncOriginal();
         $this->assertEmpty($user->getDirty());
 
         $user->birthday = new DateTime('19 august 1989');
         $this->assertEmpty($user->getDirty());
+    }
+
+    public function testGetDirty(): void
+    {
+        $ids = [
+            new ObjectId(),
+            new ObjectId(),
+        ];
+        $item = new Item([
+            'numbers' => [1, 2, 3],
+            'number' => 4,
+            'ids' => $ids,
+            'nullable' => 'value',
+            'fix' => 'fix',
+        ]);
+        $item->date = $item->fromDateTime(new DateTime('19 august 1989'));
+        $item->dates = [$item->fromDateTime(new DateTime('19 august 1989'))];
+        $item->save();
+
+        $item = $item->fresh();
+
+        $item->numbers = [1, 2, '3'];
+        $item->nullable = null;
+        $item->new_val = 'new';
+        $item->number = '4';
+        $item->ids = [
+            new ObjectId((string) $ids[0]),
+            new ObjectId((string) $ids[1]),
+        ];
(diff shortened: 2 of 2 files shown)


jenssegers:
# [Codecov](https://codecov.io/gh/jenssegers/laravel-mongodb/pull/1990?src=pr&el=h1) Report
> Merging [#1990](https://codecov.io/gh/jenssegers/laravel-mongodb/pull/1990?src=pr&el=desc) into [develop](https://codecov.io/gh/jenssegers/laravel-mongodb/commit/b379bd39cb9499e778a30ca4ad7f4b4558e3d39c&el=desc) will **decrease** coverage by `0.08%`.
> The diff coverage is `33.33%`.

[Impacted file tree graph](https://codecov.io/gh/jenssegers/laravel-mongodb/pull/1990?src=pr&el=tree)

```diff
@@              Coverage Diff              @@
##             develop    #1990      +/-   ##
=============================================
- Coverage      82.80%   82.72%   -0.09%     
+ Complexity       531      526       -5     
=============================================
  Files             27       27              
  Lines           1210     1204       -6     
=============================================
- Hits            1002      996       -6     
  Misses           208      208              
```



------

[Continue to review full report at Codecov](https://codecov.io/gh/jenssegers/laravel-mongodb/pull/1990?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/jenssegers/laravel-mongodb/pull/1990?src=pr&el=footer). Last update [b379bd3...9e7866a](https://codecov.io/gh/jenssegers/laravel-mongodb/pull/1990?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

src/Jenssegers/Mongodb/Eloquent/Model.php
```diff
@@ -239,20 +241,11 @@ public function originalIsEquivalent($key, $current)
...
         }
 
-        if ($this->isDateAttribute($key)) {
-            $current = $current instanceof UTCDateTime ? $this->asDateTime($current) : $current;
-            $original = $original instanceof UTCDateTime ? $this->asDateTime($original) : $original;
-
-            return $current == $original;
-        }
-
-        if ($this->hasCast($key)) {
```
divine: The only problem I see is a casted attributes, probably we need to check after keys was casted back to their values.

halaei: @divine  Looking at the code, I don't see it is required to check for casts. The input values passed by getDirty() are uncasted:
```php
    public function getDirty()
    {
        $dirty = [];

        foreach ($this->getAttributes() as $key => $value) {
            if (! $this->originalIsEquivalent($key, $value)) {
                $dirty[$key] = $value;
            }
        }

        return $dirty;
    }
    public function getAttributes()
    {
        return $this->attributes;
    }
```
Besides, I don't see why one should use Laravel <=6 casting system for MongoDB, I don't see a test for it in the repository either. Laravel 7 custom casts are a different story. I hope they won't make things go wrong. I've already had my custom cast implementation along side this PR in Larave 5.8 production without an issue.

halaei:
@jenssegers Nice coverage tool. But I feel I decreased the source code complexity in terms of both lines of code and branches. I also added new tests. How come the coverage is decreased? I think the report is somehow wrong.

halaei:
I see the report now. I didn't test for when exception is thrown. I don't quite remember. Maybe it has to do with my next pull request regarding unset field #1667

src/Jenssegers/Mongodb/Eloquent/Model.php
divine: Well, we're dropping support for laravel 6 (no other way to drop object id auto-casting) and this branch will be focused for versions > 7.0, so that's the reason why I've asked this question.

This might be an issue with Laravel 7 when it wouldn't check dirtiness for casted objects.

Yes, there are no tests, I think a lot of tests are missing, might take a look myself (for tests).

halaei: @divine Then I should send this PR on top of your own PR #1988 ?

divine: Probably no, it's not ready yet, still have something to do. I will let you know, ok?

halaei: @divine I have already done some checks and coding for Laravel 7. Sounds like casting in Laravel 7 also doesn't break things.
I'll send you that PR anyway. I'll wait for you to let me work on my unset-field PR #1667 after this is done.

tests/ModelTest.php
```diff
@@ -555,13 +555,52 @@ public function testMultipleLevelDotNotation(): void
...
         $this->assertEmpty($user->getDirty());
 
         $user->birthday = new DateTime('19 august 1989');
         $this->assertEmpty($user->getDirty());
     }
 
+    public function testGetDirty(): void
+    {
+        $ids = [
+            new ObjectId(),
```
Smolevich: Add value for ObjectId please

tests/ModelTest.php
```diff
@@ -555,13 +555,52 @@ public function testMultipleLevelDotNotation(): void
...
+        $item->dates = [$item->fromDateTime(new DateTime('19 august 1989'))];
+        $item->save();
+
+        $item = $item->fresh();
+
+        $item->numbers = [1, 2, '3'];
+        $item->nullable = null;
+        $item->new_val = 'new';
+        $item->number = '4';
+        $item->ids = [
```
Smolevich: Looks like strange, i suggest to replace on 
```
$item->ids = $ids;
``` or 
```
$item->ids = [
    new ObjectId('576c25db6118fd406e6e6471'),
    new ObjectId('576c25db6118fd406e6e6472')
];
```

divine:
Closing in favor of #2515 after 3 years...

Thank you @halaei !

PR #1997, opened 2020-03-11
[3.x] fix cast paginate per page to int

### LINKED ISSUE #1750: Pagination does not accept string number
## Summary

When calling the `->paginate()` method with the first (`perPage`) argument, mongodb errors if the user passes in a string.\

## Steps to reproduce

1. pass `per_page=20` as a query parameter on an application running this library 
2. use `->paginate($request->input('per_page'))` in the called method
3. get a Mongodb error `the limit must be specified as a number`

## Fix

We should cast the `$perPage` argument to an integer before passing to mongodb; this works in the default Laravel drivers, and should work with this driver. This can be easily solved by a simple `(int)` cast

Yahatix:
Cast per_page parameter Collection::paginate to an integer
Added additional test for this case.
Resolves #1750

### DIFF
--- a/tests/QueryTest.php
+++ b/tests/QueryTest.php
@@ -327,6 +327,20 @@ class QueryTest extends TestCase
         $this->assertEquals(1, $results->currentPage());
     }
 
+    public function testThatPaginateCastsPerPageParameter()
+    {
+        $results = User::paginate('2');
+        $this->assertEquals(2, $results->count());
+        $this->assertNotNull($results->first()->title);
+        $this->assertEquals(9, $results->total());
+
+        $results = User::paginate('2', ['name', 'age']);
+        $this->assertEquals(2, $results->count());
+        $this->assertNull($results->first()->title);
+        $this->assertEquals(9, $results->total());
+        $this->assertEquals(1, $results->currentPage());
+    }
+
     public function testUpdate(): void
     {
         $this->assertEquals(1, User::where(['name' => 'John Doe'])->update(['name' => 'Jim Morrison']));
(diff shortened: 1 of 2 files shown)


tests/QueryTest.php
```diff
@@ -327,6 +327,20 @@ public function testPaginate(): void
         $this->assertEquals(1, $results->currentPage());
     }
 
+    public function testThatPaginateCastsPerPageParameter()
+    {
+        $results = User::paginate('2');
```
Smolevich: What is the reason pass $perPage as string, if original [method](https://github.com/laravel/framework/blob/7.x/src/Illuminate/Database/Eloquent/Builder.php#L732-L741) expect int or null ?

Yahatix: See issue: #1750

Smolevich: Tests in laravel don't have case with pass string `$perPage` https://github.com/laravel/framework/blob/0b12ef19623c40e22eff91a4b48cb13b3b415b25/tests/Integration/Database/QueryBuilderTest.php#L181
https://github.com/laravel/framework/blob/0b12ef19623c40e22eff91a4b48cb13b3b415b25/tests/Database/DatabaseQueryBuilderTest.php#L3205

Yahatix: So... Whats the problem with adding additional tests?

Smolevich: I don't understand why we  pass not int value

Yahatix: Because it's possible when using the default laravel model without mongodb

Smolevich: i saw in [docs](https://laravel.com/docs/7.x/pagination) and tests and didn't find examples with passing string `$perPage`

Yahatix: To make you happy, i just tested it in a stock laravel installation and confirmed that it actually does work in a stock installation.
And to no surprise it does.
Granted, when i created the PR i didn't fact check the original issue

Issue #2016, opened 2020-04-08
Like query not working on integer fields.


ultimateakash:
- Laravel-mongodb Version: ^3.5
- PHP Version: 7.3
- Database Driver & Version: 3.4 (mongo)

### Description:
In my project i need to implement search functionality on different columns. on string column it works fine but it didn't work on integer fields.

### Steps to reproduce
Just for example
```
$code = 256;
        
$branch = Branch::where('code','like',"%".$code."%")->first();
        return response()->json([
            'branch' => $branch
        ]);
```

Smolevich:
@ultimateakash What is the type for `code` in mongodb?

divine:
Closing. PR with the fix has been merged and released.

Thanks!

PR #2020, opened 2020-04-11
query like on integer fields support

ilyasokay:
Fix for like query on integer fields which is discussed #2016 issue is ready.


src/Jenssegers/Mongodb/Query/Builder.php
```diff
@@ -1003,17 +1004,23 @@ protected function compileWhereBasic(array $where)
             }
 
             // Convert to regular expression.
-            $regex = preg_replace('#(^|[^\\\])%#', '$1.*', preg_quote($value));
+            $regex          = preg_replace('#(^|[^\\\])%#', '$1.*', preg_quote($value));
```
Smolevich, line 1007: Please delete indentation, let's to use common style for code

Issue #2026, opened 2020-04-22
Reading UTCDateTime from before 1970 results in an exception.


DFurnes:
- Laravel-mongodb Version: 3.6.3
- PHP Version: 7.3.15
- Database Driver & Version: v4.0.16

### Description:

Reading a date field with a value before January 1st 1970 results in a fatal exception:

```
ErrorException:  A non well formed numeric value encountered in vendor/nesbot/carbon/src/Carbon/Traits/Timestamp.php on line 50
```

### Steps to reproduce
Create a record with a pre-1970 date in a date-casted field (like `created_at`):

```js
db.users.insert({"email": "broken@example.com", "created_at": ISODate("1968-01-17T14:19:49.324Z")});
```

Then, try to read it as an Eloquent model:

```php
$user = User::where('email', 'broken@example.com')->first();

print $user->created_at; // <-- ErrorException!
```

### Expected behaviour
Dates, before or after 1970, should be correctly parsed into `Carbon` objects!

### Actual behaviour

When trying to read a record with an `UTCDateTime` before 1970, the millisecond timestamp will be formed incorrectly. This causes the following error when [trying to create the Carbon instance](https://github.com/jenssegers/laravel-mongodb/blob/master/src/Jenssegers/Mongodb/Eloquent/Model.php#L102):

```
ErrorException:  A non well formed numeric value encountered in vendor/nesbot/carbon/src/Carbon/Traits/Timestamp.php on line 50
```

This seems to be because, for dates prior to 1970, both the timestamp (from `U`) and the milliseconds (from `v`) are provided as negative numbers, and so when concatenated they don't make sense (e.g. `"-61724410-676"` for January 1st 1968 rather than `"-61724410676"`).

<details><summary><b>Logs</b>:</summary>

```
[2020-04-22 19:45:50] local.ERROR: A non well formed numeric value encountered {"exception":"[object] (ErrorException(code: 0): A non well formed numeric value encountered at vendor/nesbot/carbon/src/Carbon/Traits/Timestamp.php:50)
[stacktrace]
#0 vendor/nesbot/carbon/src/Carbon/Traits/Timestamp.php(50): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError(8, 'A non well form...', '/home/vagrant/S...', 50, Array)
#1 vendor/laravel/framework/src/Illuminate/Support/DateFactory.php(217): Carbon\\Carbon::createFromTimestampMs('-61724410-676')
[...]

PR #2028, opened 2020-04-22
Fix issue parsing millisecond-precision dates before 1970.

DFurnes:
This pull request adds some tests & a fix for an issue I ran into where pre-1970 dates with millisecond-precision could not be properly cast into a `Carbon` instance.

References #2026.


Smolevich:
Thanks for contributing!

PR #2038, opened 2020-05-22
BUGFIX: Can not 'call to a member function first() on array' in non-l…

lesteryou:
Can not 'call to a member function first() on array' in non-laravel project.

  <b>0</b>: Call to a member function first() on array in <b>E:\new\yzc\storage\vendor\illuminate\database\Concerns\BuildsQueries.php</b> on line <b>77</b>

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,12 @@ $capsule->getDatabaseManager()->extend('mongodb', function($config, $name) {
     return new Jenssegers\Mongodb\Connection($config);
 });
 ```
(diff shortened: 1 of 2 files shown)


lesteryou:
Reviewed

lesteryou:
ok

divine:
We've removed `shouldusecollection` function. It was done to support Laravel 5.3 and has nothing to do with the non Laeavel projects.

PR #2039, opened 2020-05-26
tests: for sorting

willtj:
This PR adds tests for sorting results from MongoDB - first a test to ensure that the results are unsorted as expected when no sort is applied, then tests for single and multiple sorts with/without a sort order.

https://github.com/jenssegers/laravel-mongodb/issues/1508 (and the fact that this was added to the v4 project board) seems to suggest that applying multiple sorts isn't supported, but everything seems to be working as expected unless I've totally misunderstood the issue?


Smolevich:
@willtj thanks for contribution!

PR #2069, opened 2020-07-22
always call connection logQuery function to broadcast QueryExecuted event

StevePorter92:
fixes https://github.com/jenssegers/laravel-mongodb/issues/2068

Queries *will only* be added to the in memory connection query log if
the query log is enabled using `DB::connection('mongodb')->enableQueryLog();`


divine:
Thanks!

PR #2082, opened 2020-08-19
[3.x] Fix guarded to return always true

### LINKED ISSUE #2078: After update laravel to 6.18.35, i got error in model that has $guarded attribute
- Laravel-mongodb Version: 3.6.4
- PHP Version: 7.3.21
- Database Driver & Version: mongodb 1.5.5

### Description:
Recently, was merged https://github.com/laravel/framework/pull/33777. Since then, I'm getting an error when I try to use the create method with a model that has $guarded attribute

### Steps to reproduce
1. Have a Model that has $guarded attribute:
```
class Tester extends \Moloquent
{
    protected $guarded = [
        'test',
    ];
}
```
2. call Tester::create(['foo' => 'bar']);

### Expected behaviour
It should create a object with property 'foo' in database

### Actual behaviour
PHP Error:  Call to a member function compileColumnListing() on null in /home/project/vendor/laravel/framework/src/Illuminate/Database/Schema/Builder.php on line 136

fish3046:
An update to Laravel has caused a breakdown when using guarded models.  See this for more info:

https://blog.laravel.com/security-release-laravel-61834-7232
https://github.com/laravel/framework/pull/33777

One error this caused is: `Call to a member function compileColumnListing() on null`

This is because `\Illuminate\Database\Schema\Builder::__construct()` now creates a grammar object, and our overloaded constructor does not.  I've removed our constructor to fall back to the parent's, as we aren't doing anything special in ours.

The next error is `BadMethodCallException : Method Jenssegers\Mongodb\Schema\Grammar::compileColumnListing does not exist`

This is because the new default behavior on guarded models is to query the schema for a column listing to ensure you are guarding actual fields.  Because we are a document model, I have overloaded `Model::isGuardableColumn()` to return true.  This aides in preventing all manner of checks and complications that don't really apply to Mongo models.

fixes #2078

### DIFF
--- a/tests/ModelTest.php
+++ b/tests/ModelTest.php
@@ -722,4 +722,11 @@ class ModelTest extends TestCase
 
         $this->assertEquals(0, User::count());
     }
+
+    public function testGuardedModel()
+    {
+        Guarded::create(['var' => 'val']);
+
+        $this->assertEquals(1, Guarded::count());
+    }
 }
--- a/tests/ModelTest.php
+++ b/tests/ModelTest.php
@@ -726,8 +726,24 @@ class ModelTest extends TestCase
 
     public function testGuardedModel()
     {
-        Guarded::create(['var' => 'val']);
+        $model = new Guarded();
 
-        $this->assertEquals(1, Guarded::count());
+        // foobar is properly guarded
+        $model->fill(['foobar' => 'ignored', 'name' => 'John Doe']);
+        $this->assertFalse(isset($model->foobar));
+        $this->assertSame('John Doe', $model->name);
+
+        // foobar is guarded to any level
+        $model->fill(['foobar->level2' => 'v2']);
+        $this->assertNull($model->getAttribute('foobar->level2'));
+
+        // multi level statement also guarded
+        $model->fill(['level1->level2' => 'v1']);
+        $this->assertNull($model->getAttribute('level1->level2'));
+
+        // level1 is still writable
+        $dataValues = ['array', 'of', 'values'];
+        $model->fill(['level1' => $dataValues]);
+        $this->assertEquals($dataValues, $model->getAttribute('level1'));
     }
 }
--- a/tests/ModelTest.php
+++ b/tests/ModelTest.php
@@ -19,6 +19,7 @@ class ModelTest extends TestCase
         Soft::truncate();
         Book::truncate();
         Item::truncate();
+        Guarded::truncate();
     }
 
     public function testNewModel(): void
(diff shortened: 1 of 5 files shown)


tests/ModelTest.php
```diff
@@ -722,4 +723,11 @@ public function testTruncateModel()
 
         $this->assertEquals(0, User::count());
     }
+
+    public function testGuardedModel()
+    {
+        Guarded::create(['var' => 'val']);
+
+        $this->assertEquals(1, Guarded::count());
```
Smolevich, line 731: It's not related test case, i think that we should test that model Guarded doesn't set fields in guarded attribute of Model
`
$model = Guarded::create(['var' => 'val', 'foobar' => 'bar']);
$this->assertFalse(isset($model->foobar));
` 
See tests in [laravel](https://github.com/laravel/framework/blob/e351423804a816d1acd86e732a425dabddeadd3a/tests/Database/DatabaseEloquentModelTest.php#L1097-L1115)

divine:
@fish3046 I've talked few days ago with @Smolevich about this:
- we need to update readme that `$guarded` doesn't work as expected (as a recent security fix) because this is NoSql database.
- use `$fillable` instead of `$guarded` as suggested by Taylor himself and add a link to his post with explanation https://blog.laravel.com/security-release-laravel-61835-7240

Thanks!

tests/ModelTest.php
roelofr: The error occurs when Laravel checks using an `isGuardedColumn` method. This method however doesn't perform the actual guard checks, it checks if " if the given column is a valid, guardable column" (whatever that may mean).

It's easy to add the test, but it's not strictly neccessary as the `isGuarded` check runs this statement (on Laravel 7.25.0)

```php
return $this->getGuarded() == ['*'] ||
        ! empty(preg_grep('/^'.preg_quote($key).'$/i', $this->getGuarded())) ||
        ! $this->isGuardableColumn($key);
```

([source](https://github.com/laravel/framework/blob/0d1ac3282e019ed4608b715ac4f8c57a0a042b9b/src/Illuminate/Database/Eloquent/Concerns/GuardsAttributes.php#L204-L206))

Smolevich: Change test case, please, we can show that test will be broken if laravel changes behavior with guarded fields

fish3046:
Hopefully these test and readme changes address the concerns.  Thanks for reviewing this so quickly.

Smolevich (approved):

Smolevich:
@fish3046 thanks for contributing!

PR #2086, opened 2020-08-28
Fix styleci risky error

divine:
The risky 'empty_return' fixer cannot be enabled because risky mode is not enabled.

See https://github.styleci.io/analyses/WNGwke

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,1 +1,2 @@
+risky: true
 preset: laravel


divine:
@jenssegers sorry for writing it here, could you please enable discussions in repo settings?

Thanks!

Smolevich (approved):

divine:
@jenssegers ping ^^

PR #2090, opened 2020-09-02
Remove question template

divine:
We've a new discussions where everyone could ask questions.

Not sure about feature requests, maybe we could remove it too!

Thanks!


divine:
@jenssegers could you please enable **empty_return** in styleci configuration https://docs.styleci.io/configuration?

Basically web based configuration overrides the default one.

Thanks!

Smolevich (approved):

PR #2092, opened 2020-09-09
Cursor support for laravel 5.8

skuio:
Added compatible `->cursor()` for query builder without using LazyCollection that supported in Laravel 6
by this PR Laravel 5.8 can use the cursor function


Smolevich (approved):

divine:
Laravel 5.8 EOL August 26th, 2019, Security Fixes Until February 26th, 2020.

There is no point to support that at all.

skuio:
@divine 
Yes, but why we need to use LazyCollection to return Generator?
The Generator is built-in in the PHP!

divine:
Closing this PR as there is no point to backport something that wasn't in Laravel 5.8.

Thanks!

PR #2149, opened 2020-12-16
[3.6] Remove Facade in Queue Service Provider

CurosMJ:
I use this package in an environment where Facades are turned off and I don't want to turn them on. 

This PR removes the Facade use and replaces with plain old 'db' accessor on the $app

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


CurosMJ:
Also, I guess this should be merged with other branches as well

divine (approved):
Hello,

Looks good to me.

> I guess this should be merged with other branches as well

Can you do that and update the changelogs as well here and there?

Thanks!

PR #2159, opened 2020-12-21
Created A Docs about the Laravel MongoDB

BroJenuel:
When we are working on a project, and reading the readme file in the repository takes a lot of time for me to go and find what I'm looking for so for me to be able to at least easily find what I am looking for I decided to create docs for this so that it is simple to navigate using side navigations or tabs.

So I made documentation using vue press, to see what it looks like. It's something like this.
https://laravel-mongodb.netlify.app/


divine:
Hello,

Thank you for your work but why not [nuxt content](https://content.nuxtjs.org/)? It's much better.

Also, I don't think this ever will be merged, there is no web-site even adding a Github page might take months to deploy...

Sorry about that!

Thanks!

BroJenuel:
> Hello,
> 
> Thank you for your work but why not [nuxt content](https://content.nuxtjs.org/)? It's much better.
> 
> Also, I don't think this ever will be merged, there is no web-site even adding a Github page might take months to deploy...
> 
> Sorry about that!
> 
> Thanks!

Thanks, Its ok I understand. and Thanks for suggesting nuxt content, I've seen it and it's really beautiful might as well use it in the future for documenting. :)

PR #2178, opened 2021-01-10
cast foreign key to string in BelongsTo class

amirimatin:
cast foreign key to string in BelongsTo class to fix the bug `Illegal offset type in isset or empty` when the foreign key and other Key are instance of ObjectId class


divine:
Hello,

Please kindly add tests.

Thanks!

divine:
Closing due to inactivity, feel free to reply here if you are still willing to add tests.

Thanks!

PR #2180, opened 2021-01-11
Fix job releasing back to queue

lukaszaleckas:
This pull request fixes job releasing and issue while using queue "--tries" option.

- Overrides "release" parent function. Uses "findOneAndUpdate" for job releasing (updates queue name, resets "reserved" and "reserved_at" and so on)

- Overrides "deleteAndRelease" parent function (to skip transactions) - calls "release" method.

### DIFF
--- a/src/Jenssegers/Mongodb/Queue/MongoQueue.php
+++ b/src/Jenssegers/Mongodb/Queue/MongoQueue.php
@@ -4,6 +4,7 @@ namespace Jenssegers\Mongodb\Queue;
 
 use Carbon\Carbon;
 use Illuminate\Queue\DatabaseQueue;
+use Illuminate\Queue\Jobs\DatabaseJob;
 use Jenssegers\Mongodb\Connection;
 use MongoDB\Operation\FindOneAndUpdate;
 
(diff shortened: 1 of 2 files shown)


Smolevich:
What is the root bug for creating this fix?

Smolevich:
> What is the root bug for creating this fix?

As i can see it related with branch 8.x of laravel-framework and this [changes](https://github.com/laravel/framework/commit/5006bdc244f98691adbdda5c2fe7805880c9c316)

lukaszaleckas:
> > What is the root bug for creating this fix?
> 
> As i can see it related with branch 8.x of laravel-framework and this [changes](https://github.com/laravel/framework/commit/5006bdc244f98691adbdda5c2fe7805880c9c316)

Yes, you're correct - issue is related to these changes. When you run queue worker with "--tries" option and job fails, "deleteAndRelease" function is called and it uses transaction ("Call to a member function beginTransaction() on null" is thrown). Also, release method on the job calls the same function, so it does not work.

Transaction is used because job is deleted and then reinserted, but honestly I can't find a reason why this is being done instead of simple update on a row. @Smolevich Maybe you know the reason?

divine:
> Transaction is used because job is deleted and then reinserted, but honestly I can't find a reason why this is being done instead of simple update on a row

Basically when a simple update fails transaction will revert back changes so no job will be lost.

It's actually a good idea to use transactions but MongoDB is a little bit different for that in terms of requiring sharded/replicas clusters.

lukaszaleckas:
> > Transaction is used because job is deleted and then reinserted, but honestly I can't find a reason why this is being done instead of simple update on a row
> 
> Basically when a simple update fails transaction will revert back changes so no job will be lost.
> 
> It's actually a good idea to use transactions but MongoDB is a little bit different for that in terms of requiring sharded/replicas clusters.
> 
> Forgot to mention that MongoDB on its own handles this well.

Yeah, transaction is understandable, "delete & insert" part I don't understand. Why not just do update query? In that case you would not even need transactions because it would be only one query instead of two separate.

divine:
> Yeah, transaction is understandable, "delete & insert" part I don't understand. Why not just do update query? In that case you would not even need transactions because it would be only one query instead of two separate.

Well, hard to say why they're doing this. 🤔

divine:
Closed in favor of https://github.com/jenssegers/laravel-mongodb/pull/2229

Thank you for your contribution @lukaszaleckas !

PR #2186, opened 2021-01-15
Make Dockerfile compatible with PHPUnit 9

stephandesouza:
PHPUnit 9 only runs with PHP >=7.3, also updated to Composer 2


Smolevich:
ARG is param can be passed through `docker build --build-arg PHP_VERSION=8.0 ...` I think that changing default params is can be realized later, after big refactoring code base

stephandesouza:
Oh, thanks!! I really forgot about `build-args` on Docker.

divine:
Docker file already doesn't run correctly tests. Bumping two variables doesn't change anything.

Thanks!

stephandesouza:
Well, ok, but I could run standalone test correctly on master with this bump.

So others might run this kind of tests too.

divine:
> Well, ok, but I could run standalone test correctly on master with this bump.
> 
> So others might run this kind of tests too.

Docker file doesn't contain any code to run tests. Could you please show me where exactly it does otherwise?

Thanks!

stephandesouza:
Sure, with docker-compose it runs smoothly: `docker-compose run tests ./vendor/bin/phpunit`
image

divine:
> Sure, with docker-compose it runs smoothly: `docker-compose run tests ./vendor/bin/phpunit`
> image

What was I meaning that when you start it with docker-compose the PHP container should run tests automatically.

There is no point in running it manually each time. @Smolevich what do you think?

Thanks!

Dockerfile
```diff
@@ -1,5 +1,5 @@
-ARG PHP_VERSION=7.2
-ARG COMPOSER_VERSION=1.8
+ARG PHP_VERSION=7.3
```
divine, line 1:
```suggestion
ARG PHP_VERSION=7.4
```
7.4 is currently active version

divine (approved):

divine:
Thank you for your contribution! Much appreciated!

PR #2187, opened 2021-01-15
Allow `has` method with numeric indexes

### LINKED ISSUE #2182: `getConstrainedRelatedId` method not returning array_keys as string if numeric keys
- Laravel-mongodb Version: 3.6.6
- PHP Version: 7.3.18
- Database Driver & Version: mongodb/mongodb 1.6.1

### Description:
When using a whereHas filter, if the field under the whereHas clause is a string with a numeric value, `getConstrainedRelatedIds` on the `QueriesRelationships` trait is returning the keys as numeric instead of string. With that, if the column in the where clause is a string, it fails.
### Steps to reproduce
1. Create a parent register with a `_id` PK as string and another field for testing purposes, e.g: `active: true`.
2. Create two child associated with the parent register
3. Created Example:
```
Parent: {
      _id: "12345",
      "active": true
},

Child: {
   _id:"id1",
  "parent_id": "12345"
},
Child: {
  _id: "id2,
  "parent_id": "12345"
}
```
4. Run the query: 
```
Child::whereHas('parent', function ($query) {
       return $query->where('active', true);
       })
```

### Expected behaviour
It should return the two childs

### Actual behaviour
None child is returned.

Insert log.txt here (if necessary)
</details>

stephandesouza:
Fix #2182


divine:
@stephandesouza taking look at this PR it can create some regression like this one https://github.com/jenssegers/laravel-mongodb/issues/2203

> No problem, we're here to find the best solution for everyone!

What solution we could create? Any thoughts?

Thanks!

stephandesouza:
This workaround does not modify the query used by `whereHas()`, only how the IDs are returned by `getConstrainedRelatedIds`.

And, for now, we can't run away from `array_count_values` and `array_keys`. But we can avoid using `array_map` when no numeric ID is used, with a "hasNumericIndex" variable.

divine (changes requested):
Hello,

I still think that this fix doesn't make sense at all.

`_id: "12345"` is a type of string, why this library should take care of logical problems in your project?

Is defining `$keyType = 'int';` not enough?

I'm just trying to understand and this type of code that we're still having issues with.

Thanks!

stephandesouza:
Sorry, the test was wrong after suggested review.. This is a problem with the "reverse" foreign key used on this method, and also happens with primaryKey as int

**Situation:**
Author "id:1" has two books Book "id:A", and book "id:B".
Author "id:'2'" has two books Book "id:C"

**Today:**
`getConstrainedRelatedId()` returns
> ["1", "2"]

**Fixes:**
`getConstrainedRelatedId()` returns
> [1, 2]

And yes, even a "string integer" needs to be converted to int,

divine:
Hello,

I'll take a look tomorrow with a blank project to understand better this issue.

Thanks!

Issue #2002, opened 2020-03-24
[Question] belongsToMany, hybrid relation and detach


RaGreen:
I have two models (yep I noticed about belongsToMany and HybridRelation):
1. User - MySQL model with use HybridRelation and relation belongsToMany of Role
2. Role - MongoDB model without HybridRelation and relation belongsToMany of User

$user is attached to multiple ->roles(). Ids are stored in Role model only. 

I try to detach $role from $user via:
`$user->roles()->detach($role->_id)`

It throws an exception:
`BadMethodCallException: Call to undefined method App\Models\User::pull()` because Illuminate\Database\Eloquent\Model doesn't contain pull() method, when Jenssegers\Mongodb\Eloquent\Model has it.

Attach (`$user->roles()->attach($role->_id)`) works correctly. What i'm doing wrong?

Lumen: 6.0
Jenssegers/mongodb: 3.6

Smolevich:
@RaGreen , describe full version of laravel-mongodb, for example `3.6.1`

minhlong:
Same problem, so i use `saveMany` instead.

PR #2206, opened 2021-02-22
add a check to see if the model is MySQL or Mongo collection

NourAraar91:
right now when implementing the relationship Many-to-Many with MySql, all is working but when trying to detach from the MySQL model an error
BadMethodCallException with message 'Call to undefined method App\Models\User::pull()'
is showing because that MySQL model is not a collection and can't use pull on it

so I added a check to see if the model is a mongo collection or not and then will call the pull method
similar issues: #2002


divine:
Hello,

Tests are failing. Please investigate and let us know.

Thanks!

divine:
Closing due to inactivity, feel free to reply here if you are still willing to fix failing tests.

Thanks!

PR #2236, opened 2021-04-22
Add setDatabase to Jenssegers\Mongodb\Connection

ThomasWestrelin:
We have a use case where the database is determined dynamically and is not known in advance.
Previously, we used `DB::purge()` which wasn't ideal as this would cause the previous connection to be closed.

This change would allow us to change the database on demand.

### DIFF
--- a/src/Jenssegers/Mongodb/Connection.php
+++ b/src/Jenssegers/Mongodb/Connection.php
@@ -266,6 +266,16 @@ class Connection extends BaseConnection
     {
         return new Schema\Grammar();
     }
+    
+    /**
+     * Set database.
+     * @param \MongoDB\Database $db
+     */
+    public function setDatabase(\MongoDB\Database $db)
+	{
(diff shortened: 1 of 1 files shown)


Smolevich (approved):

divine:
@Smolevich should we backport this?

Thanks!

src/Jenssegers/Mongodb/Connection.php
```diff
@@ -266,6 +266,16 @@ protected function getDefaultSchemaGrammar()
...
     }
+    
+    /**
+     * Set database.
+     * @param \MongoDB\Database $db
+     */
+    public function setDatabase(\MongoDB\Database $db)
+	{
+		$this->db = $db;
+	}
```
divine, line 277: Can we keep tabs consistent?

Thanks

divine (changes requested):

PR #2239, opened 2021-04-28
Check dates against `DateTimeInterface` instead of `DateTime`

jeromegamez:
Greeting! 👋 

I noticed that `where()` queries in which I used `DateTimeImmutable` objects returned empty results.

This change would allow the usage of `DateTimeImmutable` or `CarbonImmutable` in addition to `DateTime`/`Carbon` 🤞 

:octocat:

PR #2246, opened 2021-04-30
Ignore local PHPUnit overrides

jeromegamez:
It's me again 👋

while working on #2239, I copied the `phpunit.xml.dist` to `phpunit.xml` so that I could use my locally installed MongoDB instance with different connection parameters.

Since the repo already has the `phpunit.xml.dist`, I'd like to suggest ignoring `phpunit.xml` so that local overrides can be applied without risking to commit them to the repo by accident 🤞.

Long story short: see PR title 😂

:octocat:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ composer.lock
 *.project
 .idea/
 .phpunit.result.cache
+phpunit.xml


divine (approved):

divine:
Thank you for your contribution @jeromegamez !

PR #2271, opened 2021-06-09
Fix nested dates format

alexsmko:
https://github.com/jenssegers/laravel-mongodb/issues/2129

### DIFF
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -216,7 +216,9 @@ abstract class Model extends BaseModel
         // Convert dot-notation dates.
         foreach ($this->getDates() as $key) {
             if (Str::contains($key, '.') && Arr::has($attributes, $key)) {
-                Arr::set($attributes, $key, (string) $this->asDateTime(Arr::get($attributes, $key)));
+                Arr::set($attributes, $key, $this->serializeDate(
+                    $this->asDateTime(Arr::get($attributes, $key))
+                ));
             }
         }
 
--- a/tests/ModelTest.php
+++ b/tests/ModelTest.php
@@ -577,7 +577,7 @@ class ModelTest extends TestCase
 
         $data = $user->toArray();
         $this->assertNotInstanceOf(UTCDateTime::class, $data['entry']['date']);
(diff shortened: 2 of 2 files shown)


divine:
Hello,

This looks like a breaking change for the projects depending on the old behavior, however, on the other hand, this is definitely a fix.

@Smolevich what do you think?

Thanks!

src/Eloquent/Model.php
```diff
@@ -216,7 +216,9 @@ public function attributesToArray()
         // Convert dot-notation dates.
         foreach ($this->getDates() as $key) {
```
shaedrich, line 217: Your change works for `EmbedsOne` but not for `EmbedsMany` since `Arr::has()`/`Arr:set()` doesn't support dot-syntax on nested arrays (`entry.*.date`). Interestingly, `array_get()`/`array_set()` do. So I'd add this to make it work for both `EmbedsOne` and `EmbedsMany`:
```suggestion
        foreach ($this->getDates() as $key) {
            $res = data_get($attributes, $key);
            
            if (is_array($res)) {
                $res = array_filter($res);
            }
            if (Str::contains($key, '.') && Arr::has($attributes, $key)) {
                Arr::set($attributes, $key, $this->serializeDate(
                    $this->asDateTime(Arr::get($attributes, $key))
                ));
            } else if (Str::contains($key, '.') && !empty($res)) {
                data_set($attributes, $key, $this->serializeDate(
                    $this->asDateTime(Arr::get($attributes, $key))
                ));
            }
        }
```

GromNaN:
Fixed by https://github.com/mongodb/laravel-mongodb/pull/3105

src/Eloquent/Model.php
shaedrich: @GromNaN Does #3105 support the dot syntax?

GromNaN: In fact, the query builder will convert any `DateTimeInterface` into `UTCDatetime`, and inversely for results. Regardless of the path, no need to specify any date casting. That the benefit of having a native date type in MongoDB vs SQL database that store strings for dates.

shaedrich: Awesome 👍🏻 Thanks for the clarification

PR #2318, opened 2021-09-24
fix: morphTo when relation name is in camel case

willtj:
Commit e19e10a in this PR updates existing tests to use a 'multi-word' relation for the morph relationship. This means that:
- `imageable` is changed to `has_image` when passed as the `$name` parameter to `morphOne()` and `morphMany()`
Note that snake case is required here - the parent Eloquent [`getMorphs()`](https://github.com/laravel/framework/blob/74ad9ece6ce88dd947452cc77284087cf354916a/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php#L717) method constructs the expected column names by appending '_id' and '_type' to the name that's passed.
- `imageable` is changed to `hasImage` when accessed as the relation name

That commit causes `RelationsTest::testMorph()` to fail, then commit c02e584 fixes the issue in line with how Eloquent's `morphTo()` method works, where `$name` variable defaults to the name of the calling function and is only snake-cased when passed to `getMorphs()` ([reference](https://github.com/laravel/framework/blob/74ad9ece6ce88dd947452cc77284087cf354916a/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php#L246))

### DIFF
--- a/src/Eloquent/HybridRelations.php
+++ b/src/Eloquent/HybridRelations.php
@@ -174,10 +174,10 @@ trait HybridRelations
         if ($name === null) {
             [$current, $caller] = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2);
 
-            $name = Str::snake($caller['function']);
+            $name = $caller['function'];
         }
 
-        [$type, $id] = $this->getMorphs($name, $type, $id);
+        [$type, $id] = $this->getMorphs(Str::snake($name), $type, $id);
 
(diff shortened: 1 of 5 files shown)


divine (commented):
Hello,

Honestly, this might break some applications which were depending on the "old" behavior even though it's incorrect/buggy.

We will probably keep this PR open until a new version with breaking changes will be ready.

Sorry, but no ETA currently.

Thanks!

divine:
This was merged in #2498.

Thanks!

PR #2330, opened 2021-12-18
Update composer.json

iamfarhad:


divine:
Hello,

Just why?

Thanks!

iamfarhad:
> 

Hi
Because We use major version(1.x) so ^1.0 is better than ^1.1.0

PR #2357, opened 2022-02-23
Check if failed log storage is disabled in Queue Service Provider

robjbrain:
Fixes this issue: https://github.com/jenssegers/laravel-mongodb/issues/2356

### DIFF
--- a/src/MongodbQueueServiceProvider.php
+++ b/src/MongodbQueueServiceProvider.php
@@ -4,6 +4,7 @@ namespace Jenssegers\Mongodb;
 
 use Illuminate\Queue\QueueServiceProvider;
 use Jenssegers\Mongodb\Queue\Failed\MongoFailedJobProvider;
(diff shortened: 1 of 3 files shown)


divine (changes requested):
Shouldn't be it something like this and return `NullFailedJobProvider`?
https://github.com/laravel/framework/blob/b9203fca96960ef9cd8860cb4ec99d1279353a8d/src/Illuminate/Queue/QueueServiceProvider.php#L238-L241

robjbrain:
You're right I hadn't spotted that in the original Laravel code

robjbrain (commented):
Updated to use Illuminate\Queue\Failed\NullFailedJobProvider

robjbrain:
@divine updated that now

divine:
This won't work as $config is undefined.

divine (approved):

divine:
Thanks!

Issue #2363, opened 2022-03-10
Attribute accessors not working when accessor method is one word


CraigMonva:
- Laravel-mongodb Version: 3.9.0
- PHP Version: 8.1.3
- Database Driver & Version: irrelevant (not a database issue)

### Description:
Attribute accessors not working when accessor method is one word when using laravel 9 Attributes convention: https://laravel.com/docs/9.x/eloquent-mutators#defining-an-accessor

From the docs:
This method name should correspond to the "camel case" representation of the true underlying model attribute / database column when applicable.

e.g

- `first_name` attribute should have a `firstName` method that returns `\Illuminate\Database\Eloquent\Casts\Attribute` which is then accessible by `$model->first_name`
- `name` attribute should have a `name` method that returns `\Illuminate\Database\Eloquent\Casts\Attribute` which is then accessible by `$model->name`

When extending laravel-mongodb Model class, If model has an attribute accessor method which is one word. i.e the `name` example from above, its being treated like a relationship.

I've done some logging and believe this line is the issue:
https://github.com/jenssegers/laravel-mongodb/blob/master/src/Eloquent/Model.php#L158

Potential fix is don't call `getRelationValue()` if the return type of the method is `\Illuminate\Database\Eloquent\Casts\Attribute` like what happens in base laravel model? https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L572

mrneatly:
Any progress here? I've came across the same issue. Been investigating this for hours as the cause of the problem was not obvious at all.

mrneatly:
Made a simple fix with help of @CraigMonva's hint. Let's hope it'll be reviewed soon.

GromNaN:
This was fixed in pull request #2577. Please upgrade to a version that includes that fix. If you are still experiencing this on a current release, please re-open.

PR #2394, opened 2022-05-28
Respect Laravel's new accessors

mrneatly:
Fix getting a value from `\Illuminate\Database\Eloquent\Casts\Attribute`-returning accessors (issue #2363).

### DIFF
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -155,8 +155,12 @@ abstract class Model extends BaseModel
         }
 
         // This checks for embedded relation support.
-        if (method_exists($this, $key) && ! method_exists(self::class, $key)) {
-            return $this->getRelationValue($key);
+        if (
+			method_exists($this, $key)
+			&& ! method_exists(self::class, $key)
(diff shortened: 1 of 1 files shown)


CraigMonva:
Whats the status of this PR?

mrneatly:
I've managed to fix the issue, but the cs-fixer denied my code for some reason. As the code in `php-cs-fixer` details doesn't give any clue, I would be very happy if anyone could explain me, what exactly I have to fix in order to have my pull request accepted :)

divine:
This is a breaking change and is this related to embedded relation?

mrneatly:
@divine this change isn't specifically related to an embedded relation as this case may also happen when using other types of relations. I'm not sure if it's really a breaking change because all this code does is performs an additional check whether requested attribute returns an `Attribute` object or not. As having accessors/mutators and relations of the same name within a single model is a bad idea anyway (the latter will be ignored in this case), I wouldn't consider it as such.

RosemaryOrchard (approved):

divine:
Closed in favor of #2438.

PR #2419, opened 2022-07-13
Typecast $column to string when ordering

apeisa:
Since `$column` parameter in [Illuminate/Database/Query/Builder.php](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Query/Builder.php#L2204) can take other types than strings, it makes sense to typecast this into a string.

This came up when using [Filament Admin panel](https://github.com/filamentphp/filament/) with Mongo, where orderBy doesn't work since the [$column](url) passed to orderBy isn't a string, but a Stringable object. There are probably other places also with a similar need and I will provide more PR:s as I find them.

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -512,7 +512,7 @@ class Builder extends BaseBuilder
         if ($column == 'natural') {
             $this->orders['$natural'] = $direction;
         } else {
-            $this->orders[$column] = $direction;
+            $this->orders[(string) $column] = $direction;
         }
 
         return $this;


divine:
Hello,

Tests are failing, please check and let me know.

Thanks!

apeisa:
Hi

Thanks for your comment. That [one test failing in Mongo v4.2](https://github.com/jenssegers/laravel-mongodb/runs/7319975603?check_suite_focus=true) cannot be related to this PR. For me, it looks like some kind of race condition since it is comparing timestamps and there is a one-second difference. So if possible, can you re-run the tests to see them pass?

apeisa:
Hi @divine 

I think the tests were re-run now and all passed.

divine:
Hello,

I've rerun tests but not sure how this could be the problem with that package as they state no support for this package https://github.com/filamentphp/filament/discussions/1365?

Can you add a test to see what values pass this package and do indeed solve the problem?

Thanks!

apeisa:
Hi

Thanks for the link. Filament works actually very well with Mongo, just few minor issues and this fixes the other one.

apeisa:
I will add tests for this.

apeisa:
Hi @divine! I created another PR (with better naming), which has tests for this also. You can find it from https://github.com/jenssegers/laravel-mongodb/pull/2420

PR #2420, opened 2022-07-16
[3.9] Fix/stringable sort

apeisa:
This PR replaces an [older one](https://github.com/jenssegers/laravel-mongodb/pull/2419). This is a safe fix, where we type cast column value to `string` when ordering. This is more inline in how the Eloquent method that is overwritten works too.

There is a simple test to show how you can now pass objects that implement `Stringable` as a `$column` variable for `orderBy` method.

I wasn't sure where to add that mockable class, so I did put it directly into QueryTest.php file.

### DIFF
--- a/tests/QueryTest.php
+++ b/tests/QueryTest.php
@@ -486,9 +486,9 @@ class QueryTest extends TestCase
  * Mockup class to test stringable objects.
  */
 class stringableObject implements Stringable {
-    private String $string;
+    private string $string;
 
-    public function __construct(String $string)
+    public function __construct(string $string)
     {
         $this->string = $string;
     }
--- a/tests/QueryTest.php
+++ b/tests/QueryTest.php
@@ -486,9 +486,9 @@ class QueryTest extends TestCase
  * Mockup class to test stringable objects.
  */
 class stringableObject implements Stringable {
-    private $string;
+    private String $string;
 
-    public function __construct($string)
+    public function __construct(String $string)
     {
         $this->string = $string;
     }
--- a/tests/QueryTest.php
+++ b/tests/QueryTest.php
@@ -241,7 +241,7 @@ class QueryTest extends TestCase
 
     public function testStringableOrder(): void
     {
-        $age = new stringableObject("age");
+        $age = new stringableObject('age');
 
         $user = User::whereNotNull('age')->orderBy($age, 'asc')->first();
         $this->assertEquals(13, $user->age);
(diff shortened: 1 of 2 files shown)


apeisa:
You can also see all tests passing from my own workflows here https://github.com/apeisa/laravel-mongodb/actions/runs/2681478769

apeisa:
@divine Is it possible to merge this?

tests/QueryTest.php
```diff
@@ -239,6 +239,17 @@ public function testOrder(): void
         $this->assertEquals(35, $user->age);
     }
 
+    public function testStringableOrder(): void
+    {
+        $age = new stringableObject('age');
```
divine, line 244:
```suggestion
        $age = str('age');
```

See @ https://laravel.com/docs/9.x/helpers#method-str

tests/QueryTest.php
```diff
@@ -470,3 +481,20 @@ public function testMultipleSortOrder(): void
...
+    public function __construct(string $string)
+    {
+        $this->string = $string;
+    }
+
+    public function __toString()
+    {
+        return $this->string;
+    }
+}
```
divine, line 500:
```suggestion
```

divine (changes requested):
Also, point me where exactly that library converts this to the stringable object.

Thanks!

apeisa:
Thanks for the suggestion, didn't know about the `str()` helper, makes the test much cleaner!

Not sure I understand your request:

> Also, point me where exactly that library converts this to the stringable object.

If you are referring to the Filament library - I don't think it matters (this fix is not about Filament only). I mean the Eloquent [orderBy](https://github.com/laravel/framework/blob/ac56c4f9a388855eb6bce4c0ba04ba8d559d3064/src/Illuminate/Database/Query/Builder.php#L2201-L2232) method that we are overriding allows multiple objects (in addition to string). Most of them are Queryable-objects, but it also accepts Illuminate\Database\Query\Expression -objects, which are stringable ones.

So we could use also `$age = new \Illuminate\Database\Query\Expression('age');` where we now have `str('age');` to make that point more clear.

So this fix simply make it possible to give any stringable object for `orderBy` method, instead of pure string. Of course, we could check also if the object is `Expression`, only then we do type cast to string, but I think this code is cleaner and more future-proof this way.

divine (approved):

apeisa:
@divine @Smolevich any hope to get this merged for the next release?

Smolevich (approved):

apeisa:
Thanks!

PR #2443, opened 2022-09-07
removing unused dependency illuminate events

devzer01:
issue is described here https://github.com/jenssegers/laravel-mongodb/issues/2442

i ran the tests and had to modify few things in terms of test configuration because the docker-compose was not giving a clear indication that tests passed because it was simply executing "php -a " as default from the php:cli container, this confusion for me took the longest than fixing the codel 

if you don't merge it i will add it to another packagist and publish the package, this is not a threat just simple dissapointment on how a popular opensource code has been managed, 

a major blocker to upgrade laravel 8 to 9 was cuased by this oversight, what would have been 15 minutes of a upgrade caused over 2 hours and over an hour out of the confusion regarding how phpunit was executed in docker-compose 

now next i am goin g to pushblish on packagist when you merge and if you merge thats upto you

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,3 +13,5 @@ RUN apt-get update && \
 COPY --from=composer /usr/bin/composer /usr/local/bin/composer
 
 WORKDIR /code
+
+CMD [ "phpunit"]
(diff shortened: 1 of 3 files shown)


devzer01:
also if you want to complain about how the new configuration is going to break from default setups then remove the config and update the repo later i don't have time to fix all your problems when your test setup was unclear

devzer01:
i am closing this PR as this is no longer mergable due to package name change, i will however when have time clear up the config a big and raise another PR

Issue #2451, opened 2022-09-21
Laravel 9.31.0 does not generate IDs anymore


fosron:
None of my models seem to generate mongodb IDs anymore. I'm trying to figure out what's happening, but to no avail at the moment.

```php
<?php

class Message extends \Jenssegers\Mongodb\Eloquent\Model
{ 
    use \Illuminate\Database\Eloquent\Factories\HasFactory, \Jenssegers\Mongodb\Eloquent\HybridRelations;
    protected $connection = 'mongodb';

    protected $collection = 'messages';
}
```

```php
$message = new Message();
$message->save();

$message->getKey();
// null

$message->getKeyName();
// _id
```

fosron:
I have to add this trait for the ID to generate, can not yet make sure if it works properly - like does it add it to relations, does Mongo get proper ObjectID type when data is inserted.

```
<?php

use Illuminate\Support\Str;
use MongoDB\BSON\ObjectId;

trait HasMongoIds
{
    /**
     * Generate a primary MongoID for the model.
     *
     * @return void
     */
    public static function bootHasMongoIds()
    {
        static::creating(function (self $model) {
            if ($model->getKeyName() === '_id' && !$model->getKey()) {
                $model->setAttribute('_id', new ObjectID);
            }
        });
    }
}
```

fosron:
Issue should be fixed in the next framework release https://github.com/laravel/framework/pull/44247

PR #2452, opened 2022-09-22
Fix Laravel 9.31 breaking update

The-Hasanov:
Issue #2451

### DIFF
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -82,6 +82,14 @@ abstract class Model extends BaseModel
     /**
      * @inheritdoc
      */
+    public function getIncrementing()
+    {
+        return $this->incrementing;
+    }
+
+    /**
+     * @inheritdoc
+     */
     public function fromDateTime($value)
     {
         // If the value is already a UTCDateTime instance, we don't need to parse it.


fosron:
@The-Hasanov thank you for your solution 🙇 

@Smolevich can we get this into a release soon? It's a breaking change for L9. Thanks 🙏

Smolevich (approved):

The-Hasanov:
> Hi all. First of all apologies that things broke for this library. I'd like to know more about what happened here. Why are string keys incremental for MongoDB?

Problem is here [https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Model.php#L1209](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Model.php#L1209)
so can see model not get mongo string id after create

driesvints:
> > Hi all. First of all apologies that things broke for this library. I'd like to know more about what happened here. Why are string keys incremental for MongoDB?
> 
> Problem is here https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Model.php#L1209 model not get string uuid after create

Heya. I realize that but given it's a string based ID, the ID should be set as an attribute up front. Is MongoDB not doing that? How are ID's generated now?

The-Hasanov:
i don't know about technical side. But id or _id is generated on mongo side automatically. Also this id is unique and incremental.

driesvints:
Thanks for that. I've sent in a PR to framework to revert this change: https://github.com/laravel/framework/pull/44247

alcaeus:
> Is MongoDB not doing that? How are ID's generated now?

The MongoDB server will add an `_id` field and generate an [ObjectId](https://www.mongodb.com/docs/manual/reference/bson-types/#std-label-objectid) for it if a document being inserted does not contain one. In addition to that, the MongoDB driver for PHP will also create an ObjectId when inserting a document without an `_id` field. If an application wants to use a different key (e.g. a UUID), this has to be created on the application side before passing the document on to the driver.

fosron:
Thank you @driesvints 🙇‍♂️

PR #2454, opened 2022-10-12
fix: whereBelongsTo

The-Hasanov:
this pr focusing fix whereBelongsTo() method.

Before:
`items.find({"items.user_id":{"$in":["..."]}},{"typeMap":{"root":"array","document":"array"}})`
After:
`items.find({"user_id":{"$in":["..."]}},{"typeMap":{"root":"array","document":"array"}})`

### DIFF
--- a/src/Relations/BelongsTo.php
+++ b/src/Relations/BelongsTo.php
@@ -73,10 +73,7 @@ class BelongsTo extends \Illuminate\Database\Eloquent\Relations\BelongsTo
         return 'whereIn';
     }
 
-    /**
-     * @return string
-     */
-    public function getQualifiedForeignKeyName()
+    public function getQualifiedForeignKeyName(): string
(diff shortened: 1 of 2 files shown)


divine (dismissed):

src/Relations/BelongsTo.php
```diff
@@ -72,4 +72,12 @@ protected function whereInMethod(EloquentModel $model, $key)
     {
         return 'whereIn';
     }
+
+    /**
+     * @return string
```
Smolevich, line 77: CI supports php > 8.0 and master use illuminate packages so you can use return type hints and delete doc blocks

Smolevich: @The-Hasanov can you do this?

Smolevich (approved):

PR #2461, opened 2022-11-10
Pass timeout in milliseconds

### LINKED ISSUE #2460: Builder timeout should be in milliseconds
### Description:
In `src/Query/Builder.php` line number 158, the time-out value should be in milliseconds but it's defined as seconds.
```
  /**
     * Set the cursor timeout in seconds.
     * @param int $seconds
     * @return $this
     */
    public function timeout($seconds)
    {
        // $this->timeout = $seconds; 
          $this->timeout = $seconds * 1000; // because on line 383, it's $options['maxTimeMS'] = $this->timeout;

        return $this;
    }

alcaeus:
Fixes #2460.

I decided to convert the value where the timeout is applied, as the property is public and users may have already passed a timeout in seconds to the property. If that is not how users are expected to use the `timeout` property, please let me know.

Testing this is a bit difficult, so I decided to use a command subscriber and inspect the command sent to the server to ensure we're sending the correct timeout value.

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -380,7 +380,7 @@ class Builder extends BaseBuilder
 
             // Apply order, offset, limit and projection
             if ($this->timeout) {
-                $options['maxTimeMS'] = $this->timeout;
+                $options['maxTimeMS'] = $this->timeout * 1000;
             }
             if ($this->orders) {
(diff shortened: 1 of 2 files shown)


divine (approved):
Thanks!

PR #1904, opened 2020-01-08
Add transaction support

klinson:
requires mongodb version V4.0 or more and deployment replica sets or sharded clusters
transaction support create/insert,update,delete,etc operation
Supports infinite-level nested transactions, but outside transaction rollbacks do not affect the commit of inside transactions
```
DB::beginTransaction();
DB::collection('users')->where('name', 'klinson')->update(['age' => 18]);
DB::transaction(function () {
    DB::collection('users')->where('name', 'mongodb')->update(['age' => 30]);
});
DB::rollBack();
```

PR #2465, opened 2022-11-14
Transaction support

alcaeus:
Note: contains changes from #2462 folded into a single commit; I'd prefer merging that PR before taking on this one.

This takes the work started by @klinson in #1904 and addresses some of the feedback left in that PR:

* Transactions can't be nested. This was done as MongoDB itself can't nest transactions. Running transactions in parallel instead of nested changes the outcome of nested operations that are partially rolled back, so this is a can of worms we better not open.
* Session options are no longer hard coded. Users can specify them when calling `DB::beginTransaction` or `DB::transaction` if they need to.
* The `DB::transaction` method will respect the `$attempts` argument and not invoke the callback more often, but may also abort before invoking the callback the specified number of times. This is due to the comparable logic in the MongoDB driver aborting after two minutes. I believe this is preferable to retrying for longer, especially in a web context (where two minutes is already quite long).

Note that I have not touched the docker-compose logic. Transaction tests are skipped unless run against a deployment that supports them, and I don't think it's necessary to set up a local cluster for quick tests. People who need to debug transaction tests locally should set up their own deployment and change the connection URL via the `MONGODB_URI` environment variable when running tests.

This PR was made possible by the contributions of @klinson and @levon80999 who've worked on this in the past. Commits retain co-authorship to reflect this.

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -57,7 +57,7 @@ jobs:
             -   uses: actions/checkout@v2
             -   name: Create MongoDB Replica Set
                 run: |
-                    docker run --name mongodb -p 27017:27017 -e MONGO_INITDB_DATABASE=unittest --detach mongo:${{ matrix.mongodb }} mongod --replSet rs
+                    docker run --name mongodb -p 27017:27017 -e MONGO_INITDB_DATABASE=unittest --detach mongo:${{ matrix.mongodb }} mongod --replSet rs --setParameter transactionLifetimeLimitSeconds=5
                     until docker exec --tty mongodb mongo 127.0.0.1:27017 --eval "db.runCommand({ ping: 1 })"; do
                     sleep 1
                     done
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -58,10 +58,10 @@ jobs:
             -   name: Create MongoDB Replica Set
                 run: |
                     docker run --name mongodb -p 27017:27017 -e MONGO_INITDB_DATABASE=unittest --detach mongo:${{ matrix.mongodb }} mongod --replSet rs
-                    until docker exec --tty mongodb mongo 127.0.0.1:27017 --eval "db.serverStatus()"; do
+                    until docker exec --tty mongodb mongo 127.0.0.1:27017 --eval "db.runCommand({ ping: 1 })"; do
                     sleep 1
                     done
-                    sudo docker exec --tty mongodb mongo 127.0.0.1:27017 --eval "rs.initiate({\"_id\":\"rs\",\"members\":[{\"_id\":0,\"host\":\"127.0.0.1:27017\" }]})"
+                    sudo docker exec --tty mongodb mongo 127.0.0.1:27017 --eval "rs.initiate()"
             -   name: "Installing php"
                 uses: shivammathur/setup-php@v2
                 with:
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -44,10 +44,6 @@ jobs:
                     - '8.0'
                     - '8.1'
         services:
-            mongo:
-                image: mongo:${{ matrix.mongodb }}
-                ports:
-                    - 27017:27017
             mysql:
                 image: mysql:5.7
                 ports:
@@ -59,6 +55,13 @@ jobs:
 
         steps:
             -   uses: actions/checkout@v2
+            -   name: Create MongoDB Replica Set
+                run: |
+                    docker run --name mongodb -p 27017:27017 -e MONGO_INITDB_DATABASE=unittest --detach mongo:${{ matrix.mongodb }} mongod --replSet rs
+                    until docker exec --tty mongodb mongo 127.0.0.1:27017 --eval "db.serverStatus()"; do
+                    sleep 1
+                    done
+                    sudo docker exec --tty mongodb mongo 127.0.0.1:27017 --eval "rs.initiate({\"_id\":\"rs\",\"members\":[{\"_id\":0,\"host\":\"127.0.0.1:27017\" }]})"
             -   name: "Installing php"
                 uses: shivammathur/setup-php@v2
                 with:
@@ -88,7 +91,7 @@ jobs:
                 run: |
                     ./vendor/bin/phpunit --coverage-clover coverage.xml
                 env:
-                    MONGODB_URI: 'mongodb://127.0.0.1/'
+                    MONGODB_URI: 'mongodb://127.0.0.1/?replicaSet=rs'
                     MYSQL_HOST: 0.0.0.0
                     MYSQL_PORT: 3307
             -   uses: codecov/codecov-action@v1
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -61,7 +61,7 @@ jobs:
                     until docker exec --tty mongodb mongo 127.0.0.1:27017 --eval "db.runCommand({ ping: 1 })"; do
                     sleep 1
                     done
-                    sudo docker exec --tty mongodb mongo 127.0.0.1:27017 --eval "rs.initiate()"
+                    sudo docker exec --tty mongodb mongo 127.0.0.1:27017 --eval "rs.initiate({\"_id\":\"rs\",\"members\":[{\"_id\":0,\"host\":\"127.0.0.1:27017\" }]})"
             -   name: "Installing php"
                 uses: shivammathur/setup-php@v2
                 with:
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -62,6 +62,9 @@ jobs:
                     sleep 1
                     done
                     sudo docker exec --tty mongodb mongo 127.0.0.1:27017 --eval "rs.initiate({\"_id\":\"rs\",\"members\":[{\"_id\":0,\"host\":\"127.0.0.1:27017\" }]})"
+            -   name: Show MongoDB server status
+                run: |
+                    docker exec --tty mongodb mongo 127.0.0.1:27017 --eval "db.runCommand({ serverStatus: 1 })"
             -   name: "Installing php"
                 uses: shivammathur/setup-php@v2
                 with:
--- a/README.md
+++ b/README.md
@@ -970,16 +970,14 @@ If you are familiar with [Eloquent Queries](http://laravel.com/docs/queries), th
 To see the available operations, check the [Eloquent](#eloquent) section.
 
 Transactions
--------
+------------
 Transactions require MongoDB version ^4.0 as well as deployment of replica set or sharded clusters. You can find more information [in the MongoDB docs](https://docs.mongodb.com/manual/core/transactions/)
 
 ### Basic Usage
 
-Transaction supports all operations.
-
 ```php
 DB::transaction(function () {
-    User::create(['name' => 'john', 'age' => 19, 'title' => 'admin', 'email' => 'klinsonup@gmail.com']);
+    User::create(['name' => 'john', 'age' => 19, 'title' => 'admin', 'email' => 'john@example.com']);
     DB::collection('users')->where('name', 'john')->update(['age' => 20]);
     DB::collection('users')->where('name', 'john')->delete();
 });
@@ -988,24 +986,32 @@ DB::transaction(function () {
 ```php
 // begin a transaction
 DB::beginTransaction();
-User::create(['name' => 'john', 'age' => 19, 'title' => 'admin', 'email' => 'klinsonup@gmail.com']);
+User::create(['name' => 'john', 'age' => 19, 'title' => 'admin', 'email' => 'john@example.com']);
 DB::collection('users')->where('name', 'john')->update(['age' => 20]);
 DB::collection('users')->where('name', 'john')->delete();
 
-// you can commit your changes
+// commit changes
 DB::commit();
-
-// you can also rollback them
-//DB::rollBack();
 ```
+
+To abort a transaction, call the `rollBack` method at any point during the transaction:
+```php
+DB::beginTransaction();
+User::create(['name' => 'john', 'age' => 19, 'title' => 'admin', 'email' => 'john@example.com']);
+
+// Abort the transaction, discarding any data created as part of it
+DB::rollBack();
+```
+
 **NOTE:** Transactions in MongoDB cannot be nested. DB::beginTransaction() function will start new transactions in a new created or existing session and will raise the RuntimeException when transactions already exist. See more in MongoDB official docs [Transactions and Sessions](https://www.mongodb.com/docs/manual/core/transactions/#transactions-and-sessions)
 ```php
-// This code will raise a RuntimeException
 DB::beginTransaction();
-    User::create(['name' => 'john', 'age' => 20, 'title' => 'admin']);
-    DB::beginTransaction()
-        DB::collection('users')->where('name', 'john')->update(['age' => 20]);
-    DB::commit()
+User::create(['name' => 'john', 'age' => 20, 'title' => 'admin']);
+
+// This call to start a nested transaction will raise a RuntimeException
+DB::beginTransaction();
+DB::collection('users')->where('name', 'john')->update(['age' => 20]);
+DB::commit();
 DB::rollBack();
 ```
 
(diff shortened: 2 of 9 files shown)


src/Concerns/ManagesTransactions.php
```diff
@@ -0,0 +1,105 @@
...
+     *
+     * In version 4.0, MongoDB supports multi-document transactions on replica sets.
+     * In version 4.2, MongoDB introduces distributed transactions, which adds support for multi-document transactions on sharded clusters and incorporates the existing support for multi-document transactions on replica sets.
+     *
+     * @see https://docs.mongodb.com/manual/core/transactions/
+     */
+    public function beginTransaction(array $options = []): void
+    {
+        $this->getSessionOrCreate()->startTransaction($options);
+        $this->transactions = 1;
```
jmikola, line 56: Where did this property come from? I don't recall seeing it in the previous PRs and it's not declared in this trait.

As-is, this looks like you're creating a public property at runtime ([deprecated in 8.2](https://php.watch/versions/8.2/dynamic-properties-deprecated)).

src/Query/Builder.php
```diff
@@ -720,17 +733,19 @@ public function from($collection, $as = null)
      */
     public function truncate(): bool
     {
-        $result = $this->collection->deleteMany([]);
+        $options = $this->inheritConnectionOptions();
+        $result = $this->collection->deleteMany($options);
```
jmikola, line 737: Subtle bug that would have gone unnoticed by type checking:

```suggestion
        $result = $this->collection->deleteMany([], $options);
```

tests/ConnectionTest.php
```diff
@@ -37,11 +37,101 @@ public function testDb()
...
+    public function testConnectionConfig(string $expectedUri, string $expectedDatabaseName, array $config): void
+    {
+        $connection = new Connection($config);
+        $client = $connection->getMongoClient();
+
+        $this->assertSame($expectedUri, (string) $client);
+        $this->assertSame($expectedDatabaseName, $connection->getMongoDB()->getDatabaseName());
+    }
+
+    public function testConnectionWithoutConfiguredDatabase(): void
```
jmikola, line 129: This likely answers my previous question about the purpose of `database` in the config struct.

tests/TestCase.php
```diff
@@ -56,8 +56,6 @@ protected function getEnvironmentSetUp($app)
         $app['config']->set('database.connections.mysql', $config['connections']['mysql']);
         $app['config']->set('database.connections.mongodb', $config['connections']['mongodb']);
         $app['config']->set('database.connections.mongodb2', $config['connections']['mongodb']);
```
jmikola, line 58: Are `database.connections.mysql` and `database.connections.mongodb2` actually used for anything?

tests/TransactionTest.php
```diff
@@ -0,0 +1,430 @@
...
+    {
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessage('There is no active session.');
+
+        DB::rollback();
+    }
+
+    private function getPrimaryServerType(): int
+    {
+        return DB::getMongoClient()->getManager()->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY))->getType();
```
jmikola, line 428: `composer.json` currently requires `mongodb/mongodb: ^1.11`, which requires `ext-mongodb: ^1.10`. If you'd like to make this more concise, you can bump the PHPLIB requirement to `^1.12` and call [`selectServer()`](https://www.php.net/manual/en/mongodb-driver-manager.selectserver.php) without any parameter for the same effect.

tests/config/database.php
```diff
@@ -1,35 +1,22 @@
...
+                'connectTimeoutMS'         => 100,
+                'serverSelectionTimeoutMS' => 250,
+            ],
         ],
 
         'mysql' => [
             'driver' => 'mysql',
             'host' => env('MYSQL_HOST', 'mysql'),
-            'port' => $mysqlPort,
+            'port' => env('MYSQL_PORT') ? (int) env('MYSQL_PORT') : 3306,
```
jmikola, line 19: This change seems unrelated to the PR.

tests/config/database.php
```diff
@@ -1,35 +1,22 @@
...
 return [
-
     'connections' => [
-
         'mongodb' => [
             'name' => 'mongodb',
             'driver' => 'mongodb',
-            'host' => $mongoHost,
+            'dsn' => env('MONGODB_URI', 'mongodb://127.0.0.1/'),
             'database' => env('MONGO_DATABASE', 'unittest'),
```
jmikola, line 9: Note my earlier suggestion to rename this to `MONGODB_DATABASE` for consistency.

tests/TransactionTest.php
```diff
@@ -0,0 +1,430 @@
...
+        $timesRun = 0;
+
+        DB::transaction(function () use (&$timesRun): void {
+            User::where(['name' => 'klinson'])->update(['age' => 21]);
+
+            // Update user during transaction to simulate a simultaneous update
+            DB::getCollection('users')->updateOne(['name' => 'klinson'], ['$inc' => ['age' => 2]]);
+
+            $timesRun++;
+        }, 2);
```
jmikola, line 366: I noted you said this [fails in CI](https://github.com/jenssegers/laravel-mongodb/actions/runs/3461141290/jobs/5778375858#step:11:23), but I'm having trouble following how this is executed. I understand that `DB::transaction()` uses a facade and is going to create an active session, but which of the operations within this callback end up using the same session (i.e. updating _within_ the transaction)?

.github/workflows/build-ci.yml
```diff
@@ -59,6 +55,13 @@ jobs:
 
         steps:
             -   uses: actions/checkout@v2
+            -   name: Create MongoDB Replica Set
+                run: |
+                    docker run --name mongodb -p 27017:27017 -e MONGO_INITDB_DATABASE=unittest --detach mongo:${{ matrix.mongodb }} mongod --replSet rs
```
jmikola, line 60: Remember to use `--setParameter transactionLifetimeLimitSeconds=5` (or a lower value) to minimize deadlock time in CI.

README.md
```diff
@@ -143,47 +144,36 @@ Keep in mind that these traits are not yet supported:
...
+Instead of using a connection string, you can also use the `host` and `port` configuration options to have the connection string created for you.
 
 ```php
 'mongodb' => [
     'driver' => 'mongodb',
-    'host' => ['server1', 'server2', ...],
-    ...
+    'host' => env('DB_HOST', '127.0.0.1'),
+    'port' => env('DB_PORT', 27017),
+    'database' => env('DB_DATABASE', 'homestead'),
```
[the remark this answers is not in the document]
alcaeus: The `database` option is used as the default database name for models that don't specify a database name. It is taken from the configuration as above, or extracted from the connection string if the option is missing.

README.md
```diff
@@ -979,6 +969,46 @@ If you are familiar with [Eloquent Queries](http://laravel.com/docs/queries), th
 ### Available operations
 To see the available operations, check the [Eloquent](#eloquent) section.
 
+Transactions
+-------
+Transactions require MongoDB version ^4.0 as well as deployment of replica set or sharded clusters. You can find more information [in the MongoDB docs](https://docs.mongodb.com/manual/core/transactions/)
+
+### Basic Usage
+
+Transaction supports all operations.
```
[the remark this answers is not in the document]
alcaeus: Agree, removed.

README.md
```diff
@@ -979,6 +969,46 @@ If you are familiar with [Eloquent Queries](http://laravel.com/docs/queries), th
...
+
+```php
+DB::transaction(function () {
+    User::create(['name' => 'john', 'age' => 19, 'title' => 'admin', 'email' => 'klinsonup@gmail.com']);
```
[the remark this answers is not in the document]
alcaeus: That is true, updated to use `john@example.com`.

README.md
```diff
@@ -979,6 +969,46 @@ If you are familiar with [Eloquent Queries](http://laravel.com/docs/queries), th
...
+DB::beginTransaction();
+User::create(['name' => 'john', 'age' => 19, 'title' => 'admin', 'email' => 'klinsonup@gmail.com']);
+DB::collection('users')->where('name', 'john')->update(['age' => 20]);
+DB::collection('users')->where('name', 'john')->delete();
+
+// you can commit your changes
+DB::commit();
+
+// you can also rollback them
+//DB::rollBack();
```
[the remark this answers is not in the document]
alcaeus: I've removed the `rollBack` call from this example and added a separate example on how to abort a transaction.

README.md
```diff
@@ -979,6 +969,46 @@ If you are familiar with [Eloquent Queries](http://laravel.com/docs/queries), th
...
+```
+**NOTE:** Transactions in MongoDB cannot be nested. DB::beginTransaction() function will start new transactions in a new created or existing session and will raise the RuntimeException when transactions already exist. See more in MongoDB official docs [Transactions and Sessions](https://www.mongodb.com/docs/manual/core/transactions/#transactions-and-sessions)
+```php
+// This code will raise a RuntimeException
+DB::beginTransaction();
+    User::create(['name' => 'john', 'age' => 20, 'title' => 'admin']);
+    DB::beginTransaction()
+        DB::collection('users')->where('name', 'john')->update(['age' => 20]);
+    DB::commit()
```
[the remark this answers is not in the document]
alcaeus: I fixed those and also moved the comment about a exceptions down to where the exception is raised.

src/Concerns/ManagesTransactions.php
alcaeus: This is declared in the original `Connection` class: https://github.com/laravel/framework/blob/213e07ef363bb6f2355254a5a9a0da33dcd692ad/src/Illuminate/Database/Connection.php#L127. I've re-declared the property in the trait to make it more explicit, especially since the trait could be used outside of the connection class context.

src/Query/Builder.php
alcaeus: Good catch, updated.

tests/TransactionTest.php
alcaeus: The first method call in the transactional closure updates the `user` collection within the transaction. The second call fetches the raw collection class, bypassing the internal transaction logic.

This is the same logic I'm using in the test above this one, with the small difference that in that test, the conflicting update is only done on the first execution of the callback, which means that the transaction will commit successfully after the second callback execution. In this test, I want to specifically cause the transaction to fail every time the callback is run to test whether the repetition limit is observed. This works fine locally but for some reason fails on CI.

tests/config/database.php
alcaeus: Good point, changed.

phpunit.xml.dist
```diff
@@ -35,9 +38,8 @@
         </testsuite>
     </testsuites>
     <php>
-        <env name="MONGO_HOST" value="mongodb"/>
+        <env name="MONGODB_URI" value="mongodb://127.0.0.1/" />
```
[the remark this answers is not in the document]
alcaeus: Changed to `MONGODB_DATABASE`.

src/Concerns/ManagesTransactions.php
```diff
@@ -0,0 +1,105 @@
...
+        return $session;
+    }
+
+    /**
+     * Use the existing or create new session and start a transaction in session.
+     *
+     * In version 4.0, MongoDB supports multi-document transactions on replica sets.
+     * In version 4.2, MongoDB introduces distributed transactions, which adds support for multi-document transactions on sharded clusters and incorporates the existing support for multi-document transactions on replica sets.
+     *
+     * @see https://docs.mongodb.com/manual/core/transactions/
```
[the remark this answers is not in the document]
alcaeus: Moved the `@see` to the trait itself and used your suggested wording as comment for `beginTransaction`.

src/Concerns/ManagesTransactions.php
```diff
@@ -0,0 +1,105 @@
...
+     * Commit transaction in this session.
+     */
+    public function commit(): void
+    {
+        $this->getSessionOrThrow()->commitTransaction();
+        $this->transactions = 0;
+    }
+
+    /**
+     * Rollback transaction in this session.
```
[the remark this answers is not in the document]
alcaeus: I've left the comments in place for the sake of completeness, but I did change the wording to use "Abort".

src/Connection.php
```diff
@@ -285,7 +289,7 @@ protected function getDefaultSchemaGrammar()
     /**
      * Set database.
      *
-     * @param \MongoDB\Database $db
```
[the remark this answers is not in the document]
alcaeus: Not sure why there's a mix of these, but I'd like to clean those up in a separate PR.

src/Connection.php
```diff
@@ -295,8 +299,8 @@ public function setDatabase(\MongoDB\Database $db)
     /**
      * Dynamically pass methods to the connection.
      *
-     * @param string $method
-     * @param array $parameters
+     * @param  string  $method
+     * @param  array  $parameters
```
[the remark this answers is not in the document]
alcaeus: StyleCI suggested all these changes - not sure why they weren't applied to begin with. Unfortunately I was "smart" enough to squash the changes into my original commit, but I can undo them if you'd like.

src/Eloquent/Model.php
```diff
@@ -402,7 +402,7 @@ public function getForeignKey()
     /**
      * Set the parent relation.
      *
-     * @param \Illuminate\Database\Eloquent\Relations\Relation $relation
+     * @param  \Illuminate\Database\Eloquent\Relations\Relation  $relation
```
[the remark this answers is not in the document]
alcaeus: See above - I'd clean this up in a separate PR.

src/Query/Builder.php
```diff
@@ -720,17 +733,19 @@ public function from($collection, $as = null)
...
 
         return 1 === (int) $result->isAcknowledged();
```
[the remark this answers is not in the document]
alcaeus: Deferring this to a separate cleanup PR.

tests/TransactionTest.php
alcaeus: It's a good thing that we've decided to align the version numbers of the library and the extension. The change to `selectServer()` was made in `ext-mongodb` 1.11, which is required as of `mongodb/mongodb` 1.10. Somehow the fact that we default to a primary read preference had escaped my memory. I've updated the call.

src/Connection.php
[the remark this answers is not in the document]
jmikola: No worries. If that's part of the repo it's fine. I suppose folks forgot to apply it to existing source files when integrating.

tests/config/database.php
jmikola: Probably cleaner to do that, and it shouldn't conflict with anything in this PR since the lines aren't adjacent.

tests/TransactionTest.php
jmikola: Are you running a single node replica set with the same server version (I don't think patch is relevant) locally?

alcaeus: This test was now fixed 🎉

alcaeus:
@Smolevich thanks for your patience! This PR is now ready for review. I'll make a couple of minor changes to the CI setup to ensure that any aborted or stale transaction doesn't block the CI build for an unnecessary period of time (30 seconds by default), but the functionality is in place and ready to go!

.github/workflows/build-ci.yml
```diff
@@ -59,6 +55,13 @@ jobs:
...
+                    until docker exec --tty mongodb mongo 127.0.0.1:27017 --eval "db.runCommand({ ping: 1 })"; do
```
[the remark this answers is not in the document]
alcaeus: Done - I added an additional build step for this to avoid people having to look through the entire setup output for this.

.github/workflows/build-ci.yml
alcaeus: Done.

src/Concerns/ManagesTransactions.php
```diff
@@ -93,11 +95,22 @@ public function transaction(Closure $callback, $attempts = 1, array $options = [
                 return;
             }
 
-            $callbackResult = $callback($this);
+            // Catch, store and re-throw any exception thrown during execution
```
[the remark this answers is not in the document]
alcaeus: One day, I'll stop forgetting Oxford commas...

divine (approved):
Thanks!

axgalache:
Any clue on when this feature may be released?

PR #2476, opened 2022-12-13
Support for search query to have \ as the last character in a LIKE clause

RichuJohn:
updated regex on query builder to support a search query with \ as last characters inside LIKE clause where search query is wrapped with %

Understood the alternator used here is to avoid replacing escaped %. So added another alternator that replaces any character that is immediately followed by a % and end of string. so the backslashes in the middle of the search query is left untouched as before.

Fixes https://github.com/jenssegers/laravel-mongodb/issues/2458 and https://github.com/jenssegers/laravel-mongodb/issues/2475

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -1043,7 +1043,7 @@ class Builder extends BaseBuilder
             }
 
             // Convert to regular expression.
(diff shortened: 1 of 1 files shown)


RichuJohn:
closing this as fix might leave out scenarios where search query is ending with escaped % in a LIKE operation which is not left bounded.

Issue #2233, opened 2021-04-13
Does laravel-mongodb has laravel sanctum support?


rhasani:
Hello, I am using laravel sanctum with mongodb.

but when I am runnung:
`$authToken = $user->createToken('auth-token')->plainTextToken();`

I am getting the following error:
`"message": "Call to a member function prepare() on null",
    "exception": "Error",
    "file": "/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Connection.php",
    "line": 465,
    "trace": [
        {
            "file": "/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Connection.php",
            "line": 671,
            "function": "Illuminate\\Database\\{closure}",
            "class": "Illuminate\\Database\\Connection",
            "type": "->"
        },`

and I think this is because of the lack of pre-built columns of personal_access_token table due to using mongodb.
is there any solution make it work with laravel sanctum?

birakdar:
in File  **vendor/laravel/sanctum/src/PersonalAccessToken.php** Replace 
use Illuminate\Database\Eloquent\Model; to
` use Jenssegers\Mongodb\Eloquent\Model;`

and write `protected $connection = 'mongodb' ;`


and in User Model don't forget to replace use Illuminate\Foundation\Auth\User as Authenticatable; 
to `use Jenssegers\Mongodb\Auth\User as Authenticatable;`

and write 
protected $connection = 'mongodb' ;
    protected $collection = 'users';

kudjieRaymond:
Create a custom PersonalAccessToken Model 
`use Laravel\Sanctum\Contracts\HasAbilities;
use Jenssegers\Mongodb\Eloquent\Model;

class PersonalAccessToken extends Model implements HasAbilities
{


    /**
     * The attributes that should be cast to native types.
     *
     * @var array
     */
    protected $casts = [
        'abilities' => 'json',
        'last_used_at' => 'datetime',
    ];

    /**
     * The attributes that are mass assignable.
     *
     * @var array
     */
    protected $fillable = [
        'name',
        'token',
        'abilities',
    ];

    /**
     * The attributes that should be hidden for serialization.
     *
     * @var array
     */
    protected $hidden = [
        'token',
    ];

    /**
     * Get the tokenable model that the access token belongs to.
     *
     * @return \Illuminate\Database\Eloquent\Relations\MorphTo
     */
    public function tokenable()
    {
        return $this->morphTo('tokenable');
    }

    /**
[...]

PR #2497, opened 2023-01-13
This trait will provide you correct implementation of Laravel/Sanctum…

malhanafi:
Fix issues #2233

This commit aims to fix: **Call to a member function prepare() on null** when using on [Laravel/Sanctum](https://github.com/Laravel/Sanctum).
Which occurs when calling `createToken` function.

So, to fix the issue, use `Jenssegers\Mongodb\Sanctum\HasApiTokens` rather than `Laravel\Sanctum\HasApiTokens`:
```php
use Jenssegers\Mongodb\Sanctum\HasApiTokens;

class User
{
    use HasApiTokens;
}
```

And then change default personal token to this package by `AuthServiceProvider`:

```php
use Laravel\Sanctum\Sanctum;
use Jenssegers\Mongodb\Sanctum\PersonalAccessToken;
//...

class AuthServiceProvider extends ServiceProvider
{
    public function boot()
    {
        //...
        Sanctum::usePersonalAccessTokenModel(PersonalAccessToken::class);
    }
}
```

### DIFF
--- a/src/Sanctum/HasApiTokens.php
+++ b/src/Sanctum/HasApiTokens.php
@@ -1,0 +1,78 @@
+<?php
+
+namespace Jenssegers\Mongodb\Sanctum;
+
+use Illuminate\Support\Str;
+use Jenssegers\Mongodb\Sanctum\Contracts\HasAbilities;
+
+trait HasApiTokens
+{
+    /**
+     * The access token the user is using for the current request.
+     *
+     * @var HasAbilities
+     */
+    protected $accessToken;
+
+    /**
+     * Get the access tokens that belong to model.
+     *
+     * @return \Illuminate\Database\Eloquent\Relations\MorphMany
+     */
+    public function tokens()
+    {
+        return $this->morphMany(PersonalAccessToken::class, 'tokenable');
+    }
+
+    /**
+     * Determine if the current API token has a given scope.
+     *
+     * @param string $ability
+     * @return bool
+     */
+    public function tokenCan(string $ability)
+    {
+        return $this->accessToken && $this->accessToken->can($ability);
+    }
+
+    /**
+     * Create a new personal access token for the user.
+     *
+     * @param string $name
+     * @param array $abilities
+     * @return NewAccessToken
+     */
+    public function createToken(string $name, array $abilities = ['*'])
+    {
+        $token = $this->tokens()->create([
+            'name' => $name,
+            'token' => hash('sha256', $plainTextToken = Str::random(40)),
+            'abilities' => $abilities,
+        ]);
+
+        return new NewAccessToken($token, $token->getKey() . '|' . $plainTextToken);
+    }
+
+    /**
+     * Get the access token currently associated with the user.
+     *
+     * @return HasAbilities
+     */
+    public function currentAccessToken()
+    {
+        return $this->accessToken;
+    }
+
+    /**
+     * Set the current access token for the user.
+     *
+     * @param HasAbilities $accessToken
+     * @return $this
+     */
+    public function withAccessToken($accessToken)
+    {
+        $this->accessToken = $accessToken;
+
+        return $this;
+    }
+}
--- a/README.md
+++ b/README.md
@@ -259,6 +259,18 @@ class User extends Model
 
 For more information check [Laravel Docs about Soft Deleting](http://laravel.com/docs/eloquent#soft-deleting).
 
+### [Laravel Sanctum](https://laravel.com/docs/9.x/sanctum)
+This trait will provide you correct implementation of Sanctum with this package. If your model is already using the `Laravel\Sanctum\HasApiTokens` trait, you may replace it with this `Jenssegers\Mongodb\Sanctum\HasApiTokens`
+
+```php
+use Jenssegers\Mongodb\Sanctum\HasApiTokens;
+
+class User
+{
+    use HasApiTokens;
+}
+```
+
 ### Guarding attributes
 
 When choosing between guarding attributes or marking some as fillable, Taylor Otwell prefers the fillable route.
--- a/README.md
+++ b/README.md
@@ -270,6 +270,21 @@ class User
     use HasApiTokens;
 }
 ```
+And then you have to set This package PersonalAccessToken as default to Sanctum in `AuthServiceProvider`
+```php
+use Laravel\Sanctum\Sanctum;
+use Jenssegers\Mongodb\Sanctum\PersonalAccessToken;
+//...
+
+class AuthServiceProvider extends ServiceProvider
+{
+    public function boot()
+    {
+        //...
+        Sanctum::usePersonalAccessTokenModel(PersonalAccessToken::class);
+    }
+}
+```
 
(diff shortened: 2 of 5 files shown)


src/Sanctum/HasApiTokens.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+namespace Jenssegers\Mongodb\Sanctum;
+
+use Illuminate\Support\Str;
+use Jenssegers\Mongodb\Sanctum\Contracts\HasAbilities;
+
+trait HasApiTokens
```
Smolevich, line 8: Let's add tests for this functionality

divine:
I'm against this type of feature where we just add a library from Laravel with one line changed and then should follow up on every change with upstream.

Just no.

SharonSpider:
how to get this feature? or @malhanafi  do you have any solution? thank you in advance.

malhanafi:
I'm very sorry, I was willing to support this issue and fix until it get accepted so no one will ever face this issue, but they aren't willing to accept it, so, please as I'm using now in every project I'm working on, is to create a directory in \App with name of Sanctum and create these three files inside it.

**HasApiTokens.php**
```php
namespace App\Sanctum;

use Illuminate\Support\Str;
use Laravel\Sanctum\Contracts\HasAbilities;

trait HasApiTokens
{
    /**
     * The access token the user is using for the current request.
     *
     * @var HasAbilities
     */
    protected $accessToken;

    /**
     * Get the access tokens that belong to model.
     *
     * @return \Illuminate\Database\Eloquent\Relations\MorphMany
     */
    public function tokens()
    {
        return $this->morphMany(PersonalAccessToken::class, 'tokenable');
    }

    /**
     * Determine if the current API token has a given scope.
     *
     * @param string $ability
     * @return bool
     */
    public function tokenCan(string $ability)
    {
        return $this->accessToken && $this->accessToken->can($ability);
    }

    /**
     * Create a new personal access token for the user.
     *
     * @param string $name
     * @param array $abilities
     * @param string|null $plainTextToken
     * @return NewAccessToken
     */
    public function createToken(string $name, array $abilities = ['*'], ?string $plainTextToken = null)
    {
        if ($plainTextToken == null) {
            $plainTextToken = Str::random(40);
        }
        $token = $this->tokens()->create([
            'name' => $name,
            'token' => hash('sha256', $plainTextToken),
            'abilities' => $abilities,
        ]);

        return new NewAccessToken($token, $token->getKey() . '|' . $plainTextToken);
    }

    /**
     * Get the access token currently associated with the user.
     *
     * @return HasAbilities
     */
    public function currentAccessToken()
    {
        return $this->accessToken;
    }

    /**
     * Set the current access token for the user.
     *
     * @param HasAbilities $accessToken
     * @return $this
     */
    public function withAccessToken($accessToken)
    {
        $this->accessToken = $accessToken;

        return $this;
    }
}

```

**NewAccessToken.php**
```php
namespace App\Sanctum;

use Illuminate\Contracts\Support\Arrayable;
use Illuminate\Contracts\Support\Jsonable;

class NewAccessToken implements Arrayable, Jsonable
{
    /**
     * The access token instance.
     *
     * @var PersonalAccessToken
     */
    public $accessToken;

    /**
     * The plain text version of the token.
     *
     * @var string
     */
    public $plainTextToken;

    /**
     * Create a new access token result.
     *
     * @param PersonalAccessToken $accessToken
     * @param string $plainTextToken
     * @return void
     */
    public function __construct(PersonalAccessToken $accessToken, string $plainTextToken)
    {
        $this->accessToken = $accessToken;
        $this->plainTextToken = $plainTextToken;
    }

    /**
     * Get the instance as an array.
     *
     * @return array
     */
    public function toArray()
    {
        return [
            'accessToken' => $this->accessToken,
            'plainTextToken' => $this->plainTextToken,
        ];
    }

    /**
     * Convert the object to its JSON representation.
     *
     * @param int $options
     * @return string
     */
    public function toJson($options = 0)
    {
        return json_encode($this->toArray(), $options);
    }
}

```

**PersonalAccessToken.php**
```php
namespace App\Sanctum;

use Jenssegers\Mongodb\Eloquent\Model;
use Laravel\Sanctum\Contracts\HasAbilities;

class PersonalAccessToken extends Model implements HasAbilities
{
    /**
     * The attributes that should be cast to native types.
     *
     * @var array
     */
    protected $casts = [
        'abilities' => 'json',
        'last_used_at' => 'datetime',
    ];

    /**
     * The attributes that are mass assignable.
     *
     * @var array
     */
    protected $fillable = [
        'name',
        'token',
        'abilities',
    ];

    /**
     * The attributes that should be hidden for serialization.
     *
     * @var array
     */
    protected $hidden = [
        'token',
    ];

    /**
     * Get the tokenable model that the access token belongs to.
     *
     * @return \Illuminate\Database\Eloquent\Relations\MorphTo
     */
    public function tokenable()
    {
        return $this->morphTo('tokenable');
    }

    /**
     * Find the token instance matching the given token.
     *
     * @param string $token
     * @return static|null
     */
    public static function findToken($token)
    {
        if (stripos($token, 'bearer ') === 0) {
            $token = trim(substr($token, strlen('bearer ')));
        }

        if (strpos($token, '|') === false) {
            return self::where('token', hash('sha256', $token))->first();
        }

        [$id, $token] = explode('|', $token, 2);

        if ($instance = self::find($id)) {
            return hash_equals($instance->token, hash('sha256', $token)) ? $instance : null;
        }

        return null;
    }

    /**
     * Determine if the token has a given ability.
     *
     * @param string $ability
     * @return bool
     */
    public function can($ability)
    {
        return in_array('*', $this->abilities) || array_key_exists($ability, array_flip($this->abilities));
    }

    /**
     * Determine if the token is missing a given ability.
     *
     * @param string $ability
     * @return bool
     */
    public function cant($ability)
    {
        return !$this->can($ability);
    }
}

```

Then in boot method which exists in **\App\Providers\AuthServiceProvider**.
Add this line.

```php
use Laravel\Sanctum\Sanctum;
use Jenssegers\Mongodb\Sanctum\PersonalAccessToken;
//...
class AuthServiceProvider extends ServiceProvider
{
    public function boot()
    {
        //...
        Sanctum::usePersonalAccessTokenModel(PersonalAccessToken::class);
    }
}
```

And everything should work as excepted.

Note: in multiple implementation Authorization header doesn't work as excepted, so this way should work.
Update **\App\Http\Middleware\Authenticate** override **unauthenticated** method to like so.

```php
namespace App\Http\Middleware;

use App\Sanctum\PersonalAccessToken;
use Illuminate\Auth\Middleware\Authenticate as Middleware;

class Authenticate extends Middleware
{
    protected function unauthenticated($request, array $guards)
    {
        $user = PersonalAccessToken::findToken($request->header('authorization'));
        if ($user != null) {
            $this->auth->guard()->setUser($user->tokenable);
            return;
        }
        parent::unauthenticated($request, $guards);
    }
}
```

This method called only when authorization has failed, so we did last check by authorization header, and it works as I expect.
Good luck.

Sincerely.

PR #2501, opened 2023-01-30
Laravel 10.x Compatibility

laravel-shift:
This is an automated pull request from [Shift](https://laravelshift.com) to update your package code and dependencies to be compatible with Laravel 9.x.

**Before merging**, you need to:

- Checkout the `l10-compatibility` branch
- Review **all** comments for additional changes
- Thoroughly test your package

If you do find an issue, please report it by commenting on this PR to help improve future automation.

PR #2376, opened 2022-03-28
Fix Query on whereDate, whereDay, whereMonth, whereYear

Davpyu:
Fix issues #2334 

So basically, current whereDate, whereDay, whereMonth and whereYear is query on different column with specific need (like whereDate you need column that has value YYYY-MM-DD etc) using basic comparison.

As we all know that isnt how whereDate works on sql, this PR will generate query using built in ```$dayOfMonth, $month, $year``` on whereDay, whereMonth, and whereYear, while whereDate basically generate range date(since mongodb doesnt have date() equivalent like on sql).

Example:
Collection on mongodb
```js
db.birthdays.insert([
    { '_id': 1, 'name': 'John Doe', 'birthday_at': new Date('2002-02-01') },
    { '_id': 2, 'name': 'Nash Doe', 'birthday_at': new Date('1992-06-20') },
    { '_id': 3, 'name': 'Jane Doe', 'birthday_at': new Date('2000-01-15') },
]);
```

Eloquent if this merged
```php
Birthday::whereDate('birthday_at', '2002-02-01')->get();
// generated query: birthdays.find({ 'birthday_at' => { '$gte' => '2002-02-01T00:00:00.000Z', '$lte' => '2002-02-01T23:59:59.999Z' } })

Birthday::whereDay('birthday_at', 15)->get();
// generated query: birthdays.find({ '$expr' => { '$eq' => [{ '$dayOfMonth' => 'birthday_at' }, 15] } })

Birthday::whereMonth('birthday_at', 6)->get();
// generated query: birthdays.find({ '$expr' => { '$eq' => [{ '$month' => 'birthday_at' }, 6] } })

Birthday::whereYear('birthday_at', 2002)->get();
// generated query: birthdays.find({ '$expr' => { '$eq' => [{ '$year' => 'birthday_at' }, 2002] } })

```

Note:
1. whereTime isnt possible to fix rn since i cant find logic to query like time() on sql

PR #2506, opened 2023-02-09
Laravel 10 compatibility

alcaeus:
This PR replaces #2501 and adapts the library to accommodate the BC breaks introduced in Laravel 10.

The most impactful change is the handling of dates, which requires people to update their models to no longer use the `$dates` property, but instead define a cast for date fields in the `$casts` property (except for the `created_at` and `updated_at` fields).

Note that we will use this BC break as an opportunity to remove auto-casting for `ObjectId` and `Binary` instances, which has been previously planned for version 4.0 of this library. This will require people that want to use auto-casting for identifiers and binary fields to update the `$casts` property accordingly. Since this is a significant BC break (along with other BC breaks introduced by Laravel), we may decide to release Laravel 10 support in a new major version. This will not necessarily all of the changes previously prepared for 4.0 to accelerate the timeline for Laravel 10 support.

Any testing users can do with Laravel 10 to alert us to potential problems with this library are highly appreciated; we value your feedback to let us know what we need to fix or can improve.

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,8 +256,6 @@ use Jenssegers\Mongodb\Eloquent\SoftDeletes;
 class User extends Model
 {
     use SoftDeletes;
-
-    protected $dates = ['deleted_at'];
 }
 ```
 
@@ -279,7 +277,7 @@ use Jenssegers\Mongodb\Eloquent\Model;
 
 class User extends Model
 {
-    protected $dates = ['birthday'];
+    protected $casts = ['birthday' => 'datetime'];
 }
 ```
 
--- a/src/Auth/PasswordResetServiceProvider.php
+++ b/src/Auth/PasswordResetServiceProvider.php
@@ -7,29 +7,6 @@ use Illuminate\Auth\Passwords\PasswordResetServiceProvider as BasePasswordResetS
 class PasswordResetServiceProvider extends BasePasswordResetServiceProvider
 {
     /**
-     * Register the token repository implementation.
-     *
-     * @return void
-     */
-    protected function registerTokenRepository()
-    {
-        $this->app->singleton('auth.password.tokens', function ($app) {
-            $connection = $app['db']->connection();
-
-            // The database token repository is an implementation of the token repository
-            // interface, and is responsible for the actual storing of auth tokens and
-            // their e-mail addresses. We will inject this table and hash key to it.
-            $table = $app['config']['auth.password.table'];
-
-            $key = $app['config']['app.key'];
-
-            $expire = $app['config']->get('auth.password.expire', 60);
-
-            return new DatabaseTokenRepository($connection, $table, $key, $expire);
-        });
-    }
-
-    /**
      * @inheritdoc
      */
     protected function registerPasswordBroker()
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -187,7 +187,7 @@ abstract class Model extends BaseModel
     /**
      * @inheritdoc
      */
-    public function setAttribute($key, $value)
+    public function setAttribute($key, $value): static
     {
         // Convert _id to ObjectID.
         if ($key == '_id' && is_string($value)) {
@@ -516,13 +516,9 @@ abstract class Model extends BaseModel
     }
 
     /**
-     * Add the casted attributes to the attributes array.
-     *
-     * @param  array  $attributes
-     * @param  array  $mutatedAttributes
-     * @return array
+     * @inheritdoc
      */
-    protected function addCastAttributesToArray(array $attributes, array $mutatedAttributes)
+    protected function addCastAttributesToArray(array $attributes, array $mutatedAttributes): array
     {
         foreach ($this->getCasts() as $key => $castType) {
             if (! Arr::has($attributes, $key) || Arr::has($mutatedAttributes, $key)) {
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -2,18 +2,23 @@
 
 namespace Jenssegers\Mongodb\Eloquent;
 
+use function array_key_exists;
 use DateTimeInterface;
+use function explode;
 use Illuminate\Contracts\Queue\QueueableCollection;
 use Illuminate\Contracts\Queue\QueueableEntity;
+use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Database\Eloquent\Model as BaseModel;
 use Illuminate\Database\Eloquent\Relations\Relation;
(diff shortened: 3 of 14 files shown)


src/Eloquent/Model.php
```diff
@@ -191,13 +196,14 @@ public function setAttribute($key, $value)
...
+            // Store to a temporary key, then move data to the actual key
+            $uniqueKey = uniqid($key);
+            parent::setAttribute($uniqueKey, $value);
 
-            Arr::set($this->attributes, $key, $value);
+            Arr::set($this->attributes, $key, $this->attributes[$uniqueKey] ?? null);
+            unset($this->attributes[$uniqueKey]);
 
-            return;
+            return $this;
```
Smolevich, line 206: Let's add return type hint if return will change

alcaeus: Good point, done.

src/Eloquent/Model.php
```diff
@@ -515,4 +514,62 @@ public function __call($method, $parameters)
...
     }
+
+    /**
+     * Add the casted attributes to the attributes array.
+     *
+     * @param  array  $attributes
+     * @param  array  $mutatedAttributes
+     * @return array
+     */
+    protected function addCastAttributesToArray(array $attributes, array $mutatedAttributes)
```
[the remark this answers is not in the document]
alcaeus: Done, also replaced the docblock with `@inheritdoc` for brevity.

src/Auth/PasswordResetServiceProvider.php
```diff
@@ -6,29 +6,6 @@
 
 class PasswordResetServiceProvider extends BasePasswordResetServiceProvider
 {
-    /**
```
divine, line 9: Do we need this file at all now? AFAIK it was used only for this function to work properly.

src/Eloquent/Model.php
```diff
@@ -191,13 +196,14 @@ public function setAttribute($key, $value)
             $value = $builder->convertKey($value);
         } // Support keys in dot notation.
         elseif (Str::contains($key, '.')) {
```
divine, line 198: Well, if we do remove dot notation support in dates, we need to remove this here as well.

src/Eloquent/Model.php
```diff
@@ -222,13 +228,6 @@ public function attributesToArray()
             }
         }
 
-        // Convert dot-notation dates.
```
divine, line 225: This was discussed and the decision was to keep dot notation support, anyways I'd prefer removing it honestly but some discussion is needed. @Smolevich what do you think?

tests/AuthTest.php
```diff
@@ -10,55 +9,56 @@ public function tearDown(): void
...
-        DB::collection('password_reminders')->truncate();
+        DB::collection('password_reset_tokens')->truncate();
     }
 
     public function testAuthAttempt()
     {
         User::create([
             'name' => 'John Doe',
-            'email' => 'john@doe.com',
+            'email' => 'john.doe@example.com',
```
divine, line 19: Is there any reason why this was test changed?

divine (commented):
Honestly, we probably will need to separate each BC with a new PR.

README.md
```diff
@@ -279,7 +277,7 @@ use Jenssegers\Mongodb\Eloquent\Model;
 
 class User extends Model
 {
-    protected $dates = ['birthday'];
+    protected $casts = ['birthday' => 'datetime'];
```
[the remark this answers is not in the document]
alcaeus: This is a BC break made by Laravel - I would highlight this in the release notes and point out that this is the same change that is required for all other models.

src/Auth/PasswordResetServiceProvider.php
alcaeus: I'll check that.

src/Eloquent/Model.php
alcaeus: This is still supported, in fact this PR expands support for dot notation. Since dates are now handled via a cast, they will be covered by `addCastAttributesToArray`. This method was added to this class to support dot-notation for any cast instead of only magically casting date fields. In the future, it will be possible to have something like this:
```php
protected $casts = [
    'entry.date' => 'datetime',
    'entry.list' => 'array',
];
```

src/Eloquent/Model.php
alcaeus: Dot notation support wasn't removed, see below.

tests/AuthTest.php
alcaeus: It is not a required change, but just something I made a habit of: `@example.com` was explicitly reserved as a domain for testing purposes. This helps guard against accidental misconfiguration in the test environment sending emails to live domains.

Aside: this is a big problem in Germany, where lots of developers use `@test.de` as test email addresses, not realising that they are actually using the domain of "[Stiftung Warentest](https://en.wikipedia.org/wiki/Stiftung_Warentest)" to test their emails. This has led to some companies getting cease and desist notices from Stiftung Warentest because of abuse of their email system.

src/Auth/PasswordResetServiceProvider.php
alcaeus: It is still needed: this class ensures that the `auth.password` service is our own `PasswordBrokerManager` instance, which creates the MongoDB `DatabaseTokenRepository` instead of the default one. That has always been that way, and I couldn't find any usage of this method, which is why I removed it.

divine (approved):

divine:
Hello,

Since this PR affects dates maybe it will make sense to merge this PR #2376 here as well?

Thanks!

alcaeus:
I'm not familiar with that PR and will have to check that separately, but it isn't near the top of my list at the moment.

src/Eloquent/Model.php
```diff
@@ -515,4 +514,58 @@ public function __call($method, $parameters)
...
+            if ($castValue !== null && in_array($castType, ['date', 'datetime', 'immutable_date', 'immutable_datetime'])) {
+                $castValue = $this->serializeDate($castValue);
+            }
+
+            if ($castValue !== null && ($this->isCustomDateTimeCast($castType) ||
+                    $this->isImmutableCustomDateTimeCast($castType))) {
+                $castValue = $castValue->format(explode(':', $castType, 2)[1]);
+            }
+
+            if ($castValue instanceof DateTimeInterface &&
```
Smolevich, line 549: https://www.php-fig.org/psr/psr-2/ is deprecated and not fully described all rules with if conditions, i suggest to create task for using psr-12

Smolevich (approved):

PR #2507, opened 2023-02-09
Report package version when establishing connection

alcaeus:
This adds this package name to the driver info, which is reported as part of the [driver handshake metadata](https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.rst#client-driver-name). Server administrators can use this to get information on connecting applications and their drivers.

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -2,16 +2,21 @@
 
 namespace Jenssegers\Mongodb;
 
+use function class_exists;
+use Composer\InstalledVersions;
 use Illuminate\Database\Connection as BaseConnection;
 use Illuminate\Support\Arr;
 use InvalidArgumentException;
 use Jenssegers\Mongodb\Concerns\ManagesTransactions;
 use MongoDB\Client;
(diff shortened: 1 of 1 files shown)


src/Connection.php
```diff
@@ -169,6 +174,11 @@ protected function createConnection($dsn, array $config, array $options): Client
             $driverOptions = $config['driver_options'];
         }
 
+        $driverOptions['driver'] = [
+            'name' => 'laravel-mongodb',
+            'version' => self::getVersion(),
```
divine, line 179: Is this affects performance in any case?

alcaeus: Not that I'm aware of. This is composer functionality and it is called once per request if a connection is established. Composer itself optimises this to a significant degree, so I wouldn't consider it performance critical. We've been using this both in the PHP library as well as the Doctrine MongoDB ODM without any reports of performance degradation.

divine (approved):

alcaeus:
That sounds good. Thanks!

PR #2511, opened 2023-02-16
Use correct package name when looking up version

### LINKED ISSUE #2510: Package "jenssegers/laravel-mongodb" is not installed in ^3.9
- Laravel-mongodb Version: ^3.9
- PHP Version: 8.1
- Database Driver & Version: 1.15, pecl install

### Description: 
I get a laravel exception:
"Package "jenssegers/laravel-mongodb" is not installed"

image

its worth noting, the package name in composer is "jenssegers/mongodb" where the git repo is "jenssegers/laravel-mongodb" I'm pretty sure something got swapped around.

### Steps to reproduce
1. composer require jenssegers/mongodb "^3.9"
3. try to query a model

### Expected behaviour
The plugin should work without exception

### Actual behaviour
The plugin throws a module not found error

alcaeus:
Fixes #2510.


Smolevich:
@alcaeus master branch has changes with laravel 10, maybe to change destination branch from master to 3.9 ?

src/Connection.php
```diff
@@ -328,7 +328,7 @@ private static function lookupVersion(): string
     {
         if (class_exists(InstalledVersions::class)) {
             try {
-                return self::$version = InstalledVersions::getPrettyVersion('jenssegers/laravel-mongodb');
+                return self::$version = InstalledVersions::getPrettyVersion('jenssegers/mongodb');
```
Smolevich, line 331: I suggest don't use assignment in return block, and separate operation assignment and return value

alcaeus:
@Smolevich that's a good point, but there is no 3.9 branch. If you could create one based on the latest 3.9 release and push it I'll rebase and re-target this PR.

Smolevich:
> @Smolevich that's a good point, but there is no 3.9 branch. If you could create one based on the latest 3.9 release and push it I'll rebase and re-target this PR.

Let's do it

alcaeus:
@Smolevich I can't push the branch myself, but once it's pushed this PR can be changed to target 3.9. I've rebased on top of `v3.9.4` and dropped all commits related to Laravel 10 support that should not be in 3.9.

src/Connection.php
alcaeus: Updated the code accordingly 👍

divine:
We need to update the master branch as well.

PR #2513, opened 2023-02-18
split sql from mongodb model

abbasudo:
this will allow developers to switch between MongoDB and SQL database models easily.
by checking the configuration option in the Laravels database configuration file, we can specify which model class to use for the Developer model.

the developer doesn't care about the default database driver. he extends MongoDB model in his model. MongoDB model on the other hand has to decide which model to use based on the default database.
this way the developer doesn't have to change the extended model in his model if he decides to change the driver for any reason.


src/Eloquent/Model.php
```diff
@@ -2,570 +2,19 @@
...
-use Jenssegers\Mongodb\Query\Builder as QueryBuilder;
-use MongoDB\BSON\Binary;
-use MongoDB\BSON\ObjectID;
-use MongoDB\BSON\UTCDateTime;
-use function uniqid;
 
-abstract class Model extends BaseModel
-{
-    use HybridRelations, EmbedsRelations;
+$connection = config('database.default');
```
Smolevich, line 7: Why don't use eloquent model for MySQL for example directly, this package is created for work with mongodb, not different drivers

llabbasmkhll: this way the package acts like a plugin to add the ability for the developer to use MongoDB.
think of it as a laravel with MongoDB compatibility.

alcaeus:
I'm a strong 👎 on this PR, mainly because it is based on completely wrong assumptions.

First of all, there is this assumption:
> this will allow developers to switch between MongoDB and SQL database models easily.

This is not how you should build applications. Migrating between MongoDB and a relational database is no easy task that can be accomplished with a single change of the model. A model for a typical relational database does not translate to a model for MongoDB 1:1. In relational models, you often need foreign key references to model relationships, whereas in MongoDB you have other tools. Consider the simple example of a list of tags associated with a blog post: while you'd need to extract the list of tags to a separate table in a relational database to fulfil normalisation requirements, in MongoDB you'd store a list of tags in the original record and be done with it. Same with "richer" data, where in MongoDB you'd embed an entire document whereas you'd have to extract data in a relational database because you'd want "atomic values" (exceptions apply, but more on those later).

> by checking the configuration option in the Laravels database configuration file, we can specify which model class to use for the Developer model.

This is a horrible approach at solving the problem above. You're looking at the default database adapter and changing the behaviour of the class, and all of it to serve a use case that isn't one. I'm a big believer in using the right tool for the job, which includes using the right database. In some applications this is only a relational database, in others this is only MongoDB, but in many this will be a combination of the two. This is perfectly fine, and TBH the way this package integrates the two (by having hybrid references that point to models in relational databases) is better than a number of other approaches I've seen. This solution does nothing but complicate that.

> the developer doesn't care about the default database driver.

The developer should start caring about the default database driver. Remember, while SQL may be standardised across databases, features aren't. It's a big difference whether I use MySQL or PostgreSQL, because the latter allows me to build difference schemas (e.g. by using array fields, composite types, built-in validation constraints, and much more). The same goes for MongoDB: your schemas are designed completely different from relational databases. So while the developer may not care about the database, this is not what a library like Eloquent is there to solve. It's designed to make working with model objects easier, not to hide your database from you.

> he extends MongoDB model in his model. MongoDB model on the other hand has to decide which model to use based on the default database. this way the developer doesn't have to change the extended model in his model if he decides to change the driver for any reason.

There is another reason for why you'll want a different model: MongoDB offers an entirely different query engine. It doesn't understand SQL (some tools do, but that's outside of the scope here), but if understands aggregation pipeline which SQL databases don't understand. There are big advantages to both, and knowing how to create optimised queries for your database is a skill that engineers should have. Writing efficient queries for a database also requires you to know what database you are using, so hiding that from the engineer isn't a good call.

While I do want to make it easier to integrate MongoDB into Laravel applications, this absolutely isn't the way to go. We don't want to limit an ORM like Eloquent to only offer a subset of features that's common to every database it supports, but we rather want engineers to be able to take advantages of the features each database brings, and allow them (or rather, require them), to choose the database based on their use-case.

abbasudo:
@alcaeus Thank you for providing such detailed information. I had the idea of creating an laravel app that would offer real-time benchmarks for both MySQL and MongoDB. Initially, I thought it would be a good idea to propose this to the package maintainers themselves. However, after learning from your explanation, I understand that it is not possible to add this kind of functionality to the core database packages.

Nonetheless, I appreciate your guidance, and please let me know if you have any further suggestions or insights.

PR #2514, opened 2023-02-19
chore: improve docker tests

divine:
Currently it's not possible to run test when running `docker-compose up` as the test container just stops silently and it's possible to run it again.

After this merge it would be possible to run tests just by starting up the container and it'll just work.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,11 @@ COPY --from=composer:${COMPOSER_VERSION} /usr/bin/composer /usr/local/bin/compos
 
 WORKDIR /code
 
-COPY . .
+COPY composer.* ./
+
+RUN composer install
+
+COPY ./ ./
 
 RUN composer install
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
 ARG PHP_VERSION=8.1
 ARG COMPOSER_VERSION=2.5.4
 
-FROM composer:${COMPOSER_VERSION}
 FROM php:${PHP_VERSION}-cli
 
 RUN apt-get update && \
@@ -10,7 +9,7 @@ RUN apt-get update && \
     pecl install xdebug && docker-php-ext-enable xdebug && \
     docker-php-ext-install -j$(nproc) pdo_mysql zip
 
-COPY --from=composer /usr/bin/composer /usr/local/bin/composer
+COPY --from=composer:${COMPOSER_VERSION} /usr/bin/composer /usr/local/bin/composer
 
 WORKDIR /code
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -13,3 +13,9 @@ RUN apt-get update && \
 COPY --from=composer /usr/bin/composer /usr/local/bin/composer
 
 WORKDIR /code
+
+COPY . .
```
Smolevich, line 17: Better for optimal using docker cache 
```
COPY composer.* ./
RUN composer install
COPY ./ ./
```

Dockerfile
```diff
@@ -1,5 +1,5 @@
-ARG PHP_VERSION=8.0
-ARG COMPOSER_VERSION=2.0
+ARG PHP_VERSION=8.1
+ARG COMPOSER_VERSION=2.5.4
 
 FROM composer:${COMPOSER_VERSION}
```
Smolevich, line 4: Better delete and use in COPY section where binary file is copied

docker-compose.yml
```diff
@@ -3,6 +3,8 @@ version: '3'
 services:
     tests:
         container_name: tests
+        #platform: linux/arm64
```
Smolevich, line 6: Docker-compose cli supports option platform https://github.com/docker/compose/pull/8147

divine: That's a good point, thanks!

Dockerfile
```diff
@@ -1,8 +1,5 @@
-ARG PHP_VERSION=8.0
-ARG COMPOSER_VERSION=2.0
-
-FROM composer:${COMPOSER_VERSION}
-FROM php:${PHP_VERSION}-cli
+FROM composer:2.5.4
```
Smolevich, line 1: Why did you delete `ARG` ?

Dockerfile
```diff
@@ -13,3 +13,9 @@ RUN apt-get update && \
 COPY --from=composer /usr/bin/composer /usr/local/bin/composer
```
Smolevich, line 13: Better change on 
```
COPY --from=composer:${COMPOSER_VERSION} /usr/bin/composer /usr/local/bin/composer
``` 
And delete first `FROM`

Smolevich:
Add minimal description, please
I understand your motivation but another contributors don't understand reason of your changes

divine:
> Add minimal description, please I understand your motivation but another contributors don't understand reason of your changes

I've added description,

Thank you for your review.

Dockerfile
Smolevich: @divine can you change ?

divine: @Smolevich done.

Smolevich (approved):

PR #2515, opened 2023-02-19
feat!: use mongodb function to check for dirtiness

divine:
This is rework of test for PR #1990

> To check if a field is dirty we can simply use fromPHP() function of the mongodb driver. Besides, we shouldn't use the SQL-specific logic of Illuminate\Eloquent\Model::originalIsEquivalent() for mongodb. In sql, fields keep their schema data types and changing integer 1 to string '1' actually does nothing. However, in mongodb fields accept their types as they are inserted/updated. Therefore, if we change value of a field from (int)1 to (string)'1' we actually must regard it as dirty. This change is backward incompatible. Don't know if anyone likes the current behaviour more.
> --

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ All notable changes to this project will be documented in this file.
 - Change signature of `Query\Builder::__constructor` to match the parent class [#26](https://github.com/GromNaN/laravel-mongodb-private/pull/26) by [@GromNaN](https://github.com/GromNaN).
 - Fix Query on `whereDate`, `whereDay`, `whereMonth`, `whereYear`, `whereTime` to use MongoDB operators [#2570](https://github.com/jenssegers/laravel-mongodb/pull/2376) by [@Davpyu](https://github.com/Davpyu) and [@GromNaN](https://github.com/GromNaN).
 - `Model::unset()` does not persist the change. Call `Model::save()` to persist the change [#2578](https://github.com/jenssegers/laravel-mongodb/pull/2578) by [@GromNaN](https://github.com/GromNaN).
-- Use mongodb function to check for dirtiness [#2515](https://github.com/jenssegers/laravel-mongodb/pull/2515) by [@halaei](https://github.com/halaei) and [@GromNaN](https://github.com/GromNaN).
+- Use `MongoDB\BSON\fromPHP` to detect changes [#2515](https://github.com/jenssegers/laravel-mongodb/pull/2515) by [@halaei](https://github.com/halaei) and [@divine](https://github.com/divine).
 
 ## [3.9.2] - 2022-09-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ All notable changes to this project will be documented in this file.
 - Change signature of `Query\Builder::__constructor` to match the parent class [#26](https://github.com/GromNaN/laravel-mongodb-private/pull/26) by [@GromNaN](https://github.com/GromNaN).
 - Fix Query on `whereDate`, `whereDay`, `whereMonth`, `whereYear`, `whereTime` to use MongoDB operators [#2570](https://github.com/jenssegers/laravel-mongodb/pull/2376) by [@Davpyu](https://github.com/Davpyu) and [@GromNaN](https://github.com/GromNaN).
 - `Model::unset()` does not persist the change. Call `Model::save()` to persist the change [#2578](https://github.com/jenssegers/laravel-mongodb/pull/2578) by [@GromNaN](https://github.com/GromNaN).
+- Use mongodb function to check for dirtiness [#2515](https://github.com/jenssegers/laravel-mongodb/pull/2515) by [@halaei](https://github.com/halaei) and [@GromNaN](https://github.com/GromNaN).
 
 ## [3.9.2] - 2022-09-01
 
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -16,6 +16,7 @@ use Illuminate\Support\Str;
 use function in_array;
 use Jenssegers\Mongodb\Query\Builder as QueryBuilder;
 use MongoDB\BSON\Binary;
+use function MongoDB\BSON\fromPHP;
 use MongoDB\BSON\ObjectID;
 use MongoDB\BSON\UTCDateTime;
 use function uniqid;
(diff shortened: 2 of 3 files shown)


src/Eloquent/Model.php
```diff
@@ -259,20 +261,11 @@ public function originalIsEquivalent($key)
...
-
-            return $attribute == $original;
-        }
-
-        if ($this->hasCast($key, static::$primitiveCastTypes)) {
-            return $this->castAttribute($key, $attribute) ===
-                $this->castAttribute($key, $original);
+        try {
+            return fromPHP([$attribute]) === fromPHP([$original]);
+        } catch (UnexpectedValueException $e) {
```
GromNaN, line 266: Do you have cases of `fromPHP` throwing a `MongoDB\Driver\Exception\UnexpectedValueException`? If the attribute can't be converted to BSON, I suspect the `update` operation will fail.

GromNaN (changes requested):
This change will leverage the MongoDB driver to detect attribute changes. It will reduce bugs and make the maintenance easier.
Could you rebase onto master and add a changelog entry?

src/Eloquent/Model.php
jmikola: > If the attribute can't be converted to BSON, I suspect the update operation will fail.

What is "update operation" referring to? I'm missing the context from this PR.

Looking through [`php_phongo_zval_to_bson_internal()`](https://github.com/mongodb/mongo-php-driver/blob/1.16.2/src/phongo_bson_encode.c#L402) there are some cases that come to mind:

 * Attempting to encode prohibited values as root-level documents (e.g. enum, types)
 * Keys contain null bytes
 * Unexpected return value from a `bsonSerialize()` method (if not caught by PHP itself)
 * Encountering an unexpected BSON type instance
 * Encoding a non-backed enum
 * Encoding a string that fails UTF-8 validation
 * Encountering recursion during encoding

None of these should happen under normal operation (there's a joke about "unexpected" there somewhere).

divine:
> This change will leverage the MongoDB driver to detect attribute changes. It will reduce bugs and make the maintenance easier. Could you rebase onto master and add a changelog entry?

Hello,

Done! Can you please review it?

Thanks!

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ All notable changes to this project will be documented in this file.
 - Change signature of `Query\Builder::__constructor` to match the parent class [#26](https://github.com/GromNaN/laravel-mongodb-private/pull/26) by [@GromNaN](https://github.com/GromNaN).
 - Fix Query on `whereDate`, `whereDay`, `whereMonth`, `whereYear`, `whereTime` to use MongoDB operators [#2570](https://github.com/jenssegers/laravel-mongodb/pull/2376) by [@Davpyu](https://github.com/Davpyu) and [@GromNaN](https://github.com/GromNaN).
 - `Model::unset()` does not persist the change. Call `Model::save()` to persist the change [#2578](https://github.com/jenssegers/laravel-mongodb/pull/2578) by [@GromNaN](https://github.com/GromNaN).
+- Use mongodb function to check for dirtiness [#2515](https://github.com/jenssegers/laravel-mongodb/pull/2515) by [@halaei](https://github.com/halaei) and [@GromNaN](https://github.com/GromNaN).
```
GromNaN, line 22: It's so kind of you to want to involve me in this work. I think you deserve the credit.

```suggestion
- Use `MongoDB\BSON\fromPHP` to detect changes [#2515](https://github.com/jenssegers/laravel-mongodb/pull/2515) by [@halaei](https://github.com/halaei) and [@divine](https://github.com/divine).
```

GromNaN (dismissed):
LGTM.

src/Eloquent/Model.php
```diff
@@ -294,20 +295,11 @@ public function originalIsEquivalent($key)
...
-        if ($this->hasCast($key, static::$primitiveCastTypes)) {
-            return $this->castAttribute($key, $attribute) ===
-                $this->castAttribute($key, $original);
+        if (is_scalar($attribute) || is_scalar($original)) {
+            return false;
         }
 
-        return is_numeric($attribute) && is_numeric($original)
-            && strcmp((string) $attribute, (string) $original) === 0;
+        return fromPHP([$attribute]) === fromPHP([$original]);
```
alcaeus, line 302: A couple of notes on this:
* `fromPHP` will be deprecated and is replaced by `Document::fromPHP`
* Root-Level encoding is only supported for documents (i.e. any object or array with string keys)
* If `$attribute` happens to be an object (e.g. an embedded document), this conversion will not yield the correct results as only public properties are encoded, which is most likely wrong for any Model
* This may not be an issue, but I'll note that using this comparison `['foo' => 'bar', 'bar' => 'foo']` and `['bar' => 'foo', 'foo' => 'bar']` will not be treated as equal. That said, we don't have a good way to check BSON equality, and one may argue that two BSON documents with the same key/value pairs in different order shouldn't be considered equal.

For the time being, using this comparison may be feasible as long as we can ensure that it doesn't have side effects for embedded documents.

tests/ModelTest.php
```diff
@@ -876,13 +876,48 @@ public function testMultipleLevelDotNotation(): void
...
+        $user->name = 'John Doe';
+        $user->birthday = new DateTime('19 august 1989');
+        $user->syncOriginal();
         $this->assertEmpty($user->getDirty());
 
         $user->birthday = new DateTime('19 august 1989');
         $this->assertEmpty($user->getDirty());
     }
 
+    public function testGetDirty()
```
alcaeus, line 888: Please add a test using embedded documents and arrays to ensure correct behaviour.

alcaeus (changes requested):

GromNaN:
Now that some transformations have been moved to the query builder (`DateTimeInterface` <=> `UTCDateTime` in https://github.com/mongodb/laravel-mongodb/pull/3119 and https://github.com/mongodb/laravel-mongodb/pull/3105), we might adjust to convert all dates before comparison.

GromNaN:
Thank you for the contribution. The approach is correct and the bug is real: changing a field from int(1) to string('1') is not detected as dirty, even though MongoDB stores types as-is.

The fix has been reworked for the current codebase (DocumentModel trait, MongoDB\Laravel namespace, Document::fromPHP() instead of the deprecated global function) and merged in #3501.

PR #2522, opened 2023-03-14
fix Enums not being cast when calling `Model::toArray()`

jeromegamez:
Hey there 👋,

while testing the migration of a project to Laravel 10, calling `toArray()` on a model caused a warning when attributes are defined to be cast as Enums.

```
Attempt to read property "my_property" on string in vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php on line 1181
```

(and it didn't work). After diving a bit into the code, I noticed that the previously created `$castValue` was not used when converting the Enum to a string in the `addCastAttributesToArray()` method.

I didn't know where to put the Enum for the test, so I just added it to the `models` directory although it felt wrong 😅.

~~The changed order in the `use` statements is due to StyleCI complaining.~~

Let me know if you'd like me to change the Enum location or anything else! 🙏🏻

:octocat:

### DIFF
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -556,7 +556,7 @@ abstract class Model extends BaseModel
             }
 
             if ($this->isEnumCastable($key) && (! $castValue instanceof Arrayable)) {
-                $castValue = $castValue !== null ? $this->getStorableEnumValue($attributes[$key]) : null;
+                $castValue = $castValue !== null ? $this->getStorableEnumValue($castValue) : null;
             }
 
             if ($castValue instanceof Arrayable) {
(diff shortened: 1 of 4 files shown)


jeromegamez:
Alright, I think StyleCI and PHP CS Fixer are in conflict about whether the use statements should be ordered alphanumerically or separately 😅.

If you'd like, I could update and apply the PHP CS Fixer config to be in sync with StyleCI in an additional commit or in a new PR.

alcaeus (approved):
The change looks good, thanks for the fix and test 🎉 

As for StyleCI vs. PHP CS Fixer, I'd let @divine make a decision on that.

divine:
Hello,

We can’t disable styleci and the preferred is always php-cs-fixer.

Thanks!

divine (approved):

PR #2530, opened 2023-04-06
Added support for Laravel v10.x

subhamchbty:
As `laravel/framework` replaces `illuminate/events` and thus cannot coexist with it, we need to remove `illuminate/events` from `composer.json` file. Otherwise, it is failing to install the package using `composer require`.

I hope the PR will be approved and will be merged soon.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,6 @@
         "illuminate/support": "^10.0",
         "illuminate/container": "^10.0",
         "illuminate/database": "^10.0",
-        "illuminate/events": "^10.0",
         "mongodb/mongodb": "^1.15"
     },
     "require-dev": {


mareckigit (approved):
There is no reference to illuminate/events inside the code.

mareckigit (approved):

mareckigit (approved):

cinco (approved):

acrispim (approved):

apeisa:
This is not needed, you can use the current dev-master with Laravel 10 just fine since it will match the same version for events.

Although in my opinion, it would make much more sense to just add the Laravel framework as a dependency, so instead of 

```json
"illuminate/support": "^10.0",
"illuminate/container": "^10.0",
"illuminate/database": "^10.0",
"illuminate/events": "^10.0",
```

just defining

```json
"laravel/framework": "^10.0",
```

mareckigit:
@alcaeus You're right, but this PR is about illuminate/events. Why do you want to keep reference to this package if it's not used? Namaspace Illuminate\Events is not used in jenssegers/laravel-mongodb.

PR #2543, opened 2023-06-02
Feat/mongo db retries

AlexNesvat:
This PR made for mongodb retries which may happen during Mongo Primary Node step down and query execution was interrupted one of error exception. To prevent of loosing data `Try/Catch` statement was added to repeat failed query execution.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 ARG PHP_VERSION=8.0
-ARG COMPOSER_VERSION=2.0
+ARG COMPOSER_VERSION=2.5.4
 
 FROM composer:${COMPOSER_VERSION}
 FROM php:${PHP_VERSION}-cli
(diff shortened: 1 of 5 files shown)


alcaeus:
@AlexNesvat I would suggest taking a look at the [Retryable Writes](https://www.mongodb.com/docs/v6.0/core/retryable-writes/) and [Retryable Reads](https://www.mongodb.com/docs/v6.0/core/retryable-reads/) features in MongoDB. These features are handled in the drivers as described in the specifications ([retryable writes](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst), retryable reads](https://github.com/mongodb/specifications/blob/master/source/retryable-reads/retryable-reads.rst)). IMO there is no reason to re-implement something like this in a client library, as the driver will already have retried an operation if it was save to retry. This is the case for these two errors, as they are listed as retryable in the driver's specification.

Note that on newer server versions, the server itself will decide whether clients can retry an operation by adding a `RetryableWriteError` label to the error response, eliminating the need for hardcoding server error codes in client libraries. With that in mind, I would suggest closing this pull request and taking a look at what is causing the errors on your end and dealing with them through the already existing mechanisms.

PR #2550, opened 2023-07-01
Added two methods incrementEach and decrementEach

SmallRuralDog:
add incrementEach
add decrementEach

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -791,7 +791,7 @@ class Builder extends BaseBuilder
     {
         $query = ['$inc' => $columns];
 
-        if (! empty($extra)) {
+        if ($extra) {
             $query['$set'] = $extra;
         }
 
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -787,10 +787,44 @@ class Builder extends BaseBuilder
         return $this->performUpdate($query, $options);
     }
 
+    public function incrementEach(array $columns, array $extra = [], array $options = [])
+    {
+        $query = ['$inc' => $columns];
+
+        if (! empty($extra)) {
+            $query['$set'] = $extra;
+        }
+
+        // Protect
+        foreach ($columns as $column => $amount) {
+            $this->where(function ($query) use ($column) {
+                $query->where($column, 'exists', false);
+
+                $query->orWhereNotNull($column);
+            });
+        }
+
+        $options = $this->inheritConnectionOptions($options);
+
+        return $this->performUpdate($query, $options);
+    }
+
     /** @inheritdoc */
     public function decrement($column, $amount = 1, array $extra = [], array $options = [])
     {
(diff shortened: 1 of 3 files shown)


GromNaN:
Hello, thanks for the suggestion. For reference, this methods were added in Laravel 9 by https://github.com/laravel/framework/pull/45577.

Would you like to rebase the PR, add tests in [`QueryBuilderTest`](https://github.com/jenssegers/laravel-mongodb/blob/master/tests/QueryBuilderTest.php) and update the changelog?

src/Query/Builder.php
```diff
@@ -647,6 +647,29 @@ public function increment($column, $amount = 1, array $extra = [], array $option
...
+        if (! empty($extra)) {
+            $query['$set'] = $extra;
+        }
+
+        // Protect
+        foreach ($columns as $column => $amount) {
+            $this->where(function ($query) use ($column) {
+                $query->where($column, 'exists', false);
+
+                $query->orWhereNotNull($column);
```
GromNaN, line 664: With cumulative conditions on all fields, no field will be incremented if at least 1 of the fields is `null`.

We'd be wise not to set any conditions at all and let the server return an error.

alcaeus (changes requested):
Please address the comment by @GromNaN and add tests for the new functionality.

GromNaN (approved):

GromNaN:
PR updated with tests.

src/Query/Builder.php
```diff
@@ -789,24 +789,18 @@ public function increment($column, $amount = 1, array $extra = [], array $option
...
-        // Protect
+        // Not using $inc for each column, because it would fail if one column is null.
         foreach ($columns as $column => $amount) {
-            $this->where(function ($query) use ($column) {
-                $query->where($column, 'exists', false);
-
-                $query->orWhereNotNull($column);
-            });
+            $stage['$addFields'][$column] = [
+                '$add' => [['$ifNull' => ['$' . $column, 0]], $amount],
```
alcaeus: According to the [docs](https://www.mongodb.com/docs/manual/reference/operator/aggregation/add/), there's an optimisation we can leverage in MongoDB 6.1 and later:
> Starting in MongoDB 6.1 you can optimize the `$add` operation. To improve performance, group references at the end of the argument list.

```suggestion
                '$add' => [$amount, ['$ifNull' => ['$' . $column, 0]]],
```

alcaeus (commented):
Minor suggestion to optimise the `$add` operator, but I like that the behaviour now better matches the expectation from Laravel that all documents are updated and we gracefully handle `null` values.

src/Query/Builder.php
GromNaN: Thanks. It's strange that the server isn't able to apply this kind of optimisation itself.

alcaeus (approved):

PR #2556, opened 2023-07-10
Change `illuminate support` version to include version ^9.0.

bishoyromany:
I've made this update to clear the error in the following screenshot.
It has a conflict with [termwind](https://github.com/nunomaduro/termwind) package.

image

image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     ],
     "license": "MIT",
     "require": {
-        "illuminate/support": "^10.0",
+        "illuminate/support": "^9.0|^10.0",
         "illuminate/container": "^10.0",
         "illuminate/database": "^10.0",
         "illuminate/events": "^10.0",


alcaeus:
The 3.9.x releases support Laravel 9.0, but for some reason they were not considered eligible. Please run `composer show illuminate/support` to show the version you are currently using, and `composer why-not jenssegers/mongodb 3.9.5` to find out more about the package conflict.

I'll also note that using `--ignore-platform-reqs` in `composer require` can have serious side effects and you shouldn't typically need it. Please run `composer require` without that flag to ensure it's not responsible for the errors.

bishoyromany:
I'm using Laravel 10, maybe that's the reason of that conflict, But anyways seems that all is going good after version 9.0 for `illuminate support`

alcaeus:
For Laravel 10 support, you'll have to use `dev-master` at this time. Laravel 10 release will be part of an upcoming release.

bishoyromany:
Got it, thanks man

PR #2568, opened 2023-08-22
PHPORM-68 Fix partial value un exist validator

GromNaN:
Fix [PHPORM-68](https://jira.mongodb.org/browse/PHPORM-68)

Missed in d5f1bb901f3e3c6777bc604be1af0a8238dc089a.

### DIFF
--- a/src/Validation/DatabasePresenceVerifier.php
+++ b/src/Validation/DatabasePresenceVerifier.php
@@ -43,8 +43,13 @@ class DatabasePresenceVerifier extends \Illuminate\Validation\DatabasePresenceVe
      */
     public function getMultiCount($collection, $column, array $values, array $extra = [])
     {
-        // Generates a regex like '/(a|b|c)/i' which can query multiple values
-        $regex = '/('.implode('|', $values).')/i';
+        // Nothing can match an empty array. Return early to avoid matching an empty string.
+        if ($values === []) {
+            return 0;
+        }
+
+        // Generates a regex like '/^(a|b|c)$/i' which can query multiple values
+        $regex = new Regex('^('.implode('|', $values).')$', 'i');
 
         $query = $this->table($collection)->where($column, 'regex', $regex);
 
--- a/src/Validation/DatabasePresenceVerifier.php
+++ b/src/Validation/DatabasePresenceVerifier.php
@@ -49,7 +49,7 @@ class DatabasePresenceVerifier extends \Illuminate\Validation\DatabasePresenceVe
         }
 
         // Generates a regex like '/^(a|b|c)$/i' which can query multiple values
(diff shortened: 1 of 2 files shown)


src/Validation/DatabasePresenceVerifier.php
```diff
@@ -43,8 +43,13 @@ public function getCount($collection, $column, $value, $excludeId = null, $idCol
...
     {
-        // Generates a regex like '/(a|b|c)/i' which can query multiple values
-        $regex = '/('.implode('|', $values).')/i';
+        // Nothing can match an empty array. Return early to avoid matching an empty string.
+        if ($values === []) {
+            return 0;
+        }
+
+        // Generates a regex like '/^(a|b|c)$/i' which can query multiple values
+        $regex = new Regex('^('.implode('|', $values).')$', 'i');
```
alcaeus, line 52: I'm aware it wasn't done before, but shouldn't each individual value be sent through `preg_quote`?

GromNaN: Indeed they should. I'll add tests also.

src/Validation/DatabasePresenceVerifier.php
```diff
@@ -43,8 +43,13 @@ public function getCount($collection, $column, $value, $excludeId = null, $idCol
...
     {
-        // Generates a regex like '/(a|b|c)/i' which can query multiple values
-        $regex = '/('.implode('|', $values).')/i';
+        // Nothing can match an empty array. Return early to avoid matching an empty string.
+        if ($values === []) {
+            return 0;
+        }
+
+        // Generates a regex like '/^(a|b|c)$/i' which can query multiple values
+        $regex = new Regex('^('.implode('|', array_map(preg_quote(...), $values)).')$', 'i');
```
alcaeus, line 52: The first-class callable syntax is just beautiful...

tests/ValidationTest.php
```diff
@@ -103,5 +103,31 @@ public function testExists(): void
             ['name' => 'required|exists:users']
         );
         $this->assertFalse($validator->fails());
+
+        $validator = Validator::make(
+            ['name' => ['test name', 'john']], // Part of an existing value
+            ['name' => 'required|exists:users']
+        );
+        $this->assertTrue($validator->fails());
```
alcaeus, line 111: Just to confirm, `$validator->fails()` is expected to return `true` here because we're checking that a user with that exact name exists?

alcaeus (approved):
LGTM if my assumption about `$validator->fails()` is correct.

tests/ValidationTest.php
GromNaN: Yes, according to the previous tests, the validation must fail when the value is **not** found in the collection.

PR #2571, opened 2023-08-22
Remove calls to `Str::contains` and `Arr::get` when not necessary

GromNaN:
- [`Str::contains`](https://laravel.com/docs/10.x/helpers#method-str-contains) is identical to `str_contains` but support `array`. This is not necessary in several places.
- [`Arr::get`](https://laravel.com/docs/10.x/helpers#method-array-get) is identical to `$array[$key] ?? $default` but accept dot notation for nested arrays. This is not necessary when the key does not contain a dot.

Replacing this method calls with native functions improve performance (less method calls) and simplifies step-by-step debugging.

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -5,7 +5,6 @@ namespace Jenssegers\Mongodb;
 use function class_exists;
 use Composer\InstalledVersions;
 use Illuminate\Database\Connection as BaseConnection;
-use Illuminate\Support\Arr;
 use InvalidArgumentException;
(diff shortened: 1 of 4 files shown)


jmikola (approved):
LGTM.

Per our earlier conversation in https://github.com/GromNaN/laravel-mongodb/pull/27#discussion_r1300660009, I created [PHPORM-76](https://jira.mongodb.org/browse/PHPORM-76) to propose adding an explicit PHP language requirement to this package's `composer.json` file.

alcaeus (approved):

PR #2577, opened 2023-08-22
fix attribute vs relation priorty

mshamaseen:
Attributes should be prioritized over relationship methods; this is how Laravel does it and what the developer anticipates.

## The issue
The issue was (and I added a test for it), using this package, if you have a method name same as an attribute name, and you call the attribute, this package will return an error instead of returning the attribute.

for example:

```
/**
* @property string foo
*/
class model {
  function foo() {
    // this is not a relation, it is an action, do anything here
  }
}
```

Calling model->foo will throw an exception

## The solution

This package was trying to call the method before the attribute because of embedded relations and (strangely) the many-to-many relation.

What I did in this PR is, Instead of calling the methods first, check if there is a method that has a return type of `EmbedsOne` or `EmbedsMany` and return it, otherwise continue using Laravel attribute/relation workflow (the same logic Laravel team applied for Attribute accessor functions).

For the BelongsToMany relation, I noticed that in the tests, the foreign key was called the same as the relation, let us say `group`.

So when `$model->group` was called, the tests expected the relation instead of the attribute, which is wrong, because attributes should be prioritized over relations.

The solution was so easy, just don't name the foreign key the same as the relation; that is the default behavior if you don't pass any foreign keys to  `$this->belongsToMany(related , /* Don't pass anything else */)` , and that is also the documented way to do it.

## Extra
While I was working I updated MySql/MongoDB versions in the dockerfile and github action file. I also dropped the support for mongoDB version 4 as starting from version 5 and above Mongosh is used instead of Mongo.


README.md
```diff
@@ -856,19 +856,23 @@ class User extends Model
...
 ```
+**Warning:** naming the foreign key same as the relation name will prevent the relation for being called on dynamic property, i.e. in the example above if you replaced `group_ids` with `groups` calling `$user->groups` will return the column instead of the relation.
 
 ### EmbedsMany Relationship
 
 If you want to embed models, rather than referencing them, you can use the `embedsMany` relation. This relation is similar to the `hasMany` relation but embeds the models inside the parent object.
 
 **REMEMBER**: These relations return Eloquent collections, they don't return query builder objects!
 
+**Breaking changes** starting from v4.0 you need to define the return type of EmbedsOne and EmbedsMany relation for it to work
```
GromNaN, line 867: > What I did in this PR is, Instead of calling the methods first, check if there is a method that has a return type of EmbedsOne or EmbedsMany and return it, otherwise continue using Laravel attribute/relation workflow (the same logic Laravel team applied for Attribute accessor functions).

Very smart. Can you please provide a link to where it's done like this in laravel.

.github/workflows/build-ci.yml
```diff
@@ -36,16 +36,14 @@ jobs:
...
-                    - '4.4'
                     - '5.0'
+                    - '6.0'
                 php:
                     - '8.1'
                     - '8.2'
         services:
             mysql:
-                image: mysql:5.7
+                image: mysql:8.0
```
GromNaN, line 46: > While I was working I updated MySql/MongoDB versions in the dockerfile and github action file. I also dropped the support for mongoDB version 4 as starting from version 5 and above Mongosh is used instead of Mongo.

Sorry to ask, can you extract this CI changes into a dedicated PR for a separated review.

GromNaN (changes requested):
Thank you for contributing your change back to the original project.

I think the changes you've suggested are useful, and will help us better stick to the way Eloquent works.

It needs to be review by @alcaeus and @divine.

README.md
mshamaseen: `hasAttributeMutator` method on https://github.com/laravel/framework/blob/7f287ed2ee8aa0067b66c002b582f8ff7eedc00c/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L440C20-L440C39

src/Eloquent/Model.php
```diff
@@ -161,16 +168,38 @@ public function getAttribute($key)
...
     }
 
+    /**
+     * Determine if an attribute is an embedded relation.
+     *
+     * @param string $key
+     * @return bool
+     * @throws ReflectionException
+     */
+    public function hasEmbeddedRelation(string $key): bool
```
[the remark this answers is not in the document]
mshamaseen: @GromNaN Done.

src/Eloquent/Model.php
```diff
@@ -161,16 +168,38 @@ public function getAttribute($key)
...
+            return false;
+        }
+
+        if (isset(static::$embeddedCache[get_class($this)][$key])) {
+            return static::$embeddedCache[get_class($this)][$key];
+        }
+
+        $returnType = (new ReflectionMethod($this, $method))->getReturnType();
+
+        return $returnType && static::$embeddedCache[get_class($this)][$key] =
```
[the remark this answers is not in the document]
mshamaseen: @GromNaN Done.

.github/workflows/build-ci.yml
GromNaN: I did extracted the CI changes into #2579. So you can remove this changes from the PR.

mshamaseen: @GromNaN Great! Thank you! I'll remove it soon, inshallah.

divine:
Hello,

Any decision regarding embedded relations was made for v4.0 @GromNaN? I mean will be it rewritten completely or just "fixed" when a problem occurs?

We talked about it a lot and the way it currently "works" can lead to performance and logical problems.

My suggestion is that there should be a final decision made before v4.0 otherwise it can lead to Achilles's heel.

Thanks!

GromNaN:
> Hello,
> 
> Any decision regarding embedded relations was made for v4.0 @GromNaN? I mean will be it rewritten completely or just "fixed" when a problem occurs?
> 
> We talked about it a lot and the way it currently "works" can lead to performance and logical problems.
> 
> My suggestion is that there should be a final decision made before v4.0 otherwise it can lead to Achilles's heel.
> 
> Thanks!

To meet the need for a rapid release of a Laravel 10-compatible version, we won't be reworking the relationships for the time being.

mshamaseen:
@GromNaN any news about when this issue will be solved? your work is much appreciated!

PR #2578, opened 2023-08-23
PHPORM-75 Defer `Model::unset($field)` to the `save()`

### LINKED ISSUE #2566: weird unset functionality
As far as I can see, laravel-mongodb overrides `unset` behavior to trigger `drop` functionality that makes a database query each time you run `$moded->unset($field)`.  That means if you have some model that goes through multiple services and making unsets in some of the services for example 
In service 1:
```
$model->unset('blocked');
```
in service 2: 
```
if ($model->price > 2) {
    $model->unset('margin');
}
```
in service 3 you are doing an actual save of the model
```
$model->save();
```

all this will lead to 3 queries, right? For small projects, it maybe not be an issue, but for big projects, it's a huge overhead for the database.  and even worse, it's not a single transaction, and if execution interrupts somewhere it can bring side effects.

I tried to change the behavior of `unset` to not trigger the `drop` method but instead just to unset the attribute from the model, but in that case, during the model save, the model does not see changes because `isDirty` and `getDirty` methods do not compare with original data to find missing data and does not see changes. 

It means you either you killing DB with multiple queries or can't unset the field with eloquent and only set it to `null`. 

Is there any way to fix it???

GromNaN:
Fix #2566 and [PHPORM-75](https://jira.mongodb.org/browse/PHPORM-75).

- **Breaking change**: All calls to `$model->unset($field)`, `$model->drop($field)`, `unset($model[$field])`, `unset($model->field)` must be followed by `$model->save()` so they are persisted.
```diff
  $user = User::find($key);
  unset($user->age);
+ $user->save();
```

- Deprecate `Model::drop($field)`, it's a duplicate of `Model::unset($field)`. The native `unset($model->field)` is recommended.

- `Query\Builder::update()` accept a mix of field names and [update operators](https://www.mongodb.com/docs/manual/reference/operator/update/).

```php
DB::collection('users')->where(...)->update([
    '$inc' => ['age' => 1],
    'last_birthday' => new UTCDateTime(),
]);
```

Todo:
- [x] changelog
- [ ] relations
- [ ] nested model / objects

### DIFF
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -20,9 +20,6 @@ use MongoDB\BSON\ObjectID;
 use MongoDB\BSON\UTCDateTime;
 use function uniqid;
 
-/**
- * @method void unset(string|string[] $columns) Remove one or more fields.
- */
 abstract class Model extends BaseModel
 {
     use HybridRelations, EmbedsRelations;
@@ -341,8 +338,21 @@ abstract class Model extends BaseModel
      *
      * @param  string|string[]  $columns
      * @return void
+     *
+     * @deprecated Use unset() instead.
      */
     public function drop($columns)
+    {
+        $this->unset($columns);
+    }
+
+    /**
+     * Remove one or more fields.
+     *
+     * @param  string|string[]  $columns
+     * @return void
+     */
+    public function unset($columns)
     {
         $columns = Arr::wrap($columns);
 
@@ -558,19 +568,6 @@ abstract class Model extends BaseModel
     protected function isGuardableColumn($key)
     {
         return true;
-    }
-
-    /**
-     * @inheritdoc
-     */
-    public function __call($method, $parameters)
-    {
-        // Unset method
-        if ($method == 'unset') {
-            return $this->drop(...$parameters);
-        }
-
-        return parent::__call($method, $parameters);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/Eloquent/Model.php
```diff
@@ -502,19 +570,6 @@ protected function isGuardableColumn($key)
         return true;
     }
 
-    /**
-     * @inheritdoc
-     */
-    public function __call($method, $parameters)
```
alcaeus, line 508: It's nice to see this gone! :)

tests/ModelTest.php
```diff
@@ -488,9 +492,56 @@ public function testUnset(): void
         $this->assertTrue(isset($user2->note2));
 
         $user2->unset(['note1', 'note2']);
+        $user2->save();
```
alcaeus, line 495: I believe there was a separate issue where the model was not marked as dirty - maybe assert on `isDirty` as well?

GromNaN: It was not dirty since the change was done directly in the database.
I added the assertions on `isDirty` for exhautivity.

src/Eloquent/Model.php
GromNaN: I'm not sure why the method `unset` was not declared. It might be to allow a relation named `unset`.

alcaeus (approved):

PR #2579, opened 2023-08-24
Fix tests for MongoDB 7.0

GromNaN:
- Add MongoDB server 6.0 and 7.0 in CI.
- Remote [unmaintained MongoDB versions < 4.4](https://www.mongodb.com/support-policy/lifecycles)
- Upgrade CI to MySQL 8.0

Error message have changed. We don't need to assert on the error message as we already assert there is a `BulkWriteException`.

```
Caused by :: Write conflict during plan execution and yielding is disabled. :: Please retry your operation or multi-document transaction.
```

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -36,16 +36,16 @@ jobs:
                 os:
                     - ubuntu-latest
                 mongodb:
-                    - '4.0'
-                    - '4.2'
                     - '4.4'
                     - '5.0'
+                    - '6.0'
+                    - '7.0'
                 php:
                     - '8.1'
                     - '8.2'
         services:
             mysql:
-                image: mysql:5.7
+                image: mysql:8.0
                 ports:
                     - 3307:3306
                 env:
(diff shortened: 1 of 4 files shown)


mshamaseen:
@GromNaN what about the docker-composer file? in my case the tests were trying to run before mongoDB is fully initiated, so I made this change:
https://github.com/jenssegers/laravel-mongodb/pull/2577/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3

GromNaN:
> @GromNaN what about the docker-composer file? in my case the tests were trying to run before mongoDB is fully initiated, so I made this change:

Thanks for the remainder. I imported your changes.

GromNaN:
The transaction tests require a replicaset to work. We might add this in config in docker-compose in a future PR.

.github/workflows/build-ci.yml
```diff
@@ -58,13 +58,16 @@ jobs:
             -   name: Create MongoDB Replica Set
                 run: |
                     docker run --name mongodb -p 27017:27017 -e MONGO_INITDB_DATABASE=unittest --detach mongo:${{ matrix.mongodb }} mongod --replSet rs --setParameter transactionLifetimeLimitSeconds=5
-                    until docker exec --tty mongodb mongo 127.0.0.1:27017 --eval "db.runCommand({ ping: 1 })"; do
+
+                    if [ "${{ matrix.mongodb }}" = "4.4" ]; then MONGOSH_BIN="mongo"; else MONGOSH_BIN="mongosh"; fi
```
alcaeus, line 62: Where does `mongosh` come from? IIRC it's not distributed with the server package, so if we download it separately we should be able to use it with 4.4 as well.

Dockerfile
```diff
@@ -1,24 +1,21 @@
...
 
 RUN apt-get update && \
-    apt-get install -y autoconf pkg-config libssl-dev git libzip-dev zlib1g-dev && \
+    apt-get install -y autoconf pkg-config libssl-dev git unzip libzip-dev zlib1g-dev && \
     pecl install mongodb && docker-php-ext-enable mongodb && \
     pecl install xdebug && docker-php-ext-enable xdebug && \
     docker-php-ext-install -j$(nproc) pdo_mysql zip
 
-COPY --from=composer:${COMPOSER_VERSION} /usr/bin/composer /usr/local/bin/composer
+COPY --from=composer:2.5.8 /usr/bin/composer /usr/local/bin/composer
```
alcaeus, line 11: Having it at the top of the file made it easier to locate the version for an update, but no objection to inlining it.

alcaeus (commented):
LGTM besides the question about mongosh

Dockerfile
GromNaN: It doesn't work with my setup.
```
Docker version 24.0.5, build ced0996600
Docker Compose version 2.20.3
colima version 0.5.5
```

There is a workaroud if I alias the composer image.
https://github.com/docker/for-mac/issues/2155#issuecomment-338619432

.github/workflows/build-ci.yml
alcaeus: Oh, I see. Fine with me :)

PR #2580, opened 2023-08-28
Move model to a trait

### LINKED ISSUE #2120: Update model to implement a trait to set mongo functionality
In a project I'm working on we use laravel passport for our authentication. Up until now we've been using the [designmynight/laravel-mongodb-passport](https://github.com/designmynight/laravel-mongodb-passport) package to update the passport models to work with mongo. This package adds copies of each of the passport models but extends the mongodb base model instead.

We've just updated to version 10 of passport and you can no longer do this as a lot of the classes now type hint the default passport models, this means you now have to extend the passport models.

For example you could override the default token model easily by doing `Passport::useTokenModel(CustomTokenModel::class)`, but in the `Laravel\Passport\TokenRepository` class it type hints the default token model in save method.

```php
/**
 * Store the given token instance.
 *
 * @param  \Laravel\Passport\Token  $token
 * @return void
 */
public function save(Token $token)
{
    $token->save();
}
```

This leaves us in a situation where we either have to go through and make custom versions of every passport class that has a type hint in, or to extend every passport model and copy the contents of the mongodb model into it.

I'm proposing to change the mongo db model so it implements a trait to set the mongo specific functionality.

The model would then end up looking something like this:

```php
namespace Jenssegers\Mongodb\Eloquent;

use Illuminate\Database\Eloquent\Model as BaseModel;

abstract class Model extends BaseModel
{
    use Moloquent;

    /**
     * The primary key for the model.
     * @var string
     */
    protected $primaryKey = '_id';

    /**
     * The primary key type.
     * @var string
     */
    protected $keyType = 'string';
}
```

The `Moloquent` trait would then have all of the current methods and mongo specific properties from the current model.

This would allow us to extend models from other projects more easily and then just implement the `Moloquent` trait to make it work with mongo. So our token class could then become:

```php
class Token extends \Laravel\Passport\Token
{
    use Moloquent;
}
```

I've tried cloning the project and implementing the trait instead and all of the tests still pass. 

I'm happy to submit a merge request to show how the trait would work but let me know your thoughts.

### LINKED ISSUE #2445: [Doc Improvement Request] Laravel Sanctum Configuration
I believe that the comment in the link below should be at least referenced in the docs. 

[https://github.com/jenssegers/laravel-mongodb/discussions/2242#discussioncomment-678732](https://github.com/jenssegers/laravel-mongodb/discussions/2242)

It basically creates a new PersonalAccessToken class that can be used to make Sanctum work with MongoDB. I struggled with the same problem that the issue author had and I believe that this solution could help more people if it becomes easier to find or becomes part of the official docs.

### LINKED ISSUE #2991: [Feature Request]  SANCTUM SUPPORT
For now I check all the issue and found a way to add sanctum support but I think instead of manual, the library could make this integrated with test like the cache and queue by @GromNaN made to support.  This has status backlog in Jira, but I think with the quick support of cache in 4.3 version it should easily added and add also in docs. Thanks

Copy original PersonalAccessToken
`cp ./vendor/laravel/sanctum/src/PersonalAccessToken.php ./app/Models/PersonalAccessToken.php`

```php
<?php

namespace App\Models;

use Laravel\Sanctum\Contracts\HasAbilities;

class PersonalAccessToken extends Model implements HasAbilities
{
    /**
     * The attributes that should be cast to native types.
     *
     * @var array
     */
    protected $casts = [
        'abilities' => 'json',
        'last_used_at' => 'datetime',
        'expires_at' => 'datetime',
    ];

    /**
     * The attributes that are mass assignable.
     *
     * @var array
     */
    protected $fillable = [
        'name',
        'token',
        'abilities',
        'expires_at',
    ];

    /**
     * The attributes that should be hidden for serialization.
     *
     * @var array
     */
    protected $hidden = [
        'token',
    ];

    /**
     * Get the tokenable model that the access token belongs to.
     *
     * @return \Illuminate\Database\Eloquent\Relations\MorphTo
     */
    public function tokenable()
    {
        return $this->morphTo('tokenable');
    }

    /**
     * Find the token instance matching the given token.
     *
     * @param  string  $token
     * @return static|null
     */
    public static function findToken($token)
    {
        if (strpos($token, '|') === false) {
            return static::where('token', hash('sha256', $token))->first();
        }

        [$id, $token] = explode('|', $token, 2);

        if ($instance = static::find($id)) {
            return hash_equals($instance->token, hash('sha256', $token)) ? $instance : null;
        }
    }

    /**
     * Determine if the token has a given ability.
     *
     * @param  string  $ability
     * @return bool
     */
    public function can($ability)
    {
        return in_array('*', $this->abilities) ||
            array_key_exists($ability, array_flip($this->abilities));
    }

    /**
     * Determine if the token is missing a given ability.
     *
     * @param  string  $ability
     * @return bool
     */
    public function cant($ability)
    {
        return !$this->can($ability);
    }
}

```

Then in AppServiceProvider

```php
<?php

namespace App\Providers;

use Illuminate\Foundation\AliasLoader;
use Illuminate\Support\ServiceProvider;

class AppServiceProvider extends ServiceProvider
{
    /**
     * Register any application services.
     */
    public function register(): void
    {
        //
    }

    /**
     * Bootstrap any application services.
     */
    public function boot(): void
    {
        $loader = AliasLoader::getInstance();
        // using custom PersonalAccessToken model instead of Sanctum's
        $loader->alias(\Laravel\Sanctum\PersonalAccessToken::class, \App\Models\PersonalAccessToken::class);
    }
}
```

All references
https://github.com/mongodb/laravel-mongodb/issues/2444
https://github.com/mongodb/laravel-mongodb/issues/2445
https://jira.mongodb.org/browse/PHPORM-88

GromNaN:
Fix #2120
Fix #2445
Fix #2991

By moving all the logic to a trait, it makes it possible to reuse entities provided by community projects. Example: https://github.com/z-song/laravel-admin/blob/master/src/Auth/Database/Administrator.php

The properties `$primaryKey` and `$keyType` can't be in the trait because PHP forbid declaring a property in a trait that is different (type and value) than the property declared in the parent class.

For Laravel's `User` model:
```php
namespace App\Models;

use Illuminate\Foundation\Auth\User as Authenticatable;
use MongoDB\Laravel\Eloquent\DocumentModel;

class User extends Authenticatable
{
    protected $collection = 'users';
    protected $primaryKey = '_id';
    protected $keyType = 'string';
}
```

For Laravel Sanctum (with [default model override](https://laravel.com/docs/11.x/sanctum#overriding-default-models)):
```php
namespace App\Models;

use Laravel\Sanctum\PersonalAccessToken as SanctumToken;
use MongoDB\Laravel\Eloquent\DocumentModel;

class PersonalAccessToken extends SanctumToken
{
    use DocumentModel;

    protected $collection = 'personal_access_tokens';
    protected $primaryKey = '_id';
    protected $keyType = 'string';
}

```

~I have mixed feelings about this change, as it breaks the link with the overloaded parent methods. It would probably be better to make the change in each community-package when necessary.~


GromNaN:
Currently, the `Model` class is a marker for MongoDB models. If a trait is used, an other marker will be necessary.
https://github.com/mongodb-php/laravel-mongodb/blob/f06f7b3dca88cfd5cc6a823ff01051b4a46ca10a/src/Eloquent/HybridRelations.php#L145

src/Eloquent/Model.php
```diff
@@ -76,699 +28,10 @@ abstract class Model extends BaseModel
      */
     protected $keyType = 'string';
```
GromNaN, line 31: When the trait is used, the properties `$keyType` and `$primaryKey` must be defined, because they can't be redefined by the trait.

src/Eloquent/DocumentModel.php
```diff
@@ -0,0 +1,749 @@
...
+    /** @inheritdoc */
+    public function freshTimestamp()
+    {
+        return new UTCDateTime(Date::now());
+    }
+
+    /** @inheritdoc */
+    public function getTable()
+    {
+        return $this->collection ?? parent::getTable();
```
GromNaN, line 144: This property is not defined in the trait, otherwise it's allowed to redefine in the class that uses the trait.

src/Eloquent/DocumentModel.php
```diff
@@ -0,0 +1,749 @@
...
+
+trait DocumentModel
+{
+    use HybridRelations;
+    use EmbedsRelations;
+
+    /**
+     * The parent relation instance.
+     */
+    private Relation $parentRelation;
```
GromNaN, line 59: Changing the visibility to `private`. There is a getter and an setter for this method. It doesn't need to be protected. Since it's a trait, it can still be accessed directly in a class that uses the trait.

src/Eloquent/Model.php
```diff
@@ -77,698 +29,15 @@ abstract class Model extends BaseModel
...
-     * @param  string|string[] $columns
+     * Indicates if the given model class is a MongoDB document model.
+     * It must be a subclass of {@see BaseModel} and use the
+     * {@see DocumentModel} trait.
      *
-     * @return void
+     * @param class-string|object $related
      */
-    public function unset($columns)
+    public static function isDocumentModel(string|object $related): bool
```
GromNaN: This method replaces `$class instanceof MongoDB\Laravel\Eloquent\Model` to check if the model class is for a MongoDB connection. It is public and can be used in projects.

tests/Models/User.php
```diff
@@ -30,22 +30,24 @@
  * @property string $username
  * @property MemberStatus member_status
  */
-class User extends Eloquent implements AuthenticatableContract, CanResetPasswordContract
+class User extends Model implements AuthenticatableContract, CanResetPasswordContract
 {
+    use DocumentModel;
```
GromNaN, line 35: All the test models have been updated to use the trait, so that we are sure all the features works without extending `MongoDB\Laravel\Eloquent\Model`.

GromNaN (commented):
The trait `DocumentModel` is extracted from the `MongoDB\Laravel\Eloquent\Model` class, the only changes are commented.

src/Eloquent/Model.php
```diff
@@ -77,698 +29,15 @@ abstract class Model extends BaseModel
...
-    public function refresh()
-    {
-        parent::refresh();
-
-        // Clear list of unset fields
-        $this->unset = [];
-
-        return $this;
+        return is_subclass_of($related, BaseModel::class)
+            && array_key_exists(DocumentModel::class, class_uses_recursive($related));
```
alcaeus: `class_uses_recursive` looks like it has significant performance implications. Using a marker interface and checking using `is_subclass_of` would probably be more efficient.

Alternatively, checking whether the class is a subclass of this model class first, and only checking for traits in other cases would probably solve the performance issue for most cases.

tests/Models/User.php
alcaeus: Having one test with a model that extends our document model class instead of using the trait will help us prevent regressions in our model class.

src/Eloquent/Model.php
GromNaN: Added a shortcut to check our Model class + a memory cache for when the same class is checked twice in the same request.

tests/Models/Casting.php
```diff
@@ -5,12 +5,12 @@
 namespace MongoDB\Laravel\Tests\Models;
 
 use MongoDB\Laravel\Eloquent\Casts\BinaryUuid;
-use MongoDB\Laravel\Eloquent\Model as Eloquent;
+use MongoDB\Laravel\Eloquent\Model;
 
-class Casting extends Eloquent
+class Casting extends Model
```
GromNaN, line 10: This class continues to extend `MongoDB\Laravel\Eloquent\Model`

alcaeus (approved):

masterbater:
@GromNaN Hopefully in documentation can add a guide for third party packages like spatie's laravel-permissions, bouncer, tenancy etc, sanctum, etc. Or create a community for laravel packages being modified to work with laravel mongodb. This way all popular packages will have counterpart that is made to work with mongodb thanks.

GromNaN:
@masterbater The documentation is going to be written by MongoDB's doc team.

Issue #1506, opened 2018-05-01
Unset array replacing with null


prafful-panwar:
` foreach ($event->volunteers as $key => $volunteer) {
                if ($volunteer['user_id'] == $userId) {
                    $event->where('volunteers.user_id', intval($userId))
                        ->unset("volunteers.$key");
                }
            }`

FabianoLothor:
Up!

FabianoLothor:
@prafful-panwar try to use `$pull` instead of `$unset`.

GromNaN:
Fixed by #2578

PR #2582, opened 2023-08-28
Fix `Model::unset` with dot field name

GromNaN:
Doing #2578 I did not test subfields.

Related to #1506


alcaeus (dismissed):
Change LGTM. Is this also expected to work for lists?

PR #2584, opened 2023-08-29
PHPORM-127 Fix priority of embeds vs attributes

GromNaN:
Fix #2577 (continuing)
Fix PHPORM-127

Moving the code from the `Model` class to the `EmbedsRelations` trait.

### DIFF
--- a/README.md
+++ b/README.md
@@ -805,6 +805,7 @@ class User extends Model
     }
 }
 ```
+**Warning:** naming the foreign key same as the relation name will prevent the relation for being called on dynamic property, i.e. in the example above if you replaced `group_ids` with `groups` calling `$user->groups` will return the column instead of the relation.
 
 ### EmbedsMany Relationship
 
@@ -812,12 +813,15 @@ If you want to embed models, rather than referencing them, you can use the `embe
 
 **REMEMBER**: These relations return Eloquent collections, they don't return query builder objects!
 
+**Breaking changes** starting from v4.0 you need to define the return type of EmbedsOne and EmbedsMany relation for it to work
+
 ```php
 use MongoDB\Laravel\Eloquent\Model;
+use MongoDB\Laravel\Relations\EmbedsMany;
 
 class User extends Model
 {
-    public function books()
+    public function books(): EmbedsMany
     {
         return $this->embedsMany(Book::class);
     }
@@ -886,10 +890,11 @@ Like other relations, embedsMany assumes the local key of the relationship based
 
 ```php
 use MongoDB\Laravel\Eloquent\Model;
+use MongoDB\Laravel\Relations\EmbedsMany;
 
 class User extends Model
 {
-    public function books()
+    public function books(): EmbedsMany
     {
         return $this->embedsMany(Book::class, 'local_key');
     }
@@ -902,12 +907,15 @@ Embedded relations will return a Collection of embedded items instead of a query
 
 The embedsOne relation is similar to the embedsMany relation, but only embeds a single model.
 
+**Breaking changes** starting from v4.0 you need to define the return type of EmbedsOne and EmbedsMany relation for it to work
+
 ```php
 use MongoDB\Laravel\Eloquent\Model;
+use MongoDB\Laravel\Relations\EmbedsOne;
 
 class Book extends Model
 {
-    public function author()
+    public function author(): EmbedsOne
     {
         return $this->embedsOne(Author::class);
     }
--- a/src/Eloquent/EmbedsRelations.php
+++ b/src/Eloquent/EmbedsRelations.php
@@ -5,12 +5,19 @@ namespace MongoDB\Laravel\Eloquent;
 use Illuminate\Support\Str;
 use MongoDB\Laravel\Relations\EmbedsMany;
 use MongoDB\Laravel\Relations\EmbedsOne;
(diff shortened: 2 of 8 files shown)


src/Eloquent/HybridRelations.php
```diff
@@ -249,13 +251,18 @@ public function belongsToMany(
...
+
         // First, we'll need to determine the foreign key and "other key" for the
         // relationship. Once we have determined the keys we'll make the query
         // instances as well as the relationship instances we need for this.
         $foreignKey = $foreignKey ?: $this->getForeignKey().'s';
 
         $instance = new $related;
 
+        if ($otherKey === $relation) {
+            throw new \LogicException(sprintf('In %s::%s(), the key cannot be identical to the relation name "%s". The default key is "%s".', static::class, $relation, $relation, $instance->getForeignKey().'s'));
```
GromNaN, line 263: @mshamaseen [wrote](https://github.com/mongodb/laravel-mongodb/pull/2577):
> The solution was so easy, just don't name the foreign key the same as the relation; that is the default behavior if you don't pass any foreign keys to $this->belongsToMany(related , /* Don't pass anything else */) , and that is also the documented way to do it.

I propose to throw an exception in this case to avoid any error.

README.md
```diff
@@ -805,19 +805,23 @@ class User extends Model
     }
 }
 ```
+**Warning:** naming the foreign key same as the relation name will prevent the relation for being called on dynamic property, i.e. in the example above if you replaced `group_ids` with `groups` calling `$user->groups` will return the column instead of the relation.
```
jmikola, line 808: I think "e.g." would be more appropriate because you're providing one possible example.

See: https://www.merriam-webster.com/grammar/ie-vs-eg-abbreviation-meaning-usage-difference

README.md
```diff
@@ -805,19 +805,23 @@ class User extends Model
...
 
 ### EmbedsMany Relationship
 
 If you want to embed models, rather than referencing them, you can use the `embedsMany` relation. This relation is similar to the `hasMany` relation but embeds the models inside the parent object.
 
 **REMEMBER**: These relations return Eloquent collections, they don't return query builder objects!
 
+**Breaking changes** starting from v4.0 you need to define the return type of EmbedsOne and EmbedsMany relation for it to work
```
jmikola, line 816: I see this PR targets 5.x. Should this documentation change be backported?

Also, is this heading better written as:

```
**Breaking Change:** starting from...
```

You're only noting a single breaking change, and it appears to be a label (vs. part of the sentence) so a colon could help differentiate. This applies below as well.

src/Eloquent/EmbedsRelations.php
```diff
@@ -76,4 +83,27 @@ protected function embedsOne($related, $localKey = null, $foreignKey = null, $re
...
+    /**
+     * Determine if an attribute is an embedded relation.
+     *
+     * @param string $key
+     * @return bool
+     * @throws \ReflectionException
+     */
+    private function hasEmbeddedRelation(string $key): bool
+    {
+        if (! method_exists($this, $method = Str::camel($key))) {
```
jmikola, line 96: I've never bothered to do this myself, but assumed CS for PHPLIB prohibited such inline assignments in the interest of readability. Is this common within the Laravel package?

Happy to defer to you.

src/Eloquent/HybridRelations.php
```diff
@@ -272,6 +272,10 @@ public function belongsToMany(
 
         $instance = new $related;
 
+        if ($relatedPivotKey === $relation) {
+            throw new \LogicException(sprintf('In %s::%s(), the key cannot be identical to the relation name "%s". The default key is "%s".', static::class, $relation, $relation, $instance->getForeignKey().'s'));
```
jmikola, line 276: You can optionally use argument numbering to remove the duplicate `$relation` argument.

 ```suggestion
            throw new \LogicException(sprintf('In %1$s::%2$s(), the key cannot be identical to the relation name "%2$s". The default key is "%3$s".', static::class, $relation, $instance->getForeignKey().'s'));
```

File: src/Eloquent/Model.php [hunk 6]
```diff
@@ -172,11 +174,7 @@ public function getAttribute($key)
         }
 
         // This checks for embedded relation support.
-        if (
-            method_exists($this, $key)
-            && ! method_exists(self::class, $key)
-            && ! $this->hasAttributeGetMutator($key)
```
jmikola, line 178: Not sure what this method does, but it seems the fix here is replace this with checking that the method returns either EmbedsOne or EmbedsMany (per your EmbedsRelations trait).

tests/Models/Group.php
```diff
@@ -15,6 +15,6 @@ class Group extends Eloquent
 
     public function users(): BelongsToMany
     {
-        return $this->belongsToMany(User::class, 'users', 'groups', 'users', '_id', '_id', 'users');
+        return $this->belongsToMany(User::class, 'users', 'groupIds', 'userIds', '_id', '_id', 'users');
```
jmikola, line 18: Is renaming to "userIds" the meaningful change here, so that it doesn't conflict with the relationship name? Is the renaming of "groupIds" just for consistency?

jmikola (approved):
Some optional suggestions and questions. If everything is easily resolved/answered, feel free to merge.

PR #2588, opened 2023-08-30
Make `EmbedsRelations::embedsMany` and `EmbedsRelations::embedsOne` public

### LINKED ISSUE #2555: Make relationship methods public [Feature Request]
When trying to insert relationships into models via a service provider, it is not possible because the mongo specific methods are protected. Laravel has these methods public for this reason.

Swap out protected for public methods

Other than forking the package I don't see a work around.

GromNaN:
Fix #2555

Other relation methods are public.

### DIFF
--- a/src/Eloquent/EmbedsRelations.php
+++ b/src/Eloquent/EmbedsRelations.php
@@ -53,7 +53,7 @@ trait EmbedsRelations
     }
 
     /**
-     * Define an embedded one-to-many relationship.
+     * Define an embedded one-to-one relationship.
      *
      * @param class-string $related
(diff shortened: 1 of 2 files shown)


jmikola (approved):

src/Eloquent/EmbedsRelations.php
alcaeus:
```suggestion
     * Define an embedded one-to-one relationship.
```

alcaeus (commented):
Changes LGTM, and while I think we can skip test updates, fixing the description of `embedsOne` would be good.

alcaeus (approved):

PR #2591, opened 2023-08-31
Support delete one document with the query builder

### LINKED ISSUE #2502: Deleting With Limit Doesn't Work
- Laravel Version: 9
- PHP Version: 8.0
- MongoDB Version: 3.9

### Description:
Hello, i have problem when deleting data using where + limit. Even using limit is not working. The command didn't use the limit parameter.

### Steps to reproduce
BalanceModel::where('time', '<=', 1670707849677)->take(1)->delete();

### Expected behaviour
Delete BalanceModel where time < xxxx limit 1

### Actual behaviour
When i run that command, it delete using time parameter and delete document without limiting the delete command

The Current Operation Image Status
image
image

GromNaN:
Fix #2502

MongoDB PHP Library supports deleting [one](https://github.com/mongodb/mongo-php-library/blob/a7a768252cc7509e8d5b053b3c90f799b2e55257/src/Operation/DeleteMany.php#L71) or [all](https://github.com/mongodb/mongo-php-library/blob/a7a768252cc7509e8d5b053b3c90f799b2e55257/src/Operation/DeleteMany.php#L71) documents only. So we cannot accept other limits than `null` (unlimited) or `1`.

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -703,7 +703,14 @@ class Builder extends BaseBuilder
         $wheres = $this->compileWheres();
         $options = $this->inheritConnectionOptions();
 
-        $result = $this->collection->deleteMany($wheres, $options);
+        if (is_int($this->limit)) {
+            if ($this->limit !== 1) {
+                throw new \LogicException(sprintf('Delete limit can be 1 or null (unlimited). Got %d', $this->limit));
+            }
+            $result = $this->collection->deleteOne($wheres, $options);
+        } else {
+            $result = $this->collection->deleteMany($wheres, $options);
+        }
 
         if (1 == (int) $result->isAcknowledged()) {
             return $result->getDeletedCount();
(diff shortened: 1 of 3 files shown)


src/Query/Builder.php
```diff
@@ -703,7 +703,14 @@ public function delete($id = null)
         $wheres = $this->compileWheres();
         $options = $this->inheritConnectionOptions();
 
-        $result = $this->collection->deleteMany($wheres, $options);
+        if ($this->limit) {
+            if ($this->limit !== 1) {
```
jmikola, line 707: Should `limit === 0` be supported? Or would Eloquent expect that to be a NOP?

I suppose the notion of `limit: 0` meaning "delete all" is an implementation detail of MongoDB's [`delete` command](https://www.mongodb.com/docs/manual/reference/command/delete/#std-label-deletes-array-limit) and not a general concept, so I'm fine with not supporting that.

I do wonder if we should have a separate exception message if `0` is specified that advises users to specify `null` instead (or not call `limit()` at all), as this message might confuse a user with more experience with MongoDB than Eloquent.

GromNaN: You're absolutely right. I've also banned 0.

jmikola (approved):

enkhjargal (approved):

GromNaN:
Related issue: #2597

PR #2592, opened 2023-09-01
Add tests on null date casts

### LINKED ISSUE #2553: Laravel 9 Unexpected datetime behaviour on Null datetime field
- Laravel-mongodb Version: 9
- PHP Version: 8.2
- Database Driver & Version:

### Steps to reproduce
1. Create/store null or empty datetime field in collection
2. add this datetime field in collection model in protected $cast array 
3. try to fetch data from collection
4. and check null date will show as Current date

### Expected behaviour
After fetching data from collection the Null or empty date fields should return null value or empty value - not current datetime 

### Actual behaviour
Insted of return null or empty field it returning current datetime

GromNaN:
Fix #2553

Adding a test on the described scenario. The error found.

### DIFF
--- a/tests/ModelTest.php
+++ b/tests/ModelTest.php
@@ -6,7 +6,6 @@ namespace MongoDB\Laravel\Tests;
 
 use Carbon\Carbon;
 use DateTime;
-use DateTimeImmutable;
 use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Support\Facades\Date;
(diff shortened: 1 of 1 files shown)


tests/ModelTest.php
```diff
@@ -657,161 +656,80 @@ public function testDates(): void
...
-        $this->assertInstanceOf(Carbon::class, $user->birthday);
-
-        $user = User::create(['name' => 'Jane Doe', 'birthday' => new DateTime('1965-08-08 04.08.37')]);
-        $this->assertInstanceOf(Carbon::class, $user->birthday);
-
-        $user = User::create(['name' => 'Jane Doe', 'birthday' => new DateTime('2010-08-08 04.08.37.324')]);
-        $this->assertInstanceOf(Carbon::class, $user->birthday);
+    public static function provideDate(): \Generator
+    {
+        yield [time()];
```
alcaeus, line 663: I suggest defining names for each data set:
```suggestion
        yield 'int timestamp' => [time()];
```

tests/ModelTest.php
```diff
@@ -657,161 +656,80 @@ public function testDates(): void
...
+        yield ['Monday 8th August 1960 03:12:46 PM'];
+        yield ['2005-08-08'];
+        yield ['1965-08-08'];
+        yield [new DateTime('2010-08-08')];
+        yield [new DateTime('1965-08-08')];
+        yield [new DateTime('2010-08-08 04.08.37')];
+        yield [new DateTime('1965-08-08 04.08.37')];
+        yield [new DateTime('2010-08-08 04.08.37.324')];
+        yield [new DateTime('1965-08-08 04.08.37.324')];
+        yield [new DateTime('1965-08-08 04.08.37.324')];
```
alcaeus, line 675: The last data set is duplicated:
```suggestion
        yield [new DateTime('1965-08-08 04.08.37.324')];
```

tests/ModelTest.php
```diff
@@ -657,161 +656,80 @@ public function testDates(): void
...
-        $this->assertInstanceOf(Carbon::class, $user->birthday);
-
-        $user->setAttribute('birthday', new DateTime('1965-08-08 04.08.37'));
-        $this->assertInstanceOf(Carbon::class, $user->birthday);
-
-        $user->setAttribute('birthday', new DateTime('2010-08-08 04.08.37.324'));
-        $this->assertInstanceOf(Carbon::class, $user->birthday);
-
-        $user->setAttribute('birthday', new DateTime('1965-08-08 04.08.37.324'));
+        $user->setAttribute('birthday', null);
```
alcaeus, line 689: Is this call necessary? If so, please add a comment to explain why it must remain - this looks like something I'd "optimise away".

GromNaN: Yes, to ensure the value modified. Because the value is set to the same previous value. I can add an `assertNull`.

alcaeus: Ah, I see. A comment to remind people that the value has to be modified might be useful then, but an assertion would probably do as well. I'll leave it up to you.

alcaeus (approved):

PR #2593, opened 2023-09-01
make sure $column is string

wivaku:
Native Eloquent allows $column to be stringable(). Not possible when using as array key.

Is issue when using e.g. Filament column search.


GromNaN:
Hello @wivaku, thanks for your contribution. Would you like to add a test for this feature to avoid regression.

wivaku:
Hi @GromNaN , you're welcome. Do you feel a test is needed for casting to a string?  
The result should be a string, and now we're making sure of it.

The casting is similar to other places in the code, e.g. https://github.com/mongodb/laravel-mongodb/blob/ad79fb19faf6be454bc22e54b05346e4829f38db/src/Query/Builder.php#L530

GromNaN:
This is a feature and we need to be sure this feature keep working when we work on the codebase. It's not about testing `$column` is stringified, but testing that other types are accepted. The error could happen somewhere else.
So yes, a test is necessary.

wivaku:
I don't mind adding a test, but it seems I am not exactly sure what + how.

I tried running the tests using `docker compose up`, but the `tests` container exits with:
```
tests    | /usr/local/bin/docker-php-entrypoint: 9: exec: ./vendor/bin/phpunit: not found
tests exited with code 127
```

This is the test that could be added to `tests/QueryBuilderTest.php`:
```
use Illuminate\Support\Str;

// ...

    public function testStringableColumn()
    {
        DB::collection('users')->insert([
            ['name' => 'Jane Doe'],
            ['name' => 'John Doe'],
        ]);

        $stringableColumn = Str::of('name');
        $this->assertEquals('John Doe', DB::collection('users')->where($stringableColumn, 'John Doe')->first());
    }
```

GromNaN (approved):

src/Query/Builder.php
```diff
@@ -1004,8 +1004,14 @@ protected function compileWheres(): array
                 $where['boolean'] = 'or'.(str_ends_with($where['boolean'], 'not') ? ' not' : '');
             }
 
+            // Column name can be a Stringable object.
+            if (isset($where['column']) && $where['column'] instanceof \Stringable) {
```
alcaeus, line 1008: Why the check for `Stringable` here but not in other places? Is there a reason why we can't always cast to `string`?

GromNaN: I needed a `if` statement to check if `column` is set, so I added the `instanceof` to avoid reassigning if the value is already a string. In other places, we know `$column` is set and we are not assigning the casted value.

alcaeus (approved):

GromNaN:
Thank you @wivaku

PR #2595, opened 2023-09-04
Support IPv6 in host config

### LINKED ISSUE #2528: Unable to parse ipv6 host or dsn
Using the latest package version, I can't use ipv6 for host.
It can't parse the host and return this error : 
`No suitable servers found (`serverSelectionTryOnce` set): [Failed to resolve 'PARTIAL_IPV6']`

It doesn't parse properly the address, only selecting the first part before the ":".
If I put the address into [], it doesn't work too.

Is there any solution ?

GromNaN:
Fix #2528 

Response from MongoDB AI chat (internal experimentation):
> To connect to an IPv6 server with the MongoDB PHP driver, you can specify the IPv6 address or hostname in the connection string. For example, if your server is listening on the IPv6 address `2001:0db8:85a3:0000:0000:8a2e:0370:7334` and port `27017`, you can use the following connection string:
>```
>mongodb://[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:27017
>```
> Note that the IPv6 address must be enclosed in square brackets. For more information on connection strings, please refer to the "Connection String URI Format" section of the MongoDB PHP driver documentation in my knowledge.

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -234,9 +234,17 @@ class Connection extends BaseConnection
         $hosts = is_array($config['host']) ? $config['host'] : [$config['host']];
 
         foreach ($hosts as &$host) {
-            // Check if we need to add a port to the host
-            if (strpos($host, ':') === false && ! empty($config['port'])) {
-                $host = $host.':'.$config['port'];
+            // ipv6
+            if (filter_var($host, \FILTER_VALIDATE_IP, \FILTER_FLAG_IPV6)) {
+                $host = '['.$host.']';
+                if (! empty($config['port'])) {
+                    $host = $host.':'.$config['port'];
+                }
+            } else {
+                // Check if we need to add a port to the host
+                if (! str_contains($host, ':') && ! empty($config['port'])) {
+                    $host = $host.':'.$config['port'];
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


alcaeus (approved):

src/Connection.php
```diff
@@ -234,9 +234,17 @@ protected function getHostDsn(array $config): string
         $hosts = is_array($config['host']) ? $config['host'] : [$config['host']];
 
         foreach ($hosts as &$host) {
-            // Check if we need to add a port to the host
-            if (strpos($host, ':') === false && ! empty($config['port'])) {
-                $host = $host.':'.$config['port'];
+            // ipv6
+            if (filter_var($host, \FILTER_VALIDATE_IP, \FILTER_FLAG_IPV6)) {
+                $host = '['.$host.']';
```
jmikola, line 239: I had to check that libmongoc supports this notation, and it is handled in [`_mongoc_host_list_from_string_with_err`](https://github.com/mongodb/mongo-c-driver/blob/1.24.4/src/libmongoc/src/mongoc/mongoc-host-list.c#L221).

GromNaN: I did test before doing the PR 😉

PR #2598, opened 2023-09-04
Implement `MassPrunable` without chunks limit

GromNaN:
Fix https://github.com/mongodb/laravel-mongodb/discussions/2597

Custom implementation of `MassPrunable` is required to prevent using the limit. #2591 added an exception when limited is used because MongoDB Delete operation doesn't support it.

- `MassPrunable::pruneAll()` is called by the command [`model:prune`](https://github.com/laravel/framework/blob/ae93199f5349fd8a02364424249aad2a90185a95/src/Illuminate/Database/Console/PruneCommand.php#L101).
- Using the parent trait is required because it's used to [detect prunable models](https://github.com/laravel/framework/blob/ae93199f5349fd8a02364424249aad2a90185a95/src/Illuminate/Database/Console/PruneCommand.php#L164-L169).

Prunable feature was introducted in Laravel 8.x by laravel/framework#37889.

Users have to be aware that MassPrunable can break relationships as it doesn't call model methods to remove links.

### DIFF
--- a/src/Eloquent/MassPrunable.php
+++ b/src/Eloquent/MassPrunable.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace MongoDB\Laravel\Eloquent;
+
+use Illuminate\Database\Eloquent\MassPrunable as EloquentMassPrunable;
+use Illuminate\Database\Events\ModelsPruned;
+
+trait MassPrunable
+{
+    use EloquentMassPrunable;
+
+    /**
+     * Prune all prunable models in the database.
+     *
+     * @see \Illuminate\Database\Eloquent\MassPrunable::pruneAll()
+     */
+    public function pruneAll(): int
+    {
+        $query = $this->prunable();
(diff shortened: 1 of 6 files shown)


src/Eloquent/MassPrunable.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Illuminate\Database\Eloquent\MassPrunable as EloquentMassPrunable;
+use Illuminate\Database\Events\ModelsPruned;
+
+trait MassPrunable
+{
+    use EloquentMassPrunable;
+
+    /**
+     * Prune all prunable models in the database.
```
GromNaN, line 13: This wording comes Frome Laravel, but I don't like that model=document in their terminology. 
```suggestion
     * Prune all prunable documents in the database.
```

amm834: In my opinion models would be good choice because we all Laravel dev are stick with that, when we think about document where should our documents be stored? It's still under the `models/` directory right?

alcaeus (approved):

PR #2599, opened 2023-09-05
Check if $primaryKey is present before execute unset

FormigTeen:
# Fix: Exception on Dissociate Operation for Embed Models in MongoDB

## Problem:
Previously, while executing a dissociate operation on Embed Models in MongoDB, an Exception was thrown that prevented the removal of Models. This happened specifically when a Model in the list lacked a **$propertyKey**.

## Solution:
This Pull Request adds a validation check to the dissociate operation for Embed Models in MongoDB. With this fix, the application will now verify if a propertyKey is present in each Model in the list before proceeding with the removal. This resolves the issue of an Exception being thrown during the operation.

## Testing
The fix has been thoroughly tested to ensure that the Exception no longer occurs. Now, a model will be successfully dissociated even if another model in the list lacks the **$propertyKey**.

### DIFF
--- a/src/Relations/EmbedsMany.php
+++ b/src/Relations/EmbedsMany.php
@@ -152,7 +152,7 @@ class EmbedsMany extends EmbedsOneOrMany
 
         // Remove the document from the parent model.
         foreach ($records as $i => $record) {
-            if (in_array($record[$primaryKey], $ids)) {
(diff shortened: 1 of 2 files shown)


GromNaN (approved):

FormigTeen:
Done @GromNaN , Thanks, I closed the other request but if you want I can open it by sending only the test case

GromNaN:
Thank you @FormigTeen.

PR #2602, opened 2023-09-06
PHPORM-86: Add support links to New Issue page

alcaeus:
PHPORM-86

This PR adds additional issue types to redirect users to support forums instead of creating issues for support questions.


GromNaN (approved):
LGTM. It defines which channel to use depending on the subject.

PR #2608, opened 2023-09-08
add MorphToMany

ithuis:


src/Eloquent/HybridRelations.php
```diff
@@ -279,7 +281,7 @@ public function belongsToMany(
         }
 
         // Check if it is a relation with an original model.
-        if (! is_subclass_of($related, MongoDBModel::class)) {
+        if (!is_subclass_of($related, MongoDBModel::class)) {
```
GromNaN, line 284: Could you revert this changes using `php vendor/bin/phpcbf`.

GromNaN (commented):
Thank you for working on this major feature. 

There is already a [test for the morph feature](https://github.com/mongodb/laravel-mongodb/blob/9bc8999b79279e46bc92a4e1951a0dc70ce2d63a/tests/RelationsTest.php#L357). If you've found a specific issue that requires overloading this 2 methods, could you add new tests?

ithuis:
sorry, this is not good. Will be back with a working one

PR #2616, opened 2023-09-12
Switch tests to SQLite to remove the need for MySQL server

GromNaN:
All SQL databases are the same, right?


tests/Models/SqlUser.php
```diff
@@ -32,24 +32,21 @@ public function role(): HasOne
...
-        assert($schema instanceof MySqlBuilder);
+        $schema = Schema::connection('sqlite');
+        assert($schema instanceof SQLiteBuilder);
 
-        if ($schema->hasTable('users')) {
-            return;
-        }
-
-        Schema::connection('mysql')->create('users', function (Blueprint $table) {
+        $schema->dropIfExists('users');
```
GromNaN, line 48: Recreate the tables, in case we changed the schema between test runs.

alcaeus (approved):

Issue #2418, opened 2022-07-12
[Feature Request] 


Spartaques:
### Is your feature request related to a problem?
There are no any possibility to do negation.

### Describe the solution you'd like
I would like to use whereNot or orWhereNot to negate everything that is in the query.

### Additional context
For example i have a query:
where('something', '123')->where('something2', '456')

To negate this i can use whereNot instead of where or use it like:
where('something','!=', '123')->where('something2','!=', '456')

as you see whereNot is simpler

PR #2624, opened 2023-09-18
PHPORM-90 Fix whereNot to use `$nor`

### LINKED ISSUE #2623: Error executing whereNot
- Laravel-mongodb Version: 4.0.0-rc1
- PHP Version: 9.1
- Database Driver & Version: 6.0.4

### Description:
Executing whereNot query throws error: `unknown top level operator: $not. If you are trying to negate an entire expression, use $nor.`

GromNaN:
Fix #2623 #2418 [PHPORM-90](https://jira.mongodb.org/browse/PHPORM-90)

[`$not`](https://www.mongodb.com/docs/manual/reference/operator/query/not/) cannot be used with an expression. I switched to [`$nor`](https://www.mongodb.com/docs/manual/reference/operator/query/nor/), that negates a list of expressions.

The query could be optimized, but I expect the mongo server to be smart enough to do it.

Example of optimization I chose not to do because that would require to detect when there is a single field expression (related to https://github.com/GromNaN/laravel-mongodb/pull/13#discussion_r1268289478)

This expression:
```ts
{ field: { operator: value } }
```

Is negated with:
```ts
{ $nor: [ { field: { operator: value } } ] }
```
While it could also be:
```ts
{ field: { $not: { operator: value } }
```

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -1053,8 +1053,9 @@ class Builder extends BaseBuilder
             $method = 'compileWhere' . $where['type'];
             $result = $this->{$method}($where);
 
+            // Negate the expression
             if (str_ends_with($where['boolean'], 'not')) {
-                $result = ['$not' => $result];
+                $result = ['$nor' => [$result]];
             }
 
             // Wrap the where with an $or operator.
--- a/tests/QueryTest.php
+++ b/tests/QueryTest.php
@@ -170,12 +170,17 @@ class QueryTest extends TestCase
         $users = User::whereNot('title', '!=', 'admin')->get();
         $this->assertCount(3, $users);
 
+        // nested negation
+        $users = User::whereNot(fn (Builder $builder) => $builder
+            ->whereNot('title', 'admin'))->get();
+        $this->assertCount(3, $users);
+
         // explicit equality operator
         $users = User::whereNot('title', '=', 'admin')->get();
         $this->assertCount(6, $users);
 
         // custom query operator
-        $users = User::whereNot('title', ['$eq' => 'admin'])->get();
+        $users = User::whereNot('title', ['$in' => ['admin']])->get();
         $this->assertCount(6, $users);
 
         // regex
--- a/tests/QueryTest.php
+++ b/tests/QueryTest.php
@@ -6,6 +6,8 @@ namespace MongoDB\Laravel\Tests;
 
 use DateTimeImmutable;
 use LogicException;
+use MongoDB\BSON\Regex;
+use MongoDB\Laravel\Eloquent\Builder;
 use MongoDB\Laravel\Tests\Models\Birthday;
 use MongoDB\Laravel\Tests\Models\Scoped;
 use MongoDB\Laravel\Tests\Models\User;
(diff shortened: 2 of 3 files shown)


src/Query/Builder.php
```diff
@@ -1053,8 +1053,9 @@ protected function compileWheres(): array
             $method = 'compileWhere' . $where['type'];
             $result = $this->{$method}($where);
 
+            // Negate the expression
             if (str_ends_with($where['boolean'], 'not')) {
-                $result = ['$not' => $result];
+                $result = ['$nor' => [$result]];
```
jmikola, line 1058: I don't see [PHPORM-90](https://jira.mongodb.org/browse/PHPORM-90) linked to anything. This this bug predate us taking over the library, or was it introduced during some recent query refactoring? If the latter, we should link it up.

src/Query/Builder.php
```diff
@@ -1053,8 +1053,9 @@ protected function compileWheres(): array
             $method = 'compileWhere' . $where['type'];
             $result = $this->{$method}($where);
 
+            // Negate the expression
```
jmikola, line 1056: Is `$result` always going to be an `<expression>` at this point, and never an `<operator-expression>` as it expected by [$not](https://www.mongodb.com/docs/manual/reference/operator/query/not/)?

This is related to my other comment, in that I'm curious how this wasn't discovered sooner unless it was only recently introduced.

tests/QueryTest.php
```diff
@@ -154,6 +156,43 @@ public function testSelect(): void
         $this->assertNull($user->age);
     }
 
+    public function testWhereNot(): void
+    {
+        // implicit equality operator
+        $users = User::whereNot('title', 'admin')->get();
+        $this->assertCount(6, $users);
```
GromNaN, line 163: This test is failing if I run it on version 3.9.3: the generated query was `{ title: 'admin' }`, which is wrong because it ignore the negation.
In the current master, the query is `{ $not: { title: 'admin' } }`, which is invalid and rejected by the server.
With this patch, the working query is `{ $nor: [ { title: 'admin' } ] }`.

src/Query/Builder.php
GromNaN: Yes, `compileWhere*` function always return expressions with `$and`, `$or` or a field name, and now `$nor`.

src/Query/Builder.php
GromNaN: I implemented `whereNot` in https://github.com/GromNaN/laravel-mongodb/pull/13. The issue [PHPORM-49](https://jira.mongodb.org/browse/PHPORM-49) is now linked in Jira.

tests/QueryTest.php
jmikola: >  reviewing the generated query was { title: 'admin' }, which is wrong because it ignore the negation.

I'm not sure if I realized this when originally reviewing [PHPORM-49](https://jira.mongodb.org/browse/PHPORM-49), but the fact that earlier versions completely ignored negations with no feedback to the user sounds like pretty disastrous bug.

Issue #2634, opened 2023-10-05
Mutator can not handle array


ithuis:
- Laravel-mongodb Version: 10
- PHP Version: 8.1
- Database Driver & Version: 4

### Steps to reproduce
create a mutator in a modal class that handles an array

```
  protected function roles(): Attribute
    {
        return Attribute::make(
            set: fn (array $value) => array_unique($value)
       );
}
```

### Actual behaviour
it throws an error in update methode  in file MongoDB\Laravel\Query\Builder :: 664

`str_starts_with(): Argument #1 ($haystack) must be of type string, int given`

the current code is 
`
 if (str_starts_with($key, '$')) {
                continue;
            }
`

maybe?
`
 if (is_string($key) && str_starts_with($key, '$')) {
                continue;
            }
`

GromNaN:
Hello, your patch looks good. Could you create a PR and add a test?

PR #2635, opened 2023-10-07
Mutator can not handle array

hans-thomas:
Hi,
As mentioned in #2634, creating a mutator for the `array` type attribute will throw an error while setting the value to the attribute. This error will be thrown because of the invalid value prepared for the update action. So I override `normalizeCastClassResponse` method from `Illuminate\Database\Eloquent\Concerns\HasAttributes` trait and force it to add the key name all the time.

So if we have something like this on our model class
```php
protected function skills(): Attribute
{
    return Attribute::make(
        get: fn ($value) => $value,
        set: fn ($values) => array_unique($values)
    );
}
```
we do not get an error anymore.

### DIFF
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -20,6 +20,7 @@ use MongoDB\Laravel\Query\Builder as QueryBuilder;
 
 use function abs;
 use function array_key_exists;
+use function array_key_first;
 use function array_keys;
 use function array_unique;
 use function array_values;
@@ -254,7 +255,11 @@ abstract class Model extends BaseModel
     /** @inheritdoc */
     protected function normalizeCastClassResponse($key, $value)
     {
-        return [$key => $value];
+        if (! is_array($value) || (count($value) && is_numeric(array_key_first($value)))) {
+            return [$key => $value];
+        }
+
+        return $value;
     }
 
     /** @inheritdoc */
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -252,37 +252,6 @@ abstract class Model extends BaseModel
     }
 
     /** @inheritdoc */
-    protected function setAttributeMarkedMutatedAttributeValue($key, $value)
-    {
-        $attribute = $this->{Str::camel($key)}();
-
-        $callback = $attribute->set ?: function ($value) use ($key) {
-            $this->attributes[$key] = $value;
-        };
-
-        $normalizeCastClassResponse = $this->normalizeCastClassResponse(
-            $key, $callback($value, $this->attributes)
-        );
-
-        if (count($normalizeCastClassResponse) > 1) {
-            $normalizeCastClassResponse = [$key => $normalizeCastClassResponse];
-        }
-
-        $this->attributes = array_merge(
-            $this->attributes,
-            $normalizeCastClassResponse
-        );
-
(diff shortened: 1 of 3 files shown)


src/Eloquent/Model.php
```diff
@@ -251,6 +251,12 @@ public function attributesToArray()
         return $attributes;
     }
 
+    /** @inheritdoc */
+    protected function normalizeCastClassResponse($key, $value)
+    {
+        return [$key => $value];
```
GromNaN: At first I was in favor of applying your fix, but it would mean no longer supporting casters that return `[ key => value ]` to change the key name. That would break for people relying on this behavior. For example, if they implement the alternative solution proposed in the other comment.

tests/Models/User.php
```diff
@@ -109,6 +109,14 @@ protected function username(): Attribute
         );
     }
 
+    protected function skills(): Attribute
+    {
+        return Attribute::make(
+            get: fn ($value) => $value,
+            set: fn ($values) => array_unique($values)
```
GromNaN, line 118: If you set the field name as key, the patch is not needed.

```suggestion
            set: fn ($values) => ['skills' => array_unique($values)]
```

GromNaN (changes requested):

src/Eloquent/Model.php
hans-thomas: When I was changing it, I was in doubt that it could be a breaking change but I didn't get any error when I ran unit tests.
First, can you give me a sample code for the feature that you mentioned? so I can write tests for that to prevent a breaking change in the future. (I really didn't get what you mean).
Second, when I write tests for that feature, I will figure out this issue.
I think I can solve this issue but first I want to test the feature that you mentioned.

tests/Models/User.php
hans-thomas: Yes I know, but it's not the best solution and it can lead to a bug when the developer renames the `skills` field. I prefer to fix this issue in the right way. Since this package is made on top of eloquent, it would be better to avoid making unexpected behavior.

GromNaN: For the risk in refactoring, the method could be:
```suggestion
            set: fn ($values) => [__FUNCTION__ => array_unique($values)]
```

I'll have a new look at the solution.

hans-thomas: I want to fix this issue without determining a key because I believe it's not the correct way. I know I can set the key dynamically but it's not what I want to.

hans-thomas:
@GromNaN I think this would be fix this issue. Now it's supporting mutators that return `[ key => value ]` to change the key name and also mutators that return an array of values. I write tests for both to prevent any breaking changes in the future.

tests/Models/User.php
GromNaN: Unfortunately, Laravel designed the functionality this way, without considering that a field can contain an array. Considering on the one hand that a field name can be an int (see #2642), and on the other that the array you want to store can be anything (a sequential list, disordered int or string keys), you can't detect properly without introducing random behavior depending on the value.

I understand that this is not an ideal experience for the developer. But why should `[1, 2, 3]` be treated as a single value and `['foo' => 1, 2, 9 => 3]` as a definition of the root fields `foo`, `0` and `9`?

GromNaN (changes requested):

hans-thomas:
So if the field's key can be an integer, this can't fix the problem and I think there is no way to solve this. Thank you for your time❤️

PR #2642, opened 2023-10-15
PHPORM-100 Support query on numerical field names

### LINKED ISSUE #2639: setAttribute does not accept key as integer
- Laravel-mongodb Version: 10.25.2
- PHP Version: 8.2.11
- Database Driver & Version: 4

### Description:
When triyng to set a key of object as integer on update it throws a type error

### Steps to reproduce
Set a key as integer on updateOrCreate
`Model::updateOrCreate(['_id' => 1], [ 1 => "value"])`

### Actual behaviour
It throws an error in MongoDB\Laravel\Eloquent\Model::216

`str_contains(): Argument #1 ($haystack) must be of type string, int given`

My sugesstion is change the code from 

`if (str_contains($key, '.')) {`

to

`if (is_string($key) && str_contains($key, '.')) {`

GromNaN:
Fix PHPORM-100
Fix #2639
Related to #2634

- Fix filter on numeric field names.
- Expand the "Too few arguments" exception in `Query::where()` to all scalar values instead of only string as column.
- Cast to field name to string when calling `str_*` functions.

This bug was existing before 4.0.0, this change could be considered a new feature for 4.1 or merged as a bugfix in 4.0

### DIFF
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -209,14 +209,14 @@ abstract class Model extends BaseModel
     /** @inheritdoc */
     public function setAttribute($key, $value)
     {
+        $key = (string) $key;
+
         // Convert _id to ObjectID.
         if ($key === '_id' && is_string($value)) {
             $builder = $this->newBaseQueryBuilder();
 
             $value = $builder->convertKey($value);
         }
-
-        $key = (string) $key;
 
         // Support keys in dot notation.
         if (str_contains($key, '.')) {
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -169,6 +169,8 @@ abstract class Model extends BaseModel
             return null;
         }
 
+        $key = (string) $key;
+
         // An unset attribute is null or throw an exception.
         if (isset($this->unset[$key])) {
             return $this->throwMissingAttributeExceptionIfApplicable($key);
@@ -194,6 +196,8 @@ abstract class Model extends BaseModel
     /** @inheritdoc */
     protected function getAttributeFromArray($key)
     {
+        $key = (string) $key;
+
         // Support keys in dot notation.
         if (str_contains($key, '.')) {
             return Arr::get($this->attributes, $key);
@@ -211,6 +215,8 @@ abstract class Model extends BaseModel
 
             $value = $builder->convertKey($value);
         }
+
+        $key = (string) $key;
 
         // Support keys in dot notation.
         if (str_contains($key, '.')) {
@@ -314,6 +320,8 @@ abstract class Model extends BaseModel
     /** @inheritdoc */
     public function offsetUnset($offset): void
     {
+        $offset = (string) $offset;
+
         if (str_contains($offset, '.')) {
             // Update the field in the subdocument
             Arr::forget($this->attributes, $offset);
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -953,7 +953,20 @@ class Builder extends BaseBuilder
         return $id;
     }
 
-    /** @inheritdoc */
+    /**
+     * Add a basic where clause to the query.
+     *
+     * If 1 argument, the signature is: where(array|Closure $where)
+     * If 2 arguments, the signature is: where(string $column, mixed $value)
+     * If 3 arguments, the signature is: where(string $colum, string $operator, mixed $value)
+     *
+     * @param  Closure|string|array $column
+     * @param  mixed                $operator
+     * @param  mixed                $value
+     * @param  string               $boolean
+     *
+     * @return $this
+     */
     public function where($column, $operator = null, $value = null, $boolean = 'and')
     {
         $params = func_get_args();
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -45,8 +45,8 @@ use function get_debug_type;
 use function implode;
 use function in_array;
 use function is_array;
+use function is_callable;
 use function is_int;
-use function is_scalar;
 use function is_string;
 use function md5;
 use function preg_match;
@@ -967,8 +967,8 @@ class Builder extends BaseBuilder
             }
         }
 
-        if (func_num_args() === 1 && is_scalar($column)) {
-            throw new ArgumentCountError(sprintf('Too few arguments to function %s(%s), 1 passed and at least 2 expected when the 1st is not an array.', __METHOD__, var_export($column, true)));
+        if (func_num_args() === 1 && ! is_array($column) && ! is_callable($column)) {
+            throw new ArgumentCountError(sprintf('Too few arguments to function %s(%s), 1 passed and at least 2 expected when the 1st is not an array or a callable', __METHOD__, var_export($column, true)));
         }
 
         return parent::where(...$params);
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -971,6 +971,10 @@ class Builder extends BaseBuilder
             throw new ArgumentCountError(sprintf('Too few arguments to function %s(%s), 1 passed and at least 2 expected when the 1st is not an array or a callable', __METHOD__, var_export($column, true)));
         }
 
+        if (! is_int($column) && ! is_string($column)) {
+            throw new InvalidArgumentException(sprintf('First argument of %s must be a column name as "string". Got "%s"', __METHOD__, get_debug_type($column)));
+        }
+
         return parent::where(...$params);
     }
 
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -23,13 +23,11 @@ use MongoDB\BSON\Regex;
 use MongoDB\BSON\UTCDateTime;
 use MongoDB\Driver\Cursor;
 use RuntimeException;
-use Stringable;
 
 use function array_fill_keys;
 use function array_is_list;
 use function array_key_exists;
 use function array_merge;
-use function array_merge_recursive;
 use function array_values;
 use function array_walk_recursive;
 use function assert;
@@ -47,6 +45,7 @@ use function implode;
 use function in_array;
 use function is_array;
 use function is_int;
+use function is_scalar;
 use function is_string;
 use function md5;
 use function preg_match;
@@ -60,6 +59,7 @@ use function str_starts_with;
 use function strlen;
 use function strtolower;
 use function substr;
+use function var_export;
 
 class Builder extends BaseBuilder
 {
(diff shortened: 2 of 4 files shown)


src/Query/Builder.php
```diff
@@ -1071,7 +1071,14 @@ protected function compileWheres(): array
             }
 
             // Merge the compiled where with the others.
-            $compiled = array_merge_recursive($compiled, $result);
+            // array_merge_recursive can't be used here because it converts int keys to sequential int.
```
GromNaN, line 1098: `array_merge_recursive` get any int-key as a list and reset the value to `0`.

The merge logic have to be fine-coded to only merge lists of queries in logical operator (`$and`, `$or`, `$nor`).

src/Eloquent/Model.php
```diff
@@ -169,6 +169,8 @@ public function getAttribute($key)
             return null;
         }
 
+        $key = (string) $key;
```
GromNaN, line 172: These type casts can be removed once we add types on the method argument.

GromNaN:
Test failure to investigate: `MongoDB\Driver\Exception\UnexpectedValueException: Detected recursion for field path "_id.$in.0"`

tests/Query/BuilderTest.php
```diff
@@ -1175,10 +1180,16 @@ public static function provideExceptions(): iterable
...
         yield 'find with single string argument' => [
             ArgumentCountError::class,
-            'Too few arguments to function MongoDB\Laravel\Query\Builder::where("foo"), 1 passed and at least 2 expected when the 1st is a string',
+            'Too few arguments to function MongoDB\Laravel\Query\Builder::where(\'foo\'), 1 passed and at least 2 expected when the 1st is a scalar',
             fn (Builder $builder) => $builder->where('foo'),
         ];
 
+        yield 'find with single numeric argument' => [
+            ArgumentCountError::class,
+            'Too few arguments to function MongoDB\Laravel\Query\Builder::where(123), 1 passed and at least 2 expected when the 1st is a scalar',
```
GromNaN: Should be `is not an array`.

tests/ModelTest.php
```diff
@@ -957,4 +957,20 @@ public function testEnumCast(): void
...
+        $user      = new User();
+        $user->{1} = 'one';
+        $user->{2} = ['2' => 'two.two'];
+        $user->save();
+
+        $found = User::where(1, 'one')->first();
+        $this->assertInstanceOf(User::class, $found);
+        $this->assertEquals('one', $found[1]);
+
+        $found = User::where('2.2', 'two.two')->first();
```
GromNaN: Will change to `2.3`.

alcaeus: I believe it's worth leaving this test in place but adding a separate test that calls `where` with a float value:
```php
$found = User::where(2.2, 'two.two')->first();
```

I believe it's utter madness to expect this to work, but at least we'll have the tests in place for when we start slapping deprecation warnings on this kind of type juggling.

src/Query/Builder.php
```diff
@@ -993,17 +994,15 @@ protected function compileWheres(): array
...
-            if (isset($where['column']) && $where['column'] instanceof Stringable) {
+            if (isset($where['column'])) {
                 $where['column'] = (string) $where['column'];
             }
 
             // Convert id's.
             if (isset($where['column']) && ($where['column'] === '_id' || str_ends_with($where['column'], '._id'))) {
                 if (isset($where['values'])) {
                     // Multiple values.
-                    foreach ($where['values'] as &$value) {
```
GromNaN, line 1009: This passing of the value by reference confused me.

Before the new loop, `$result = ['$in' => [ObjectId, ObjectId, '']]`.
Screenshot 2023-10-17 at 23 38 41

When the loop starts, `$result = ['$in' => [ObjectId, ObjectId, &$result]]`. With recursion.
Screenshot 2023-10-17 at 23 39 10

src/Query/Builder.php
```diff
@@ -961,8 +962,8 @@ public function where($column, $operator = null, $value = null, $boolean = 'and'
             }
         }
 
-        if (func_num_args() === 1 && is_string($column)) {
-            throw new ArgumentCountError(sprintf('Too few arguments to function %s("%s"), 1 passed and at least 2 expected when the 1st is a string.', __METHOD__, $column));
+        if (func_num_args() === 1 && is_scalar($column)) {
```
alcaeus: If we're expecting an array, I'd specifically check for that:

```suggestion
        if (func_num_args() === 1 && ! is_array($column)) {
```

src/Query/Builder.php
alcaeus: Yeah, foreach with references is always dodgy, I appreciate the refactoring 👍

src/Query/Builder.php
GromNaN: Fixed.

src/Eloquent/Model.php
```diff
@@ -212,6 +216,8 @@ public function setAttribute($key, $value)
             $value = $builder->convertKey($value);
         }
 
+        $key = (string) $key;
```
jmikola: Do you want to move this to the top of the function for consistency with other methods? It won't affect the `// Convert _id to ObjectID` logic at all, and my understanding is this is the best we can do in lieu of adding a `string` type hint to the `$key` argument, which would be a BC break due to our use of strict types.

Also, perhaps that comment should be changed since `Builder::convertKey()` also converts strings to Binary UUIDs.

src/Query/Builder.php
```diff
@@ -961,8 +962,8 @@ public function where($column, $operator = null, $value = null, $boolean = 'and'
             }
         }
 
-        if (func_num_args() === 1 && is_string($column)) {
-            throw new ArgumentCountError(sprintf('Too few arguments to function %s("%s"), 1 passed and at least 2 expected when the 1st is a string.', __METHOD__, $column));
+        if (func_num_args() === 1 && ! is_array($column) && ! is_callable($column)) {
```
jmikola, line 986: I realize this is unrelated to the change, but why does the preceding `// Remove the leading $ from operators` condition require 3+ arguments if it only interacts with the second `$operator` param?

Why would the absence of an explicit `$value` argument result in `$operator` being left as-is?

tests/ModelTest.php
```diff
@@ -957,4 +957,24 @@ public function testEnumCast(): void
         $this->assertSame(MemberStatus::Member->value, $check->getRawOriginal('member_status'));
         $this->assertSame(MemberStatus::Member, $check->member_status);
     }
+
+    public function testNumericFieldName(): void
+    {
+        $user      = new User();
```
jmikola, line 977: Does `phpcs` enforce this alignment or is this something you did on your own? Just curious, as I've only seen this enforced by `clang-format` in PHPC.

tests/ModelTest.php
```diff
@@ -957,4 +957,24 @@ public function testEnumCast(): void
...
+
+        $found = User::where(1, 'one')->first();
+        $this->assertInstanceOf(User::class, $found);
+        $this->assertEquals('one', $found[1]);
+
+        $found = User::where('2.3', 'two.three')->first();
+        $this->assertInstanceOf(User::class, $found);
+        $this->assertEquals([3 => 'two.three'], $found[2]);
+
+        $found = User::where(2.3, 'two.three')->first();
```
jmikola: I hate everything about this. I understand the original request was to support integers, but do we really need to accept floats in this fashion? That seems like something we could outright reject, unless it's problematic.

This made me think of how MongoDB now supports field names with [dots and dollars](https://www.mongodb.com/docs/manual/core/dot-dollar-considerations/) (related spec change in https://github.com/mongodb/specifications/commit/851ca10388cced05117ceace0ced4b7e1aac02c4). Querying still requires special operators, but inserting does not. I'm not sure this comes into play with this library, but I wouldn't want to allow someone to use a float to insert a field this way and would much rather force them to format it as a string.

src/Eloquent/Model.php
GromNaN: Cast moved.

> Also, perhaps that comment should be changed since `Builder::convertKey()` also converts strings to Binary UUIDs.

I don't understand what you mean.

src/Query/Builder.php
GromNaN: Because we want to convert `Model::where('foo', '$gt', 5)` to `Model::where('foo', 'gt', 5)`, but not `Model::where('foo', '$dollar')` which will result in an equality `{ foo: '$dollar' }`

tests/ModelTest.php
GromNaN: Yes, phpcs: `Generic.Formatting.MultipleStatementAlignment.NotSame`.

tests/ModelTest.php
GromNaN: @alcaeus [you asked](https://github.com/mongodb/laravel-mongodb/pull/2642#discussion_r1360263305) for this test case.

I can't add an exception here for something that "works". But sure, using a float here is totally missleading as the dot is understood as a subfields separator.

GromNaN: I changed the logic to throw an exception when the column is not a string or an int.

src/Eloquent/Model.php
jmikola: The comment here said "Convert _id to ObjectID" but `convertKey()` _also_ converts UUIDs. Small detail (and most people won't be using UUID identifiers), so no objection if it remains as-is.

src/Query/Builder.php
jmikola: Ah, so the signature is really `where($column, $value)` when two arguments are supplied, and `where($column, $operator, $value)` when receiving three arguments. That was lost on me.

I realize this is probably an Eloquent API, but a comment in this method might serve as a helpful reminder.

src/Query/Builder.php
```diff
@@ -966,8 +967,12 @@ public function where($column, $operator = null, $value = null, $boolean = 'and'
...
+            throw new ArgumentCountError(sprintf('Too few arguments to function %s(%s), 1 passed and at least 2 expected when the 1st is not an array or a callable', __METHOD__, var_export($column, true)));
+        }
+
+        if (! is_int($column) && ! is_string($column)) {
+            throw new InvalidArgumentException(sprintf('First argument of %s must be a column name as "string". Got "%s"', __METHOD__, get_debug_type($column)));
```
jmikola: I assume you intentionally do not want to tell people that an "int" is accepted, since you silently convert it. SGTM.

jmikola (approved):

PR #2647, opened 2023-10-20
fix GitHub workflow badge URL

shoito:
fix GitHub workflow badge URL.

See https://github.com/badges/shields/issues/8671


GromNaN (approved):
Thank you.

PR #2396, opened 2022-05-28
Allow to sync a single model instance

mrneatly:
Fix `sync()` method not handling properly single model's instances passed via `$ids` argument

PR #2648, opened 2023-10-21
Handling single model in sync method

### LINKED ISSUE #2395: `syncWithoutDetaching()` isn't accepting a single model as an argument
- Laravel Version: 9.13.0
- Laravel-mongodb Version: 3.9.0
- PHP Version: 8.0.17
- Database Driver & Version: MongoDB PHP, v5.0.6

### Description:
When passing a model instance to a `syncWithoutDetaching()` method, an exception is thrown.
This behaviour differs from cases with Laravel's Eloquent models, when a single model is accepted by this method.

### Steps to reproduce
*Let's use a `User` and a `Post` entity for this example*
1. Create a `Post` model
2. Create a `User` model with a one-to-many relation to `Post` one
3. Store a record of a single user in DB (without any related posts)
4. Store a record of a single post in DB
5. Pick a post record from a previous step via `Post` model
6. Try to assign this post to a user like this: `$user->posts()->syncWithoutDetaching($post);`

### Expected behaviour
An existing post record should be put in a relationship with a user record

### Actual behaviour
An exception is thrown instead:
```
Jenssegers\Mongodb\Relations\BelongsToMany::formatSyncList(): Argument #1 ($records) must be of type array, App\Models\Representative given
```

hans-thomas:
Hi, `sync` and `syncWithoutDetaching` methods now accept a single model as an argument.
Related issue (fix #2395) and PR (fix #2396).

### DIFF
--- a/src/Relations/BelongsToMany.php
+++ b/src/Relations/BelongsToMany.php
@@ -117,6 +117,8 @@ class BelongsToMany extends EloquentBelongsToMany
 
         if ($ids instanceof Collection) {
             $ids = $ids->modelKeys();
+        } elseif ($ids instanceof Model) {
+            $ids = $this->parseIds($ids);
         }
 
         // First we need to attach any of the associated models that are not currently
--- a/tests/RelationsTest.php
+++ b/tests/RelationsTest.php
@@ -255,6 +255,36 @@ class RelationsTest extends TestCase
         $this->assertCount(1, $client->users);
     }
 
+    public function testSyncBelongsToMany()
+    {
+        $user = User::create(['name' => 'John Doe']);
+
+        $first  = Client::query()->create(['name' => 'Hans']);
+        $second = Client::query()->create(['name' => 'Thomas']);
+
+        $user->load('clients');
+        self::assertEmpty($user->clients);
+
+        $user->clients()->sync($first);
+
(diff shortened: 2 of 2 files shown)


hans-thomas:
@GromNaN Can you please review this PR?

src/Relations/BelongsToMany.php
```diff
@@ -117,6 +117,8 @@ public function sync($ids, $detaching = true)
 
         if ($ids instanceof Collection) {
             $ids = $ids->modelKeys();
```
GromNaN, line 119: I guess there is an other issue with this call. It should be `$ids = $this->parseIds($ids);` also. Otherwise it's getting the model `_id` instead of the `relatedKey`.

Do you want to add an other test with a customized `relatedKey`? That will require a new model class for tests, but will ensure we use the correct ID field.

GromNaN (approved):
Looks good. Thanks.

GromNaN (approved):
Thank you @hans-thomas

src/Relations/BelongsToMany.php
hans-thomas: Sorry for the delay, I was super busy with other issues. I will add your requested tests in another PR later.

Issue #2651, opened 2023-10-24
Fix cast issue


rbruhn:
As reported in #2257 , #2199 , #1580 there is an issue with casting and saving to the database. @alcaeus asked for a test example in a new issue. A sample test was provided by @hans-thomas but I figured I would add one here that's more inline with current tests.
The pull request from #2257 works.

First, change User model to make age and integer. This should not effect any other tests as an integer is used in those tests.
```
protected $casts            = [
    'birthday' => 'datetime',
    'entry.date' => 'datetime',
    'member_status' => MemberStatus::class,
    'age' => 'int'
];
```

In ModelTest, add the following test. Here we set age as string. It will fail because the cast type is not being saved to the database.
```
public function testCasting(): void
{
    $user        = new User();
    $user->name  = 'John Doe';
    $user->title = 'admin';
    $user->age   = '35';

    $user->save();

    $this->assertTrue($user->exists);
    $this->assertEquals(1, User::count());

    $raw = $user->getAttributes();
    $this->assertIsInt($raw['age']);
}
```

hans-thomas:
To merge #2257 PR, I think we might need to override some methods used in [castAttribute](https://github.com/laravel/framework/blob/61ea0bc68c4f293194e3132b0bae8af01f41aec7/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L733-L795) method and writing tests to cover every data type declared in `castAttribute` method.

rbruhn:
@hans-thomas Or, if the ```parent::setAttribute``` is not casting (which it's not) perhaps it's the wrong method to extend.

hans-thomas:
The `setAttribute` method casts attributes on retrieving from DB. Because of that, attributes were not cast in creating or updating actions. That's why we need this PR #2257. However, as I said, we need to write tests to cover all supported data types.

rbruhn:
@hans-thomas I did some reading. Learn something new everyday. Some info states casts are used used in saving/updating. Some info states the opposite. I know this: When testing an eloquent model, ```castAttribute()``` is never called when saving or updating.

This changes my mind on this issue. If Laravel doesn't cast when saving/updating to the database, neither should this package. It should be left up to the user to use Attribute (which I originally did to solve this issue).

Since I opened this issue, I'm now closing it. If someone wants this functionality changed, they can open a new issue.

hans-thomas:
[...]

Issue #2393, opened 2022-05-26
Unable to cast to native type on unknown schema migration


sts-ryan-holton:
- Laravel-mongodb Version: 3.9.0
- PHP Version: 8.0.2
- Database Driver & Version: MongoDB community version 5.0

### Description:

I've chosen Mongo DB as a database for my Laravel project for the primary reason of it supporting schema-less data, which is what my project is using.

I'm saving data in my project from another system and have hundreds of fields, so using Mongo DB is perfect for this.

For this reason, I cannot create a migration as such with every possible field with their types as I do not know them, so, my migration simply contains the Laravel specifics with the id, and timestamps, maybe I'll add a few more:

My migration is called: **2022_05_25_133123_create_applications_table.php**

```php
/**
 * Run the migrations.
 *
 * @return void
 */
public function up()
{
    Schema::create('applications', function (Blueprint $collection) {
        $collection->id();
        $collection->timestamps();
    });
}
```

My model, is called `Application`, and it's here where I'm trying to cast a field that's been saved:

```php
<?php

namespace App\Models;

use Illuminate\Database\Eloquent\Factories\HasFactory;
use Jenssegers\Mongodb\Eloquent\Model;

class Application extends Model
{
    use HasFactory;

    /**
    * The collection associated with the model.
    *
    * @var string
    */
    protected $collection = 'applications';

    /**
     * The attributes that aren't mass assignable.
     *
     * @var array
     */
    protected $guarded = [
        'id',
        'created_at',
        'updated_at'
    ];

    /**
     * The attributes that should be cast.
     *
     * @var array
     */
    protected $casts = [
        'processing_duration_secs' => 'integer'
    ];

    /*
    ** Linked application payday
    */
    public function payday()
    {
        return $this->hasOne(ApplicationPayday::class);
    }

    /*
    ** Linked application response
    */
    public function response()
    {
        return $this->hasOne(ApplicationResponse::class);
    }

    /*
    ** Linked application api links
    */
    public function apiLinks()
    {
        return $this->hasMany(ApplicationApiLink::class);
    }
}
```

Data is saved like:

```php
Application::create($this->source['Application']);
```

But my casting when, for instance, looking through tinker, I see that it's still a string?

### Steps to reproduce

1. Set up basic migration without any specific fields
2. Use `Model::create(['processing_duration_secs'=>'5'])` to save a field as a string
3. Add the cast for the field to an integer
4. Open tinker, query the first model and observe that it's not an integer

[...]

PR #2653, opened 2023-10-29
Fix casting issue

hans-thomas:
Hello friends,
If you had a look at this repository, you would find that one of the most reported issues is **failing to cast attributes** on creating/saving data on the database. There are some issues that have been opened since 2018: #1580, #2199, #2257, #2393, #2651.
I applied #2257 PR and specifically wrote tests for each data type that might had an impact.
In this PR, I had to override `asDecimal` and `fromJson` methods from the `Illuminate\Database\Eloquent\Concerns\HasAttributes` trait.


src/Eloquent/Model.php
```diff
@@ -4,16 +4,22 @@
 
 namespace MongoDB\Laravel\Eloquent;
 
+use Brick\Math\BigDecimal;
```
GromNaN, line 7: I discover that [Laravel requires this lib](https://github.com/laravel/framework/blob/10.x/composer.json#L27), which has [no stable version](https://packagist.org/packages/brick/math) (>1).

src/Eloquent/Model.php
```diff
@@ -231,6 +242,28 @@ public function setAttribute($key, $value)
...
+        } catch (BrickMathException $e) {
+            throw new MathException('Unable to cast value to a decimal.', previous: $e);
+        }
+    }
+
+    /** @inheritdoc */
+    public function fromJson($value, $asObject = false)
+    {
+        if (! is_string($value)) {
+            $value = Json::encode($value ?? '');
```
GromNaN, line 267: Why not directly returning the value, instead of encoding and decoding?

The object to array conversion can be done with `array_walk_recursive`.

tests/Casts/BooleanTest.php
```diff
@@ -0,0 +1,58 @@
...
+
+        Casting::truncate();
+    }
+
+    public function testBool(): void
+    {
+        $model = Casting::query()->create(['booleanValue' => true]);
+
+        self::assertIsBool($model->booleanValue);
+        self::assertEquals(true, $model->booleanValue);
```
GromNaN: It is preferable to use the strict `assertSame` when possible.

tests/Casts/BooleanTest.php
```diff
@@ -0,0 +1,58 @@
...
+
+        $model->update(['booleanValue' => false]);
+        $check = Casting::query()->find($model->_id);
+
+        self::assertIsBool($check->booleanValue);
+        self::assertEquals(false, $check->booleanValue);
+
+        $model->update(['booleanValue' => 1]);
+        $check = Casting::query()->find($model->_id);
```
GromNaN: You should check the value just after the update, and before refetching it.

tests/Casts/CollectionTest.php
```diff
@@ -0,0 +1,35 @@
...
+        $model = Casting::query()->create(['collectionValue' => ['g' => 'G-Eazy']]);
+
+        self::assertInstanceOf(Collection::class, $model->collectionValue);
+        self::assertEquals(collect(['g' => 'G-Eazy']), $model->collectionValue);
+
+        $model->update(['collectionValue' => ['Dont let me go' => 'Even the longest of nights turn days']]);
+        $check = Casting::query()->find($model->_id);
+
+        self::assertInstanceOf(Collection::class, $check->collectionValue);
+        self::assertEquals(collect(['Dont let me go' => 'Even the longest of nights turn days']), $check->collectionValue);
```
GromNaN: What's the behavior in Eloquent when the value is `null`? I think we should cast to an empty collection.

tests/Casts/StringTest.php
```diff
@@ -0,0 +1,32 @@
...
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        Casting::truncate();
+    }
+
+    public function testString(): void
+    {
+        $model = Casting::query()->create(['stringContent' => "If I'ma shoot, I shoot to kill"]);
```
GromNaN: Fun but please choose another sentence.

GromNaN (commented):
Thank you for reviewing these issues and proposing a fix. I have a few remarks, and I still need to test your branch. But overall it looks great!
Big up for all the tests.

tests/Models/Casting.php
```diff
@@ -0,0 +1,40 @@
...
+namespace MongoDB\Laravel\Tests\Models;
+
+use MongoDB\Laravel\Eloquent\Model as Eloquent;
+
+class Casting extends Eloquent
+{
+    protected $connection = 'mongodb';
+    protected $collection = 'casting';
+
+    protected $fillable = [
```
GromNaN, line 15: It seems more correct to have used `$fillable` instead of `$unguarded`. Could you merge `CastBinaryUuid` into the generic `Casting` model. I see no reason to separate them.

tests/Models/Casting.php
```diff
@@ -0,0 +1,40 @@
...
+        'jsonValue',
+        'collectionValue',
+        'dateField',
+        'datetimeField',
+    ];
+
+    protected $casts = [
+        'intNumber' => 'int',
+        'floatNumber' => 'float',
+        'decimalNumber' => 'decimal:2',
```
GromNaN, line 34: Is there a case with `decimal` without the prevision part?

tests/Casts/CollectionTest.php
GromNaN: Actually, it's null. That's fine.

tests/Casts/IntegerTest.php
```diff
@@ -0,0 +1,46 @@
...
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        Casting::truncate();
+    }
+
+    public function testInt(): void
+    {
+        $model = Casting::query()->create(['intNumber' => 1]);
```
GromNaN, line 21: A test with a float casted to `int` would be useful for exhautivity.

tests/Casts/JsonTest.php
```diff
@@ -0,0 +1,34 @@
...
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        Casting::truncate();
+    }
+
+    public function testJson(): void
+    {
+        $model = Casting::query()->create(['jsonValue' => ['g' => 'G-Eazy']]);
```
GromNaN, line 23: I don't think JSON cast should be used with MongoDB. It's surprising to store stringified JSON in a MongoDB field.
But since Eloquent offers it, we also want it to work with this package.

tests/Casts/DateTest.php
```diff
@@ -0,0 +1,56 @@
...
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        Casting::truncate();
+    }
+
+    public function testDate(): void
+    {
+        $model = Casting::query()->create(['dateField' => now()]);
```
GromNaN, line 25: Would it make sense to add a test with `\DateTime`? It actually works.

GromNaN (commented):
Do you think there is some error cases that could be added? Like data that can’t be casted.

src/Eloquent/Model.php
hans-thomas: > This library requires PHP 8.0 or later. For PHP 7.4 compatibility, you can use version 0.10. For PHP 7.1, 7.2 & 7.3, you can use version 0.9.

It seems the package is not using semantic versioning. However, it is used by 1.2m other repos.

src/Eloquent/Model.php
hans-thomas: To prevent passing value to `json_encode` every time on creating/updating actions.
Before
```php
Casting::query()->create(['jsonValue' => json_encode(['key' => 'value'])]);
```
After
```php
Casting::query()->create(['jsonValue' => ['key' => 'value']]);
```

GromNaN: It's a good thing you don't force storage in json strings. 

To understand my point, here's what happens when the data is not encoded in JSON:
```php
if (! is_string($value)) {
    return  Json::decode(Json::encode($value ?? ''), ! $asObject);
}
```

Why not return it directly:
```php
if (! is_string($value)) {
    return $value;
}
```

Now I see that a custom json encoder can be used. Feel free to resolve.

tests/Casts/BooleanTest.php
hans-thomas: When we will use refetched data for testing?

GromNaN: Adding an assertion, just like this:
```suggestion
        self::assertSame(true, $model->booleanValue);
        $check = Casting::query()->find($model->_id);
```

hans-thomas: Done in fc504ee.

tests/Models/Casting.php
hans-thomas: There is also `CastObjectId` model that I think can be merged into `Casting`.

GromNaN: There is specific behaviors with the `_id` field. It's safer to keep it separate.

tests/Models/Casting.php
hans-thomas: As mentioned in the #2146 issue, an error will be thrown.

tests/Casts/IntegerTest.php
hans-thomas: Done in 6b81529.

tests/Casts/DateTest.php
hans-thomas: Done in c715194.

tests/Models/Casting.php
hans-thomas: Done in 8d9736b.

tests/Casts/StringTest.php
hans-thomas: Done in 187cfc0.

tests/Casts/BooleanTest.php
hans-thomas: Done in fdaa9dc.

hans-thomas:
> Do you think there is some error cases that could be added? Like data that can’t be casted.

I tried my best to cover any possible failure in casting values and nothing new came to my mind.

GromNaN (approved):
LGTM.

GromNaN:
Thank you @hans-thomas. I merged into 4.1 because the behavior changes and some users may rely on the current behavior.

apeisa:
Thanks @hans-thomas for this! It works, but not in all cases. When using date casts, you can also define the format in the string format, like this ([docs](https://laravel.com/docs/10.x/eloquent-mutators#date-casting)):

```PHP
   protected $casts = [
        'mydate' => 'datetime:j.n.Y H:i',
    ];
```

When using cast like that, it doesn't work. Without formatting set, it does work as expected:

```PHP
   protected $casts = [
        'mydate' => 'datetime',
    ];
```

hans-thomas:
Hello @apeisa, I had some tests on `datetime` cast with custom formatting, It's not from my changes because obviously, I did not change a thing on date casting. Please open a new issue and I will try to fix that. Thank you for your feedback.

apeisa:
Sure thing, here you are https://github.com/mongodb/laravel-mongodb/issues/2655

jmikola:
> I merged into 4.1 because the behavior changes and some users may rely on the current behavior.

@GromNaN: Did you create a PHPORM issue to track this? If not, is it worth doing so for generating release notes down the line, or is that not necessary for this project?

rbruhn:
Coming back to this after some time away.
Am I to assume this fixes "set" when saving data? For example:
```
// Inside Subject Model
protected $casts = [
    'project_id' => 'integer'
];

// Script
$subject = new Subject();
$subject->project_id = "1000";
$subject->save();
```
Because it doesn't and I'm on v4.3.0.
When viewing the record in Mongo, the project_id is Type = string, not integer.

Issue #2655, opened 2023-10-31
Custom datetime casting works, but not if you define custom format


apeisa:
Thanks to #2653 casting works now, but not in all cases. When using date casts, you can also define the format in the string format, like this ([docs](https://laravel.com/docs/10.x/eloquent-mutators#date-casting)):

```PHP
   protected $casts = [
        'mydate' => 'datetime:j.n.Y H:i',
    ];
```

When using cast like that, it doesn't work. Without formatting set, it does work as expected:

```PHP
   protected $casts = [
        'mydate' => 'datetime',
    ];
```

Here is simple class that can be used for testing:

```PHP
<?php

namespace App\Models;

use Carbon\Carbon;
use MongoDB\Laravel\Eloquent\Model;

class Car extends Model
{
    protected $guarded = [];

    protected $casts = [
        'sold_at' => 'datetime:j.n.Y H:i',
    ];

    public function sell()
    {
        $this->sold_at = Carbon::now();
        $this->save();
        return $this;
    }
}
```

And to test this run these commands on Tinker:

```
> $car = App\Models\Car::create();
= App\Models\Car {#8065
    updated_at: MongoDB\BSON\UTCDateTime {#8071
      +"milliseconds": "1698741994126",
    },
    created_at: MongoDB\BSON\UTCDateTime {#8071},
    _id: MongoDB\BSON\ObjectId {#8075
      +"oid": "6540beea13e41d27cd036b52",
    },
  }

> $car->sell();
= App\Models\Car {#8065
    updated_at: MongoDB\BSON\UTCDateTime {#8063
      +"milliseconds": "1698742000466",
    },
    created_at: MongoDB\BSON\UTCDateTime {#8071
      +"milliseconds": "1698741994126",
    },
    _id: MongoDB\BSON\ObjectId {#8075
      +"oid": "6540beea13e41d27cd036b52",
    },
    sold_at: Illuminate\Support\Carbon @1698742000 {#8066
      date: 2023-10-31 08:46:40.461974 UTC (+00:00),
    },
  }
```

As you can see the `sold_at` has Carbon object in database, it should be MongoDB\BSON\UTCDateTime. If you remove the `:j.n.Y H:i` part from `$casts` then it works as expected. Here is tinker output without the date formatting:

```
App\Models\Car {#8065
    updated_at: MongoDB\BSON\UTCDateTime {#8063
      +"milliseconds": "1698742105625",
    },
[...]

PR #2658, opened 2023-10-31
Datetime casting with custom format

hans-thomas:
Hi again,
As mentioned in the #2655 issue, casting date types with a custom format did not work properly. So I decided to cover all date types and fix this issue.
I tested these date types:

 - date with custom formatting
 - immutable_date
 - immutable_date with custom formatting
 - datetime
 - immutable_datetime
 - immutable_datetime with custom formatting

To achieve this goal, I override one of the cases in `castAttribute` method to handle `immutable_date` date type. Also, I override `transformModelValue` method to check if the result is a `DateTimeInterface` or not for applying defined custom formatting on the Carbon instance and resetting time in a custom `date` type.
In the end, I should say that all date values will be stored in DB as `Date` type. (not as an object or string anymore)


hans-thomas:
I have no idea why pipeline is failing.

src/Eloquent/Model.php
```diff
@@ -199,6 +201,33 @@ public function getAttribute($key)
         return parent::getAttribute($key);
     }
 
+    /** @inheritdoc */
+    protected function transformModelValue($key, $value)
+    {
+        $value = parent::transformModelValue($key, $value);
+        if ($this->hasCast($key) && $value instanceof DateTimeInterface) {
+            $value->settings(array_merge($value->getSettings(), ['toStringFormat' => $this->getDateFormat()]));
```
GromNaN, line 213: `settings` is a Carbon method. You can't assume any `DateTimeInterface` has this method.

src/Eloquent/Model.php
```diff
@@ -270,6 +299,24 @@ public function fromJson($value, $asObject = false)
         return Json::decode($value ?? '', ! $asObject);
     }
 
+    /** @inheritdoc */
+    protected function castAttribute($key, $value)
+    {
+        $castType = $this->getCastType($key);
+
+        switch ($castType) {
```
GromNaN: The `match` operator would be perfect for this usage.

src/Eloquent/Model.php
```diff
@@ -270,6 +299,24 @@ public function fromJson($value, $asObject = false)
...
+            case 'immutable_custom_datetime':
+            case 'immutable_datetime':
+                if (str_starts_with($this->getCasts()[$key], 'immutable_date:')) {
```
GromNaN: I don't quite understand how the cast works here: why can you have `immutable_date:` and `immutable_datetime` at the same time?

GromNaN (changes requested):
It's a good start

src/Eloquent/Model.php
hans-thomas: After I added [this](https://github.com/mongodb/laravel-mongodb/blob/698711ce63f8662e1579944ba47854fa0c5366e8/src/Eloquent/Model.php#L221-L223) to Model class, All date attributes in every type such as `\DateTime` or `string` and anything else are casting into a `Carbon` instance. If you have a date attribute but you don't cast it, it will be stored as an object and never reach this line.
However, should I check that `$value` is an instance of `Carbon` instead of `DateTimeInterface` to make sure the `settings` method is available?

src/Eloquent/Model.php
hans-thomas: I used `switch` because of its parent method that used `switch` instead of `match`. I'm going to replace this with `match` though.

src/Eloquent/Model.php
hans-thomas: If you cast a data attribute like `immutable_date`, it's totally okay and works fine, but if you cast it to `immutable_date` with custom formatting like `immutable_date:j.n.Y H:i`, It will be returned as an `immutable_custom_datetime` from `getCastType` method. It's because Laravel doesn't handle the `immutable_custom_date` cast type. I might open a PR later on Laravel but until that time, it will be fixed in this way.

src/Eloquent/Model.php
GromNaN: Yes, please change the type check for `$value instanceof Carbon` to be specific. If you can add a code comment that would be fantastic.

hans-thomas: What should I say in the comment?

src/Eloquent/Model.php
```diff
@@ -199,6 +202,33 @@ public function getAttribute($key)
         return parent::getAttribute($key);
     }
 
+    /** @inheritdoc */
+    protected function transformModelValue($key, $value)
+    {
+        $value = parent::transformModelValue($key, $value);
+        if ($this->hasCast($key) && $value instanceof CarbonInterface) {
```
hans-thomas, line 212: I used `CarbonInterface` to cover both `Carbon` and `CarbonImmutable` instances.

src/Eloquent/Model.php
GromNaN: > "date" cast convert values to Carbon.

GromNaN (approved):

src/Eloquent/Model.php
```diff
@@ -199,6 +202,35 @@ public function getAttribute($key)
         return parent::getAttribute($key);
     }
 
+    /** @inheritdoc */
+    protected function transformModelValue($key, $value)
+    {
+        $value = parent::transformModelValue($key, $value);
+        // Casting attributes to any of date types, will convert that attribute
+        // to a Carbon or CarbonImmutable instance. It's done in setAttribute method L254.
```
GromNaN:
```suggestion
        // to a Carbon or CarbonImmutable instance.
        // @see Model::setAttribute()
```

alcaeus (approved):

GromNaN:
Thank you @hans-thomas

hans-thomas:
@GromNaN Always a Pleasure❤️
Could you please close the mentioned issue?

Issue #2656, opened 2023-10-31
Call to undefined method compileWhereSub()


mehmetgulbilge:
- Laravel-mongodb Version: 4.1
- PHP Version: 8.2
- Database Driver & Version:

BadMethodCallException
Call to undefined method MongoDB\Laravel\Query\Builder::compileWhereSub()

How can I solve this?

mehmetgulbilge:
trace: 
#0 /application/vendor/illuminate/database/Query/Builder.php(3967): Illuminate\\Database\\Query\\Builder::throwBadMethodCallException()\n#1 /application/vendor/mongodb/laravel-mongodb/src/Query/Builder.php(1353): Illuminate\\Database\\Query\\Builder->__call()\n#2 /application/vendor/mongodb/laravel-mongodb/src/Query/Builder.php(1059): MongoDB\\Laravel\\Query\\Builder->__call()\n#3 /application/vendor/mongodb/laravel-mongodb/src/Query/Builder.php(926): MongoDB\\Laravel\\Query\\Builder->compileWheres()\n#4 /application/vendor/mongodb/laravel-mongodb/src/Query/Builder.php(678): MongoDB\\Laravel\\Query\\Builder->performUpdate()\n#5 /application/vendor/mongodb/laravel-mongodb/src/Eloquent/Builder.php(65): MongoDB\\Laravel\\Query\\Builder->update()

hans-thomas:
@GromNaN I reverted my changes in #2658 but the pipeline is still failing.

<details><summary><b>Pipeline's first error</b>:</summary>

```
There were 93 errors:

1) MongoDB\Laravel\Tests\Casts\BooleanTest::testBool
BadMethodCallException: Call to undefined method MongoDB\Laravel\Query\Builder::compileWhereSub()

/home/runner/work/laravel-mongodb/laravel-mongodb/vendor/laravel/framework/src/Illuminate/Support/Traits/ForwardsCalls.php:67
/home/runner/work/laravel-mongodb/laravel-mongodb/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php:3967
/home/runner/work/laravel-mongodb/laravel-mongodb/src/Query/Builder.php:1378
/home/runner/work/laravel-mongodb/laravel-mongodb/src/Query/Builder.php:1077
/home/runner/work/laravel-mongodb/laravel-mongodb/src/Query/Builder.php:929
[...]

PR #2659, opened 2023-11-01
Fix Call to undefined method compileWhereSub()

hans-thomas:
It fixes the #2656 issue. I believe some updates in dependencies caused this break.

### DIFF
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -92,7 +92,7 @@ class Builder extends EloquentBuilder
             return $this->model->getKey();
         }
 
-        return parent::insertGetId($values, $sequence);
+        return $this->toBase()->insertGetId($values, $sequence);
     }
 
     /** @inheritdoc */


src/Eloquent/Builder.php
```diff
@@ -92,7 +92,7 @@ public function insertGetId(array $values, $sequence = null)
             return $this->model->getKey();
         }
 
-        return parent::insertGetId($values, $sequence);
```
hans-thomas, line 95: By calling `parent::insertGetId`, the `insertGetId` will be executed from [here](https://github.com/laravel/framework/blob/d0a6080db78a7e8a3316ff9c41eb86aa0866742d/src/Illuminate/Database/Eloquent/Builder.php#L1967-L1969) on [`Illuminate\Database\Query\Builder`](https://github.com/laravel/framework/blob/d0a6080db78a7e8a3316ff9c41eb86aa0866742d/src/Illuminate/Database/Query/Builder.php#L3379-L3388) but we override this method in our [`MongoDB\Laravel\Query\Builder`](https://github.com/mongodb/laravel-mongodb/blob/698711ce63f8662e1579944ba47854fa0c5366e8/src/Query/Builder.php#L649-L664). So, I fix this problem by calling this method on our `MongoDB\Laravel\Query\Builder`.

GromNaN: It seems to be fixed by [#2661](https://github.com/mongodb/laravel-mongodb/pull/2661/files#diff-5ea252bbf9d6b385257af07ce12346559d3d4e1727eef473c50b081e28e10230R40). Both solutions will work, yours removes a layer of magics but it is more fragile if the implementation changes in Laravel.

PR #2661, opened 2023-11-01
Fix compatibility with Laravel 10.30 — lowercase the `$passthru` array values

### LINKED ISSUE #2656: Call to undefined method compileWhereSub()

### LINKED ISSUE #2660: model id property returns the builder and not the id
- Laravel-mongodb Version: 4.0.0
- PHP Version: 8.2
- Database Driver & Version: Mongo & `db.version()` gives me 7.0.0
- Laravel 10.30.0

I've created https://github.com/mongodb/laravel-mongodb/pull/2661 for this issue.

### Description:

https://github.com/laravel/framework/pull/48852 updated the way method's are called/proxied.
It seems that it requires that the list with methods should be lowercase, as the Mongo Builder class doesn't have that the Builder is returned instead of the method that should have been called.

### Steps to reproduce
1. Laravel app version 10.30.0
2. create connection / model which uses Mongo as backend
3. `$model->id` would return the Builder

### Expected behaviour
When calling `$model->id` I expect a string with the id for the model.

### Actual behaviour
Calling `$model->id` returns a Builder instance.

Treggats:
fixes #2660
fixes #2656

As explained in the issue, Laravel updated the way Builder methods are called. The list of methods in the `$passthru` property now should be lowercase.

Currently they are not, and this causes the methods to not be called, but passed on and thus returning the Builder instance.

This PR makes them all lowercase, which will fix this issue. I have tested this in our application and it works.


src/Eloquent/Builder.php
```diff
@@ -30,24 +30,24 @@ class Builder extends EloquentBuilder
...
+        'insertusing',
         'max',
         'min',
         'pluck',
         'pull',
         'push',
         'raw',
         'sum',
-        'toSql',
+        'tosql',
```
GromNaN: This should be removed. `SQL` doesn't make sense here.

GromNaN (commented):
Hello @Treggats. Thank you for identifying and fixing this bug.
Could you add a test for each method to ensure we don't have regression?

src/Eloquent/Builder.php
```diff
@@ -30,24 +30,24 @@ class Builder extends EloquentBuilder
...
-        'getConnection',
-        'getGrammar',
+        'getbindings',
+        'getconnection',
+        'getgrammar',
         'insert',
-        'insertGetId',
-        'insertOrIgnore',
-        'insertUsing',
+        'insertgetid',
```
GromNaN, line 40: If #2659 is merged, this method will never be called. Would it be better to remove it? Possibly by leaving a comment.
Or maybe discard #2659 and keep this PR as a fix.

Treggats:
> Hello @Treggats. Thank you for identifying and fixing this bug. Could you add a test for each method to ensure we don't have regression?

@GromNaN thanks for the review, I will add tests and address the review comments.

src/Eloquent/Builder.php
Treggats: What you said in the [comment](https://github.com/mongodb/laravel-mongodb/pull/2659/files#r1379745121) in #2659 makes sense, the latter part.
Personally I like to be specific/remove magic, but that's the way Laravel works and we need to align with that.

In short, close #2659 in favour of this one.

> It seems to be fixed by [#2661](https://github.com/mongodb/laravel-mongodb/pull/2661/files#diff-5ea252bbf9d6b385257af07ce12346559d3d4e1727eef473c50b081e28e10230R40). Both solutions will work, yours removes a layer of magics but it is more fragile if the implementation changes in Laravel.

src/Eloquent/Builder.php
Treggats: Do you mean the whole entry (`toSql`) or to leave it camel cased?

Treggats:
a little sidenote, locally I made the following change. I won't commit it as its a personal preference though it doesn't seem to impact the current codebase.
```diff
diff --git a/phpcs.xml.dist b/phpcs.xml.dist
index 36cc870..23bc44a 100644
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -34,5 +34,7 @@
         <exclude name="SlevomatCodingStandard.TypeHints.ParameterTypeHint"/>
         <exclude name="SlevomatCodingStandard.TypeHints.PropertyTypeHint"/>
         <exclude name="SlevomatCodingStandard.TypeHints.ReturnTypeHint"/>
+
+        <exclude name="Generic.Formatting.MultipleStatementAlignment" />
     </rule>
 </ruleset>
```

src/Eloquent/Builder.php
Treggats:
```php
User::query()->toSql();
```
Gives the following error
> BadMethodCallException: This method is not supported by MongoDB. Try "toMql()" instead.

So I guess it should be removed completely. Maybe replace it with the mentioned method, but that's your call.

GromNaN: Yes, remove the method `tosql` from the list.
Adding `tomql` would be helpful.

GromNaN:
You can commit this change. It annoys me too.
It's [excluded from mongodb-php-library](https://github.com/mongodb/mongo-php-library/blob/22e99b88068c88862a63aa65dfe6355db2553a88/phpcs.xml.dist#L42).

Treggats:
> You can commit this change. It annoys me too. It's [excluded from mongodb-php-library](https://github.com/mongodb/mongo-php-library/blob/22e99b88068c88862a63aa65dfe6355db2553a88/phpcs.xml.dist#L42).

I will follow suit then.

hans-thomas:
@GromNaN the #2656 issue is related to this PR too. After merging, don't forget to close it.

Treggats:
> @GromNaN the #2656 issue is related to this PR too. After merging, don't forget to close it.

updated the description so that it closes after merging of this PR

src/Eloquent/Builder.php
```diff
@@ -30,24 +30,24 @@ class Builder extends EloquentBuilder
...
+        'insertusing',
         'max',
         'min',
         'pluck',
         'pull',
         'push',
         'raw',
         'sum',
-        'toSql',
+        'tomql',
```
Treggats, line 50: replaced as it is not supported

tests/Eloquent/CallBuilderTest.php
```diff
@@ -0,0 +1,101 @@
...
+    #[Dataprovider('provideFunctionNames')]
+    public function callingABuilderMethodDoesNotReturnTheBuilderInstance(string $method, string $className, $parameters = []): void
+    {
+        $builder = User::query()->newQuery();
+        assert($builder instanceof Builder);
+
+        self::assertNotInstanceOf(expected: $className, actual: $builder->{$method}(...$parameters));
+    }
+
+    public static function provideFunctionNames(): Generator
```
Treggats, line 35: provide supported methods, which should not return a Builder instance.

tests/Eloquent/CallBuilderTest.php
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    #[Test]
+    #[DataProvider('provideUnsupportedMethods')]
+    public function callingUnsupportedMethodThrowsAnException(string $method, string $exceptionClass, string $exceptionMessage, $parameters = []): void
+    {
+        $builder = User::query()->newQuery();
+        assert($builder instanceof Builder);
+
+        $this->expectException($exceptionClass);
```
Treggats, line 64: verify that certain methods are not supported, checking both the exception type and its message

Treggats:
Currently the `MassPrunableTest::testPruneWithQuery` test is failing locally, I'm trying to figure out why that is.

The `$this->assertEquals(1, User::count());` assertion fails, two models were not pruned. Thus the `count()` query returns 3

GromNaN:
While testing, I just realized that it breaks compatibility with `illuminate/database < 10.30.0`.

One strategy for maintaining compatibility would be to modify this list in `__construct`, adding only the specific methods, in lowercase and non-lowercase versions (leaving a comment that non-lowercase versions are kept for compatibility with Laravel < 10.30).

tests/Eloquent/CallBuilderTest.php
```diff
@@ -0,0 +1,185 @@
...
+
+        yield 'to sql' => [
+            'toSql',
+            BadMethodCallException::class,
+            'This method is not supported by MongoDB. Try "toMql()" instead',
+            [[['name' => 'Jane']], fn (QueryBuilder $builder) => $builder],
+        ];
+
+        yield 'dd' => [
+            'dd',
```
GromNaN, line 94: We can create a separate issue to support this, once this PR get merged.

GromNaN:
> Currently the `MassPrunableTest::testPruneWithQuery` test is failing locally, I'm trying to figure out why that is.
> 
> The `$this->assertEquals(1, User::count());` assertion fails, two models were not pruned. Thus the `count()` query returns 3

Some test may not clean the collection correctly. You can clean them in the `setUp`:

```php
    public function setUp(): void
    {
        User::truncate();
        Soft::truncate();
    }
```

tests/Eloquent/CallBuilderTest.php
```diff
@@ -0,0 +1,185 @@
...
+use MongoDB\Laravel\Tests\TestCase;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Test;
+use RuntimeException;
+
+use function assert;
+
+final class CallBuilderTest extends TestCase
+{
+    #[Test]
```
GromNaN: Some methods may insert data in the collection. You need to add a cleanup:
```php
    {
        User::truncate();
    }
```

src/Eloquent/Builder.php
```diff
@@ -30,24 +30,33 @@ class Builder extends EloquentBuilder
...
+        // Kept for compatibility with Laravel < 10.3
```
Treggats, line 51: And here I was trying to conditionally set the method names casing 😆

GromNaN (approved):
Thank you @Treggats for the time you've devoted to correcting the breaking changes coming from Laravel.

Treggats:
> Thank you @Treggats for the time you've devoted to correcting the breaking changes coming from Laravel.

@GromNaN You're welcome, we had to pin our `laravel/framework` version because of it. So we needed a fix for this.

src/Eloquent/Builder.php
GromNaN: Simple and verbose solutions are often good enough.

GromNaN:
I'm going to make a release tomorrow.

PR #2663, opened 2023-11-02
Test with lowest dependencies in CI

GromNaN:
As seen with the [recent breaking change in Laravel 10.30](https://github.com/mongodb/laravel-mongodb/pull/2661#issuecomment-1790830940), we need to test for highest and lowest dependencies.

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -9,7 +9,7 @@ on:
 jobs:
     build:
         runs-on: ${{ matrix.os }}
-        name: PHP v${{ matrix.php }} with MongoDB ${{ matrix.mongodb }}
+        name: PHP v${{ matrix.php }} with MongoDB ${{ matrix.mongodb }} ${{ matrix.mode }}
         strategy:
             matrix:
                 os:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-ci.yml
```diff
@@ -23,6 +23,10 @@ jobs:
                     - '8.1'
                     - '8.2'
                     - '8.3'
+                include:
+                    - php: '8.1'
+                      mongodb: '5.0'
+                      mode: 'low-deps'
```
GromNaN, line 29: @alcaeus this modifies an existing matrix combination instead of creating an extra one. I think it's perfect. 
https://github.com/mongodb/laravel-mongodb/actions/runs/6735955088/job/18310279364?pr=2663

.github/workflows/build-ci.yml
```diff
@@ -63,7 +67,7 @@ jobs:
                     restore-keys: ${{ matrix.os }}-composer-
             -   name: Install dependencies
                 run: |
-                    composer install --no-interaction
+                    composer update --no-interaction $([[ "${{ matrix.mode }}" == low-deps ]] && echo ' --prefer-lowest --prefer-stable')
```
jmikola, line 70: This is a bit different than what we do in PHPLIB IIRC, where the CI parameter is just the args to inject (and defaults to an empty string).

I'm sure this works though so no objection. The approach in PHPLIB also doesn't change the build title, which you've achieved here.

jmikola (approved):

PR #2471, opened 2022-11-27
Add PHPStan

divine:
Add PHPStan

Closes: https://github.com/jenssegers/laravel-mongodb/issues/1930

Fixes will be added with another PR to keep everything simple as possible.

PR #2664, opened 2023-11-02
[feature] add static analysis tool

### LINKED ISSUE #1930: use static analyzer
Use static analyzer 
Variants: 
https://github.com/phpstan/phpstan
https://github.com/vimeo/psalm

Treggats:
Introduce PHPStan as static analyser. Continues the work done on #2471

resolves #1930

This PR does the following
1. add a basic config for PHPStan
2. resolve issues found for level 1
3. add a baseline for issues found in level 2
4. add a job to the coding standard Github Workflow

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,3 +7,6 @@ insert_final_newline = true
 indent_style = space
 indent_size = 4
 trim_trailing_whitespace = true
+
+[*.yml]
+indent_size = 2
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -53,7 +53,7 @@ jobs:
             -   name: Show Docker version
                 run: if [[ "$DEBUG" == "true" ]]; then docker version && env; fi
                 env:
-                    DEBUG: ${{secrets.DEBUG}}
+                    DEBUG: ${{ secrets.DEBUG }}
             -   name: Download Composer cache dependencies from cache
                 id: composer-cache
                 run: echo "dir=$(composer config cache-files-dir)" >> $GITHUB_OUTPUT
@@ -64,11 +64,9 @@ jobs:
                     key: ${{ matrix.os }}-composer-${{ hashFiles('**/composer.json') }}
                     restore-keys: ${{ matrix.os }}-composer-
             -   name: Install dependencies
-                run: |
-                    composer update --no-interaction $([[ "${{ matrix.mode }}" == low-deps ]] && echo ' --prefer-lowest --prefer-stable')
+                run: composer update --no-interaction $([[ "${{ matrix.mode }}" == low-deps ]] && echo ' --prefer-lowest --prefer-stable')
             -   name: Run tests
-                run: |
-                    ./vendor/bin/phpunit --coverage-clover coverage.xml
+                run: ./vendor/bin/phpunit --coverage-clover coverage.xml
                 env:
                     MONGODB_URI: 'mongodb://127.0.0.1/?replicaSet=rs'
             -   uses: codecov/codecov-action@v3
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -69,6 +69,3 @@ jobs:
                 run: ./vendor/bin/phpunit --coverage-clover coverage.xml
                 env:
                     MONGODB_URI: 'mongodb://127.0.0.1/?replicaSet=rs'
-            -   uses: codecov/codecov-action@v3
-                with:
-                    token: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -2,8 +2,6 @@ name: CI
 
 on:
     push:
-        branches:
-        tags:
     pull_request:
 
 jobs:
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -72,4 +72,3 @@ jobs:
             -   uses: codecov/codecov-action@v3
                 with:
                     token: ${{ secrets.CODECOV_TOKEN }}
-                    fail_ci_if_error: false
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -54,6 +54,7 @@ jobs:
 
   analysis:
     runs-on: "ubuntu-22.04"
+    continue-on-error: true
     strategy:
       matrix:
         php:
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -13,6 +13,11 @@ jobs:
     name: "phpcs"
     runs-on: "ubuntu-22.04"
 
+    permissions:
+      # Give the default GITHUB_TOKEN write permission to commit and push the
+      # added or changed files to the repository.
+      contents: write
+
     steps:
       - name: "Checkout"
         uses: "actions/checkout@v4"
@@ -51,6 +56,14 @@ jobs:
       # The -q option is required until phpcs v4 is released
       - name: "Run PHP_CodeSniffer"
         run: "vendor/bin/phpcs -q --no-colors --report=checkstyle | cs2pr"
+
+      - name: "Format the code"
+        run: ./vendor/bin/phpcbf
+
+      - name: "Commit the changes"
+        uses: stefanzweifel/git-auto-commit-action@v5
+        with:
+          commit_message: "apply phpcbf formatting"
 
   analysis:
     runs-on: "ubuntu-22.04"
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -53,3 +53,28 @@ jobs:
       # The -q option is required until phpcs v4 is released
       - name: "Run PHP_CodeSniffer"
         run: "vendor/bin/phpcs -q --no-colors --report=checkstyle | cs2pr"
+
+  analysis:
+    runs-on: "ubuntu-22.04"
+    strategy:
+      matrix:
+        php:
+          - '8.1'
+          - '8.2'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          extensions: curl, mbstring
+          tools: composer:v2
+          coverage: none
+
+      - name: Install dependencies
+        run: composer install
+
+      - name: Run PHPStan
+        run: ./vendor/bin/phpstan analyse --no-interaction --no-progress --ansi
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -53,12 +53,14 @@ jobs:
         with:
           composer-options: "--no-suggest"
 
+      - name: "Format the code"
+        run: |
+          mkdir .cache
+          ./vendor/bin/phpcbf
+
       # The -q option is required until phpcs v4 is released
       - name: "Run PHP_CodeSniffer"
         run: "vendor/bin/phpcs -q --no-colors --report=checkstyle | cs2pr"
-
-      - name: "Format the code"
-        run: ./vendor/bin/phpcbf
 
       - name: "Commit the changes"
         uses: stefanzweifel/git-auto-commit-action@v5
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -2,8 +2,6 @@ name: "Coding Standards"
 
 on:
   push:
-    branches:
-    tags:
   pull_request:
 
 env:
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -54,6 +54,7 @@ jobs:
           composer-options: "--no-suggest"
 
       - name: "Format the code"
+        continue-on-error: true
         run: |
           mkdir .cache
           ./vendor/bin/phpcbf
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -72,8 +72,32 @@ jobs:
           tools: composer:v2
           coverage: none
 
+      - name: Cache dependencies
+        id: composer-cache
+        uses: actions/cache@v3
+        with:
+          path: ./vendor
+          key: composer-${{ hashFiles('**/composer.lock') }}
+
       - name: Install dependencies
         run: composer install
 
+      - name: Restore cache PHPStan results
+        id: phpstan-cache-restore
+        uses: actions/cache/restore@v3
+        with:
+          path: .cache
+          key: "phpstan-result-cache-${{ github.run_id }}"
+          restore-keys: |
+            phpstan-result-cache-
+
       - name: Run PHPStan
         run: ./vendor/bin/phpstan analyse --no-interaction --no-progress --ansi
+
+      - name: Save cache PHPStan results
+        id: phpstan-cache-save
+        if: always()
+        uses: actions/cache/save@v3
+        with:
+          path: .cache
+          key: ${{ steps.phpstan-cache-restore.outputs.cache-primary-key }}
--- a/.gitignore
+++ b/.gitignore
@@ -3,11 +3,9 @@
 *.sublime-workspace
 .DS_Store
 .idea/
-.phpunit.cache/
-.phpcs-cache
 /vendor
 composer.lock
 composer.phar
 phpunit.xml
 phpstan.neon
-/.cache/phpstan/
+/.cache/
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@
 composer.lock
 composer.phar
 phpunit.xml
+phpstan.neon
+/.cache/phpstan/
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,8 @@
         "orchestra/testbench": "^8.0",
         "mockery/mockery": "^1.4.4",
         "doctrine/coding-standard": "12.0.x-dev",
-        "spatie/laravel-query-builder": "^5.6"
+        "spatie/laravel-query-builder": "^5.6",
+        "phpstan/phpstan": "^1.10"
     },
     "replace": {
         "jenssegers/mongodb": "self.version"
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,11 +1,6 @@
 parameters:
 	ignoreErrors:
 		-
-			message: "#^Cannot call method modelKeys\\(\\) on array\\|Illuminate\\\\Support\\\\Collection\\.$#"
-			count: 1
-			path: src/Relations/BelongsToMany.php
-
-		-
 			message: "#^Method Illuminate\\\\Database\\\\Eloquent\\\\Model\\:\\:push\\(\\) invoked with 3 parameters, 0 required\\.$#"
 			count: 3
 			path: src/Relations/BelongsToMany.php
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,0 +1,46 @@
+parameters:
+	ignoreErrors:
+		-
+			message: "#^Method MongoDB\\\\Laravel\\\\Connection\\:\\:transaction\\(\\) has invalid return type Illuminate\\\\Database\\\\Concerns\\\\TCallbackReturnType\\.$#"
+			count: 1
+			path: src/Connection.php
+
+		-
+			message: "#^Parameter \\$callback of method MongoDB\\\\Laravel\\\\Connection\\:\\:transaction\\(\\) has invalid type Illuminate\\\\Database\\\\Concerns\\\\TCallbackReturnType\\.$#"
+			count: 1
+			path: src/Connection.php
+
+		-
+			message: "#^Method Illuminate\\\\Database\\\\Query\\\\Builder\\:\\:update\\(\\) invoked with 2 parameters, 1 required\\.$#"
+			count: 1
+			path: src/Eloquent/Builder.php
+
+		-
+			message: "#^Parameter \\$offset of method MongoDB\\\\Laravel\\\\Eloquent\\\\Model\\:\\:offsetSet\\(\\) has invalid type Illuminate\\\\Database\\\\Eloquent\\\\TKey\\.$#"
+			count: 1
+			path: src/Eloquent/Model.php
+
+		-
+			message: "#^Parameter \\$offset of method MongoDB\\\\Laravel\\\\Eloquent\\\\Model\\:\\:offsetUnset\\(\\) has invalid type Illuminate\\\\Database\\\\Eloquent\\\\TKey\\.$#"
+			count: 1
+			path: src/Eloquent/Model.php
+
+		-
+			message: "#^Parameter \\$value of method MongoDB\\\\Laravel\\\\Eloquent\\\\Model\\:\\:offsetSet\\(\\) has invalid type Illuminate\\\\Database\\\\Eloquent\\\\TValue\\.$#"
+			count: 1
+			path: src/Eloquent/Model.php
+
+		-
+			message: "#^Cannot call method modelKeys\\(\\) on array\\|Illuminate\\\\Database\\\\Eloquent\\\\Model\\|Illuminate\\\\Support\\\\Collection\\.$#"
+			count: 1
+			path: src/Relations/BelongsToMany.php
+
+		-
+			message: "#^Method Illuminate\\\\Database\\\\Eloquent\\\\Model\\:\\:push\\(\\) invoked with 3 parameters, 0 required\\.$#"
+			count: 2
+			path: src/Relations/BelongsToMany.php
+
+		-
+			message: "#^Method Illuminate\\\\Database\\\\Schema\\\\Blueprint\\:\\:create\\(\\) invoked with 1 parameter, 0 required\\.$#"
+			count: 1
+			path: src/Schema/Builder.php
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,12 +1,22 @@
 parameters:
 	ignoreErrors:
 		-
+			message: "#^Variable \\$value on left side of \\?\\? always exists and is always null\\.$#"
+			count: 1
+			path: src/Eloquent/Model.php
+
+		-
+			message: "#^Variable \\$value on left side of \\?\\? always exists and is not nullable\\.$#"
+			count: 1
+			path: src/Eloquent/Model.php
+
+		-
 			message: "#^Call to an undefined method Illuminate\\\\Database\\\\Eloquent\\\\Builder\\|Illuminate\\\\Database\\\\Eloquent\\\\Model\\:\\:pull\\(\\)\\.$#"
 			count: 1
 			path: src/Relations/BelongsToMany.php
 
 		-
-			message: "#^Cannot call method modelKeys\\(\\) on array\\|Illuminate\\\\Database\\\\Eloquent\\\\Model\\|Illuminate\\\\Support\\\\Collection\\.$#"
+			message: "#^Cannot call method modelKeys\\(\\) on array\\|Illuminate\\\\Support\\\\Collection\\.$#"
 			count: 1
 			path: src/Relations/BelongsToMany.php
 
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,16 +1,6 @@
 parameters:
 	ignoreErrors:
 		-
-			message: "#^Variable \\$value on left side of \\?\\? always exists and is always null\\.$#"
-			count: 1
-			path: src/Eloquent/Model.php
-
-		-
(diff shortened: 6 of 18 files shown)


.editorconfig
```diff
@@ -6,4 +6,7 @@ end_of_line = lf
 insert_final_newline = true
 indent_style = space
 indent_size = 4
-trim_trailing_whitespace = true
\ No newline at end of file
+trim_trailing_whitespace = true
+
+[*.yml]
+indent_size = 2
```
Treggats: EOL final newline
```suggestion
indent_size = 2

```

.github/workflows/build-ci.yml
```diff
@@ -2,8 +2,6 @@ name: CI
 
 on:
     push:
-        branches:
```
Treggats, line 5: redundant, and not according to the Github Workflow schema

.github/workflows/coding-standards.yml
```diff
@@ -53,3 +51,29 @@ jobs:
       # The -q option is required until phpcs v4 is released
       - name: "Run PHP_CodeSniffer"
         run: "vendor/bin/phpcs -q --no-colors --report=checkstyle | cs2pr"
+
+  analysis:
+    runs-on: "ubuntu-22.04"
+    continue-on-error: true
```
Treggats, line 73: This will allow the job to continue even if PHPStan finds errors.

For now mostly a convenience to see the output of runs for both PHP 8.1 and 8.2

.github/workflows/coding-standards.yml
```diff
@@ -53,3 +51,29 @@ jobs:
       # The -q option is required until phpcs v4 is released
       - name: "Run PHP_CodeSniffer"
         run: "vendor/bin/phpcs -q --no-colors --report=checkstyle | cs2pr"
+
+  analysis:
```
Treggats, line 71: Borrowed from the changes that @divine made in #2471

.github/workflows/coding-standards.yml
```diff
@@ -53,3 +51,29 @@ jobs:
...
+        uses: actions/checkout@v4
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          extensions: curl, mbstring
+          tools: composer:v2
+          coverage: none
+          # ini-values: memory_limit=-1
```
Treggats: No need to increase the memory limit, runs just fine without.
```suggestion

```

composer.json
```diff
@@ -31,10 +31,12 @@
         "mongodb/mongodb": "^1.15"
     },
     "require-dev": {
+        "roave/security-advisories": "dev-latest",
```
Treggats: Nice to have, will cause `composer install` to halt if there is a security advisory for one of the dependencies

phpstan-baseline.neon
```diff
@@ -0,0 +1,46 @@
+parameters:
```
Treggats, line 1: The baseline provides PHPStan with a list of errors to ignore, kind of saying 
> I know these issues exist, will fix it later

phpstan.neon.dist
```diff
@@ -0,0 +1,17 @@
...
+
+    paths:
+        - src
+
+    level: 2
+
+    editorUrl: 'phpstorm://open?file=%%file%%&line=%%line%%'
+
+    ignoreErrors:
+        - '#Unsafe usage of new static#'
```
Treggats, line 15: These issues were the remaining errors from level 1, I wasn't sure how to fix them. So I chose to ignore them.

phpunit.xml.dist
```diff
@@ -1,13 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.4/phpunit.xsd"
-         backupGlobals="false"
```
Treggats, line 4: removed options that had the default value.

src/Relations/BelongsToMany.php
```diff
@@ -82,11 +82,11 @@ protected function setWhere()
     }
 
     /** @inheritdoc */
-    public function save(Model $model, array $joining = [], $touch = true)
+    public function save(Model $model, array $pivotAttributes = [], $touch = true)
```
Treggats, line 85: The parent method argument name was changed, updated it to match it.

.editorconfig
Treggats, line 12: Match the current indenting.

.github/workflows/build-ci.yml
```diff
@@ -2,8 +2,6 @@ name: CI
 
 on:
     push:
```
Treggats, line 4: @divine should this always run on push? I think it makes more sense to only run for a pull request event.

In our application we have the following defined in our code quality Github Workflow. We run on certain pull request event and push to the develop branch. Reduces the Github Action minutes, and the double runs of the workflow.

```yml
on:

  pull_request:
    types:
      - opened
      - synchronize
      - ready_for_review

  workflow_dispatch:

  push:
    branches:
      - develop
```

phpstan.neon.dist
```diff
@@ -0,0 +1,17 @@
+includes:
+    - ./vendor/nunomaduro/larastan/extension.neon
+    - ./phpstan-baseline.neon
+
+parameters:
+    tmpDir: .cache/phpstan
```
Treggats, line 5: Will speedup subsequent runs, in a Github Workflow we'd need to use the cache action to store this cache

.github/workflows/coding-standards.yml
```diff
@@ -53,3 +56,61 @@ jobs:
       # The -q option is required until phpcs v4 is released
       - name: "Run PHP_CodeSniffer"
         run: "vendor/bin/phpcs -q --no-colors --report=checkstyle | cs2pr"
+
+      - name: "Format the code"
```
Treggats: Nice to have, removes the need to remember to run this locally

composer.json
GromNaN: This project is not directly deployed. We will not restrict versions based on known security issues. Please remove this job.

src/Relations/EmbedsMany.php
```diff
@@ -195,9 +196,11 @@ public function destroy($ids = [])
     /**
      * Delete all embedded models.
      *
+     * @param null $id
```
GromNaN, line 201: Can we remove this argument, since it's not used.

src/Relations/EmbedsOne.php
```diff
@@ -133,9 +133,11 @@ public function dissociate()
     /**
      * Delete all embedded models.
      *
+     * @param ?string $id
+     *
      * @return int
      */
-    public function delete()
+    public function delete($id = null)
```
GromNaN: Same here, can we remove this argument, since it's not used.

src/Relations/EmbedsOneOrMany.php
```diff
@@ -42,6 +43,8 @@ abstract class EmbedsOneOrMany extends Relation
      */
     public function __construct(Builder $query, Model $parent, Model $related, string $localKey, string $foreignKey, string $relation)
     {
+        parent::__construct($query, $parent);
```
GromNaN, line 49: I think parent constructor should be called last. The affectation of `$this->query` and `$this->parent` can be removed, as the duplicate `$this->addConstraints();` at the end.

src/Relations/EmbedsOneOrMany.php
```diff
@@ -101,9 +104,11 @@ public function get($columns = ['*'])
     /**
      * Get the number of embedded models.
      *
+     * @param Expression|string $columns
+     *
      * @return int
      */
-    public function count()
+    public function count($columns = '*')
```
GromNaN: Not used.

composer.json
```diff
@@ -31,10 +31,12 @@
...
         "phpunit/phpunit": "^10.3",
         "orchestra/testbench": "^8.0",
         "mockery/mockery": "^1.4.4",
-        "doctrine/coding-standard": "12.0.x-dev"
+        "doctrine/coding-standard": "12.0.x-dev",
+        "nunomaduro/larastan": "^2.6"
```
GromNaN: I understand that larastan's interest is in its ability to boot the container to analyze the facades. We don't have any applications to boot in this library. I'd rather use phpstan directly instead of depending on an extra abstraction layer and its potential breaking changes.

.github/workflows/coding-standards.yml
GromNaN: I can see the benefit, but I'm not sure...

Treggats: Just an idea, I'm fine with removing it if need be

composer.json
Treggats: 👌 
```suggestion
```

composer.json
Treggats: Alright, I'll replace it with "native" PHPStan

src/Relations/EmbedsMany.php
Treggats: I know that it's not used, but it is part of the methods heading (name, arguments, etc.) as defined in the [parent method](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Query/Builder.php#L3147).

Should we not keep the inherited methods heading the same, even though we might not use some arguments?

src/Relations/EmbedsOneOrMany.php
Treggats: Do you mean that those can be removed?

GromNaN: Yes, it we call the parent constructor, we don't need to repeat all it does.

src/Eloquent/Builder.php
```diff
@@ -71,7 +73,10 @@ public function update(array $values, array $options = [])
             return 1;
         }
 
-        return $this->toBase()->update($this->addUpdatedAtColumn($values), $options);
+        $builder = $this->toBase();
+        assert($builder instanceof MongoDBQueryBuilder);
+
+        return $builder->update($this->addUpdatedAtColumn($values), $options);
```
Treggats: So this is an interesting case. Both the IDE and PHPStan think that `Illuminate\Database\Query\Builder` is used, while actually it's `MongoDB\Laravel\Query\Builder`.
Adding the assertion is quickest/eastiest way to make sure the correct one is used.

src/Eloquent/Builder.php
```diff
@@ -228,16 +233,15 @@ protected function ensureOrderForCursorPagination($shouldReverse = false)
         }
 
         if ($shouldReverse) {
-            $this->query->orders = collect($this->query->orders)->map(function ($direction) {
-                return $direction === 1 ? -1 : 1;
-            })->toArray();
+            $this->query->orders = collect($this->query->orders)
```
Treggats, line 223: Thought that this was a bit cleaner.

src/Relations/EmbedsOneOrMany.php
```diff
@@ -42,8 +43,8 @@ abstract class EmbedsOneOrMany extends Relation
      */
     public function __construct(Builder $query, Model $parent, Model $related, string $localKey, string $foreignKey, string $relation)
     {
-        $this->query      = $query;
-        $this->parent     = $parent;
+        parent::__construct($query, $parent);
+
         $this->related    = $related;
```
Treggats, line 51: This property is also being set in the parent constructor, but when I tried to change it a few mocks broke and I'm not sure how to update those. So I left it alone 😅

Treggats:
Putting in draft while I figure out why the tests are faling

.github/workflows/build-ci.yml
divine: As this is an open-source project there is no limit on minutes, so it should be fine.
The reason from what I remember is that sometimes I've pushed commits directly to the branches which of course is unnecessary.
Your suggested change looks great.

Thanks!

.github/workflows/coding-standards.yml
divine: I think formatting the code should happen after each commit but this might make things ugly.

Treggats: @divine keep it or remove it?

.github/workflows/coding-standards.yml
```diff
@@ -53,3 +56,61 @@ jobs:
...
+        run: ./vendor/bin/phpcbf
```
GromNaN: Some errors can't be fixed. We need to run `./vendor/bin/phpcs` after this command to avoid committing partially fixed code.

.cache/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
```
Treggats: make sure the directory for the cache is present

.github/workflows/build-ci.yml
```diff
@@ -39,41 +38,47 @@ jobs:
...
-                run: |
-                    ./vendor/bin/phpunit --coverage-clover coverage.xml
+                run: ./vendor/bin/phpunit --coverage-clover coverage.xml
                 env:
                     MONGODB_URI: 'mongodb://127.0.0.1/?replicaSet=rs'
+
             -   uses: codecov/codecov-action@v3
                 with:
                     token: ${{ secrets.CODECOV_TOKEN }}
-                    fail_ci_if_error: false
```
Treggats, line 79: This is the default value, so it can be omitted

.gitignore
```diff
@@ -3,9 +3,9 @@
 *.sublime-workspace
 .DS_Store
 .idea/
-.phpunit.cache/
```
Treggats, line 6: I introduced the `.cache` directory for caching phpunit/phpcs/phpstan

phpunit.xml.dist
```diff
@@ -20,10 +18,15 @@
         <env name="MONGODB_DATABASE" value="unittest"/>
         <env name="SQLITE_DATABASE" value=":memory:"/>
         <env name="QUEUE_CONNECTION" value="database"/>
+
+        <ini name="memory_limit" value="-1"/>
```
Treggats, line 22: While running the tests locally in Docker I ran into the issue that there wasn't enough memory. Disabling the limit here, seems like the most logical place

phpunit.xml.dist
```diff
@@ -20,10 +18,15 @@
...
     </php>
-    <source>
+
+    <source restrictDeprecations="true"
+        restrictNotices="true"
+        restrictWarnings="true">
         <include>
-            <directory suffix=".php">./src</directory>
```
Treggats, line 26: the suffix was redundant so it could be removed

.cache/.gitignore
GromNaN: Why can't this directory be created automatically? I would prefer not to have an empty folder in the project

.github/workflows/build-ci.yml
```diff
@@ -30,6 +28,7 @@ jobs:
 
         steps:
             -   uses: actions/checkout@v4
+
```
GromNaN: Could you revert this empty newlines, it forces you to scroll more than necessary. The `-` already make each task legible.

src/Eloquent/Builder.php
```diff
@@ -228,16 +233,15 @@ protected function ensureOrderForCursorPagination($shouldReverse = false)
...
+                ->map(fn (int $direction) => $direction === 1 ? -1 : 1)
```
GromNaN: Good use of an arrow function.
```suggestion
                ->map(static fn (int $direction) => $direction === 1 ? -1 : 1)
```

src/Eloquent/Builder.php
```diff
@@ -228,16 +233,15 @@ protected function ensureOrderForCursorPagination($shouldReverse = false)
...
+                ->toArray();
         }
 
-        return collect($this->query->orders)->map(function ($direction, $column) {
-            return [
+        return collect($this->query->orders)
+            ->map(fn ($direction, $column) => [
```
GromNaN:
```suggestion
            ->map(static fn ($direction, $column) => [
```

src/Relations/EmbedsMany.php
GromNaN: Ok, then we need to add a comment to mention that the id is not used. And throw a LogicException if it is not null. We don't want to give the illusion that embeds can be deleted by Id. Even though it's an interesting feature to implement.

.cache/.gitignore
Treggats: I'll see if I can make that happen

src/Relations/EmbedsOneOrMany.php
```diff
@@ -265,17 +273,17 @@ protected function toCollection(array $records = [])
...
-    protected function toModel($attributes = [])
+    protected function toModel(array $attributes = []): Model
     {
-        if ($attributes === null) {
-            return;
-        }
-
         $connection = $this->related->getConnection();
 
+        if (empty($attributes)) {
```
Treggats: Return a actual model instance

src/Relations/EmbedsMany.php
```diff
@@ -82,7 +86,7 @@ public function performUpdate(Model $model)
         // Get the correct foreign key value.
         $foreignKey = $this->getForeignKeyValue($model);
 
-        $values = $this->getUpdateValues($model->getDirty(), $this->localKey . '.$.');
+        $values = self::getUpdateValues($model->getDirty(), $this->localKey . '.$.');
```
Treggats, line 88: The method is static, so using a static call here

src/Relations/EmbedsMany.php
GromNaN: The note must be on the argument description.
```suggestion
     * @param null $id Not used. Inherited from parent method.
```

src/Relations/EmbedsMany.php
```diff
@@ -195,10 +199,16 @@ public function destroy($ids = [])
     /**
      * Delete all embedded models.
      *
-     * @return int
+     * @param null $id
+     *
+     * @throws LogicException|Throwable
```
GromNaN: I never use `@throw` annotation for logic exception, as they are made for the developer and should not be caught.

Treggats: Should I remove them?

src/Relations/EmbedsMany.php
```diff
@@ -195,10 +199,16 @@ public function destroy($ids = [])
...
+     *
+     * @note The $id is not used to delete embedded models.
```
Treggats, line 203:
```suggestion
```

src/Relations/EmbedsOne.php
```diff
@@ -133,10 +137,16 @@ public function dissociate()
     /**
      * Delete all embedded models.
      *
-     * @return int
+     * @param ?string $id
+     *
+     * @throws LogicException|Throwable
+     *
+     * @note The $id is not used to delete embedded models.
```
Treggats, line 144:
```suggestion
     * @param ?string $id Not used. Inherited from parent method.
     *
     * @throws LogicException|Throwable
```

src/Relations/EmbedsOneOrMany.php
```diff
@@ -101,10 +103,16 @@ public function get($columns = ['*'])
     /**
      * Get the number of embedded models.
      *
-     * @return int
+     * @param Expression|string $columns
+     *
+     * @throws LogicException|Throwable
+     *
+     * @note The $column parameter is not used to count embedded models.
```
Treggats, line 110:
```suggestion
     * @param Expression|string $columns Not used. Inherited from parent method.
     *
     * @throws LogicException|Throwable
```

.github/workflows/build-ci.yml
```diff
@@ -66,14 +64,11 @@ jobs:
...
+                run: composer update --no-interaction $([[ "${{ matrix.mode }}" == low-deps ]] && echo ' --prefer-lowest --prefer-stable')
             -   name: Run tests
-                run: |
-                    ./vendor/bin/phpunit --coverage-clover coverage.xml
+                run: ./vendor/bin/phpunit --coverage-clover coverage.xml
                 env:
                     MONGODB_URI: 'mongodb://127.0.0.1/?replicaSet=rs'
             -   uses: codecov/codecov-action@v3
                 with:
                     token: ${{ secrets.CODECOV_TOKEN }}
```
alcaeus: @GromNaN I noticed we still have this in. Given that we're no longer using codecov.io, should we remove this step?

.github/workflows/coding-standards.yml
```diff
@@ -50,6 +53,67 @@ jobs:
         with:
           composer-options: "--no-suggest"
 
+      - name: "Format the code"
+        continue-on-error: true
+        run: |
+          mkdir .cache
+          ./vendor/bin/phpcbf
+
```
alcaeus, line 61: @GromNaN what are your thoughts on this? I'm usually not a fan of workflows that commit changes, but since it's "just" checkstyle it may be fine.

alcaeus (commented):
I have a few comments on the CI changes, haven't yet reviewed the changes to satisfy PHPStan.

@Treggats can you please change the target branch of this PR to 4.1 and rebase changes to resolve conflicts? I don't feel comfortable introducing these changes in a patch release.

.github/workflows/build-ci.yml
GromNaN: Yes.

.github/workflows/coding-standards.yml
GromNaN: I think it's an acceptable way of speeding up workflow when you forget to launch phpcs. It adds a commit, so it's easy to undo and will be squashed during the merge.

Treggats:
@alcaeus I've updated the base and removed the codecov.io step in the Github Action workflow

src/Relations/EmbedsMany.php
GromNaN: Yes please.

src/Eloquent/Builder.php
GromNaN: You can overload the return type of `toBase` with an annotation on this class.
```php
/** @method \MongoDB\Laravel\Query\Builder toBase() */
class Builder extends EloquentBuilder
```

phpstan-baseline.neon
```diff
@@ -0,0 +1,31 @@
...
+	ignoreErrors:
+		-
+			message: "#^Variable \\$value on left side of \\?\\? always exists and is always null\\.$#"
```
GromNaN: If we remove the null coalescence operator in `Model::fromJson`, tests still pass. 
https://github.com/mongodb/laravel-mongodb/blob/82ddb839ed2f3c027d4b810a66d056ba5221d4bb/src/Eloquent/Model.php#L297-L304
This was introduced in #2653, could you check this error @hans-thomas?

phpstan-baseline.neon
```diff
@@ -0,0 +1,31 @@
...
+
+		-
+			message: "#^Variable \\$value on left side of \\?\\? always exists and is not nullable\\.$#"
+			count: 1
+			path: src/Eloquent/Model.php
+
+		-
+			message: "#^Call to an undefined method Illuminate\\\\Database\\\\Eloquent\\\\Builder\\|Illuminate\\\\Database\\\\Eloquent\\\\Model\\:\\:pull\\(\\)\\.$#"
+			count: 1
+			path: src/Relations/BelongsToMany.php
```
GromNaN: This should be fixed by adding `assert($this->parent instanceof MongoDB\Eloquent\Model)`

phpstan-baseline.neon
```diff
@@ -0,0 +1,31 @@
...
+
+		-
+			message: "#^Cannot call method modelKeys\\(\\) on array\\|Illuminate\\\\Support\\\\Collection\\.$#"
```
GromNaN: It looks like there is an hidden bug if `$ids` is an array.
https://github.com/mongodb/laravel-mongodb/blob/82ddb839ed2f3c027d4b810a66d056ba5221d4bb/src/Relations/BelongsToMany.php#L129-L132

GromNaN (commented):
The *virtual* arguments of `Model::push()` must be documented to remove the errors with `Relations/MorphToMany`.

`MorphToMany::updateExistingPivot()` can return `0`, this returned value is documented but not used. https://github.com/laravel/framework/blob/2967d89906708cc7d619fc130e835c8002b7d3e3/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php#L187

phpstan-baseline.neon
```diff
@@ -0,0 +1,31 @@
...
+			count: 1
+			path: src/Eloquent/Model.php
+
+		-
+			message: "#^Cannot call method modelKeys\\(\\) on array\\|Illuminate\\\\Support\\\\Collection\\.$#"
+			count: 1
+			path: src/Relations/BelongsToMany.php
+
+		-
+			message: "#^Method Illuminate\\\\Database\\\\Eloquent\\\\Model\\:\\:push\\(\\) invoked with 3 parameters, 0 required\\.$#"
```
Treggats, line 4: The `push()` method uses `func_get_args()` to get its arguments, but we're passing it to the method as arguments. Not sure how to fix or write differently.

GromNaN: My idea was to add phpdoc for this arguments, but PHPStorm and phpcs both complain that "PHPDoc tag `@param` references unknown parameter". 

```
    /*
     * If no arguments are provided, save the model and all of its relationships.
     * If a $column is provided, append the $values to the $column array attribute.
     *
     * @param string $column
     * @param mixed  $values
     * @param bool   $unique
     *
     * @return bool
     */
    public function push(/* string $column = null, $values = null, bool $unique = false */)
```

I'm delaying it for a later PR. No doubt the design of this method is incorrect and needs to be changed.

src/Relations/BelongsToMany.php
```diff
@@ -229,6 +230,8 @@ public function detach($ids = [], $touch = true)
         }
 
         // Remove the relation to the parent.
+        assert($this->parent instanceof \MongoDB\Laravel\Eloquent\Model);
+        assert($query instanceof \MongoDB\Laravel\Eloquent\Model);
```
hans-thomas:
```suggestion
        assert($query instanceof \MongoDB\Laravel\Eloquent\Builder);
```
The `$query` variable is an instance of `Builder` class.

phpstan-baseline.neon
hans-thomas: @Treggats We can remove `Null Coalescing` Operator. If the `$value` was `Null`, the ` Json::encode` will handle it.

UPDATE: To solve the second error, we can also remove the second  `Null Coalescing` Operator.

The final method:
```php
public function fromJson($value, $asObject = false) 
 { 
     if (! is_string($value)) { 
         $value = Json::encode($value); 
     } 
  
     return Json::decode($value, ! $asObject); 
 } 
```

phpstan-baseline.neon
hans-thomas: https://github.com/mongodb/laravel-mongodb/blob/82ddb839ed2f3c027d4b810a66d056ba5221d4bb/src/Relations/BelongsToMany.php#L127-L132
It's not even covered in the tests. 
In what case, the `$this->parent->{$this->relatedPivotKey}` could return a `Collection`?
I think in the past, it was possible but not now. The pivot column always is an array.

UPDATE: removing this condition doesn't cause any error.

phpstan-baseline.neon
```diff
@@ -0,0 +1,31 @@
...
+		-
+			message: "#^Variable \\$value on left side of \\?\\? always exists and is always null\\.$#"
+			count: 1
+			path: src/Eloquent/Model.php
+
+		-
+			message: "#^Variable \\$value on left side of \\?\\? always exists and is not nullable\\.$#"
+			count: 1
+			path: src/Eloquent/Model.php
+
```
hans-thomas:
```suggestion

```
You can remove these two errors after applying this [change](https://github.com/mongodb/laravel-mongodb/pull/2664/files#r1403260408).

src/Relations/BelongsToMany.php
```diff
@@ -295,28 +293,6 @@ public function getQualifiedRelatedPivotKeyName()
...
 
-    /**
-     * Format the sync list so that it is keyed by ID. (Legacy Support)
-     * The original function has been renamed to formatRecordsList since Laravel 5.3.
-     *
-     * @deprecated
-     *
-     * @return array
-     */
-    protected function formatSyncList(array $records)
```
Treggats, line 306: removing in favor of `formatRecordList` as mentioned in the comment

GromNaN (approved):

src/Relations/BelongsToMany.php
```diff
@@ -126,12 +126,7 @@ public function sync($ids, $detaching = true)
         // if they exist in the array of current ones, and if not we will insert.
         $current = $this->parent->{$this->relatedPivotKey} ?: [];
 
-        // See issue #256.
-        if ($current instanceof Collection) {
-            $current = $ids->modelKeys();
-        }
```
hans-thomas, line 132: Could you please remove these lines in the[ `MorphToMany`](https://github.com/hans-thomas/laravel-mongodb/blob/d47380ccfb3cba967855191600847133726ceed3/src/Relations/MorphToMany.php#L136-L139) class too?

GromNaN:
Thank you @Treggats

Treggats:
> Thank you @Treggats

You're welcome @GromNaN was a nice experience

PR #2667, opened 2023-11-05
[Fix] BelongsToMany sync does't use configured keys

### LINKED ISSUE #2666: BelongsToMany sync does't use configured foreign key
> I guess there is an other issue with this call. It should be `$ids = $this->parseIds($ids);` also. Otherwise it's getting the model `_id` instead of the `relatedKey`.
> Do you want to add an other test with a customized `relatedKey`? That will require a new model class for tests, but will ensure we use the correct ID field.

_Originally posted by @GromNaN in https://github.com/mongodb/laravel-mongodb/pull/2648#discussion_r1376742591_

hans-thomas:
Hi there,
In this PR, I fix #2666 issue. There were other bugs in using a `BelongsToMany` relationship with custom `relatedKey` or `parentKey` that I will explain in the comments.

### DIFF
--- a/src/Relations/BelongsToMany.php
+++ b/src/Relations/BelongsToMany.php
@@ -190,10 +190,10 @@ class BelongsToMany extends EloquentBelongsToMany
 
             $query = $this->newRelatedQuery();
 
-            $query->whereIn($this->related->getKeyName(), (array) $id);
+            $query->whereIn($this->relatedKey, (array) $id);
 
             // Attach the new parent id to the related model.
-            $query->push($this->foreignPivotKey, $this->parent->getKey(), true);
+            $query->push($this->foreignPivotKey, $this->parent->{$this->parentKey}, true);
         }
 
         // Attach the new ids to the parent model.
--- a/src/Relations/BelongsToMany.php
+++ b/src/Relations/BelongsToMany.php
@@ -116,7 +116,7 @@ class BelongsToMany extends EloquentBelongsToMany
         ];
 
         if ($ids instanceof Collection) {
-            $ids = $ids->modelKeys();
+            $ids = $this->parseIds($ids);
         } elseif ($ids instanceof Model) {
             $ids = $this->parseIds($ids);
         }
--- a/src/Relations/BelongsToMany.php
+++ b/src/Relations/BelongsToMany.php
@@ -76,7 +76,7 @@ class BelongsToMany extends EloquentBelongsToMany
     {
         $foreign = $this->getForeignKey();
 
-        $this->query->where($foreign, '=', $this->parent->getKey());
+        $this->query->where($foreign, '=', $this->parent->{$this->parentKey});
 
         return $this;
     }
--- a/tests/RelationsTest.php
+++ b/tests/RelationsTest.php
@@ -6,9 +6,11 @@ namespace MongoDB\Laravel\Tests;
 
 use Illuminate\Database\Eloquent\Collection;
 use Mockery;
+use MongoDB\BSON\ObjectId;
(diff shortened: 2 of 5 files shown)


src/Relations/BelongsToMany.php
```diff
@@ -76,7 +76,7 @@ protected function setWhere()
     {
         $foreign = $this->getForeignKey();
 
-        $this->query->where($foreign, '=', $this->parent->getKey());
+        $this->query->where($foreign, '=', $this->parent->{$this->parentKey});
```
hans-thomas, line 79: When a custom `parentKey` is defined, the custom `parentKey`'s value must be searched in the foreign keys list instead of `_id`.

src/Relations/BelongsToMany.php
```diff
@@ -116,7 +116,7 @@ public function sync($ids, $detaching = true)
         ];
 
         if ($ids instanceof Collection) {
-            $ids = $ids->modelKeys();
+            $ids = $this->parseIds($ids);
```
hans-thomas, line 119: As mentioned in #2666, `$ids->modelKeys()` returns `_id` field or primary key. However, `$this->parseIds($ids)` returns `relatedKey` which is the primary key by default.

src/Relations/BelongsToMany.php
```diff
@@ -190,10 +190,10 @@ public function attach($id, array $attributes = [], $touch = true)
 
             $query = $this->newRelatedQuery();
 
-            $query->whereIn($this->related->getKeyName(), (array) $id);
+            $query->whereIn($this->relatedKey, (array) $id);
```
hans-thomas, line 193: As I said in the previous comment, `relatedKey` is `$this->related->getKeyName()` by default but respects our override.

src/Relations/BelongsToMany.php
```diff
@@ -190,10 +190,10 @@ public function attach($id, array $attributes = [], $touch = true)
...
 
             // Attach the new parent id to the related model.
-            $query->push($this->foreignPivotKey, $this->parent->getKey(), true);
+            $query->push($this->foreignPivotKey, $this->parent->{$this->parentKey}, true);
```
hans-thomas, line 196: The `$this->parentKey` is replaced with `$this->parent->getKey()` to respect the custom `parentKey` field.

tests/RelationsTest.php
```diff
@@ -255,36 +260,6 @@ public function testBelongsToMany(): void
         $this->assertCount(1, $client->users);
     }
 
-    public function testSyncBelongsToMany()
```
hans-thomas, line 258: I merged this test with `testBelongsToManySync` test. because I see no reason to keep them separate and I think it's cleaner now.

tests/RelationsTest.php
```diff
@@ -327,20 +302,27 @@ public function testBelongsToManyAttachesExistingModels(): void
     public function testBelongsToManySync(): void
     {
         // create test instances
-        $user    = User::create(['name' => 'John Doe']);
-        $client1 = Client::create(['name' => 'Pork Pies Ltd.'])->_id;
-        $client2 = Client::create(['name' => 'Buffet Bar Inc.'])->_id;
+        $user    = User::create(['name' => 'Hans Thomas']);
```
GromNaN, line 305: We removed the annoying rule [MultipleStatementAlignment](https://github.com/mongodb/laravel-mongodb/blob/4.1/phpcs.xml.dist#L38). You can modify this line.
```suggestion
        $user = User::create(['name' => 'Hans Thomas']);
```

GromNaN (approved):
LGTM.

GromNaN:
Thank you @hans-thomas.

PR #2668, opened 2023-11-06
Remove changelog. Use release notes instead

GromNaN:
[Release notes](https://github.com/mongodb/laravel-mongodb/releases) are a lot easier to update and maintain.


divine:
Sorry but I’m not sure why this type of hard decision is made by MongoDB PHP Team? This and other changes definitely needs some proper reviews instead of instant merge like what happened with other PRs without review sadly.

Just interesting what type of reaction this could get in a Symfony or Laravel repositories? Is that hard to keep the changelog for the users or just get some automation done right?

Thanks!

GromNaN:
Hello @divine, thanks for your feedback. I get that you created this changelog file some time ago and you have been the only one to maintain it. Currently we don't have a changelog in the mongodb library project and everything is documented in release notes.

GitHub provides a good automation with generated release notes, that we customize on each release. What advantage do you see in keeping this file?

alcaeus (approved):
LGTM. While it may seem convenient, the changelog file requires manual updating, whereas generating a changelog as part of the release notes can be done automatically. In addition to those automated lists (which GitHub generates based on commits to the branch since the last release) we highlight significant changes in the release notes, usually in more detail than the changelog would. I believe that negates the necessity for manually maintaining this file.

Instead, we should consider introducing an UPGRADE file containing incompatible changes to point out changes that require intervention by the user (e.g. to prepare for a major release).

GromNaN:
Closing in favor of #2684

PR #2640, opened 2023-10-12
added MorphToMany with testcases

ithuis:
Can you give this a try?

PR #2669, opened 2023-11-06
[Fix] morphTo relationship

### LINKED ISSUE #2657: Support for morphTo relations broken in 4.1
- Laravel-mongodb Version: 4.1.dev
- PHP Version: 8.2.11
- Database Driver & Version: 1.16.2 (6.0.3)

### Description: 
There has been working support for morphTo relations, but that seems to break after upgrading from 4.0 to 4.1. 

### Steps to reproduce
1. You need to have model with morphTo relationship. Here is my "Subscription" model (it does extend mongo model, leaving namespaces out for simpllicity):

```
class Subscription extends Model
{
    use HasFactory;

    public function subscribable(): MorphTo
    {
        return $this->morphTo();
    }

    public function user()
    {
        return $this->belongsTo(User::class);
    }
}
```

2. When using the relationship it fails. This is how this is used in our codebase (controller):

```
$subscription = new Subscription();
$subscription->user()->associate($user);
$subscription->subscribable()->associate($thread);
$subscription->save();
```

### Expected behaviour
New subscription should be saved to database with this kind of values:

```
{
  "_id": {
    "$oid": "647ee3c2ef95cab3850d4ee9"
  },
  "user_id": "64637d18a21fdb655d09bc83",
  "subscribable_id": "647ee3c2ef95cab3850d4ee8",
  "subscribable_type": "App\\Models\\Thread",
  "updated_at": {
    "$date": "2023-06-06T07:44:02.589Z"
  },
  "created_at": {
    "$date": "2023-06-06T07:44:02.589Z"
  }
}
```

### Actual behaviour
InvalidArgumentException is thrown: `First argument of MongoDB\Laravel\Query\Builder::where must be a field path as "string". Got "null"`

Here is more detailed stacktrace: https://flareapp.io/share/bP99dGOP

hans-thomas:
Hi, it fixes #2657.

### DIFF
--- a/src/Eloquent/HybridRelations.php
+++ b/src/Eloquent/HybridRelations.php
@@ -212,6 +212,10 @@ trait HybridRelations
 
         [$type, $id] = $this->getMorphs(Str::snake($name), $type, $id);
 
+        if ($ownerKey === null){
+            $ownerKey = $this->getKeyName();
+        }
+
         // If the type value is null it is probably safe to assume we're eager loading
         // the relationship. When that is the case we will pass in a dummy query as
         // there are multiple types in the morph and we can't use single queries.
--- a/src/Eloquent/HybridRelations.php
+++ b/src/Eloquent/HybridRelations.php
@@ -221,7 +221,7 @@ trait HybridRelations
                 $this->newQuery(),
                 $this,
                 $id,
-                $ownerKey,
+                $ownerKey ?: $this->getKeyName(),
                 $type,
                 $name,
             );
--- a/src/Eloquent/HybridRelations.php
+++ b/src/Eloquent/HybridRelations.php
@@ -238,7 +238,7 @@ trait HybridRelations
 
         // Check if it is a relation with an original model.
         if (! is_subclass_of($instance, MongoDBModel::class)) {
-            return parent::morphTo($name,$type,$id,$ownerKey);
+            return parent::morphTo($name, $type, $id, $ownerKey);
         }
 
         return new MorphTo(
--- a/src/Eloquent/HybridRelations.php
+++ b/src/Eloquent/HybridRelations.php
@@ -212,7 +212,7 @@ trait HybridRelations
 
         [$type, $id] = $this->getMorphs(Str::snake($name), $type, $id);
 
-        if ($ownerKey === null){
+        if ($ownerKey === null) {
             $ownerKey = $this->getKeyName();
         }
 
--- a/src/Eloquent/HybridRelations.php
+++ b/src/Eloquent/HybridRelations.php
@@ -236,6 +236,11 @@ trait HybridRelations
 
         $ownerKey ??= $instance->getKeyName();
 
+        // Check if it is a relation with an original model.
+        if (! is_subclass_of($instance, MongoDBModel::class)) {
+            return parent::morphTo($name,$type,$id,$ownerKey);
+        }
+
         return new MorphTo(
             $instance->newQuery(),
             $this,
--- a/src/Eloquent/HybridRelations.php
+++ b/src/Eloquent/HybridRelations.php
@@ -212,10 +212,6 @@ trait HybridRelations
 
         [$type, $id] = $this->getMorphs(Str::snake($name), $type, $id);
 
-        if ($ownerKey === null) {
-            $ownerKey = $this->getKeyName();
-        }
-
         // If the type value is null it is probably safe to assume we're eager loading
         // the relationship. When that is the case we will pass in a dummy query as
         // there are multiple types in the morph and we can't use single queries.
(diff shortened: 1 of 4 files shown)


src/Relations/MorphTo.php
```diff
@@ -16,7 +16,11 @@ public function addConstraints()
             // For belongs to relationships, which are essentially the inverse of has one
             // or has many relationships, we need to actually query on the primary key
             // of the related models matching on the foreign key that's on a parent.
-            $this->query->where($this->ownerKey, '=', $this->parent->{$this->foreignKey});
+            $this->query->where(
+                $this->ownerKey ?: $this->query->getModel()->getKeyName(),
```
hans-thomas: Queried model's `key name` sets as the default value for `ownerKey`.

GromNaN: The `$owerKey` property must be set earlier. We are not going to patch all methods that use it. It should be in `HybridRelations::morphTo`.

Also, I think the we should use `getForeignKeyFrom` for the right operand, as in Eloquent: https://github.com/laravel/framework/blob/ea4198a1dcb8dcaf5037f7f1f9cb5d9dc2e1a58e/src/Illuminate/Database/Eloquent/Relations/BelongsTo.php#L97

```php
$this->query->where($this->ownerKey, '=', $this->getForeignKeyFrom($this->child));
```

GromNaN (commented):
I think changes should be done in `HybridRelations`.

There is a previous PR that that tries to fix morphTo https://github.com/mongodb/laravel-mongodb/pull/2640, but it does not solves the issue reported in #2657.

Also,  in `HybridRelations::morphTo`, a "Check if it is a relation with an original model." with fallback to Eloquent is missing.

GromNaN (approved):
LGTM. It fixes the issue reported in https://github.com/mongodb/laravel-mongodb/issues/2657.

src/Eloquent/HybridRelations.php
```diff
@@ -236,6 +236,11 @@ public function morphTo($name = null, $type = null, $id = null, $ownerKey = null
 
         $ownerKey ??= $instance->getKeyName();
```
GromNaN, line 237: This line could go line 214 to avoid the elvis operator line 224.

hans-thomas: At line 214, The `$class` is `Null` and I can't create the instance to get its `key name`.

alcaeus (approved):

alcaeus:
Thank you @hans-thomas!

src/Eloquent/HybridRelations.php
```diff
@@ -221,7 +221,7 @@ public function morphTo($name = null, $type = null, $id = null, $ownerKey = null
                 $this->newQuery(),
                 $this,
                 $id,
-                $ownerKey,
+                $ownerKey ?: $this->getKeyName(),
```
jmikola, line 224: https://github.com/mongodb/laravel-mongodb/issues/2783 suggests that this broke MorphTo relationships when a model uses a custom primary key.

https://github.com/mongodb/laravel-mongodb/pull/3011 includes a proposed fix.

src/Relations/MorphTo.php
```diff
@@ -16,7 +16,11 @@ public function addConstraints()
             // For belongs to relationships, which are essentially the inverse of has one
             // or has many relationships, we need to actually query on the primary key
             // of the related models matching on the foreign key that's on a parent.
-            $this->query->where($this->ownerKey, '=', $this->parent->{$this->foreignKey});
+            $this->query->where(
+                $this->ownerKey,
```
jmikola, line 20: Removing the `$this->getKeyName()` fallback in `HybridRelations::morphTo` and allowing MorphTo to be constructed with a null `$ownerKey` would result in an exception here when `null` is used in place of a string field name.

The relevant fix in https://github.com/mongodb/laravel-mongodb/pull/3011 is to fall back to `$this->getForeignKeyName()` here, which respects a custom primary key name on parent.

tests/RelationsTest.php
```diff
@@ -470,6 +470,25 @@ public function testMorph(): void
         $relations = $photos[1]->getRelations();
         $this->assertArrayHasKey('hasImage', $relations);
         $this->assertInstanceOf(Client::class, $photos[1]->hasImage);
+
+        // inverse
+        $photo = Photo::query()->create(['url' => 'https://graph.facebook.com/hans.thomas/picture']);
+        $client = Client::create(['name' => 'Hans Thomas']);
```
jmikola, line 476: @hans-thomas: Was there a particular reason you used `Photo::query()->create()` above but `Client::create()` here? Best I could tell, they resolve to the same `Builder::create()` method.

`Client::create()` just ends up forwarding the call through `Model::__call()`.

PR #2615, opened 2023-09-11
Remove compatibility layers with old Laravel versions

GromNaN:
To make the library compatible with several major versions of Laravel, abstraction method had to be added by #1116 and [3e26e05b](https://github.com/mongodb/laravel-mongodb/commit/3e26e05b90cf5e207c66e30ea2021ff9ddb16bf9#diff-b815a7989006c7aaa05e7ccda0d141ca8adfa9c2d7d28bef4d38be00b772fd35R260-R267). They are no longer necessary.

PR #2670, opened 2023-11-07
[Feature] Add MorphToMany support

hans-thomas:
Hi,
I continued #2640 PR, thanks to @ithuis. I fixed the relationship and its inverse way, 
but it still needs more tests for 

- [x] `attaching` a collection of models
- [x]  `detaching` action
- [x] finally `syncing` action
- [x] There is a problem with loading relation using `load` method

It has to be done by tomorrow.


hans-thomas:
The implementation is done but to make it perfect, I want to add more tests for custom `parentKey` and `relatedKey` keys.

hans-thomas:
This error is randomly thrown in one job each time.

```
 There was 1 failure:

1) MongoDB\Laravel\Tests\QueueTest::testQueueRelease
Failed asserting that 1699792697 matches expected 1699792698.

/home/runner/work/laravel-mongodb/laravel-mongodb/tests/QueueTest.php:166
```

</details>

src/Relations/MorphToMany.php
```diff
@@ -0,0 +1,460 @@
...
+    /** @inheritdoc */
+    public function addConstraints()
+    {
+        if (static::$constraints) {
+            $this->setWhere();
+        }
+    }
+
+    /** @inheritdoc */
+    public function addEagerConstraints(array $models)
```
hans-thomas, line 53: Override this method to load the relation's data correctly in the inverse mode.

src/Relations/MorphToMany.php
```diff
@@ -0,0 +1,460 @@
...
+    {
+        $changes = [
+            'attached' => [],
+            'detached' => [],
+            'updated' => [],
+        ];
+
+        if ($ids instanceof Collection) {
+            $ids = $this->parseIds($ids);
+        } elseif ($ids instanceof Model) {
```
hans-thomas, line 123: The first argument can be a model instance.

src/Relations/MorphToMany.php
```diff
@@ -0,0 +1,459 @@
...
+    public function attach($id, array $attributes = [], $touch = true)
+    {
+        if ($id instanceof Model) {
+            $model = $id;
+
+            $id = $this->parseId($model);
+
+            if ($this->getInverse()) {
+                // Attach the new ids to the parent model.
+                $this->parent->push($this->table, [
```
hans-thomas, line 193: after any `sync` or `attach` action, I had to refresh the instance to fetch the correct relation's data and I couldn't find the problem. but finally, I figured out that when we push an array in unique mode, after the first time, only the id field was pushed into the instance pivot table (because id is unique). So, I solved this issue on the `push` and `pull` actions by wrapping the data in another array.

src/Relations/MorphToMany.php
```diff
@@ -0,0 +1,461 @@
...
+        // associations, otherwise all the association ties will be broken.
+        // We'll return the numbers of affected rows when we do the deletes.
+        $ids = (array) $ids;
+
+        // Detach all ids from the parent model.
+        if ($this->getInverse()) {
+            // Remove the relation from the parent.
+            $data = [];
+            foreach ($ids as $item) {
+                $data = array_merge($data, [
```
hans-thomas, line 275: Detach multiple ids from the parent pivot column at once.

src/Relations/MorphToMany.php
```diff
@@ -0,0 +1,461 @@
...
+        $dictionary = [];
+
+        foreach ($results as $result) {
+            if ($this->getInverse()) {
+                foreach ($result->$foreign as $item) {
+                    $dictionary[$item][] = $result;
+                }
+            } else {
+                // Collect $foreign value from pivot column of result model
+                $items = $this->extractIds($result->{$this->table} ?? [], $foreign);
```
hans-thomas, line 334: Ids extracted from the results pivot column in `MorphToMany` mode to build the dictionary.

src/Relations/MorphToMany.php
```diff
@@ -0,0 +1,461 @@
...
+
+    /**
+     * Extract ids from given pivot table data
+     *
+     * @param  array       $data
+     * @param  string|null $relatedPivotKey
+     *
+     * @return mixed
+     */
+    private function extractIds(array $data, ?string $relatedPivotKey = null)
```
hans-thomas: I created this method to extract id values from pivot columns.

src/Eloquent/HybridRelations.php
```diff
@@ -328,6 +334,139 @@ public function belongsToMany(
...
+                $relatedKey,
+                $relation,
+                $inverse,
+            );
+        }
+
+        $instance = new $related();
+
+        $foreignPivotKey = $foreignPivotKey ?: $name . '_id'; // labelled_id
+        $relatedPivotKey = $relatedPivotKey ?: $instance->getForeignKey() . 's'; // label_ids
```
GromNaN: These comments can be removed, as they only make sense in a specific case.
```suggestion
        $foreignPivotKey = $foreignPivotKey ?: $name . '_id';
        $relatedPivotKey = $relatedPivotKey ?: $instance->getForeignKey() . 's';
```

src/Eloquent/HybridRelations.php
```diff
@@ -328,6 +334,139 @@ public function belongsToMany(
...
+        $related,
+        $name,
+        $table = null,
+        $foreignPivotKey = null,
+        $relatedPivotKey = null,
+        $parentKey = null,
+        $relatedKey = null,
+        $relation = null,
+    ) {
+        $foreignPivotKey = $foreignPivotKey ?: $this->getForeignKey() . 's';
```
GromNaN: It would be more correct to use the [`Pluralizer`](https://laravel.com/api/10.x/Illuminate/Support/Pluralizer.html#method_plural).

tests/RelationsTest.php
```diff
@@ -491,6 +493,391 @@ public function testMorph(): void
...
+        $label  = Label::query()->create(['name' => 'Had the world in my palms, I gave it to you']);
+
+        $user->labels()->attach($label);
+        $client->labels()->attach($label);
+
+        $this->assertEquals(1, $user->labels->count());
+        $this->assertContains($label->_id, $user->labels->pluck('_id'));
+
+        $this->assertEquals(1, $client->labels->count());
+        $this->assertContains($label->_id, $user->labels->pluck('_id'));
```
GromNaN, line 510: I'd add this checks to ensure the data is correctly stored.

```php
        $checkUser = User::query()->find($user->_id);
        $this->assertEquals(1, $checkUser->labels->count());

        $checkLabel = Label::query()->find($label->_id);
        $this->assertEquals(2, $checkLabel->labelleds->count());
```

You can do the same for other tests.

GromNaN (commented):
Just a few comments for now, I'm still testing.

tests/RelationsTest.php
GromNaN: I think `$checkLabel->labelleds` should be a `Illuminate\Support\Collection`, instead of an `array`.

src/Eloquent/HybridRelations.php
hans-thomas: It's the way is used in other methods like `belongsToMany`. Should they change too?

tests/RelationsTest.php
hans-thomas: Why the pivot column's data should be a `Collection` instead of an `array`?

src/Eloquent/HybridRelations.php
GromNaN: We will have to change it in other place, but this will be a breaking change. Yes please `Str::plural` here to prevent this breaking change for morphTo relations.

Laravel uses it since 5.8: https://github.com/laravel/framework/commit/15161135ea7ca2a47ed07494479c878e99089d87

tests/RelationsTest.php
GromNaN: I think it should be a list of model objects, is the behavior really identical to Laravel?

hans-thomas: The `$label->labelleds` returns the pivot column's data while `$label->users` finds related users using `labelleds` pivot column and returns a collection of users. I believe it's done in the correct way.

GromNaN: Ok. I presume it's not possible to find a list of all "labelled" objects, whatever their type. That would require a list of `$related` classes, which is not implemented in Eloquent.

GromNaN:
Hello @divine would you like to review this PR, please?

tests/RelationsTest.php
hans-thomas: > I'd add this checks to ensure the data is correctly stored.
> 
> ```
>         $checkUser = User::query()->find($user->_id);
>         $this->assertEquals(1, $checkUser->labels->count());
> 
>         $checkLabel = Label::query()->find($label->_id);
>         $this->assertEquals(2, $checkLabel->labelleds->count());
> ```
> 
> You can do the same for other tests.

After any `attach`, `detach`, or `sync` actions, the relation's data is loading from the DB.

```php
$user->labels()->attach($label);
$this->assertEquals(1, $user->labels->count());
```

However, I wrote some tests to reload and validate the relation's data from DB. See lines 648 and 849.

Do you think it is still necessary to check this on every test?

src/Eloquent/HybridRelations.php
```diff
@@ -328,6 +334,139 @@ public function belongsToMany(
...
+            $this,
+            $name,
+            $table,
+            $foreignPivotKey,
+            $relatedPivotKey,
+            $parentKey ?: $this->getKeyName(),
+            $relatedKey ?: $instance->getKeyName(),
+            $relation,
+            $inverse,
+        );
```
GromNaN, line 409: I just found `Str::pluralStudly` which does exactly what we need here. I propose that we inline every default value:

```suggestion
        return new MorphToMany(
            $instance->newQuery(),
            $this,
            $name,
            $table ?? Str::pluralStudly($name);,
            $foreignPivotKey ?? $name . '_id',
            $relatedPivotKey ?? Str::plural($instance->getForeignKey()),
            $parentKey ?? $this->getKeyName(),
            $relatedKey ?? $instance->getKeyName(),
            $relation,
            $inverse,
        );
```

Do you want to open PR on the Laravel repository to propose the same change?
https://github.com/laravel/framework/blob/b79025607e4f1c8d9cec59a998d4e59a0b008b27/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php#L597-L603

src/Eloquent/HybridRelations.php
```diff
@@ -328,6 +334,139 @@ public function belongsToMany(
...
+            $related,
+            $name,
+            $table,
+            $foreignPivotKey,
+            $relatedPivotKey,
+            $parentKey,
+            $relatedKey,
+            $relatedKey,
+            true,
+        );
```
GromNaN, line 452:
```suggestion
        return $this->morphToMany(
            $related,
            $name,
            $table,
            $foreignPivotKey ?? Str::plural($this->getForeignKey()),
            $relatedPivotKey ?? $name . '_id',
            $parentKey,
            $relatedKey,
            $relatedKey,
            true,
        );
```

src/Eloquent/HybridRelations.php
```diff
@@ -328,6 +334,139 @@ public function belongsToMany(
...
+    }
+
+    /**
+     * Get the relationship name of the belongs to many.
+     *
+     * @return string
+     */
+    protected function guessBelongsToManyRelation()
+    {
+        if (method_exists($this, 'getBelongsToManyCaller')) {
```
GromNaN: I removed this method previously in #2615 because I think it is a backward compatibility layer that is no longer necessary. Why would we keep this method?

src/Relations/MorphToMany.php
```diff
@@ -0,0 +1,461 @@
...
+    {
+        // Do nothing.
+    }
+
+    /**
+     * Set the select clause for the relation query.
+     *
+     * @return array
+     */
+    protected function getSelectColumns(array $columns = ['*'])
```
GromNaN: Tests pass if you remove this method. It doesn't exist in parent class.

src/Relations/MorphToMany.php
```diff
@@ -0,0 +1,461 @@
...
+    protected function shouldSelect(array $columns = ['*'])
+    {
+        return $columns;
+    }
+
+    /** @inheritdoc */
+    public function addConstraints()
+    {
+        if (static::$constraints) {
+            $this->setWhere();
```
GromNaN, line 48: Side note: This make me realize that we don't support constraint on relational data. It could be possible using an aggregation `$lookup` stage. Should we throw an exception instead of hiding the fact that this library doesn't support it?

src/Relations/MorphToMany.php
```diff
@@ -0,0 +1,461 @@
...
+        // But on MorphedByMany relation, we collect related ids from pivot column
+        // and add to a whereIn condition
+        if ($this->getInverse()) {
+            $ids = $this->getKeys($models, $this->table);
+            $ids = $this->extractIds($ids[0] ?? []);
+            $this->query->whereIn($this->relatedKey, $ids);
+        } else {
+            parent::addEagerConstraints($models);
+
+            $this->query->where($this->qualifyPivotColumn($this->morphType), $this->morphClass);
```
GromNaN, line 65: This `qualifyPivotColumn` may require the same neutralization that what I did in #2676: In MongoDB, field names are never prefixed by collection name.

src/Relations/MorphToMany.php
```diff
@@ -0,0 +1,461 @@
...
+    /** @inheritdoc */
+    public function getQualifiedForeignPivotKeyName()
+    {
+        return str_contains($this->foreignPivotKey, '.')
+            ? $this->foreignPivotKey
+            : $this->table . '.' . $this->foreignPivotKey;
+    }
+
+    /** @inheritdoc */
+    public function getQualifiedRelatedPivotKeyName()
```
GromNaN, line 361: This method overload can be removed once `qualifyPivotColumn` have been neutralized.

src/Relations/MorphToMany.php
```diff
@@ -0,0 +1,461 @@
...
+     *
+     * @return string
+     */
+    public function getForeignKey()
+    {
+        return $this->foreignPivotKey;
+    }
+
+    /** @inheritdoc */
+    public function getQualifiedForeignPivotKeyName()
```
GromNaN: This method overload can be removed once `qualifyPivotColumn` have been neutralized.

src/Relations/MorphToMany.php
```diff
@@ -0,0 +1,461 @@
...
+     *
+     * @return mixed
+     */
+    private function extractIds(array $data, ?string $relatedPivotKey = null)
+    {
+        $relatedPivotKey = $relatedPivotKey ?: $this->relatedPivotKey;
+        return array_reduce($data, function ($carry, $item) use ($relatedPivotKey) {
+            if (is_array($item) && array_key_exists($relatedPivotKey, $item)) {
+                $carry[] = $item[$relatedPivotKey];
+            } elseif (is_string($item) && ! str_contains($item, '\\')) {
```
GromNaN: This condition is not covered by tests. What is it for?

src/Relations/MorphToMany.php
```diff
@@ -0,0 +1,461 @@
...
+    public function getQualifiedRelatedPivotKeyName()
+    {
+        return $this->relatedPivotKey;
+    }
+
+    /**
+     * Format the sync list so that it is keyed by ID. (Legacy Support)
+     * The original function has been renamed to formatRecordsList since Laravel 5.3.
+     *
+     * @deprecated
```
GromNaN: It's strange to introduce a new method that is already deprecated. We support Laravel 10.x, so I think this backward-compatibility layer can be removed.

GromNaN (commented):
Thanks again for taking on this complex subject.

My idea for this new feature is that we can omit any backward-compatibility code that exists in Laravel or for other relations in this library.

src/Eloquent/HybridRelations.php
hans-thomas: Yes, I'm going to open a PR. much thanks.

hans-thomas: My PR on Laravel closed under 6 minutes and that's kinda a new record for me😂
However, I noticed there is some difference in regex expressions used in `morphToMany` and `pluralStudly` methods.
The regex used in `pluralStudly` is `/(.)(?=[A-Z])/u` which splits the value followed by an uppercase letter. but in the `morphToMany` method we have `/(_)/u` regex which splits the value using `_` character.

GromNaN: I'm sorry about that. So let's make the MongoDB code as close as possible to Eloquent code. Thanks for investigating.

src/Relations/MorphToMany.php
hans-thomas: The `$this->table` is the representer of the pivot column's name, not the collection name.
Let's assume we have the `labelleds` pivot column. the `$this->morphType` returns `labelled_type`. So, `qualifyPivotColumn` method just prefixes the  `$this->morphType` with pivot columns. the result of `qualifyPivotColumn` is `labelleds.labelled_type`.
If I neutralize this method, it will throw errors.

src/Relations/MorphToMany.php
hans-thomas: Since the pivot column stores on `MorphedByMany` model, we just need to prefix the `$this->foreignPivotKey` with the pivot column name. On the other side, on `MorphToMany` model, the relationship's data will be stored as an array and doesn't have any nested column to be prefixed with.

src/Relations/MorphToMany.php
hans-thomas: I removed this method because the result was the same as the parent.

src/Relations/MorphToMany.php
hans-thomas: This method is also used in `BelongsToMany` relation.

src/Relations/MorphToMany.php
hans-thomas: It was related to an issue before I fixed it.

GromNaN (approved):
LGTM

src/Relations/MorphToMany.php
```diff
@@ -0,0 +1,407 @@
...
+{
+    /**
+     * Get the key for comparing against the parent key in "has" query.
+     *
+     * @return string
+     */
+    public function getHasCompareKey()
+    {
+        return $this->foreignPivotKey;
+    }
```
hans-thomas: I guess this won't work. I will write some tests for that tomorrow.

src/Relations/MorphToMany.php
```diff
@@ -0,0 +1,407 @@
...
+
+    /** @inheritdoc */
+    public function getRelationExistenceQuery(Builder $query, Builder $parentQuery, $columns = ['*'])
+    {
+        return $query;
+    }
```
hans-thomas, line 30: I think the parent method is fine and we don't need to override this. I will write test for this too.

src/Relations/MorphToMany.php
hans-thomas: It's more complicated than I thought. I think we have to make some changes [here](https://github.com/mongodb/laravel-mongodb/blob/1c2d1fe29a53278ec34a50c1a289b5e3a8f1a838/src/Helpers/QueriesRelationships.php#L117) to fix this issue.

hans-thomas: I removed this method because it has no use in `has` querying.

src/Helpers/QueriesRelationships.php
```diff
@@ -114,13 +117,46 @@ public function addHybridHas(Relation $relation, $operator = '>=', $count = 1, $
             $not = ! $not;
         }
 
-        $relations = $hasQuery->pluck($this->getHasCompareKey($relation));
+        $relations = match (true) {
+            $relation instanceof MorphToMany => $relation->getInverse() ?
+              $this->handleMorphedByMany($hasQuery, $relation) :
+              $this->handleMorphToMany($hasQuery, $relation),
+            default => $hasQuery->pluck($this->getHasCompareKey($relation))
+        };
```
hans-thomas, line 125: I added a `match` expression to handle the `MorphToMany` relations `has` query.

src/Relations/MorphToMany.php
hans-thomas: This one is related to `HybridRelations` and never called in my tests.

src/Eloquent/HybridRelations.php
```diff
@@ -328,6 +333,125 @@ public function belongsToMany(
...
+        // If no relationship name was passed, we will pull backtraces to get the
+        // name of the calling function. We will use that function name as the
+        // title of this relation since that is a great convention to apply.
+        if ($relation === null) {
+            $relation = $this->guessBelongsToManyRelation();
+        }
+
+        // Check if it is a relation with an original model.
+        if (! is_subclass_of($related, Model::class)) {
+            return parent::MorphToMany(
```
alcaeus: Should this be `morphToMany` (with a minuscule "m")? I know method calls aren't case-sensitive in PHP, but still 😉 
```suggestion
            return parent::morphToMany(
```

alcaeus (approved):
Just a minor comment. Thanks for working on this @hans-thomas, and I really appreciate the amount of comments you added to the code! ❤️

src/Helpers/QueriesRelationships.php
```diff
@@ -114,13 +117,46 @@ public function addHybridHas(Relation $relation, $operator = '>=', $count = 1, $
...
+     * @return Collection
+     */
+    private function handleMorphToMany($hasQuery, $relation)
+    {
+        // First we select the parent models that have a relation to our related model,
+        // Then extracts related model's ids from the pivot column
+        $hasQuery->where($relation->getTable() . '.' . $relation->getMorphType(), get_class($relation->getParent()));
+        $relations = $hasQuery->pluck($relation->getTable());
+        $relations = $relation->extractIds($relations->flatten(1)->toArray(), $relation->getForeignPivotKeyName());
+        return collect($relations);
```
GromNaN, line 146: Please always add an empty line before a `return`. phpcs doesn't enforce it.
```suggestion

        return collect($relations);
```

src/Helpers/QueriesRelationships.php
```diff
@@ -114,13 +117,46 @@ public function addHybridHas(Relation $relation, $operator = '>=', $count = 1, $
...
+    /**
+     * @param Builder  $hasQuery
+     * @param Relation $relation
+     *
+     * @return Collection
+     */
+    private function handleMorphedByMany($hasQuery, $relation)
+    {
+        $hasQuery->whereNotNull($relation->getForeignPivotKeyName());
+        return $hasQuery->pluck($relation->getForeignPivotKeyName())->flatten(1);
```
GromNaN, line 159:
```suggestion

        return $hasQuery->pluck($relation->getForeignPivotKeyName())->flatten(1);
```

tests/RelationsTest.php
```diff
@@ -501,10 +947,10 @@ public function testHasManyHas(): void
         User::create(['name' => 'Anonymous author']);
         Book::create(['title' => 'Anonymous book', 'rating' => 1]);
 
-        $authors = User::has('books')->get();
-        $this->assertCount(2, $authors);
-        $this->assertEquals('George R. R. Martin', $authors[0]->name);
-        $this->assertEquals('John Doe', $authors[1]->name);
+//        $authors = User::has('books')->get();
```
GromNaN: Why is it commented?

src/Relations/MorphToMany.php
GromNaN: Great 👍🏻  It seems that you have tested this case carefully.

hans-thomas:
@alcaeus My pleasure❤️

src/Relations/MorphToMany.php
hans-thomas: I'm not sure about that. If we throw an error in `addConstraints` method then we need another method to act like it. I mean this method will be called when we want to read the relation's data, not when we are working on the relation (like `sync`ing or any other actions). So, to create a method like `addConstraints`, we will need to override at least one method on the parent class. I think it requires too much effort to implement and update relationship classes while everything is working fine right now.

GromNaN (commented):
LGTM. A last question about the commented code and minor CS fixes.

tests/RelationsTest.php
hans-thomas: I have no clue😅
Reverted in 9198df9.

GromNaN (approved):

GromNaN:
Thank you so much @hans-thomas

hans-thomas:
@GromNaN I'm glad I could help❤️

PR #2671, opened 2023-11-08
Add tests to cover morphTo with hybrid MongoDB and SQL

GromNaN:
Follow #2669

Add tests for SQL model to MongoDB model morph relation in both directions.

In tests, I rename SQL tables to singular (common practice) to ensure we don't mess with collection name of similar models. It helps when debugging.

### DIFF
--- a/tests/HybridRelationsTest.php
+++ b/tests/HybridRelationsTest.php
@@ -7,8 +7,10 @@ namespace MongoDB\Laravel\Tests;
 use Illuminate\Database\SQLiteConnection;
 use Illuminate\Support\Facades\DB;
 use MongoDB\Laravel\Tests\Models\Book;
+use MongoDB\Laravel\Tests\Models\Photo;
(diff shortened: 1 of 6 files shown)


tests/HybridRelationsTest.php
```diff
@@ -210,4 +217,63 @@ public function testHybridWith()
...
+
+    public function testHybridMorphToSql(): void
+    {
+        $user = User::create(['name' => 'John Doe']);
+
+        $photo = SqlPhoto::create(['url' => 'http://graph.facebook.com/john.doe/picture']);
+        $photo->save();
+        $photo = $user->photos()->save($photo);
+
+        $this->assertEquals(1, $user->photos->count());
```
GromNaN, line 258: Fails here: 
```
Failed asserting that 0 matches expected 1.
```

But this error seems to come from Laravel itself: if I replace `User` with `SqlUser` and remove the `HybridRelations` trait from every Sql model, it still fails. I may have an error in the model definition.

PR #2674, opened 2023-11-09
Avoid time-sensible tests from failing randomly

GromNaN:
Freeze the clock for tests that use `now()` function.

Tests introduced by #2658 and #2653


alcaeus (approved):

GromNaN:
Freeze the clock for tests that use now() function.

PR #2676, opened 2023-11-10
PHPORM-119 Fix integration with Spatie Query Builder - Don't qualify field names in document models

### LINKED ISSUE #2409: Spatie's QueryBuilder support
- Laravel-mongodb Version: 3.9.9
- PHP Version: 8.1.0
- Database Driver & Version: 

### Description:
Queries, populated with Spaties's query builder are not correct. 
Spatie's package adds table prefix for WHERE filters: `select * from table 'example' ... where 'example'.'column' = ?`
MongoDB driver can not execute such queries, as it interprets this query as deep object comparison (`{ "example": { "column": "value" } }`) instead of `{ "column": "value" }`

Adding table (or collection) prefix to query is correct behavior and part of Model function: ``

Temporal fix is adding this to your model:

```php
    public function qualifyColumn($column)
    {
        return Str::after($column, $this->getTable() . '.');
    }
````

### Steps to reproduce
1. `composer require spatie/query-builder`
2. Create MongoDB model
3. ```php
    QueryBuilder::for(MongoModel::class)
      ->allowedFiilter([
        AllowedFilter::exact('column'),
      ])
      ->dd();
    ```
4. `/some-route?filter[column]=value`

### Expected behaviour
```
{
  "$and": {
    "column": "value",
  }
}
```

### Actual behaviour
```
{
  "$and": {
    "mongo-model.column": "value",
  }
}
```

### PR DESCRIPTION
**Pull request by GromNaN**
Fix #2409 PHPORM-119

As far as I know, there are no cases where you would want to prefix the field name with the collection name. Except in `$lookup` aggregation pipeline, but that's out of scope of the query builder.

Note that I'm adding the package [`spatie/laravel-query-builder`](https://packagist.org/packages/spatie/laravel-query-builder) as a dev dependency to ensure we will always support it. It's a very popular package that looks stable.

### DIFF
--- a/tests/ExternalPackageTest.php
+++ b/tests/ExternalPackageTest.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Laravel\Tests;
+
+use Composer\InstalledVersions;
+use Illuminate\Http\Request;
+use MongoDB\Laravel\Tests\Models\User;
+use Spatie\QueryBuilder\AllowedFilter;
+use Spatie\QueryBuilder\AllowedSort;
+use Spatie\QueryBuilder\QueryBuilder;
+
+/**
+ * Test integration with external packages.
+ */
+class ExternalPackageTest extends TestCase
+{
+    protected function tearDown(): void
+    {
+        parent::tearDown();
+
+        User::truncate();
+    }
+
+    public function testSpacieQueryBuilder(): void
+    {
+        if (! InstalledVersions::isInstalled('spatie/laravel-query-builder')) {
+            $this->markTestSkipped('spatie/laravel-query-builder is not installed.');
+        }
+
+        User::insert([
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: tests/ExternalPackageTest.php [hunk 1]
```diff
@@ -0,0 +1,51 @@
...
+    protected function tearDown(): void
+    {
+        parent::tearDown();
+
+        User::truncate();
+    }
+
+    public function testSpacieQueryBuilder(): void
+    {
+        if (! InstalledVersions::isInstalled('spatie/laravel-query-builder')) {
```
alcaeus: Not critical, but do we need to check whether this is installed? I'd say that the test should fail if the dependency was removed instead of silently being skipped, as we want to maintain compatibility with the package.

alcaeus (approved):
Functionality LGTM, just a question about skipping the test.

tests/ExternalPackageTest.php
GromNaN: Indeed, and the package will always be present in dev.

Issue #2129, opened 2020-10-23
Nested date format is different from top level in response


alexsmko:
- Laravel-mongodb Version: 3.7
- PHP Version: 7.4.9

### Steps to reproduce
1. create model with propery "nested.date"
2. return it as JsonResource for example 
3. you will date in "2020-10-23 09:12:19" format where other dates in "2020-10-23T09:12:19.512000Z"

It looks like $this->serializeDate() should be called in [Mongodb/Eloquent/Model.php#L219](https://github.com/jenssegers/laravel-mongodb/blob/3.7/src/Jenssegers/Mongodb/Eloquent/Model.php#L219) in order to mimic Eloquent's addDateAttributesToArray() in [Eloquent/Concerns/HasAttributes.php#L168](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L168) trait

divine:
Hello,

I see that you've already debugged and found resolution, could you please push a new PR with tests?

Thanks!

alexsmko:
Will try later. Also there is another related bug: new object of such Model fails
```
[2020-10-23 10:50:25] production.ERROR: invalid document for insert: keys cannot contain ".": "last_request.date" {"exception":"[object] (MongoDB\\Driver\\Exception\\InvalidArgumentException(code: 22): invalid document for insert: keys cannot contain \".\": \"last_request.date\" at /src/vendor/mongodb/mongodb/src/Operation/InsertOne.php:132)
```
in my case last_request is null and there is no nested date, didn't dig deep into this,
added this overrides on my model as temporary solution


```
    // Skip setting dotted dates as attribute for new instance
    public function __construct(array $attributes = [])
    {
        parent::__construct($attributes);
        foreach ($this->getDates() as $key) {
            if (Str::contains($key, '.')) {
                unset($this->attributes[$key]);
            }
        }
    }

    // Avoid issue with formatting nested dates
    public function attributesToArray()
    {
        $attributes = parent::attributesToArray();
        foreach ($this->getDates() as $key) {
            if (Str::contains($key, '.') && Arr::has($attributes, $key)) {
                Arr::set($attributes, $key, $this->serializeDate($this->asDateTime(Arr::get($attributes, $key))));
            }
        }
        return $attributes;
    }
```

MeerKatDev:
I'm having the same issue, was it merged so far?

divine:
[...]

PR #2678, opened 2023-11-14
Update docker and test configs

hans-thomas:
Hello there,
I updated the `Dockerfile` and `docker-compose` files. I also added some scripts to the `composer` that might be used very often.

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -55,12 +55,12 @@ jobs:
                     tools: "composer"
 
             -   name: "Show PHP version"
+                if: ${{ secrets.DEBUG == 'true' }}
                 run: "php -v && composer -V"
 
             -   name: "Show Docker version"
-                run: if [[ "$DEBUG" == "true" ]]; then docker version && env; fi
-                env:
-                    DEBUG: ${{ secrets.DEBUG }}
+                if: ${{ secrets.DEBUG == 'true' }}
+                run: "docker version && env"
 
             -   name: "Download Composer cache dependencies from cache"
                 id: "composer-cache"
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -80,8 +80,7 @@ jobs:
                     composer update --no-interaction $([[ "${{ matrix.mode }}" == low-deps ]] && echo ' --prefer-lowest --prefer-stable')
 
             -   name: Run tests
-                run: |
-                    ./vendor/bin/phpunit --coverage-clover coverage.xml
+                run: composer test:coverage
                 env:
                     MONGODB_URI: 'mongodb://127.0.0.1/?replicaSet=rs'
 
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -9,7 +9,9 @@ on:
 jobs:
     build:
         runs-on: ${{ matrix.os }}
+
         name: PHP v${{ matrix.php }} with MongoDB ${{ matrix.mongodb }} ${{ matrix.mode }}
+
         strategy:
             matrix:
                 os:
@@ -30,6 +32,7 @@ jobs:
 
         steps:
             -   uses: actions/checkout@v4
+
             -   name: Create MongoDB Replica Set
                 run: |
                     docker run --name mongodb -p 27017:27017 -e MONGO_INITDB_DATABASE=unittest --detach mongo:${{ matrix.mongodb }} mongod --replSet rs --setParameter transactionLifetimeLimitSeconds=5
@@ -39,10 +42,12 @@ jobs:
                     sleep 1
                     done
                     sudo docker exec --tty mongodb $MONGOSH_BIN 127.0.0.1:27017 --eval "rs.initiate({\"_id\":\"rs\",\"members\":[{\"_id\":0,\"host\":\"127.0.0.1:27017\" }]})"
+
             -   name: Show MongoDB server status
                 run: |
                     if [ "${{ matrix.mongodb }}" = "4.4" ]; then MONGOSH_BIN="mongo"; else MONGOSH_BIN="mongosh"; fi
                     docker exec --tty mongodb $MONGOSH_BIN 127.0.0.1:27017 --eval "db.runCommand({ serverStatus: 1 })"
+
             -   name: "Installing php"
                 uses: shivammathur/setup-php@v2
                 with:
@@ -50,29 +55,36 @@ jobs:
                     extensions: curl,mbstring,xdebug
                     coverage: xdebug
                     tools: composer
+
             -   name: Show PHP version
                 run: php -v && composer -V
+
             -   name: Show Docker version
                 run: if [[ "$DEBUG" == "true" ]]; then docker version && env; fi
                 env:
                     DEBUG: ${{secrets.DEBUG}}
+
             -   name: Download Composer cache dependencies from cache
                 id: composer-cache
                 run: echo "dir=$(composer config cache-files-dir)" >> $GITHUB_OUTPUT
+
             -   name: Cache Composer dependencies
                 uses: actions/cache@v3
                 with:
                     path: ${{ steps.composer-cache.outputs.dir }}
                     key: ${{ matrix.os }}-composer-${{ hashFiles('**/composer.json') }}
                     restore-keys: ${{ matrix.os }}-composer-
+
             -   name: Install dependencies
                 run: |
                     composer update --no-interaction $([[ "${{ matrix.mode }}" == low-deps ]] && echo ' --prefer-lowest --prefer-stable')
+
             -   name: Run tests
                 run: |
                     ./vendor/bin/phpunit --coverage-clover coverage.xml
                 env:
                     MONGODB_URI: 'mongodb://127.0.0.1/?replicaSet=rs'
+
             -   uses: codecov/codecov-action@v3
                 with:
                     token: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -48,7 +48,7 @@ jobs:
                     if [ "${{ matrix.mongodb }}" = "4.4" ]; then MONGOSH_BIN="mongo"; else MONGOSH_BIN="mongosh"; fi
                     docker exec --tty mongodb $MONGOSH_BIN 127.0.0.1:27017 --eval "db.runCommand({ serverStatus: 1 })"
 
-            -   name: "Installing php"
+            -   name: Installing php
                 uses: shivammathur/setup-php@v2
                 with:
                     php-version: ${{ matrix.php }}
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -1,4 +1,4 @@
-name: CI
+name: "CI"
 
 on:
     push:
@@ -8,32 +8,32 @@ on:
 
 jobs:
     build:
-        runs-on: ${{ matrix.os }}
+        runs-on: "${{ matrix.os }}"
 
-        name: PHP v${{ matrix.php }} with MongoDB ${{ matrix.mongodb }} ${{ matrix.mode }}
+        name: "PHP v${{ matrix.php }} with MongoDB ${{ matrix.mongodb }} ${{ matrix.mode }}"
 
         strategy:
             matrix:
                 os:
-                    - ubuntu-latest
+                    - "ubuntu-latest"
                 mongodb:
-                    - '4.4'
-                    - '5.0'
-                    - '6.0'
-                    - '7.0'
+                    - "4.4"
+                    - "5.0"
+                    - "6.0"
+                    - "7.0"
                 php:
-                    - '8.1'
-                    - '8.2'
-                    - '8.3'
+                    - "8.1"
+                    - "8.2"
+                    - "8.3"
                 include:
-                    - php: '8.1'
-                      mongodb: '5.0'
-                      mode: 'low-deps'
+                    - php: "8.1"
+                      mongodb: "5.0"
+                      mode: "low-deps"
 
         steps:
-            -   uses: actions/checkout@v4
+            -   uses: "actions/checkout@v4"
 
-            -   name: Create MongoDB Replica Set
+            -   name: "Create MongoDB Replica Set"
                 run: |
                     docker run --name mongodb -p 27017:27017 -e MONGO_INITDB_DATABASE=unittest --detach mongo:${{ matrix.mongodb }} mongod --replSet rs --setParameter transactionLifetimeLimitSeconds=5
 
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -1,19 +1,11 @@
-ARG PHP_VERSION=8.1
-
-FROM php:${PHP_VERSION}-cli
+FROM php:8.1-cli
```
hans-thomas: Minimum PHP version supported according to `composer.json` file.

Dockerfile
```diff
@@ -1,19 +1,11 @@
...
 
-COPY --from=composer:2.6.2 /usr/bin/composer /usr/local/bin/composer
-
-ENV COMPOSER_ALLOW_SUPERUSER=1
-
-WORKDIR /code
-
-COPY ./ ./
-
-CMD ["bash", "-c", "composer install && ./vendor/bin/phpunit --testdox"]
```
hans-thomas: These lines were ineffective. I moved some of them to the `docker-compose` file. However, To run the tests, I added scripts to the `composer` to make it easy for users to run the commands.

.github/workflows/build-ci.yml
```diff
@@ -39,40 +42,48 @@ jobs:
...
                     restore-keys: ${{ matrix.os }}-composer-
+
             -   name: Install dependencies
                 run: |
                     composer update --no-interaction $([[ "${{ matrix.mode }}" == low-deps ]] && echo ' --prefer-lowest --prefer-stable')
+
             -   name: Run tests
-                run: |
-                    ./vendor/bin/phpunit --coverage-clover coverage.xml
+                run: composer test:coverage
```
hans-thomas: Use a composer script in `build-ci` workflow.

Dockerfile
```diff
@@ -1,19 +1,11 @@
...
+# Install Composer
+RUN curl https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer
```
GromNaN: Getting the composer executable from the composer image it correct with Docker. It provides more security and allows caching.

.github/workflows/build-ci.yml
```diff
@@ -39,40 +42,48 @@ jobs:
...
                     done
                     sudo docker exec --tty mongodb $MONGOSH_BIN 127.0.0.1:27017 --eval "rs.initiate({\"_id\":\"rs\",\"members\":[{\"_id\":0,\"host\":\"127.0.0.1:27017\" }]})"
+
             -   name: Show MongoDB server status
                 run: |
                     if [ "${{ matrix.mongodb }}" = "4.4" ]; then MONGOSH_BIN="mongo"; else MONGOSH_BIN="mongosh"; fi
                     docker exec --tty mongodb $MONGOSH_BIN 127.0.0.1:27017 --eval "db.runCommand({ serverStatus: 1 })"
-            -   name: "Installing php"
+
+            -   name: Installing php
```
GromNaN: Quoting strings in Yaml is a good practice to avoid unexpected behaviors like `no` being converted to `false`. So I'm not in favor of removing this quotes everywhere.

.github/workflows/build-ci.yml
GromNaN: It makes things a little more complicated if we want to add new options to this command.

GromNaN (changes requested):
Apart from composer scripts, I don't think these changes improve anything.
Thanks for the suggestion.

.github/workflows/build-ci.yml
hans-thomas: It can be done as follows
```bash
composer test:coverage -- --new-argument
```
I don't think it's going to make things more complicated. 
Let's assume we used a command in two places with several arguments. If we want a change, we must apply that change in two places, and it increases the chance of an error.
What I suggest is, defining common commands as composer scripts and using them everywhere. If somewhere we need to add an extra argument, we can pass it like the example code I mentioned.
These scripts can make workflows cleaner and also make developing easier and faster.

hans-thomas:
> Apart from composer scripts, I don't think these changes improve anything.

I don't want to improve performance or anything else. I just want to update the docker configuration to be more effective and make developing easier.

If you look at the current [dockerfile](https://github.com/mongodb/laravel-mongodb/blob/4.1/Dockerfile), in the last line, it is installing dependencies and run the tests. which means to run the tests, I have to build the `dockerfile` each time?
It's obvious developers must build `dockerfile` the first time and then run the tests using command line inside of the `countainer`.

Dockerfile
hans-thomas: I replaced this line
```bash
COPY --from=composer:2.6.2 /usr/bin/composer /usr/local/bin/composer
```
Because the composer's version was fixed.
Also, we don't need to cache, because developers build this file one time, and the second reason is if they decide to rebuild this file, they definitely want the updates.

hans-thomas:
Another update for the project's structure can be using [StyleCI](https://styleci.io/) instead of PHP Code Sniffer. StyleCI can be configured to fix the code style using an automated merge commit.
Fixing code style after every change (after the pipeline failed mostly😅) is annoying.

GromNaN:
> If you look at the current [dockerfile](https://github.com/mongodb/laravel-mongodb/blob/4.1/Dockerfile), in the last line, it is installing dependencies and run the tests. which means to run the tests, I have to build the `dockerfile` each time? It's obvious developers must build `dockerfile` the first time and then run the tests using command line inside of the `countainer`.

The last line is the default command when you run the container.
```
docker-compose run --rm tests
```

Note that I don't use this docker image to develop because the integration with PHPStorm and xdebug set-by-step debugging is hard to setup.

So I just found that the memory limit is too low. This lines should be added after `docker-php-ext-install -j$(nproc) zip`.

```
    cp /usr/local/etc/php/php.ini-development /usr/local/etc/php/php.ini && \
    echo "memory_limit=-1" >> /usr/local/etc/php/php.ini
```

> Another update for the project's structure can be using [StyleCI](https://styleci.io/) instead of PHP Code Sniffer. StyleCI can be configured to fix the code style using an automated merge commit.
> Fixing code style after every change (after the pipeline failed mostly😅) is annoying.

Not all CS issues can be fixed automatically, but in 90% of cases, we could make it easier for external contributors to contribute by automatically committing modifications proposed by phpcbf.

hans-thomas:
> The last line is the default command when you run the container.

It had never run tests on my computer. It is even mentioned [here](https://github.com/mongodb/laravel-mongodb/issues/2129#issuecomment-858027089).

> we could make it easier for external contributors to contribute by automatically committing modifications proposed by phpcbf.

So you mean it can be done automatically? How?

GromNaN:
> > we could make it easier for external contributors to contribute by automatically committing modifications proposed by phpcbf.
> 
> So you mean it can be done automatically? How?

As proposed by @Treggats in [#2664](https://github.com/mongodb/laravel-mongodb/pull/2664/files#diff-dfd15ae70247bb45f121522ec2e50e5da9aed17e23116d94ec887a4769c44294).

hans-thomas:
> As proposed by @Treggats in [#2664](https://github.com/mongodb/laravel-mongodb/pull/2664/files#diff-dfd15ae70247bb45f121522ec2e50e5da9aed17e23116d94ec887a4769c44294).

That's interesting, I might work on that PR next. Seems an important PR and the best opportunity for me to work with `PHPStan`.

Treggats:
> As proposed by @Treggats in [#2664](https://github.com/mongodb/laravel-mongodb/pull/2664/files#diff-dfd15ae70247bb45f121522ec2e50e5da9aed17e23116d94ec887a4769c44294).

Thanks for the reminder that I needed to check the test suite. Hope that I can open it pretty soon

**Edit**
I've opened #2664 for review again.

Dockerfile
```diff
@@ -1,19 +1,18 @@
...
     docker-php-ext-install -j$(nproc) zip
 
-COPY --from=composer:2.6.2 /usr/bin/composer /usr/local/bin/composer
-
-ENV COMPOSER_ALLOW_SUPERUSER=1
-
-WORKDIR /code
+# Increase memory limit
+RUN cp /usr/local/etc/php/php.ini-development /usr/local/etc/php/php.ini && \
+    echo "memory_limit=-1" >> /usr/local/etc/php/php.ini
```
hans-thomas: Memory limit increased.

Dockerfile
```diff
@@ -1,19 +1,18 @@
...
 
-COPY ./ ./
+# Enable coverage mode in xdebug
+RUN echo "xdebug.mode=coverage" >> /usr/local/etc/php/conf.d/xdebug.ini
```
hans-thomas: Xdebug mode is set to `coverage`.

composer.json
```diff
@@ -57,6 +57,13 @@
             ]
         }
     },
+    "scripts": {
+        "test": "./vendor/bin/phpunit",
+        "test:group": "./vendor/bin/phpunit --group test",
```
GromNaN: We don't have `@group test` in the codebase. What's the purpose of this?

I use `./vendor/bin/phpunit --filter 'ModelTest::testNewModel'` to target a specific test (or PHPStorm PHPUnit integration of course!).

hans-thomas: Yes, that's true. `--group` is another option to run the tests. we can select the tests that we want to execute by adding a docblock.
```php
/** @group test */
public function testWithCountHasMany(): void {}
```
The advantage is, that you can select one or more tests from several classes.
I often write multiple tests for a PR. I add this docblock to my tests to ensure that my changes don't break the previous tests. I prefer this method because adding a docblock to the code is easier (for me at least).

hans-thomas:
I don't have any more changes to propose and I think it's the end of this PR.
Please take every feature that you find helpful and revert the rest.

hans-thomas:
If you want to merge this PR, you can do it after merging #2664. I think there are some conflicts.

hans-thomas:
Conflicts resolved.

composer.json
Treggats: Looking through the codebase, you'd want to use the attribute for this. No docblocks are used for indicating a test or dataprovider. Well there is one, which should be a attribute in my eyes, but in general you'd want to use a attribute

```php
    #[Test]
    #[Group('hide-properties')]
    public function canHideCertainProperties(): void {}
```

More on topic of the change, personally I would not hardcode something like this. As you would use group with different groups. Not just the one defined in the composer.json file

Dockerfile
Treggats: Is this still required after the phpunit ini setting?

composer.json
hans-thomas: There is no difference between the docblock and the attribute. You can use anything you like. I just suggest adding `./vendor/bin/phpunit --group test` command as a script to save developers time to write this command each time.
Developers can use this script and add the docblock/attribute to the test cases that want to run. After that, they must remove the docblock/attribute, because that's just for testing.

> personally I would not hardcode something like this.

I'm not a big fan of hardcoding but I found this command very general and useful. As I said, It's just a suggestion, and if it doesn't make any sense, we can ignore it.

Dockerfile
GromNaN: Please revert. You can set a less restrictive version constraint.

```php
COPY --from=composer:2 /usr/bin/composer /usr/local/bin/composer
```

> Also, we don't need to cache, because developers build this file one time, and the second reason is if they decide to rebuild this file, they definitely want the updates.

This cache is also cross-projects. It's more durable than the getcomposer.org url.

Dockerfile
GromNaN: The memory limit in phpunit.xml is enough.
Could you merge the `RUN` stages for php.ini copy and xdebug mode.

composer.json
GromNaN: Please revert.

.github/workflows/build-ci.yml
```diff
@@ -37,35 +40,43 @@ jobs:
...
-            -   name: Show PHP version
-                run: php -v && composer -V
-            -   name: Show Docker version
+                    extensions: "curl,mbstring,xdebug"
+                    coverage: "xdebug"
+                    tools: "composer"
+
+            -   name: "Show PHP version"
+                run: "php -v && composer -V"
+
```
Treggats: Should this always run, or only when debug is set?
If only when debug is set, then the following suggestion might be a bit simpler / easier to understand than the bash check.

```suggestion
            -   name: "Show PHP version"
                if: ${{ secrets.DEBUG == 'true' }}
                run: "php -v && composer -V"

```

Dockerfile
Treggats: I'm assuming that you removed the argument because it wasn't used? Sort of inlining the version.

Which raises the question wether or not we should give the developer the option which PHP version to use.
I can imagine the tests might pass on one version and fail on the next version. But I don't work with docker on a day to day basis, so maybe my understand of this is lacking.

.github/workflows/build-ci.yml
```diff
@@ -1,34 +1,37 @@
...
 
 on:
     push:
     pull_request:
 
 jobs:
     build:
-        runs-on: ${{ matrix.os }}
-        name: PHP v${{ matrix.php }} with MongoDB ${{ matrix.mongodb }} ${{ matrix.mode }}
+        runs-on: "${{ matrix.os }}"
```
Treggats, line 9: I believe this change is based on a comment in another PR, but I wonder if it belongs in this PR. As in my eyes it is a code style thing, which adds a lot of noise to the PR. Personally I would create a separate PR to add these changes.

Though personally I don't see a benefit to these changes, but that's me and the project style is leading.

.github/workflows/build-ci.yml
```diff
@@ -37,35 +40,43 @@ jobs:
...
+                uses: "shivammathur/setup-php@v2"
                 with:
                     php-version: ${{ matrix.php }}
-                    extensions: curl,mbstring,xdebug
-                    coverage: xdebug
-                    tools: composer
-            -   name: Show PHP version
-                run: php -v && composer -V
-            -   name: Show Docker version
+                    extensions: "curl,mbstring,xdebug"
```
Treggats, line 53: Not related to the changes, as those are CS changes.

From the action-runners docs/[readme](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md)
> Both Xdebug and PCOV extensions are installed, but only Xdebug is enabled.

Does this get overridden with setting the extensions manually?

.github/workflows/build-ci.yml
hans-thomas: The next step can be updated like this.

.github/workflows/build-ci.yml
hans-thomas: reverted.

Dockerfile
hans-thomas: We run the tests with the minimum supported PHP version locally. We can see the result of running tests with several versions of PHP and MongoDB when we push the changes. That's why we use strategies in the workflow.

I believe we should leave it to workflows because developers shouldn't run tests with several versions of PHP in combination with several MongoDB versions. It takes time while workflows can do it very accurately and fast.

GromNaN:
I updated the branch with the followings:
- Revert the `PHP_VERSION` argument, so it's easy to test with other PHP versions. Even if that's not the main usage, but it would necessary to debug errors specific to a PHP version.
- Set the ini config in `phpunit.xml.dist`, it's possible to change them by creating the `phpunit.xml`
- Composer scripts automatically resolves binaries in `vendor/bin`
- Add a default command for `docker-compose run app`, that install vendors and run tests. `docker-compose run app bash` is recommanded if you need to tests many times without restarting the container.

GromNaN:
Thank you @hans-thomas

PR #2682, opened 2023-11-19
Support renaming columns in migrations

### LINKED ISSUE #2391: Rename column in migration has no effect
- Laravel-mongodb Version: 3.8
- PHP Version: 8.1.16
- Database Driver & Version: MongoDB 4.4.14, mongodb php driver 1.13.0

### Description:

When I try to rename a field of a collection within a migration using ```$table->renameColumn()``` the migration runs without errors but the fields in the collection have the same name as before running the migration 

### Steps to reproduce

Collection users:
```
{
    _id: 62859365a1767a00c9589712,
    name: "admin",
    role: "member"
}
```
Migration file:

```
class RenameUser extends Migration
{
    public function up()
    {
		Schema::table ('users', function (Blueprint $table) {
			$table->renameColumn ('role', 'global_role');
		});
    }
}
```

```> php artisan migrate```

### Expected behaviour

I expected the same behavior as running the following monogdb command
```db.users.updateMany ({},{ $rename: { 'role': 'global_role'}} )```
The name of the field ```role``` should be ```global_role``` after running the migration.

### Actual behaviour
Nothing, no error but also no effect.

### PR DESCRIPTION
**Pull request by hans-thomas**
Fix #2391
As mentioned in the linked issue, the `renameColumn` doesn't work. I just override that to fix the method.

### DIFF
--- a/src/Schema/Blueprint.php
+++ b/src/Schema/Blueprint.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace MongoDB\Laravel\Schema;
 
 use Illuminate\Database\Connection;
+use Illuminate\Database\Schema\Blueprint as SchemaBlueprint;
 use MongoDB\Laravel\Collection;
 
 use function array_flip;
@@ -15,7 +16,7 @@ use function is_int;
 use function is_string;
 use function key;
 
-class Blueprint extends \Illuminate\Database\Schema\Blueprint
+class Blueprint extends SchemaBlueprint
 {
     /**
      * The MongoConnection object for this blueprint.
@@ -274,6 +275,14 @@ class Blueprint extends \Illuminate\Database\Schema\Blueprint
     public function drop()
     {
         $this->collection->drop();
+    }
+
+    /** @inheritdoc */
+    public function renameColumn($from, $to)
+    {
+        $this->collection->updateMany([$from => ['$exists' => true]], ['$rename' => [$from => $to]]);
+
+        return $this;
     }
 
     /** @inheritdoc */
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/Schema/Blueprint.php [hunk 1]
```diff
@@ -276,6 +277,14 @@ public function drop()
...
     }
 
+    /** @inheritdoc */
+    public function renameColumn($from, $to)
+    {
+        $this->collection->updateMany([$from => ['$exists' => true]], ['$rename' => [$from => $to]]);
+
+        return $this;
+    }
+
```
hans-thomas, line 287: First, it finds the documents that have the `$from` column and then runs a query to rename their column to the `$to`.

alcaeus (approved):

tests/SchemaTest.php
```diff
@@ -337,6 +337,46 @@ public function testSparseUnique(): void
...
+
+        $this->assertArrayHasKey('test', $check[0]);
+        $this->assertArrayNotHasKey('newtest', $check[0]);
+
+        $this->assertArrayHasKey('test', $check[1]);
+        $this->assertArrayNotHasKey('newtest', $check[1]);
+
+        $this->assertArrayHasKey('column', $check[2]);
+        $this->assertArrayNotHasKey('test', $check[2]);
+        $this->assertArrayNotHasKey('newtest', $check[2]);
```
GromNaN, line 357: The checks before the rename operation don't serve the tested feature.

```suggestion
```

GromNaN (approved):
The test should be more focused, otherwise LGTM.

GromNaN:
Thank you @hans-thomas

PR #2685, opened 2023-11-23
Update `push` and `pull` docs

### LINKED ISSUE #2387: Pushing an array to an array shows each element in the array as children, but it is inserted and fetched as expected
- Laravel-mongodb Version: 3.9.0
- PHP Version: 8.0.17
- Database Driver & Version:

### Steps to reproduce
1. Push to an array
2. Return the parent entity as json response
3. See the items pushed are shown as each pushed.
4. Do a GET for that entity, the item is shown as expected.

### Expected behaviour
It should be consistent between the after-update and the re-fetch

### Actual behaviour
The output is different from the fetch (and how it looks in the DB) from the after-update state

Example:

```
public function addProduct(ProductAddRequest $request, $purchaseRequestId): Response|JsonResponse
    {
        /** @var PurchaseRequest|null $purchaseRequest */
        $purchaseRequest = PurchaseRequest::find($purchaseRequestId);

        if (!$purchaseRequest) {
            return new Response("", HttpCodes::NOT_FOUND);
        }

        $product = new Product(
            $request->request->get('currency'),
            $request->request->get('unit_cost'),
            $request->request->get('product_id'),
            $request->request->get('quantity'),
        );

        $purchaseRequest->push('products', $product->toArray());   // <---------- here

        return new JsonResponse($purchaseRequest);
    }
```

output when adding an item:

```
{
    "id": "62862f04375eed096c6abc32",
    "products": [
            "id": "628667dccee1aa7afe760d24",
            "currency": "MXN",
            "unit_cost": 15999,
            "product_id": 1,
            "quantity": 1,
            "metadata": []
    ]
}
```

output when fetching the same item

```
{
    "id": "62862f04375eed096c6abc32",
    "products": [
        {
            "id": "628667dccee1aa7afe760d24",
            "currency": "MXN",
            "unit_cost": 15999,
            "product_id": 1,
            "quantity": 1,
            "metadata": []
        },
   ]
}
```

Note: It can be fixed by wrapping the array into a new array. Then it inserts as before, and also shows correctly in the current loaded entity.

hans-thomas:
Fixes #2387,
Hello there, I updated the `push` and `pull` methods documentation to solve the confusion about using these methods.

### DIFF
--- a/docs/query-builder.md
+++ b/docs/query-builder.md
@@ -480,7 +480,7 @@ DB::collection('users')
   ->where('name', 'John')
     // Push an embedded document as a value to the messages array
   ->push('messages', [
-      [ 'from'    => 'Jane Doe', 'message' => 'Hi John' ]
+      [ 'from' => 'Jane Doe', 'message' => 'Hi John' ]
   ]);
 // Result:
 // messages: [
@@ -490,7 +490,7 @@ DB::collection('users')
 // Or
 
 $user->push('messages', [
-    [ 'from'    => 'Jane Doe', 'message' => 'Hi John' ]
+    [ 'from' => 'Jane Doe', 'message' => 'Hi John' ]
 ]);
 ```
 
@@ -512,7 +512,7 @@ DB::collection('users')
 // Or
 
 $user->push('messages', [
-    [ 'from'    => 'Jane Doe', 'message' => 'Hi John' ]
+    [ 'from' => 'Jane Doe', 'message' => 'Hi John' ]
 ]);
 // Result:
 // messages: [
@@ -520,8 +520,8 @@ $user->push('messages', [
 //  ]
 
 $user->push('messages', [
-    [ 'from'    => 'Jess Doe', 'message' => 'Hi' ],
-    [ 'from'    => 'Jane Doe', 'message' => 'Hi John' ],
+    [ 'from' => 'Jess Doe', 'message' => 'Hi' ],
+    [ 'from' => 'Jane Doe', 'message' => 'Hi John' ],
 ], true);
 // Result:
 // messages: [
@@ -550,7 +550,7 @@ $user->pull('items', ['boots', 'jeans']);
 // items: ['hat']
 ```
 
-Embedded document and arrays values can also be pulled from the `messages` array.
+Embedded document and arrays values can also be removed from the `messages` array.
 
 ```php
 // Latest state:
@@ -561,7 +561,7 @@ Embedded document and arrays values can also be pulled from the `messages` array
 
 DB::collection('users')
   ->where('name', 'John')
-    // It will search for the exact array and pulls it
+    // Pull an embedded document from the array
   ->pull('messages', [
       [ 'from' => 'Jane Doe', 'message' => 'Hi John' ]
   ]);
@@ -570,7 +570,7 @@ DB::collection('users')
 //      { from: "Jess Doe", message: "Hi" }
 //  ]
 
-// Or pull multiple array
+// Or pull multiple embedded documents
 
 $user->pull('messages', [
     [ 'from' => 'Jane Doe', 'message' => 'Hi John' ],
--- a/docs/query-builder.md
+++ b/docs/query-builder.md
@@ -448,7 +448,7 @@ DB::collection('items')
 
 **Push**
 
-Add one or more values to the items array.
+Add one or multiple values to the items array.
 
 ```php
 // Push the value to the matched documents
@@ -492,7 +492,7 @@ $user->push('messages', [
 ]);
 ```
 
-If you **DON'T** want duplicate items, set the third parameter to `true`:
+If you **DON'T** want duplicate values, set the third parameter to `true`:
 
 ```php
 DB::collection('users')
@@ -526,7 +526,7 @@ $user->push('messages', [
 
 **Pull**
 
-Remove a value form items array.
+Remove one or multiple values form items array.
 
 ```php
 // items: ['boots', 'hat', 'jeans']
--- a/docs/query-builder.md
+++ b/docs/query-builder.md
@@ -469,7 +469,8 @@ $user->push('items', 'boots');
 $user->push('items', ['hat', 'jeans']);
 ```
 
-To add an array to the messages array.
+To add an array as a value to the messages array. value arrays must be nested into a list array, otherwise they are
+interpreted as a list of scalar values.
 
 ```php
 DB::collection('users')
--- a/docs/query-builder.md
+++ b/docs/query-builder.md
@@ -469,7 +469,7 @@ $user->push('items', 'boots');
 $user->push('items', ['hat', 'jeans']);
 ```
 
-To add an array to the items array.
+To add an array to the messages array.
 
 ```php
 DB::collection('users')
--- a/docs/query-builder.md
+++ b/docs/query-builder.md
@@ -448,7 +448,7 @@ DB::collection('items')
 
 **Push**
 
-Add one or multiple values to the items array.
+Add one or multiple values to the `items` array.
 
 ```php
 // Push the value to the matched documents
@@ -468,23 +468,19 @@ DB::collection('users')
 
 // Or
 
-// Push the value to an exact document
+// Push the values directly to a model object
 $user->push('items', 'boots');
 $user->push('items', ['hat', 'jeans']);
 ```
 
-To add an array as a value to the messages array. value arrays must be nested into a list array, otherwise they are
-interpreted as a list of scalar values.
+To add embedded document or array values to the `messages` array, those values must be specified within a list array.
 
 ```php
 DB::collection('users')
   ->where('name', 'John')
-    // Push an array as a value to the items array
+    // Push an embedded document as a value to the messages array
   ->push('messages', [
-      [
-          'from'    => 'Jane Doe',
-          'message' => 'Hi John',
-      ],
+      [ 'from'    => 'Jane Doe', 'message' => 'Hi John' ]
   ]);
 // Result:
 // messages: [
@@ -494,10 +490,7 @@ DB::collection('users')
 // Or
 
 $user->push('messages', [
-    [
-        'from'    => 'Jane Doe',
-        'message' => 'Hi John',
-    ],
+    [ 'from'    => 'Jane Doe', 'message' => 'Hi John' ]
 ]);
 ```
 
@@ -519,10 +512,7 @@ DB::collection('users')
 // Or
 
 $user->push('messages', [
-    [
-        'from'    => 'Jane Doe',
-        'message' => 'Hi John',
-    ],
+    [ 'from'    => 'Jane Doe', 'message' => 'Hi John' ]
 ]);
 // Result:
 // messages: [
@@ -530,14 +520,8 @@ $user->push('messages', [
 //  ]
 
 $user->push('messages', [
-    [
-        'from'    => 'Jess Doe',
-        'message' => 'Hi',
-    ],
-    [
-        'from'    => 'Jane Doe',
-        'message' => 'Hi John',
-    ],
+    [ 'from'    => 'Jess Doe', 'message' => 'Hi' ],
+    [ 'from'    => 'Jane Doe', 'message' => 'Hi John' ],
 ], true);
 // Result:
 // messages: [
@@ -548,7 +532,7 @@ $user->push('messages', [
 
 **Pull**
 
-Remove one or multiple values form items array.
+Remove one or multiple values from the `items` array.
 
 ```php
 // items: ['boots', 'hat', 'jeans']
@@ -566,7 +550,7 @@ $user->pull('items', ['boots', 'jeans']);
 // items: ['hat']
 ```
 
-Also, you can pull an array value from the messages array.
+Embedded document and arrays values can also be pulled from the `messages` array.
 
 ```php
 // Latest state:
@@ -579,10 +563,7 @@ DB::collection('users')
   ->where('name', 'John')
     // It will search for the exact array and pulls it
   ->pull('messages', [
-      [
-          'from' => 'Jane Doe',
-          'message' => 'Hi John',
-      ]
+      [ 'from' => 'Jane Doe', 'message' => 'Hi John' ]
   ]);
 // Result:
 // messages: [
@@ -592,14 +573,8 @@ DB::collection('users')
 // Or pull multiple array
 
 $user->pull('messages', [
-    [
-        'from' => 'Jane Doe',
-        'message' => 'Hi John',
-    ],
-    [
-        'from' => 'Jess Doe',
-        'message' => 'Hi',
-    ]
+    [ 'from' => 'Jane Doe', 'message' => 'Hi John' ],
+    [ 'from' => 'Jess Doe', 'message' => 'Hi' ]
 ]);
 // Result:
 // messages: [ ]
--- a/docs/query-builder.md
+++ b/docs/query-builder.md
@@ -464,7 +464,7 @@ DB::collection('users')
 
 // Or
 
-// Push the value to exact document
+// Push the value to an exact document
 $user->push('items', 'boots');
 $user->push('items', ['hat', 'jeans']);
 ```
--- a/docs/query-builder.md
+++ b/docs/query-builder.md
@@ -448,27 +448,47 @@ DB::collection('items')
 
 **Push**
 
-Add items to an array.
+Add one or more values to the items array.
 
 ```php
+// Push the value to the matched documents
 DB::collection('users')
-    ->where('name', 'John')
-    ->push('items', 'boots');
+  ->where('name', 'John')
+    // Push a single value to the items array
+  ->push('items', 'boots'); // items: ['boots']
 
+DB::collection('users')
+  ->where('name', 'John')
+    // Push multiple values to the items array
+  ->push('items', ['hat', 'jeans']); // items: ['boots', 'hat', 'jeans']
+
+// Or
+
+// Push the value to exact document
 $user->push('items', 'boots');
+$user->push('items', ['hat', 'jeans']);
 ```
 
+To add an array to the items array.
+
 ```php
 DB::collection('users')
-    ->where('name', 'John')
-    ->push('messages', [
-        'from' => 'Jane Doe',
-        'message' => 'Hi John',
-    ]);
+  ->where('name', 'John')
+    // Push an array as a value to the items array
+  ->push('messages', [
+      [
+          'from'    => 'Jane Doe',
+          'message' => 'Hi John',
+      ],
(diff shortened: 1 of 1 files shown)


docs/query-builder.md
```diff
@@ -448,64 +448,126 @@ DB::collection('items')
...
+  ->push('items', ['hat', 'jeans']); // items: ['boots', 'hat', 'jeans']
 
+// Or
+
+// Push the value to an exact document
 $user->push('items', 'boots');
+$user->push('items', ['hat', 'jeans']);
 ```
 
+To add an array to the messages array.
```
GromNaN: We should add a note that values arrays must be nested into a list array, otherwise they are interpreted as a list of scalar values.

docs/query-builder.md
```diff
@@ -448,64 +448,126 @@ DB::collection('items')
...
-        'message' => 'Hi John',
-    ]);
+  ->where('name', 'John')
+    // Push an array as a value to the items array
+  ->push('messages', [
+      [
+          'from'    => 'Jane Doe',
+          'message' => 'Hi John',
+      ],
+  ]); // messages: [ ['from' => 'Jane Doe', 'message' => 'Hi John'] ]
```
GromNaN: This should be presented as BSON documents: (I'm not sure about the comment style, it might be better to use a separate the code block with `ts` syntax.

```suggestion
  ]);
  // Result:
  // messages: [
  //     { from: "Jane Doe", message: "Hi John" }
  // ]
```

Same for other examples.

GromNaN (commented):
I like this improvement. It's a good step for more clarity.

docs/query-builder.md
```diff
@@ -480,7 +485,11 @@ DB::collection('users')
           'from'    => 'Jane Doe',
           'message' => 'Hi John',
       ],
-  ]); // messages: [ ['from' => 'Jane Doe', 'message' => 'Hi John'] ]
+  ]);
+// Result:
+// messages: [
+//      { 'from' => 'Jane Doe', 'message' => 'Hi John' }
```
GromNaN: This is half way to a JS format.
```suggestion
//      { from: "Jane Doe", message: "Hi John" }
```

docs/query-builder.md
```diff
@@ -448,64 +448,161 @@ DB::collection('items')
...
+  // Push multiple values to the items array
+  ->push('items', ['hat', 'jeans']);
+// Result:
+// items: ['boots', 'hat', 'jeans']
 
+// Or
+
+// Push the value to an exact document
 $user->push('items', 'boots');
+$user->push('items', ['hat', 'jeans']);
```
jmikola: > Push the value to an exact document

The "exact document" language isn't very clear to me. I assume the difference here is that it's referring to calling `push()` on a model object instead of executing an update through the collection.

Would "Push the values directly to a model object" read better?

docs/query-builder.md
```diff
@@ -448,64 +448,161 @@ DB::collection('items')
...
-    'message' => 'Hi John',
+    [
+        'from'    => 'Jane Doe',
+        'message' => 'Hi John',
+    ],
 ]);
 ```
 
-If you **DON'T** want duplicate items, set the third parameter to `true`:
+If you **DON'T** want duplicate values, set the third parameter to `true`:
```
jmikola, line 497: I haven't checked the code, but this suggests that the boolean causes [`$addToSet`](https://www.mongodb.com/docs/manual/reference/operator/update/addToSet/) to be used internally instead of [`$push`](https://www.mongodb.com/docs/manual/reference/operator/update/push/). Is that correct?

Having a boolean argument control that rubs me the wrong way (I'll add it to the growing list), but I assume there's an explanation for this behavior. Is the `push()` method here actually derived from something else in Eloquent, or was it originally created in this library specifically for the MQL operators?

docs/query-builder.md
```diff
@@ -448,64 +448,161 @@ DB::collection('items')
...
+
 ```php
 DB::collection('users')
-    ->where('name', 'John')
-    ->push('messages', [
-        'from' => 'Jane Doe',
-        'message' => 'Hi John',
-    ]);
+  ->where('name', 'John')
+    // Push an array as a value to the items array
```
jmikola:
```suggestion
    // Push an embedded document as a value to the items array
```

Using the correct term (from MongoDB's POV) here will help avoid any confusion.

docs/query-builder.md
```diff
@@ -448,64 +448,161 @@ DB::collection('items')
 
 **Push**
 
-Add items to an array.
+Add one or multiple values to the items array.
```
jmikola:
```suggestion
Add one or multiple values to the `items` array.
```

Quoting field names should make this more readable.

docs/query-builder.md
```diff
@@ -448,64 +448,161 @@ DB::collection('items')
...
 
+// Or
+
+// Push the value to an exact document
 $user->push('items', 'boots');
+$user->push('items', ['hat', 'jeans']);
 ```
 
+To add an array as a value to the messages array. value arrays must be nested into a list array, otherwise they are
+interpreted as a list of scalar values.
```
jmikola:
```suggestion
To add embedded document or array values to the `messages` array, those values must be specified within a list array.
```

> otherwise they are interpreted as a list of scalar values.

I think this raises more questions, as it's unclear what would happen if the second argument to `push()` was an associative array. I'd hope there'd be an exception thrown if the argument is not a real list, but I wouldn't be surprised if it was undefined behavior (maybe keys are ignored).

Offhand, does the method internally wrap its arguments in an [`$each`](https://www.mongodb.com/docs/manual/reference/operator/update/each/#mongodb-update-up.-each) operator?

docs/query-builder.md
```diff
@@ -448,64 +448,161 @@ DB::collection('items')
...
+$user->push('messages', [
+    [
+        'from'    => 'Jess Doe',
+        'message' => 'Hi',
+    ],
+    [
+        'from'    => 'Jane Doe',
+        'message' => 'Hi John',
+    ],
+], true);
```
jmikola, line 525: This example may open a can of worms about how MongoDB compares documents. For instance, if you happened to swap the order of `from` and `messages` here, the document would likely be appended and you'd end up with three elements in the `messages` array. [Comparison and Sort Order: Objects](https://www.mongodb.com/docs/manual/reference/bson-type-comparison-order/#objects) goes into more detail about that.

If it's not too useful to readers, I'd propose removing it. An earlier example already demonstrates that we can call `push()` on a model object directly, so I wouldn't push for repeating that multiple times (pun intended).

docs/query-builder.md
```diff
@@ -448,64 +448,161 @@ DB::collection('items')
...
+// messages: [
+//      { from: "Jane Doe", message: "Hi John" }
+//      { from: "Jess Doe", message: "Hi" }
+//  ]
 ```
 
 **Pull**
 
-Remove an item from an array.
+Remove one or multiple values form items array.
```
jmikola:
```suggestion
Remove one or multiple values from the `items` array.
```

docs/query-builder.md
```diff
@@ -448,64 +448,161 @@ DB::collection('items')
...
+
+// Or pull multiple values
 
-$user->pull('items', 'boots');
+$user->pull('items', ['boots', 'jeans']);
+// Result:
+// items: ['hat']
 ```
 
+Also, you can pull an array value from the messages array.
```
jmikola:
```suggestion
Embedded document and arrays values can also be pulled from the `messages` array.
```

Using the specific nomenclature here.

Avoiding "you" in documentation when possible is also a plus. The PHP docs are pretty good about this (see: [Style Guidelines: Personalization](http://doc.php.net/tutorial/style.php) for their reasoning).

docs/query-builder.md
```diff
@@ -448,64 +448,161 @@ DB::collection('items')
...
-        'from' => 'Jane Doe',
-        'message' => 'Hi John',
-    ]);
+  ->where('name', 'John')
+    // It will search for the exact array and pulls it
+  ->pull('messages', [
+      [
+          'from' => 'Jane Doe',
+          'message' => 'Hi John',
+      ]
```
jmikola: Consider writing all of these embedded documents as one-liners through this PR to make the examples more concise. This would be particularly helpful for the example where a third boolean argument is provided to `push()` since it can easily get lost at the end of all of the whitespace in the method call.

docs/query-builder.md
hans-thomas: Sorry, I'm not that good in English to write docs😅

docs/query-builder.md
hans-thomas: > I haven't checked the code, but this suggests that the boolean causes [`$addToSet`](https://www.mongodb.com/docs/manual/reference/operator/update/addToSet/) to be used internally instead of [`$push`](https://www.mongodb.com/docs/manual/reference/operator/update/push/). Is that correct?

Yes. It's the logged queries of pushing normally and pushing unique values.

```plain
array:2 [
  0 => array:3 [
    "query" => "users.updateMany({"_id":"65672f58551357588a07c993"},{"$push":{"items":{"$each":["boots"]}}},{"multiple":true})"
    "bindings" => []
    "time" => 1.82
  ]
  1 => array:3 [
    "query" => "users.updateMany({"_id":"65672f58551357588a07c993"},{"$addToSet":{"items":{"$each":["hat","jeans"]}}},{"multiple":true})"
    "bindings" => []
    "time" => 0.6
  ]
]
```
 

> Having a boolean argument control that rubs me the wrong way (I'll add it to the growing list), but I assume there's an explanation for this behavior. Is the `push()` method here actually derived from something else in Eloquent, or was it originally created in this library specifically for the MQL operators?

 The `push` method is originally from Eloquent and it's used for saving a model instance and its relationships to the database. this library overrides the `push` method and accepts three args. If no arg is passed, the original `push` method will be called.

docs/query-builder.md
hans-thomas: https://github.com/mongodb/laravel-mongodb/blob/973b6b90f7916916bf252ebd707fe4f4b5eaaeb1/src/Eloquent/Model.php#L470

Yes, It does.

docs/query-builder.md
hans-thomas: > If it's not too useful to readers, I'd propose removing it.

It's the reason that I've opened this PR. Users get confused when they want to use these methods (like this issue #2387). To reduce the complexity here, I suggest using arrays with only one key and value.

```suggestion
$user->push('messages', [
    [ 'message' => 'Hi' ],
    [ 'message' => 'Hi John' ],
], true);
```

docs/query-builder.md
hans-thomas: Thank you, Jeremy. I will keep this in mind.

docs/query-builder.md
jmikola: Is Eloquent's `push` method documented anywhere beyond [Recursively Saving Models & Relationships](https://laravel.com/docs/10.x/eloquent-relationships#the-push-method)? That doesn't appear to take _any_ arguments and the method seems entirely different in purpose (this doesn't have to do with relationships).

@GromNaN: A boolean flag to determine whether `$push` or `$addToSet` is used seems like a poor design as it makes the code less self-documenting. I think this is compounded by the fact that the method on the model class [uses `func_get_args()`](https://github.com/mongodb/laravel-mongodb/blob/973b6b90f7916916bf252ebd707fe4f4b5eaaeb1/src/Eloquent/Model.php#L459C23-L459C36) and doesn't define a named parameter that users could specify if they so desired.

It may be worth making a ticket to revisit this down the line. Perhaps introduce a new `addToSet()` method and then deprecate the original variation. This might also be a moot point when/if the query builder is integrated.

docs/query-builder.md
jmikola: Ah, just saw your comment in https://github.com/mongodb/laravel-mongodb/issues/2387#issuecomment-1819110263. In that case, I think it's fine to leave the embedded documents as-is with two fields. Putting myself in the reader's shoes for a moment, I think seeing an embedded document with a single string field would make me wonder why the author didn't just store `messages` as a list of strings.

docs/query-builder.md
```diff
@@ -448,64 +448,136 @@ DB::collection('items')
...
 DB::collection('users')
-    ->where('name', 'John')
-    ->push('messages', [
-        'from' => 'Jane Doe',
-        'message' => 'Hi John',
-    ]);
+  ->where('name', 'John')
+    // Push an embedded document as a value to the messages array
+  ->push('messages', [
+      [ 'from'    => 'Jane Doe', 'message' => 'Hi John' ]
```
jmikola:
```suggestion
      [ 'from' => 'Jane Doe', 'message' => 'Hi John' ]
```

docs/query-builder.md
```diff
@@ -448,64 +448,136 @@ DB::collection('items')
...
+// messages: [
+//      { from: "Jane Doe", message: "Hi John" }
+//  ]
+
+// Or
 
 $user->push('messages', [
-    'from' => 'Jane Doe',
-    'message' => 'Hi John',
+    [ 'from'    => 'Jane Doe', 'message' => 'Hi John' ]
```
jmikola:
```suggestion
    [ 'from' => 'Jane Doe', 'message' => 'Hi John' ]
```

Not sure where this extra indentation came from.

docs/query-builder.md
```diff
@@ -448,64 +448,136 @@ DB::collection('items')
...
+  ->where('name', 'John')
+  ->push('items', ['hat', 'boots', 'jeans'], true);
+// Result:
+// items: ['boots', 'hat', 'jeans']
+
+// Or
 
-$user->push('items', 'boots', true);
+$user->push('messages', [
+    [ 'from'    => 'Jane Doe', 'message' => 'Hi John' ]
```
jmikola:
```suggestion
    [ 'from' => 'Jane Doe', 'message' => 'Hi John' ]
```

docs/query-builder.md
```diff
@@ -448,64 +448,136 @@ DB::collection('items')
...
+]);
+// Result:
+// messages: [
+//      { from: "Jane Doe", message: "Hi John" }
+//  ]
+
+$user->push('messages', [
+    [ 'from'    => 'Jess Doe', 'message' => 'Hi' ],
+    [ 'from'    => 'Jane Doe', 'message' => 'Hi John' ],
```
jmikola:
```suggestion
    [ 'from' => 'Jess Doe', 'message' => 'Hi' ],
    [ 'from' => 'Jane Doe', 'message' => 'Hi John' ],
```

docs/query-builder.md
```diff
@@ -448,64 +448,136 @@ DB::collection('items')
...
+    // It will search for the exact array and pulls it
+  ->pull('messages', [
+      [ 'from' => 'Jane Doe', 'message' => 'Hi John' ]
+  ]);
+// Result:
+// messages: [
+//      { from: "Jess Doe", message: "Hi" }
+//  ]
+
+// Or pull multiple array
```
jmikola:
```suggestion
// Or pull multiple embedded documents
```

docs/query-builder.md
```diff
@@ -448,64 +448,136 @@ DB::collection('items')
...
+//  ]
+
 DB::collection('users')
-    ->where('name', 'John')
-    ->pull('messages', [
-        'from' => 'Jane Doe',
-        'message' => 'Hi John',
-    ]);
+  ->where('name', 'John')
+    // It will search for the exact array and pulls it
```
jmikola:
```suggestion
    // Pull an embedded document from the array
```

I don't think we need to say "exact" here, as `$pull` performs a match. Quoting [`$pull`: Behavior](https://www.mongodb.com/docs/manual/reference/operator/update/pull/#behavior): 

> If the specified <value> to remove is a document, [$pull](https://www.mongodb.com/docs/manual/reference/operator/update/pull/#mongodb-update-up.-pull) removes only the elements in the array that have the exact same fields and values. The ordering of the fields can differ.

MongoDB still has logic elsewhere where BSON field order factors into comparisons (as I mentioned in https://github.com/mongodb/laravel-mongodb/pull/2685/files#r1406281116), but it's thankfully not a concern with `$pull`. I believe [`$pullAll`](https://www.mongodb.com/docs/manual/reference/operator/update/pullAll/#behavior) is what actually does an exact match with respect to field order. I pray there isn't an extra boolean argument on the `pull()` method that determines which operator to use, so hopefully this won't come up :laughing: 

----

Example from the shell:

```
> db.foo.find()
[
  {
    _id: ObjectId("656758ae5b133c573d501330"),
    x: 1,
    y: [ { a: 1, b: 2, c: 3 } ]
  }

> db.foo.updateOne({ x: 1 }, { $pullAll: { y: [ { a: 1 }]}})
{
  acknowledged: true,
  insertedId: null,
  matchedCount: 1,
  modifiedCount: 0,
  upsertedCount: 0
}

> db.foo.updateOne({ x: 1 }, { $pull: { y: {a: 1}}})
{
  acknowledged: true,
  insertedId: null,
  matchedCount: 1,
  modifiedCount: 1,
  upsertedCount: 0
}

> db.foo.find()
[ { _id: ObjectId("656758ae5b133c573d501330"), x: 1, y: [] } ]
```

docs/query-builder.md
```diff
@@ -448,64 +448,136 @@ DB::collection('items')
...
 
-$user->pull('items', 'boots');
+// Or pull multiple values
+
+$user->pull('items', ['boots', 'jeans']);
+// Result:
+// items: ['hat']
 ```
 
+Embedded document and arrays values can also be pulled from the `messages` array.
```
jmikola:
```suggestion
Embedded document and arrays values can also be removed from the `messages` array.
```

I think "removed" is clearer to the reader here, since you use the same term in the text above the preceding example.

docs/query-builder.md
GromNaN: 💯 agree that the double behavior of this method is a bad design.

If deprecate the custom behavior of `Model::push` with parameters, we have to do deprecate the `pull` method too.

Let's improve the documentation for now.

docs/query-builder.md
```diff
@@ -448,64 +448,136 @@ DB::collection('items')
...
+//      { from: "Jess Doe", message: "Hi" }
+//  ]
+
+// Or pull multiple array
 
 $user->pull('messages', [
-    'from' => 'Jane Doe',
-    'message' => 'Hi John',
+    [ 'from' => 'Jane Doe', 'message' => 'Hi John' ],
+    [ 'from' => 'Jess Doe', 'message' => 'Hi' ]
```
jmikola, line 577: Again, the field order shouldn't matter here if the library is always using `$pull`.

hans-thomas: I see. Thanks.

jmikola (approved):
LGTM, so over to @GromNaN.

GromNaN (approved):
Thank you @hans-thomas for the care taken with the documentation and for your kind review @jmikola

PR #2276, opened 2021-06-21
add validate instaceof Model and fix current in sync method for Hybrid relation

juniohyago:

PR #2688, opened 2023-11-27
Hybrid support for BelongsToMany relationship

hans-thomas:
Hi, I noticed the lines that we removed in https://github.com/mongodb/laravel-mongodb/pull/2664#discussion_r1402010334 PR, were part of handling the hybrid relationship that was not covered by tests.
This PR is continuing #2276. Thanks to @juniohyago.

- [x] Updating eloquent model instance with latest relation data

### DIFF
--- a/tests/Models/SqlUser.php
+++ b/tests/Models/SqlUser.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace MongoDB\Laravel\Tests\Models;
 
 use Illuminate\Database\Eloquent\Model as EloquentModel;
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 use Illuminate\Database\Eloquent\Relations\HasMany;
 use Illuminate\Database\Eloquent\Relations\HasOne;
 use Illuminate\Database\Schema\Blueprint;
@@ -32,6 +33,11 @@ class SqlUser extends EloquentModel
         return $this->hasOne(Role::class);
     }
 
+    public function skills(): BelongsToMany
+    {
+        return $this->belongsToMany(Skill::class, relatedPivotKey: 'skills');
+    }
+
     public function sqlBooks(): HasMany
     {
         return $this->hasMany(SqlBook::class);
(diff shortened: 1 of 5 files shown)


hans-thomas:
The relation data of the current eloquent model instance doesn't update. It has to be fixed to prevent users from fetching the latest relationship data from DB each time or it's possible to make users confused.

tests/Models/SqlUser.php
```diff
@@ -51,5 +57,12 @@ public static function executeSchema(): void
             $table->string('name');
             $table->timestamps();
         });
+        if (! $schema->hasTable('skill_sql_user')) {
+            $schema->create('skill_sql_user', function (Blueprint $table) {
```
GromNaN, line 61: Do we really want to have a pivot table here? If the relations are already stored in the MongoDB document, it should not be duplicated in an SQL table.

hans-thomas: https://github.com/mongodb/laravel-mongodb/blob/fc1f9cc0b96d5d7d25fcb0eff7306d87c1a1efbc/src/Eloquent/HybridRelations.php#L290-L301

The sql models don't reach our `BelongsToMany` class. So, we can't control this behavior.
As I know, the sql models, store their relations data in a pivot tabel and mongo models, store theirs in a pivot column.

GromNaN (approved):

hans-thomas:
Hi @GromNaN, Could you please merge these two PRs?
I know you're super busy but I will be gone from next week for at least 2 months (it can be increased to one and a half years) and I don't have access to the Internet in that period.
Is there any chance to release the v4.1?

GromNaN:
It seems feasible today.

hans-thomas:
It would be awesome😁🔥
before releasing it, I want to open one more PR for `BelongsToMany` relation. I noticed some potential bugs when we set custom keys.

GromNaN:
Thank you @hans-thomas. I'm merging the other one and wait for your new PR.

PR #2690, opened 2023-11-29
Hybrid support for MorphToMany relationship

hans-thomas:
Hello everybody, this PR adds hybrid support for the MorphToMany relationship.
I tested it in two scenarios. First, an SQL model makes a MorphToMany relation to a MongoDB model. In the second one, A MongoDB Model makes a MorphToMany to an SQL model.


GromNaN (approved):

GromNaN:
Can you update this PR please. There is some conflicts.

src/Relations/MorphToMany.php
```diff
@@ -185,15 +213,19 @@ public function attach($id, array $attributes = [], $touch = true)
...
+                            $this->morphType => $model->getMorphClass(),
+                        ],
+                    ], true);
+                } else {
+                    $this->addIdToParentRelationData($id);
+                }
 
                 // Attach the new parent id to the related model.
-                $model->push($this->foreignPivotKey, $this->parseIds($this->parent), true);
+                $model->push($this->foreignPivotKey, (array) $this->parent->{$this->parentKey}, true);
```
GromNaN, line 228: `(array)` cast should not be used. If the value is an `ObjectId`, the result is not the one you expect.

```php
var_dump((array) new MongoDB\BSON\ObjectId());
/*
array(1) {
  'oid' =>
  string(24) "6577071f3fded0903a01f030"
}
*/
```

Here we know it's not a list of ids, so a simple array wrapping is fine.
```suggestion
                $model->push($this->foreignPivotKey, [$this->parent->{$this->parentKey}], true);
```

GromNaN (commented):
Thank you for the update.

GromNaN (approved):

GromNaN:
Thank you @hans-thomas

src/Relations/MorphToMany.php
hans-thomas: I will double-check it later. thanks for the review❤️

hans-thomas: The `ObjectId` fields are always `string`. I tested this with a primary key and a custom key and a custom key was cast to `ObjectId`.
In all cases, the `$this->parent->{$this->parentKey}` value is `string`. Your suggestion is more readable, but I don't see any bug here to open a PR.

GromNaN: Yes, that's something we'll have to fix if we manage to store references as ObjectId. But there are many other hurdles to overcome, and I think this notion of "all keys are strings" is too hard-wired into Eloquent anyway.

PR #2691, opened 2023-12-05
fix CI workflow

Treggats:
I noticed that the CI workflow stopped working because it didn't know about secrets somehow.

I've updated the workflow by only showing debug info when the action is re-run with debugging enabled

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -54,10 +54,6 @@ jobs:
                     coverage: "xdebug"
                     tools: "composer"
 
-            -   name: "Show PHP version"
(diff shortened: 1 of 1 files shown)


.github/workflows/build-ci.yml
```diff
@@ -54,12 +54,8 @@ jobs:
                     coverage: "xdebug"
                     tools: "composer"
 
-            -   name: "Show PHP version"
```
Treggats, line 57: this is already shown in the `setup-php` step

GromNaN (approved):
Thank you @Treggats

PR #2692, opened 2023-12-06
Avoid unnecessary data fetch for exists method

andersonls:
### The issue
The `exists()` method is fetching all document properties unnecessarily since we only want to know if a record exists.

### The solution
Changed behaviour so it only fetches the `_id` property.


GromNaN (approved):
Good catch.
For reference, Laravel uses a [specific SQL statement](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Query/Grammars/Grammar.php#L1010).

GromNaN:
Thank you @andersonls

PR #2695, opened 2023-12-12
Fix BelongsToMany relationship bugs when using custom keys

hans-thomas:
Hi there,
This PR fixes bugs in BlongsToMany relationship when custom keys defined.
I tested custom `parentKey` and `relatedKey` in #2667, but that was not enough. In this PR `foreignPivotKey` and `relatedPivotKey` keys have been tested and previous tests are improved.


src/Eloquent/HybridRelations.php
```diff
@@ -303,11 +303,11 @@ public function belongsToMany(
         // First, we'll need to determine the foreign key and "other key" for the
         // relationship. Once we have determined the keys we'll make the query
         // instances as well as the relationship instances we need for this.
-        $foreignPivotKey = $foreignPivotKey ?: $this->getForeignKey() . 's';
+        $foreignPivotKey = $foreignPivotKey ?: Str::plural($this->getForeignKey());
 
         $instance = new $related();
 
-        $relatedPivotKey = $relatedPivotKey ?: $instance->getForeignKey() . 's';
+        $relatedPivotKey = $relatedPivotKey ?: Str::plural($instance->getForeignKey());
```
hans-thomas: Pluralizing `foreignPivotKey` and `relatedPivotKey` keys in the same way in MorphToMany
https://github.com/mongodb/laravel-mongodb/blob/f0eed1a306de3688591706d04c0bf701a0ea3422/src/Eloquent/HybridRelations.php#L387

GromNaN: I'm afraid it's a breaking change. This (incorrect) naming convention is here since the beginning and it will break silently for applications that don't use schema validation (what I think is the majority). By changing this rule, they will start having 2 lists (`childs` and `children`) and that will be hard to fix.

Let's add a deprecation notice instead, if the pluralization is different.

Please revert in this PR and open an new one with this change.

GromNaN (changes requested):

src/Eloquent/HybridRelations.php
hans-thomas: I agree. this can be done in `5.x` version. devs need to override their `foreignPivotKey` and `relatedPivotKey` keys for just existing relationships to upgrade to `5.x` version.

GromNaN (approved):

GromNaN:
Thank you @hans-thomas for all your work on relationships.

hans-thomas:
@GromNaN You're welcome buddy. Thank you for sharing your knowledge with me❤️ I learned a lot.

Issue #2703, opened 2024-01-09
Fatal casting issues


stubbo:
- Laravel-mongodb Version: >= 4.1.0
- PHP Version: Any (I'm using 8.2.14)
- Database Driver & Version: Any (I'm using MongoDB, server: 6.0.6, PHP driver: 1.16.2)

### Description:
Models sometimes fail to set attributes because the [`MongoDB\Laravel\Eloquent\Model::setAttribute` calls `castAttribute`](https://github.com/mongodb/laravel-mongodb/pull/2653/files#diff-284e164f76aeb13f424d154f9208d91fdbb0112b0a50ad8b1dfa237176f03cf2R220-R224), which is intended to be called when going from database to PHP and not PHP to database.

The main cast which has this issue is any of the encryption ones, the model will attempt to decrypt something which isn't encrypted.

I have started work on a PR to fix this, hopefully I can get it finished today/tomorrow.

### Steps to reproduce
1. Create a model with an encrypted cast
```php
class EncryptedModel extends \MongoDB\Laravel\Eloquent\Model {
    public $fillable = ['fieldName'];
    public $casts = ['fieldName' => 'encrypted'];
}
```
2. Try to set the encrypted value, this is when it should error with: `Illuminate\Contracts\Encryption\DecryptException: The payload is invalid.`
```php
EncryptedModel::make(['fieldName' => 'willError']);
```
3. 😭

### Expected behaviour
The value should be set without error.
`$model->fieldName` should return a decrypted string.
`$model->getRawOriginal('fieldName')` should return an encrypted string.

### Actual behaviour
Anything which calls setAttribute will error which includes calls to: make, create, update
Here is a working test which shows this issue: https://github.com/stubbo/laravel-mongodb/actions/runs/7462021749

richardfila:
Hit this issue today when switching a Laravel project from `jenssegers/mongodb` to `mongodb/laravel-mongodb`.
Showstopping error; had to switch back.

PR #2705, opened 2024-01-10
Fix casting issues (issue: #2703)

stubbo:
### Summary
#2653 Added a cast inside of the setAttribute, intended to be used when accessing raw data from a model, not adding data.

This will cause some casts to break fully. If you have the encrypt cast it will try to decrypt a string before encrypting it which causes an exception.

I have explained this in more detail in issue #2703

### Checklist


stubbo:
I will update the changelog and see if there's any documentation to update later today

tests/Casts/StringTest.php
```diff
@@ -27,5 +29,11 @@ public function testString(): void
 
         self::assertIsString($model->stringContent);
         self::assertEquals("Losing hope, don't mean I'm hopeless And maybe all I need is time", $model->stringContent);
+
+        $now = now();
+        $model->update(['stringContent' => now()]);
```
GromNaN: You must use the variable to prevent time-related issue.
```suggestion
        $model->update(['stringContent' => $now]);
```

src/Eloquent/Model.php
```diff
@@ -250,9 +237,16 @@ public function setAttribute($key, $value)
...
+        $casts = $this->getCasts();
+        if (array_key_exists($key, $casts)) {
+            $castType = $this->getCastType($key);
+            $castOptions = Str::after($casts[$key], ':');
+
+            // Can add more native mongo type casts here.
+            $value = match (true) {
+                $castType === 'decimal' => $this->fromDecimal($value, $castOptions),
+                default                 => $value,
+            };
```
GromNaN: Thanks for the comment, what other [native types](https://www.mongodb.com/docs/manual/reference/bson-types/) do you think of?

If the goal is to have a map of `$castType` to native converter, the `match` may be rewrote like this:
```suggestion
            // Can add more native mongo type casts here.
            $value = match ($castType) {
                'decimal' => $this->fromDecimal($value, $castOptions),
                default   => $value,
            };
```

src/Eloquent/Model.php
```diff
@@ -284,23 +278,25 @@ public function setAttribute($key, $value)
     /** @inheritdoc */
     protected function asDecimal($value, $decimals)
     {
-        try {
-            $value = (string) BigDecimal::of((string) $value)->toScale((int) $decimals, RoundingMode::HALF_UP);
-
-            return new Decimal128($value);
```
GromNaN, line 290: :+1: Good catch, the method `asDecimal` must return a string, as mandated in the phpdoc of the parent method.

src/Eloquent/Model.php
```diff
@@ -284,23 +278,25 @@ public function setAttribute($key, $value)
...
-        } catch (BrickMathException $e) {
-            throw new MathException('Unable to cast value to a decimal.', previous: $e);
+        if ($value instanceof Decimal128) {
```
GromNaN: This may convert `MongoDB\BSON\Int64` also.

GromNaN (commented):
Hello @stubbo, thanks for investigating the issue and submitting a patch.

@hans-thomas you may be interested in reviewing this PR as you worked on this part previously.

src/Eloquent/Model.php
stubbo: I will implement and add a test for this after work :)

niyazialpay (approved):
I tested this changes with my project and it seems solved the problem

tests/QueueTest.php
```diff
@@ -149,7 +149,7 @@ public function testQueueRelease(): void
     {
         Carbon::setTestNow();
         $queue = 'test';
-        $delay = 123;
+        $delay = now()->addSeconds(123);
```
stubbo: This caused the tests to fail so sneaking it in with these other changes
https://github.com/stubbo/laravel-mongodb/actions/runs/7492205173/job/20395216401

GromNaN: The issue may be that `Carbon::setTestNow()` must receive a date as argument. As it is line 151, the clock is reset to real time.
https://github.com/briannesbitt/Carbon/blob/add90c4558af9bda13fcdaec10ce137c3a3e300b/readme.md?plain=1#L30C1-L39

stubbo: Thanks for mentioning that, I've not done much with tests so nice to know that exists :)

GromNaN (approved):

stubbo:
@GromNaN Do you want me to update the change log?
If so, what would you like to use for the version?

GromNaN:
Changelog updated in https://github.com/mongodb/laravel-mongodb/commit/749a2d082526706a60240d2fd712d97bb7a84b1e

Thank you @stubbo

hans-thomas:
Hi @GromNaN, I will be unavailable for next month too. sorry.

PR #2709, opened 2024-01-15
Reset `Model::$unset` when a model is saved or refreshed

### LINKED ISSUE #2708: Saving/refreshing model doesn't reset `Model::$unset`
- Laravel-mongodb Version: 4.0.2
- PHP Version: 8.1.19
- Database Driver & Version: mongo 5.0

### Steps to reproduce
Two reproductions:

1. Create/load model `$model`
2. Call `$model->unset('field')`
3. Call `$model->refresh()`
4. `$model->field` will always be null

And secondly:

1. Create/load model `$model`
2. Call `$model->unset('field')`
3. (optional) call `$model->save()`
4. Accessing `$model->field` will never call `$model->getFieldAttribute()` and will always return null.

(In the latter case, our `$model->getFieldAttribute()` would return an empty array if the value doesn't exist)

### Expected behaviour
The `Model::$unset` array should be cleared on a save/refresh, and it shouldn't prevent getters from being called.

### Actual behaviour
`Model::$unset` persists until the value is set manually, preventing any `getXAttribute()` getters from being called and always returning null.

richardfila:
Fix #2708

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
-## [4.0.3] - unreleased
+## [4.0.3] - 2024-01-17
 
+- Reset `Model::$unset` when a model is saved or refreshed [#2709](https://github.com/mongodb/laravel-mongodb/pull/2709) by [@richardfila](https://github.com/richardfila)
 
 ## [4.0.2] - 2023-11-03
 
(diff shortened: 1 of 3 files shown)


GromNaN:
Hello @richardfila, thanks for working on this fix. Could you add a test case in `ModelTest` to prevent regression on this feature?

GromNaN (approved):

GromNaN:
Thank you @richardfila

richardfila:
> Hello @richardfila, thanks for working on this fix. Could you add a test case in `ModelTest` to prevent regression on this feature?

Thanks for finishing the testing Jérôme.

PR #2714, opened 2024-01-26
DOCSP-35932: compatibility page

ccho-mongodb:
Compatibility for supported versions of Laravel and Laravel MongoDB.

JIRA - <https://jira.mongodb.org/browse/DOCSP-35932>
Staging:
- [Compatibility](https://preview-mongodbcchomongodb.gatsbyjs.io/laravel/DOCSP-35932-compatibility/compatibility/)
- [Compatibility summary](https://preview-mongodbcchomongodb.gatsbyjs.io/laravel/DOCSP-35932-compatibility/#compatibility)

### Checklist


GromNaN (approved):
SGTM

docs/includes/framework-compatibility-laravel.rst
```diff
@@ -0,0 +1,16 @@
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+
+   * - {+odm-short+} Version
+     - Laravel 10.x
+     - Laravel 9.x
```
mongoKart, line 7: Q: Is "9.x" here just for completeness? Or do we expect to use it in the future?

mongoKart (approved):
lgtm + 1 question

docs/includes/framework-compatibility-laravel.rst
GromNaN: Not expected. I think it's for completeness because we didn't import the [original table](https://github.com/mongodb/laravel-mongodb/blob/3.9/README.md#installation).

ccho-mongodb: I included it for completeness since it's currently a [supported version of Laravel](https://laravelversions.com/en), so users might wonder whether it is compatible.

GromNaN: You may add the first line of the table then. 


9.x | 3.9.x | ❌

GromNaN: Then this doesn't tell which version is supported.

ccho-mongodb: I interpreted this as 3.9.x is incompatible with 10.x, but is compatible with 9.x and added that row to the [new table](https://preview-mongodbcchomongodb.gatsbyjs.io/laravel/DOCSP-35932-compatibility/compatibility/). Is that incorrect? Would you like me to keep/remove/change anything before merging?

GromNaN: Sorry, I mean 3.9 is no longer maintained.

ccho-mongodb: Ah ok, so if I understand correctly, we should also inform the reader which versions are maintained. Do you think it would be better to omit this line since there are several unmaintained versions of Laravel MongoDB prior to 3.9? Or if you want to keep it, I can add a line of text or footnote that mentions that 3.9 isn't maintained anymore. Let me know what works best for you, thanks.

GromNaN: Lets remove this last line then. We already have a link to the old versions table.

PR #2722, opened 2024-02-05
DOCSP-35957: Retrieve guide

norareidy:
Adds a find operations page to the docs.

JIRA: https://jira.mongodb.org/browse/DOCSP-35957
Staging: https://preview-mongodbnorareidy.gatsbyjs.io/laravel/DOCSP-35957/retrieve/

### Checklist


docs/retrieve.txt
```diff
@@ -0,0 +1,248 @@
...
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+In this guide, you can learn how to retrieve data from your MongoDB
+collections using **find operations**. Find operations are commands that
```
rustagir:
```suggestion
collections by using **find operations**. Find operations are commands that
```

docs/retrieve.txt
```diff
@@ -0,0 +1,248 @@
...
+Find All Documents
+------------------
+
+To retrieve all documents in a collection, use the ``all()`` method.
+
+To retrieve all documents that match a set of criteria, use the ``where()`` method.
+Pass the field name and value for which you are querying as parameters to ``where()``.
+
+You can also pass a comparison operator as the ``where()`` method's second parameter,
+such as the ``>`` operator. If you do not include a comparison operator parameter,
```
rustagir: S: explain what `>` does
```suggestion
such as the ``>`` operator to perform a "greater than" comparison. If you do not include a comparison operator parameter,
```

docs/retrieve.txt
```diff
@@ -0,0 +1,248 @@
...
+
+The following example retrieves all documents in the ``movies`` collection:
+
+.. io-code-block::
+   :copyable: true
+
+   .. input::
+      :language: php
+
+      $movies = Movie::all();
```
rustagir: Q: does this code example also print the retrieved documents?

docs/retrieve.txt
```diff
@@ -0,0 +1,248 @@
...
+.. io-code-block::
+   :copyable: true
+
+   .. input::
+      :language: php
+
+      $movies =
+         Movie::where('year', 2010)
+            ->where('imdb.rating', '>', 8.5)
+            ->get();
```
rustagir: Q: same question here - does this code actually print the results?

docs/retrieve.txt
```diff
@@ -0,0 +1,248 @@
...
+followed by the ``first()`` method. 
+
+For more information about the ``where()`` method and its parameters, see the
+:ref:`laravel-find-all` section of this guide.
+
+Find One Example
+~~~~~~~~~~~~~~~~
+
+The following example returns one document in which the value of the ``title`` field
+is ``"The Parent Trap"``:
```
rustagir: S: could you use an example here for a query that would definitely match more than one document to show that the method matches only the first one?

rustagir (changes requested):
a few small things + questions

docs/retrieve.txt
```diff
@@ -0,0 +1,248 @@
...
+dataset into your database, see the Quick Start or the :atlas:`Get Started with Atlas Guide
+</getting-started/#atlas-getting-started>`.
+
+Before running the examples in this guide, you must create the following classes:
+
+- Model class to represent the ``movies`` collection
+- Controller class to store and run the find operations
+- View class to display the results of the find operations
+
+For instructions on creating these classes, see the Quick Start.
```
ccho-mongodb: Suggestion:
I think it would be good to add explanations or links to instructions on how to edit the files and view the results. I think the instructions could eventually be moved to a "Fundamentals" landing page. Happy to help with the content for this.

docs/retrieve.txt
```diff
@@ -0,0 +1,248 @@
...
+.. _laravel-find-one:
+
+Find One Document
+-----------------
+
+To retrieve one document that matches a set of criteria, use the ``where()`` method
+followed by the ``first()`` method. 
+
+For more information about the ``where()`` method and its parameters, see the
+:ref:`laravel-find-all` section of this guide.
```
ccho-mongodb: Suggestion:
I think it would be helpful to mention sorting, even if showing how to sort is not shown on this page so that the reader can understand, from a set of results, which document to expect as the "first".

docs/retrieve.txt
```diff
@@ -0,0 +1,248 @@
...
+- View class to display the results of the find operations
+
+For instructions on creating these classes, see the Quick Start.
+
+.. _laravel-find-all:
+
+Find All Documents
+------------------
+
+To retrieve all documents in a collection, use the ``all()`` method.
```
ccho-mongodb: Issue:

I think "use the ___ method" might not be clear as to what you're calling it on.

Suggestion:
I think it could be helpful to make the following changes:
- Describe connection between "MongoDB collection" and "Laravel Model" in the Overview section.
- Provide specifics on what the user should call the "all()" method on.

docs/retrieve.txt
norareidy: The code snippet itself doesn't, but editing the view class according to the Quick Start instructions outputs the results in the format shown

ccho-mongodb: Issue:
This relates to both "Find All Documents in a Collection Example" and "Find All Matching Documents Example".

I ran an issue retrieving large result sets in the Quick Start because the movies collection is large and the web app will time out trying to retrieve them. If it was able to retrieve them, displaying them on a webpage would probably cause memory issues for the browser.

Suggestion:
I think omitting the example could prevent users from running into the same issue. Instead of showing the example, describe the syntax of the method and alias and what it would do.

Instead of showing output, I think it would be better just to show the syntax of how to provide an empty filter (``Movie::get()``), which would help them understand how to chain other modifiers, and its alias (``Movie::all()``). Rather than showing the result sets, it would be helpful to add an admonition mentioning that sample sets such as the movies collection would be too large.

docs/retrieve.txt
norareidy: I used "The Parent Trap" because there are two versions (the original and the remake) and two matching documents - would it be better to have a query that's even more broad / matches more documents?

ccho-mongodb: I think a query filter such as "year == 1961" or "runtime == 30" could provide a stronger hint that there's more than one result (without risk of causing an overload of results since the limit function hasn't been introduced).

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+In this guide, you can learn how to retrieve data from your MongoDB
+collections by using **find operations**. Find operations are commands that
+retrieve documents from the server.
```
rustagir: S: these sentences kind of go in circles. The first sentence says that you you can retrieve data by using find operations -> find operations retrieve data. I noticed this is the wording used on our other pages too.. I think you can use the phrase "find operations" first.

What about something like:
```suggestion
In this guide, you can learn how to use the Laravel package to perform find operations on your
MongoDB collections. Find operations allow you to retrieve documents based on criteria that you specify.
```

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+
+This guide includes the following sections:
+
+- :ref:`laravel-retrieve-setup`
+- :ref:`laravel-retrieve-matching`
+- :ref:`laravel-modify-find`
+- :ref:`laravel-addtl-info`
```
rustagir: S: add some description to these sections like [here](https://www.mongodb.com/docs/drivers/rust/current/fundamentals/crud/read-operations/retrieve/)

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+
+.. _laravel-retrieve-setup:
+
+Prerequisites
+-------------
+
+Complete the :ref:`Quick Start <laravel-quick-start>` tutorial before running the examples in this
+guide.
```
rustagir: Q: just checking that you know this link doesnt work right now

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+
+After completing the Quick Start, your MongoDB deployment and ``my-app`` Laravel application
+should satisfy the following requirements:
```
rustagir: S: make this imperative
```suggestion
Ensure that your MongoDB deployment and ``my-app`` Laravel application 
satisfy the following requirements after you complete the Quick Start:
```

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+
+- Atlas MongoDB deployment stores the Atlas sample data
```
rustagir:
```suggestion
- Your MongoDB deployment contains the Atlas sample data
```

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+- ``Movie.php`` file contains a ``Movie`` model to represent the ``movies`` collection
+- ``MovieController.php`` file contains a ``show()`` function to run database operations
+- ``browse_movies.blade.php`` file contains HTML code to display the results of database
+  operations
+
+You must edit some files in your ``my-app`` directory to run the code examples in this
+guide. You can find information on required file changes in each section that contains
+a code example.
```
rustagir: S: simplify
```suggestion
The following sections describe how to edit the files in your ``my-app`` directory to run 
the code examples and achieve the desired output.
```

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+---------------------------
+
+To retrieve documents that match a set of criteria, call the ``where()`` method on a model
+instance that represents your collection. Pass the field name and value for which you are
+querying as parameters to ``where()``.
+
+You can also pass a comparison operator as the ``where()`` method's second parameter,
+such as the ``>`` operator to perform a "greater than" comparison. If you do not include
+a comparison operator parameter, the find operation queries for documents whose field
+values are equal to the specified value.
```
rustagir:
```suggestion
such as the ``>`` operator to perform a "greater than" comparison. If you do not include
a comparison operator parameter, the driver(?) performs an equality query that matches documents
with field values that are equal to the filter criteria.
```

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+      IMDB Rating: 8.6
+      IMDB Votes: 41904
+      Plot: A documentary on Brazilian Formula One racing driver Ayrton Senna, who won the
+      F1 world championship three times before his death at age 34.
+
+.. important:: Required File Changes
+
+   To run the preceding code example, you must edit the ``MovieController.php`` file in your
+   ``my-app`` application. Copy the code example and paste it into the ``show()`` function, replacing
+   any existing code inside this function.
```
rustagir: S: move this admonition to before the code example

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+
+   $movies = Movie::get();
+
+Alternatively, use the ``get()`` method's alias ``all()`` to retrieve all documents.
+
+.. warning::
+
+   The ``movies`` collection in the Atlas sample dataset stores a large amount of data.
+   Retrieving and displaying all documents in this collection may cause your web 
+   application to time out.
```
rustagir:
```suggestion
.. warning::

   The ``movies`` collection in the Atlas sample dataset stores a large amount of data.
   Retrieving and displaying all documents in this collection might cause your web 
   application to time out.
```

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+database. If you apply a sort to the find operation, ``first()`` returns the first matching
+document according to the sorted order. 
+
+To learn more about sorting, see the following resources:
+
+- :manual:`Natural order </reference/glossary/#std-term-natural-order>`
+  in the Server manual glossary
+- `Ordering, Grouping, Limit and Offset
+  <https://laravel.com/docs/10.x/queries#ordering-grouping-limit-and-offset>`__ in the Laravel
+  documentation
```
rustagir: S: move into tip at the end of this section

docs/retrieve.txt
rustagir: agreed!

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+      Plot: Teenage twin girls swap places and scheme to reunite their divorced parents.
+
+Without chaining the ``first()`` method to the ``where()`` method, the preceding example returns
+two documents in which the ``title`` value is ``"The Parent Trap"``.
+
+.. important:: Required File Changes
+
+   To run the preceding code example, you must edit the ``MovieController.php`` file in your
+   ``my-app`` application. Copy the code example and paste it into the ``show()`` function, replacing
+   any existing code inside this function.
```
rustagir: S: similar to earlier comment, move up

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+
+Use Laravel Queries to Retrieve Documents
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Rather than on a model, you can run find operations on a query builder instance that
```
rustagir: S: the first clause is a bit confusing
```suggestion
You can run find operations on a query builder instance that represents your collection instead of running them on models.
```

rustagir (changes requested):
a few things!

docs/index.txt
```diff
@@ -52,6 +53,7 @@ Fundamentals
 To learn how to perform the following tasks by using the {+odm-short+},
 see the following content:
 
+- :ref:`laravel-retrieve`
```
ccho-mongodb: Suggestion:
Not a necessary change at this moment, but eventually this page will be part of a Fundamentals section (https://docs.google.com/spreadsheets/d/1KK7IJr15i_wjlyZXoV2ze2etEgPEUaj3Wj1EXP3grio/edit#gid=374293207), so it might be good to incorporate that into the anchor name, e.g.:

laravel-fundamentals-retrieve

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+   :class: singlecol
+
+Overview
+--------
+
+In this guide, you can learn how to retrieve data from your MongoDB
+collections by using **find operations**. Find operations are commands that
+retrieve documents from the server.
+
+This guide includes the following sections:
```
ccho-mongodb: Suggestion:
I think this list could omit any non-instructional section such as Prerequisites and Additional Information. E.g.

"This guide shows you how to perform the following operations:

- Retrieve documents that match a query
- Retrieve all documents in a collection
- Retrieve the first document that matches a query
- Retrieve data without using an Eloquent model
- Modify the behavior of the query
"

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+
+      Title: Senna
+      Year: 2010
+      Runtime: 106
+      IMDB Rating: 8.6
+      IMDB Votes: 41904
+      Plot: A documentary on Brazilian Formula One racing driver Ayrton Senna, who won the
+      F1 world championship three times before his death at age 34.
+
+.. important:: Required File Changes
```
ccho-mongodb: Issue:

Currently, the syntax that's specific to the custom view is mixed into the example of how to retrieve documents and therefore confuse readers. I think this needs to serve both audiences that might have gone through the documentation sequentially and those that landed on that page directly through search.

Suggestion:

While reviewing this PR, I thought of a potentially cleaner way to separate the description of what files need to be changed to run it and showing an example of the code. I think reorganizing the info into tabs entitled "Code Example" and "Run the Code in your Quick Start App" could work well.

For example, in this section, the "Code Example" tab could show the code:

```
Movie::where('year', 2010)
            ->where('imdb.rating', '>', 8.5)
            ->get()
```

And then the other tab would explain which files they need to update, what page they would need to open, and what output they should see.

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+
+.. code-block:: php
+
+   $movies = Movie::get();
+
+Alternatively, use the ``get()`` method's alias ``all()`` to retrieve all documents.
+
+.. warning::
+
+   The ``movies`` collection in the Atlas sample dataset stores a large amount of data.
```
ccho-mongodb: Suggestion:

I think "stores" could be ambiguous and interpreted as ""actively saving data" instead of "keeping data available".

```suggestion
   The ``movies`` collection in the Atlas sample dataset contains a large amount of data.
```

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+
+.. important:: Required File Changes
+
+   To run the preceding code example, you must make the following changes to your ``my-app``
+   application:
+
+   - Copy the code example and paste it into the ``show()`` function in your ``MovieController.php`` file
+   - Add the ``use Illuminate\Support\Facades\DB`` use statement to your ``MovieController.php``
+     file
+   - Replace ``DB_CONNECTION=mysql`` with ``DB_CONNECTION=mongodb`` in your ``.env`` file
```
ccho-mongodb: Question:
Is DB_CONNECTION related to the facade and/or something that we should include in the Quick Start? I didn't run into this requirement while building and testing the sample app.

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+   - Replace the contents of your ``browse_movies.blade.php`` file with the following code:
```
ccho-mongodb: Question:
Does this differ from the Quick Start browse_movies.blade.php file? It looks the same to me in which case, either it could be linked https://raw.githubusercontent.com/mongodb-university/laravel-quickstart/main/my-app/resources/views/browse_movies.blade.php

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+
+.. _laravel-retrieve-setup:
+
+Prerequisites
+-------------
+
+Complete the :ref:`Quick Start <laravel-quick-start>` tutorial before running the examples in this
+guide.
+
+After completing the Quick Start, your MongoDB deployment and ``my-app`` Laravel application
```
ccho-mongodb: Suggestion:
I think it could potentially be less confusing to omit any mention of the app name on this page since it's not providing more information than "application" or "Laravel application".

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+guide. You can find information on required file changes in each section that contains
+a code example.
+
+.. _laravel-retrieve-matching:
+
+Retrieve Matching Documents
+---------------------------
+
+To retrieve documents that match a set of criteria, call the ``where()`` method on a model
+instance that represents your collection. Pass the field name and value for which you are
```
ccho-mongodb: Question:
Are these instances? I thought that these methods were static methods on the Model.
I think you need to declare "new <Model>()" to get an instance. For example: https://www.mongodb.com/docs/drivers/php/laravel-mongodb/current/quick-start/write-data/#implement-the-function-to-insert-data

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+  operations
+
+You must edit some files in your ``my-app`` directory to run the code examples in this
+guide. You can find information on required file changes in each section that contains
+a code example.
+
+.. _laravel-retrieve-matching:
+
+Retrieve Matching Documents
+---------------------------
```
ccho-mongodb: Issue:

I think the wording is unclear and could be misinterpreted as "retrieve the documents that match each other".

Suggestion:

I think the following wording, which is more similar to the introductory sentence, would be less ambiguous:

```suggestion
Retrieve Documents that Match a Query
-------------------------------------
```

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+
+To retrieve documents that match a set of criteria, call the ``where()`` method on a model
```
ccho-mongodb: Suggestion:
I think the suggested action, "call the where() method on a model", should fully describe how to accomplish that action. The requirements for retrieving documents only become clear in the second sentence. I think introducing MongoDB terminology could be helpful.


"To retrieve documents that match a set of criteria, call the where() method and pass it a query filter. A query filter contains..."

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+instance that represents your collection. Pass the field name and value for which you are
+querying as parameters to ``where()``.
+
+You can also pass a comparison operator as the ``where()`` method's second parameter,
```
ccho-mongodb: Suggestion:
I think providing the initial comparison point would make the comparison more apparent, e.g.:

"When you pass a name and value as the query filter, MongoDB returns documents that match the exact value for the specified field name. You can pass a comparison operator such as "greater than" in the query filter to retrieve all documents that meet the criteria."

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+      F1 world championship three times before his death at age 34.
+
+.. important:: Required File Changes
+
+   To run the preceding code example, you must edit the ``MovieController.php`` file in your
+   ``my-app`` application. Copy the code example and paste it into the ``show()`` function, replacing
+   any existing code inside this function.
+
+Retrieve All Documents
+~~~~~~~~~~~~~~~~~~~~~~
```
ccho-mongodb: Issue:
I think this subheading doesn't match the heading "Retrieve Matching Documents" and therefore shouldn't be included in this section. I think the headings should be separated by intent, and then subheadings show the different ways of accomplishing that intent.

Suggestion:
I think the following top-level heading structure could prevent subheadings that don't match the headings:

- Retrieve Documents that Match a Query
This section shows how to pass a query (set of criteria). It describes/shows how you can pass exact match queries, pass conditional operator queries, and chain how to them. This could show two examples -- one by using the model static methods and the other by using the content currently in "Use Laravel Queries to Retrieve Documents".

- Retrieve All the Documents in a Collection
This section shows how to retrieve all documents in a collection, without providing any criteria.

- Modify the Retrieved Results
This section shows how to get a specified number of documents, or just one document, and how to skip results.

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
...
+   ``my-app`` application. Copy the code example and paste it into the ``show()`` function, replacing
+   any existing code inside this function.
+
+.. _laravel-addtl-info:
+
+Additional Information
+----------------------
+
+To learn more about the methods mentioned in this guide, see the 
+:ref:`laravel-query-builder` page.
```
ccho-mongodb: Suggestion:
I think this statement implies that the reader can learn about the methods listed on this page in further depth, but I think this page provides more detail. The focus of the linked page is to show a broad range of methods available.

```suggestion
To learn more about other query methods in {+odm-short+}, see the :ref:`laravel-query-builder` page.
```

docs/retrieve.txt
```diff
@@ -0,0 +1,342 @@
+.. _laravel-retrieve:
+
+==============
+Retrieve Data
+==============
+
+.. facet::
+   :name: genre
+   :values: reference
```
ccho-mongodb: Suggestion:

Since this shows how to run the examples in the app, I think this should be a tutorial rather than reference.

ccho-mongodb (changes requested):
Thanks for picking up this prototype page. After going through the staged content, I think the information could be easier to find with descriptive headings that don't overlap and separating out "how to run this" the into tabs.

I think the introductory sentences in each section could guide the readers better if they followed a consistent structure. Let me know if the suggestions I provided make sense.

Finally, while I didn't leave any other comments on this, I think any resource/link mentioned must be accompanied by information that connects it to what the reader should learn, especially when the content of the linked page doesn't provide that specific guidance.

Let me know if you want to pair on changes!

docs/retrieve.txt
norareidy: yeah, just merged the changes from Chris' PR though so should work now!

docs/retrieve.txt
norareidy: different wording, but fixed

docs/retrieve.txt
norareidy: I like the tabs idea!

docs/retrieve.txt
norareidy: Reworded this section

docs/retrieve.txt
norareidy: changed to just "call the method on an Eloquent model," which I think is correct

docs/retrieve.txt
norareidy: There's a small difference: the DB:collection query returns an array rather than an object (I believe), so you can't use the `->` object operator and have to use brackets instead

ccho-mongodb: Ah, thanks for answering, that is good to know. Since this is a difference between using Eloquent models and the facade, could that be included in the explanation of the latter? I think including it could potentially help readers that might not expect that difference.

docs/retrieve.txt
norareidy: So this came up during my experimentation with the code; the `DB::collection` queries don't work unless you update the DB_CONNECTION variable. I do think it would be good to instruct users to update DB_CONNECTION in the Quick Start. The Eloquent model queries still work if you leave it as is, but I don't see a reason not to change the variable from mysql to mongodb.

docs/retrieve.txt
```diff
@@ -0,0 +1,360 @@
...
+-------------------------------------
+
+You can retrieve documents that match a set of criteria by passing a query filter to the ``where()``
+method. A query filter specifies field value requirements and instructs the find operation
+to only return documents that meet these requirements. To run the query, call the ``where()``
+method on an Eloquent model or query builder that represents your collection.
+
+You can pass the following types of query filters to the ``where()`` method:
+
+- ``where('<field name>', <value>)``: instructs MongoDB to return documents that match
```
GromNaN: I find it more explicit if the context of the function is provided. Otherwise they can think the function `where` is global.

```suggestion
- ``Model::where('<field name>', <value>)``: instructs MongoDB to return documents that match
```

docs/retrieve.txt
```diff
@@ -0,0 +1,360 @@
...
+  the exact value for the specified field name
+
+- ``where('<field name>', '<comparison operator>', <value>)``: instructs MongoDB to
+  return documents in which the field values meet the comparison criteria
+
+To apply multiple sets of criteria to the find operation, you can chain a series
+of ``where()`` methods together.
```
GromNaN, line 71: This is the best place to add a link to the query builder page.

docs/retrieve.txt
```diff
@@ -0,0 +1,360 @@
...
+   function, replacing any existing code inside this function.
+
+.. io-code-block::
+   :copyable: true
+
+   .. input::
+      :language: php
+
+      return view('browse_movies', [
+         'movies' => Movie::where('year', 2010)
```
GromNaN: I prefer to separate the query from the view call. Separation of concerns.

```php
$movies = Movie::where('year', 2010)
    ->where('imdb.rating', '>', 8.5)
    ->get();

return view('browse_movies', [
    'movies' => $movies,
]);
```

docs/retrieve.txt
```diff
@@ -0,0 +1,360 @@
...
+
+Use Laravel Queries to Retrieve Documents
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can use Laravel's database query builder to run find operations instead of using Eloquent
+models. To run the database query, import the ``DB`` facade into your controller file and use
+Laravel's query builder syntax.
+
+The following example uses Laravel's query builder to retrieve documents in which the value
+of the ``imdb.votes`` nested field is greater than ``1300000``.
```
GromNaN: The query must be directly after this introducing sentence. And move the "important" block after.

And explain why this changes are necessary: because the data is returned as an array instead of instances of the Model class.

docs/retrieve.txt
```diff
@@ -0,0 +1,360 @@
...
+         </html>
+
+.. io-code-block::
+   :copyable: true
+
+   .. input::
+      :language: php
+
+      return view('browse_movies', [
+         'movies' => DB::collection('movies')
```
GromNaN: Alternative to changing the default connection.
```suggestion
         'movies' => DB::connection('mongodb')->collection('movies')
```

docs/retrieve.txt
```diff
@@ -0,0 +1,360 @@
...
+
+.. code-block:: php
+
+   $movies = Movie::get();
+
+.. warning::
+
+   The ``movies`` collection in the Atlas sample dataset contains a large amount of data.
+   Retrieving and displaying all documents in this collection might cause your web 
+   application to time out.
```
GromNaN: Good warning. You can add that they should always add a limit.

docs/retrieve.txt
```diff
@@ -0,0 +1,360 @@
...
+
+   .. input::
+      :language: php
+
+      return view('browse_movies', [
+         'movies' => Movie::where('year', 1999)
+            ->skip(2)
+            ->take(3)
+            ->get()
+      ]);
```
GromNaN: Same separation of concerns. I'm not sure the `return view` makes much sense here.
```suggestion
      $movies = Movie::where('year', 1999)
          ->skip(2)
          ->take(3)
          ->get()

      return view('browse_movies', [
         'movies' => $movies,
      ]);
```

docs/retrieve.txt
```diff
@@ -0,0 +1,360 @@
...
+   function, replacing any existing code inside this function.
+
+.. io-code-block::
+   :copyable: true
+
+   .. input::
+      :language: php
+      
+      return view('browse_movies', [
+         'movies' => Movie::where('runtime', 30)
```
GromNaN: You should set the sorting order. Otherwise the order is unpredictable. Except if we know there is only one result (unique constraint).

```php
$movies = Movie::where('runtime', 30)->orderBy('_id');
```

docs/retrieve.txt
GromNaN: Here I summarize what I noted in my review: Focus on queries examples.

Splitting the `return view(...)` calls and the query allows to skip the `return view(...)` is most examples. And move the blade template change after the query change.

docs/retrieve.txt
ccho-mongodb: Ticketed this update to the Quick Start [here](https://jira.mongodb.org/browse/DOCSP-36792).

docs/retrieve.txt
norareidy: @GromNaN I was thinking of combining your feedback and Chris' feedback on this; what are your thoughts on having two tabs, one with just the query syntax and one with the query, `return view` call (still separated) and information on how to actually run the example? I reworked this code example to show what I mean.

GromNaN: Eventually, the 2nd tab provides more context for those who need it.
In this case, put the whole class.

```php
class MovieController
{
    public function show()
    {
        $movies = Movie::where('runtime', '<', 60)
            ->where('imdb.rating', '>', 8.5)
            ->orderBy('imdb.rating', 'desc')
            ->take(10)
            ->get();
        
        return view('browse_movies', [
            'movies' => $movies
        ]);
    }
}
```

docs/retrieve.txt
```diff
@@ -0,0 +1,456 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+In this guide, you can learn how to use the Laravel package to perform **find operations**
```
ccho-mongodb: Suggestion:

I think "Laravel MongoDB" (the odm-short source constant) could be more consistent with the product name:

```suggestion
In this guide, you can learn how to use the {+odm-short+} to perform **find operations**
```

docs/retrieve.txt
```diff
@@ -0,0 +1,456 @@
...
+guide.
+
+After completing the Quick Start, ensure that your MongoDB deployment and Laravel application
+satisfy the following requirements:
+
+- Your Atlas MongoDB deployment contains the Atlas sample data
+- ``Movie.php`` file contains a ``Movie`` model to represent the ``movies`` collection
+- ``MovieController.php`` file contains a ``show()`` function to run database operations
+- ``browse_movies.blade.php`` file contains HTML code to display the results of database
+  operations
```
ccho-mongodb: Suggestion:

I think the "Prerequisites" heading should describe a condition/state rather than provide instructions. One potential solution could be to replace the heading title with "Before You Get Started" and explain the following similar to this:

"To run the code in this guide, you need to complete the Quick Start tutorial, which includes setting up a MongoDB Atlas instance with sample data and the following files in your Laravel web application:
- Movie.php, which contains the Movie model that ...
"

docs/retrieve.txt
ccho-mongodb: https://jira.mongodb.org/browse/DOCSP-36792

docs/retrieve.txt
```diff
@@ -0,0 +1,456 @@
...
+of the ``imdb.votes`` nested field is greater than ``1300000``.
+
+.. tabs::
+
+   .. tab:: Query Syntax
+      :tabid: query-syntax
+
+      Use the following syntax to specify the query:
+
+      .. code-block:: go
```
ccho-mongodb: Issue:

I think "go" syntax highlighting might be compatible in this case, but it would be better to identify this as php code.

Suggestion:

I think using "php" would be more consistent.

Applies to all instances.

docs/retrieve.txt
```diff
@@ -0,0 +1,456 @@
...
+
+Retrieve Documents that Match a Query
+-------------------------------------
+
+You can retrieve documents that match a set of criteria by passing a query filter to the ``where()``
+method. A query filter specifies field value requirements and instructs the find operation
+to only return documents that meet these requirements. To run the query, call the ``where()``
+method on an Eloquent model or query builder that represents your collection.
+
+You can pass the following types of query filters to the ``where()`` method:
```
ccho-mongodb: Suggestion:

I think the description of the "types" in "types of query filters" is missing from the list items. I think describing the list items directly like the following would help guide the reader:

"You can use one of the following "where()" method calls to build a query:"

docs/retrieve.txt
```diff
@@ -0,0 +1,456 @@
...
+
+- ``Model::where('<field name>', <value>)``: instructs MongoDB to return documents that match
+  the exact value for the specified field name
+
+- ``Model::where('<field name>', '<comparison operator>', <value>)``: instructs MongoDB to
+  return documents in which the field values meet the comparison criteria
```
ccho-mongodb: Suggestion:

I think the placement of the syntax presented here ("Model::where(...") in the heading can be confusing since it does not match the syntax used in the "Use Laravel Queries to Retrieve Documents" section ("DB::connection...").

I think omitting "Model::" and describing the list items as query builder methods instead of queries could make it easier for a user to to connect the information, e.g.

"where(...) builds a query that matches documents that exactly match the value for the specified field name."

docs/retrieve.txt
```diff
@@ -0,0 +1,456 @@
...
+
+Use Laravel Queries to Retrieve Documents
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can use Laravel's database query builder to run find operations instead of using Eloquent
+models. To run the database query, import the ``DB`` facade into your controller file and use
+Laravel's query builder syntax.
+
+This example uses Laravel's query builder to retrieve documents in which the value
+of the ``imdb.votes`` nested field is greater than ``1300000``.
```
ccho-mongodb: Suggestion:

No change necessary, but I think "1300000" can be difficult to read because of how long the number is. If you happen to have a different criteria/value to filter by that uses less lengthy values, I think that might be better.

docs/retrieve.txt
```diff
@@ -0,0 +1,456 @@
...
+
+         $movies = Movie::where('year', 2010)
+            ->where('imdb.rating', '>', 8.5)
+            ->get();
+
+   .. tab:: Run the Query
+      :tabid: query-run
+
+      To run the find operation and view the results, edit the ``show()`` function
+      in the ``MovieController.php`` file to resemble the following code:
```
ccho-mongodb: Suggestion:

I think it could be confusing to describe the controller as something the user can run. Instead


"To see the query results in the ``browse_movies`` view, edit the ..."


Applies to all instances.

docs/retrieve.txt
```diff
@@ -0,0 +1,456 @@
...
+           the :ref:`Laravel Queries <laravel-retrieve-query-builder>` example. Since the example uses the query builder
+           syntax, the find operation returns data as an array rather than as instances of the Eloquent model class.
+
+.. _laravel-retrieve-all:
+
+Retrieve All Documents in a Collection
+--------------------------------------
+
+You can retrieve documents without specifying a query filter, which returns all
+documents in a collection.
```
ccho-mongodb: Suggestion:

I think it can be easier to understand the requirement by introducing the task first and then how to solve for it. E.g.

"You can retrieve all documents (task) in a collection by omitting the query filter (solution)."

docs/retrieve.txt
```diff
@@ -0,0 +1,456 @@
...
+      :tabid: query-syntax
+
+      Use the following syntax to specify the query:
+
+      .. code-block:: go
+
+         $movies = DB::connection('mongodb')
+               ->collection('movies')
+               ->where('imdb.votes', '>', 1300000)
+               ->get();
```
ccho-mongodb: Suggestion:

I think the get() method should be explained prior to the example.

docs/retrieve.txt
```diff
@@ -0,0 +1,456 @@
...
+.. _laravel-retrieve-all-example:
+
+Retrieve All Documents
+~~~~~~~~~~~~~~~~~~~~~~
+
+Use the following syntax to run a find operation that matches all documents:
+
+.. code-block:: php
+
+   $movies = Movie::get();
```
ccho-mongodb: Suggestion:

I think "To retrieve all documents" on line 271 repeats the task in the previous sentence and should be omitted. 

Related to this suggestion, I think the subheading is unnecessary because it duplicates the top-level heading and there's no other subheading (I expected to see both Eloquent Models and Laravel Queries similar to the previous top-level haeding).

docs/retrieve.txt
```diff
@@ -0,0 +1,456 @@
...
+      Use the following syntax to specify the query:
+
+      .. code-block:: go
+
+         $movies = Movie::where('year', 1999)
+               ->skip(2)
+               ->take(3)
+               ->get()
+
+   .. tab:: Run the Query
```
ccho-mongodb: Suggestion:
I think the tabs should match in form of speech. "Query Syntax" is a noun phrase, so I think this tab should be reworded to match, e.g. "Controller Method".

Applies to all instances.

docs/retrieve.txt
```diff
@@ -0,0 +1,456 @@
...
+            Year: 1999
+            Runtime: 95
+            IMDB Rating: 4
+            IMDB Votes: 9296
+            Plot: A sci-fi update of the famous 6th Century poem. In a beseiged land, Beowulf must
+            battle against the hideous creature Grendel and his vengeance seeking mother.
+
+.. _laravel-retrieve-one:
+
+Retrieve Only One Result
```
ccho-mongodb: Suggestion:

I think "Return the First Result" could be more descriptive since the section doesn't describe how to retrieve any other type of a single result (such as the last result).

Applies to the description contained in this section.

docs/retrieve.txt
```diff
@@ -0,0 +1,456 @@
...
+Modify Behavior
+---------------
+
+You can modify the results of a find operation by chaining the following methods
+to ``where()``:
+
+- ``skip()``: sets the number of documents to skip when returning results
+- ``take()``: sets the total number of documents to return
+- ``first()``: returns the first document that matches the query filter
+
```
ccho-mongodb, line 322: Suggestion:
I think the subheadings could be introduced, e.g.
"The following sections show examples of how to use these methods."

docs/retrieve.txt
```diff
@@ -0,0 +1,456 @@
...
+.. _laravel-retrieve-one:
+
+Retrieve Only One Result
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+To retrieve only one document that matches a set of criteria, use the ``where()`` method
+followed by the ``first()`` method. By default, the ``first()`` method returns the first
+matching document according to the documents' natural order, or as they appear in the
+database. If you apply a sort to the find operation, ``first()`` returns the first matching
+document according to the sorted order.
```
ccho-mongodb: Suggestion:

I think it would be better to recommend a consistent sort order before describing natural ordering. 

E.g. "Use the sort() method when you use the first() to get consistent results when you query on a unique value. If you omit the sort() method, MongoDB returns the matching documents according to the documents' natural order."

docs/retrieve.txt
```diff
@@ -0,0 +1,456 @@
...
+            Runtime: 152
+            IMDB Rating: 9
+            IMDB Votes: 1495351
+            Plot: When the menace known as the Joker wreaks havoc and chaos on the people of Gotham, the
+            caped crusader must come to terms with one of the greatest psychological tests of his ability to
+            fight injustice.
+
+      Then, make the following changes to your Laravel Quick Start application:
+
+      - Add the ``use Illuminate\Support\Facades\DB`` use statement to your ``MovieController.php``
```
ccho-mongodb: Suggestion:

I think it could be good to match this with the other mention of this import in the section description since both versions of the example use it. The other instance currently reads:
"import the DB facade into your controller file"

docs/retrieve.txt
```diff
@@ -0,0 +1,456 @@
...
+           @empty
+              <p>No results</p>
+           @endforelse
+
+           </body>
+           </html>
+
+        .. note::
+
+           This change is necessary because you cannot use the ``->`` object operator to access document fields when running
```
ccho-mongodb: Suggestion:

I think it could be better if the admonition identified which change it refers to.
E.g.

"Since the Laravel query builder returns ... the view accesses the fields by using the ... syntax instead of the "->" object operator."

ccho-mongodb (approved):
Looks pretty good, thanks for making all the changes! Most of the suggestions are straightforward, but let me know if there's anything you'd like to discuss and/or whether you want another review.

docs/retrieve.txt
norareidy: oops didn't notice that - fixed!

docs/retrieve.txt
```diff
@@ -0,0 +1,471 @@
...
+- :ref:`Modify Find Operation Behavior <laravel-modify-find>`
+
+Before You Get Started
+----------------------
+
+To run the code examples in this guide, complete the :ref:`Quick Start <laravel-quick-start>`
+tutorial. This tutorial provides instructions on setting up a MongoDB Atlas instance with
+sample data and creating the following files in your Laravel web application:
+
+- ``Movie.php`` file, which contains a ``Movie`` model to represent the ``movies`` collection
```
rustagir:
```suggestion
- ``Movie.php`` file, which contains a ``Movie`` model to represent documents in the ``movies`` collection
```

docs/retrieve.txt
```diff
@@ -0,0 +1,471 @@
...
+- ``MovieController.php`` file, which contains a ``show()`` function to run database operations
+- ``browse_movies.blade.php`` file, which contains HTML code to display the results of database
+  operations
+
+The following sections describe how to edit the files in your Laravel application to run 
+the code examples and view the expected output.
```
rustagir:
```suggestion
the find operation code examples and view the expected output.
```

docs/retrieve.txt
```diff
@@ -0,0 +1,471 @@
...
+
+Return the First Result
+~~~~~~~~~~~~~~~~~~~~~~~
+
+To retrieve the first document that matches a set of criteria, use the ``where()`` method
+followed by the ``first()`` method.
+
+Chain the ``orderBy()`` method to ``first()`` to get consistent results when you query on a unique
+value. If you omit the ``orderBy()`` method, MongoDB returns the matching documents according to
+the documents' natural order, or as they appear in the database.
```
rustagir:
```suggestion
the documents' natural order, or as they appear in the collection.
```

docs/retrieve.txt
```diff
@@ -0,0 +1,471 @@
...
+
+This example returns one document in which the value of the ``runtime`` field
+is ``30``.
```
rustagir: S: also describe that this code orders by `_id`

rustagir (approved):
some small suggestions but LGTM! great work on this long PR!

docs/retrieve.txt
```diff
@@ -0,0 +1,473 @@
...
+field.
+
+.. tabs::
+
+   .. tab:: Query Syntax
+      :tabid: query-syntax
+
+      Use the following syntax to specify the query:
+
+      .. code-block:: go
```
GromNaN:
```suggestion
      .. code-block:: php
```

GromNaN (approved):
LGTM after a small fix

GromNaN:
Thank you @norareidy

PR #2734, opened 2024-02-21
PHPORM-140: Fix documentation for inverse embed relationships

alcaeus:
PHPORM-140

The issue is that fetching books through its model isn't possible, as it's an embedded relationship. I updated the code to fetch a book through the user to have a functional example, even if the use case isn't realistic.

### Checklist

- [ ] ~Add tests and ensure they pass~
- [ ] ~Add an entry to the CHANGELOG.md file~

### DIFF
--- a/docs/eloquent-models.txt
+++ b/docs/eloquent-models.txt
@@ -168,7 +168,7 @@ Keep in mind guarding still works, but you may experience unexpected behavior.
 Schema
 ------
 
-The database driver also has (limited) schema builder support. You can 
+The database driver also has (limited) schema builder support. You can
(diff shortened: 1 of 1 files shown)


docs/eloquent-models.txt
```diff
@@ -351,7 +351,7 @@ relation definition.
 
 .. code-block:: php
 
-   $book = Book::first();
+   $book = User::first()->books()->first();
```
GromNaN, line 354: Correct, `Book::first()` sends a query to the `books` collection. Which is not expected for embeds.

GromNaN (approved):

PR #2735, opened 2024-02-21
PHPORM-150 Run CI on Laravel 11

GromNaN:
Fix [PHPORM-150](https://jira.mongodb.org/browse/PHPORM-150)

### Checklist

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -12,6 +12,8 @@ jobs:
 
         strategy:
             matrix:
+                mode:
+                    - "stable"
                 os:
                     - "ubuntu-latest"
                 mongodb:
@@ -27,7 +29,10 @@ jobs:
                     - php: "8.1"
                       mongodb: "5.0"
                       mode: "low-deps"
-
+                    # Laravel 11
+                    - php: "8.3"
+                      mongodb: "7.0"
+                      mode: "dev"
         steps:
             -   uses: "actions/checkout@v4"
 
@@ -70,8 +75,10 @@ jobs:
                     restore-keys: "${{ matrix.os }}-composer-"
 
             -   name: "Install dependencies"
-                run: composer update --no-interaction $([[ "${{ matrix.mode }}" == low-deps ]] && echo ' --prefer-lowest --prefer-stable')
-
+                run: |
+                  composer update --no-interaction \
+                  $([[ "${{ matrix.mode }}" == low-deps ]] && echo ' --prefer-lowest --prefer-stable') \
+                  $([[ "${{ matrix.mode }}" == stable ]] && echo ' --prefer-stable')
             -   name: "Run tests"
                 run: "./vendor/bin/phpunit --coverage-clover coverage.xml"
                 env:
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -12,8 +12,6 @@ jobs:
 
         strategy:
             matrix:
-                mode:
-                    - "stable"
                 os:
                     - "ubuntu-latest"
                 mongodb:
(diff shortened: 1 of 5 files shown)


GromNaN:
Static analysis found a breaking change in [`HasAttributes::getStorableEnumValue()`](https://github.com/laravel/framework/commit/8647dcf18e6e315e97daea9ed60ef79b420ad327#diff-59d24f1a8f0dd1e51ee7dffc8778133711634e928ecef4a91b63fc3851cb1579R1188), which is hard to get around to support laravel 10 and 11 at the same time. 

```
 ------ ---------------------------------------------------------------------------------- 
  Line   Eloquent/Model.php                                                                
 ------ ---------------------------------------------------------------------------------- 
  707    Method Illuminate\Database\Eloquent\Model::getStorableEnumValue() invoked with 1  
         parameter, 2 required.                                                            
         ✏️  Eloquent/Model.php                                                            
 ------ ---------------------------------------------------------------------------------- 
```

src/Eloquent/Model.php
```diff
@@ -704,7 +708,7 @@ protected function addCastAttributesToArray(array $attributes, array $mutatedAtt
             }
 
             if ($this->isEnumCastable($key) && (! $castValue instanceof Arrayable)) {
-                $castValue = $castValue !== null ? $this->getStorableEnumValue($castValue) : null;
```
GromNaN, line 707: The signature of this method have changed in a non-backward compatible way. So I duplicated this simple method to keep compatibility with both Laravel 10 and 11.
https://github.com/laravel/framework/commit/8647dcf18e6e315e97daea9ed60ef79b420ad327#diff-59d24f1a8f0dd1e51ee7dffc8778133711634e928ecef4a91b63fc3851cb1579R1188

We must keep the function `addCastAttributesToArray` introduced for Laravel 10 compatibility, as it provides casting of nested fields. https://github.com/mongodb/laravel-mongodb/commit/2ea1a7c0aee085c382349613d399be79a0bd1f1a#diff-284e164f76aeb13f424d154f9208d91fdbb0112b0a50ad8b1dfa237176f03cf2R525-R530

src/Query/Builder.php
```diff
@@ -614,11 +614,7 @@ public function orderBy($column, $direction = 'asc')
         return $this;
     }
 
-    /**
-     * @param list{mixed, mixed}|CarbonPeriod $values
```
GromNaN, line 618: This signature was causing psalm error. Param type incompatible with the parent method.

.github/workflows/build-ci.yml
```diff
@@ -27,7 +27,10 @@ jobs:
                     - php: "8.1"
                       mongodb: "5.0"
                       mode: "low-deps"
-
+                    # Laravel 11
+                    - php: "8.3"
+                      mongodb: "7.0"
+                      mode: "dev"
```
GromNaN, line 33: Once Laravel 11 will be released, this will no longer be necessary. But only `low-deps` job will run Laravel 10.

.github/workflows/build-ci.yml
```diff
@@ -70,8 +73,10 @@ jobs:
                     restore-keys: "${{ matrix.os }}-composer-"
 
             -   name: "Install dependencies"
-                run: composer update --no-interaction $([[ "${{ matrix.mode }}" == low-deps ]] && echo ' --prefer-lowest --prefer-stable')
-
+                run: |
+                  composer update --no-interaction \
+                  $([[ "${{ matrix.mode }}" == low-deps ]] && echo ' --prefer-lowest --prefer-stable') \
+                  $([[ "${{ matrix.mode }}" != dev ]] && echo ' --prefer-stable')
```
alcaeus, line 79: Nit: I believe this may result in a duplicate `prefer-stable` option when the mode is `low-deps`, as the first and second conditions will match. It should be sufficient to use `--prefer-lowest` in the first check, and add `--prefer-stable` if the second condition matches.

alcaeus (approved):
Looks good, I have a minor nit but considering the build passes it's not too relevant.

.github/workflows/build-ci.yml
GromNaN: Fixed.

PR #2738, opened 2024-02-26
PHPORM-155 Fluent aggregation builder

GromNaN:
Fix PHPORM-155

With this integration, it's possible to derivate any Query builder into an aggregation pipeline. The conditions built with the query builder are used as first stages (match, project, limit ...)

Requires https://github.com/mongodb/mongo-php-builder/pull/67

### Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,6 +39,9 @@
         "spatie/laravel-query-builder": "^5.6",
         "phpstan/phpstan": "^1.10"
     },
+    "suggest": {
+        "mongodb/builder": "Provides a fluent aggregation builder for MongoDB pipelines"
+    },
     "minimum-stability": "dev",
     "replace": {
         "jenssegers/mongodb": "self.version"
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,7 @@
         "mongodb/mongodb": "^1.15"
     },
     "require-dev": {
+        "mongodb/builder": "^0.1",
         "phpunit/phpunit": "^10.3",
         "orchestra/testbench": "^8.0|^9.0",
         "mockery/mockery": "^1.4.4",
--- a/composer.json
+++ b/composer.json
@@ -14,9 +14,6 @@
         "issues": "https://www.mongodb.com/support",
         "security": "https://www.mongodb.com/security"
     },
-    "repositories": [
-        { "type": "vcs", "url": "https://github.com/GromNaN/mongo-php-builder.git" }
-    ],
     "authors": [
         { "name": "Andreas Braun", "email": "andreas.braun@mongodb.com", "role": "Leader" },
         { "name": "Jérôme Tamarelle", "email": "jerome.tamarelle@mongodb.com", "role": "Maintainer" },
@@ -34,7 +31,7 @@
         "mongodb/mongodb": "^1.15"
     },
     "require-dev": {
-        "mongodb/builder": "dev-laravel-fluent",
+        "mongodb/builder": "^0.2",
         "phpunit/phpunit": "^10.3",
         "orchestra/testbench": "^8.0|^9.0",
         "mockery/mockery": "^1.4.4",
--- a/composer.json
+++ b/composer.json
@@ -14,6 +14,9 @@
         "issues": "https://www.mongodb.com/support",
         "security": "https://www.mongodb.com/security"
     },
+    "repositories": [
+        { "type": "vcs", "url": "https://github.com/GromNaN/mongo-php-builder.git" }
+    ],
     "authors": [
         { "name": "Andreas Braun", "email": "andreas.braun@mongodb.com", "role": "Leader" },
         { "name": "Jérôme Tamarelle", "email": "jerome.tamarelle@mongodb.com", "role": "Maintainer" },
@@ -31,7 +34,7 @@
         "mongodb/mongodb": "^1.15"
     },
     "require-dev": {
-        "mongodb/builder": "^0.1",
+        "mongodb/builder": "dev-laravel-fluent",
         "phpunit/phpunit": "^10.3",
         "orchestra/testbench": "^8.0|^9.0",
         "mockery/mockery": "^1.4.4",
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -11,6 +11,7 @@ use MongoDB\Driver\Cursor;
 use MongoDB\Laravel\Collection;
 use MongoDB\Laravel\Helpers\QueriesRelationships;
 use MongoDB\Laravel\Internal\FindAndModifyCommandSubscriber;
+use MongoDB\Laravel\Query\AggregationBuilder;
 use MongoDB\Model\BSONDocument;
 use MongoDB\Operation\FindOneAndUpdate;
 
@@ -57,7 +58,7 @@ class Builder extends EloquentBuilder
     ];
 
     /**
-     * @return ($function === null ? PipelineBuilder : self)
+     * @return ($function is null ? AggregationBuilder : self)
      *
      * @inheritdoc
      */
@@ -65,7 +66,7 @@ class Builder extends EloquentBuilder
     {
         $result = $this->toBase()->aggregate($function, $columns);
 
-        return $result ?? $this;
+        return $result ?: $this;
     }
 
     /** @inheritdoc */
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -56,6 +56,11 @@ class Builder extends EloquentBuilder
         'tomql',
     ];
 
+    /**
+     * @return ($function === null ? PipelineBuilder : self)
+     *
+     * @inheritdoc
+     */
     public function aggregate($function = null, $columns = ['*'])
     {
         $result = $this->toBase()->aggregate($function, $columns);
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -56,6 +56,13 @@ class Builder extends EloquentBuilder
         'tomql',
     ];
 
+    public function aggregate($function = null, $columns = ['*'])
+    {
+        $result = $this->toBase()->aggregate($function, $columns);
+
+        return $result ?? $this;
+    }
+
     /** @inheritdoc */
     public function update(array $values, array $options = [])
     {
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -227,7 +227,7 @@ class Builder extends EloquentBuilder
                 $document = $collection->findOneAndUpdate(
                     $attributes,
                     // Before MongoDB 5.0, $setOnInsert requires a non-empty document.
-                    // This is should not be an issue as $values includes the query filter.
+                    // This should not be an issue as $values includes the query filter.
                     ['$setOnInsert' => (object) $values],
                     [
                         'upsert' => true,
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -49,6 +49,7 @@ use function strcmp;
 use function uniqid;
 use function var_export;
 
+/** @mixin Builder */
 abstract class Model extends BaseModel
 {
     use HybridRelations;
--- a/src/Query/AggregationBuilder.php
+++ b/src/Query/AggregationBuilder.php
@@ -6,14 +6,16 @@ namespace MongoDB\Laravel\Query;
 
 use Illuminate\Support\Collection as LaravelCollection;
 use MongoDB\Builder\BuilderEncoder;
-use MongoDB\Builder\Stage\FluentFactory;
+use MongoDB\Builder\Stage\FluentFactoryTrait;
 use MongoDB\Laravel\Collection;
 
 use function array_replace;
 use function collect;
 
-final class AggregationBuilder extends FluentFactory
+final class AggregationBuilder
 {
+    use FluentFactoryTrait;
+
     public function __construct(
         array $pipeline,
         private Collection $collection,
--- a/src/Query/AggregationBuilder.php
+++ b/src/Query/AggregationBuilder.php
@@ -75,7 +75,7 @@ class AggregationBuilder
     {
         return (clone $this)
             ->limit(1)
-            ->cursor($options)
+            ->get($options)
             ->first();
     }
 
--- a/src/Query/AggregationBuilder.php
+++ b/src/Query/AggregationBuilder.php
@@ -12,7 +12,7 @@ use MongoDB\Laravel\Collection;
 use function array_replace;
 use function collect;
 
-final class PipelineBuilder extends FluentFactory
+final class AggregationBuilder extends FluentFactory
 {
     public function __construct(
         array $pipeline,
--- a/src/Query/AggregationBuilder.php
+++ b/src/Query/AggregationBuilder.php
@@ -5,24 +5,28 @@ declare(strict_types=1);
 namespace MongoDB\Laravel\Query;
 
 use Illuminate\Support\Collection as LaravelCollection;
+use Illuminate\Support\LazyCollection;
+use InvalidArgumentException;
+use Iterator;
 use MongoDB\Builder\BuilderEncoder;
 use MongoDB\Builder\Stage\FluentFactoryTrait;
 use MongoDB\Collection as MongoDBCollection;
+use MongoDB\Driver\CursorInterface;
 use MongoDB\Laravel\Collection as LaravelMongoDBCollection;
 
 use function array_replace;
 use function collect;
+use function sprintf;
+use function str_starts_with;
 
-final class AggregationBuilder
+class AggregationBuilder
 {
     use FluentFactoryTrait;
 
     public function __construct(
         private MongoDBCollection|LaravelMongoDBCollection $collection,
-        array $pipeline = [],
-        private array $options = [],
+        private readonly array $options = [],
     ) {
-        $this->pipeline = $pipeline;
     }
 
     /**
@@ -31,6 +35,10 @@ final class AggregationBuilder
      */
     public function addRawStage(string $operator, mixed $value): static
     {
+        if (! str_starts_with($operator, '$')) {
+            throw new InvalidArgumentException(sprintf('The stage name "%s" is invalid. It must start with a "$" sign.', $operator));
+        }
+
         $this->pipeline[] = [$operator => $value];
 
         return $this;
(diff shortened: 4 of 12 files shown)


src/Eloquent/Builder.php
```diff
@@ -51,6 +51,13 @@ class Builder extends EloquentBuilder
         'tomql',
     ];
 
+    public function aggregate($function = null, $columns = ['*'])
+    {
+        $result = $this->toBase()->aggregate($function, $columns);
+
+        return $result ?? $this;
```
GromNaN: This method must be overrided, because by default this is a fluent interface.

GromNaN:
1. `$near`, and `$nearSphere` query operators are not supported in `$match` stage. They must be replaced by a `$geoNear`. This is something we can't do automatically, especially when nested in a `$and` or `$or` query.
2. [`$slice`](https://www.mongodb.com/docs/manual/reference/operator/projection/slice/) projection operator has a different signature than the [`$slice`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/slice/) expression (array field as 1st argument). This needs to be translated in the context of a `$project` stage.
3. `$where` is not available within aggregation pipeline. They can be replaced by the [`$function`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/function/#example-2--alternative-to--where) expression.

List of limitations: https://www.mongodb.com/docs/manual/reference/operator/aggregation/match/#restrictions

Since this operators are currently supported by the query builder, we need to keep building a `find` command with the query builder, and only create an aggregation pipeline when necessary.

composer.json
```diff
@@ -14,6 +14,9 @@
         "issues": "https://www.mongodb.com/support",
         "security": "https://www.mongodb.com/security"
     },
+    "repositories": [
+        { "type": "vcs", "url": "https://github.com/GromNaN/mongo-php-builder.git" }
```
GromNaN: To be removed once https://github.com/mongodb/mongo-php-builder/pull/65 is merged and tagged.

src/Eloquent/Model.php
```diff
@@ -49,6 +49,7 @@
 use function uniqid;
 use function var_export;
 
+/** @mixin Builder */
```
GromNaN, line 52: Allows completions such as `Model::where()->aggregate()->addField(`

src/Query/AggregationBuilder.php
```diff
@@ -0,0 +1,52 @@
...
+        $this->pipeline = $pipeline;
+    }
+
+    /**
+     * Add a stage without using the builder. Necessary if the stage is built
+     * outside the builder, or it is not yet supported by the library.
+     */
+    public function addRawStage(string $operator, mixed $value): static
+    {
+        $this->pipeline[] = [$operator => $value];
```
GromNaN, line 42: Todo: add test on this method.

src/Eloquent/Builder.php
jmikola: Does `@return` need to be `@psalm-return`?

src/Query/Builder.php
```diff
@@ -74,7 +77,7 @@ class Builder extends BaseBuilder
     /**
      * The database collection.
      *
-     * @var \MongoDB\Collection
+     * @var \MongoDB\Laravel\Collection
```
jmikola, line 79: Should this have a `use` statement instead of referring to a FQCN?

src/Query/Builder.php
```diff
@@ -83,7 +86,7 @@ class Builder extends BaseBuilder
      *
      * @var array
      */
-    public $projections;
+    public $projections = [];
```
jmikola, line 88: Should this property utilize an `array` type, or would that conflict with the base class and/or BC?

src/Query/Builder.php
```diff
@@ -538,9 +590,13 @@ public function generateCacheKey()
         return md5(serialize(array_values($key)));
     }
 
-    /** @inheritdoc */
-    public function aggregate($function, $columns = [])
+    /** @return ($function === null ? PipelineBuilder : self) */
```
jmikola: Is `@psalm-return` needed?

src/Query/Builder.php
GromNaN: This property doesn't exist in the base class. Adding the type would be a BC break (depending on our unwritten policy).

src/Query/AggregationBuilder.php
GromNaN: Done.

src/Eloquent/Builder.php
```diff
@@ -56,6 +56,18 @@ class Builder extends EloquentBuilder
...
     ];
 
+    /**
+     * @return ($function === null ? PipelineBuilder : self)
+     *
+     * @inheritdoc
+     */
+    public function aggregate($function = null, $columns = ['*'])
+    {
+        $result = $this->toBase()->aggregate($function, $columns);
```
jmikola, line 67: Also, I have no idea what actually causes `toBase()` to return `MongoDB\Laravel\Query\Builder` instead of the Eloquent object. Best I can tell, `toBase()` is not overridden so I'd expect this to invoke [`Illuminate\Database\Eloquent\Builder::toBase()`](https://github.com/illuminate/database/blob/v11.0.6/Eloquent/Builder.php#L1760).

GromNaN: Without this change, the call stack is:

- [AggregationBuilderTest.php:39](https://github.com/GromNaN/laravel-mongodb-fork/blob/05054e5491f22a080a0d75acb03ad142b37f2a08/tests/Query/AggregationBuilderTest.php#L39), `MongoDB\Laravel\Tests\Query\AggregationBuilderTest->testCreateFromQueryBuilder()`
- [Builder.php:1974](https://github.com/laravel/framework/blob/8af24b819e55beaf440954516f926d0f5f9abafd/src/Illuminate/Database/Eloquent/Builder.php#L1974), `Illuminate\Database\Eloquent\Builder->__call()` => **Always returns the `Illuminate\Database\Eloquent\Builder`**
- [ForwardsCalls.php:23](https://github.com/laravel/framework/blob/8af24b819e55beaf440954516f926d0f5f9abafd/src/Illuminate/Support/Traits/ForwardsCalls.php#L23), `Illuminate\Database\Eloquent\Builder->forwardCallTo()`
- [Builder.php:602](https://github.com/GromNaN/laravel-mongodb-fork/blob/05054e5491f22a080a0d75acb03ad142b37f2a08/src/Query/Builder.php#L603), `MongoDB\Laravel\Query\Builder->aggregate()`

There is several methods in `Illuminate\Database\Eloquent\Builder` that a calls the `toBase` query builder and pass the result of the query.

src/Eloquent/Builder.php
GromNaN: We use phpstan on this project, and it get [promoted](https://laravel-news.com/larastan) for Laravel projects.

I have updated the annotation to `@return ($function is null ? AggregationBuilder : self)`, which is the [documented way](https://phpstan.org/writing-php-code/phpdoc-types#conditional-return-types).

src/Query/Builder.php
GromNaN: There is so many `Collection` classes:
- `MongoDB\Collection`
- `MongoDB\Laravel\Collection`
- `Illuminate\Support\Collection`

Using the FQCN is more explicit in this context.

src/Query/Builder.php
GromNaN: Updated. Same as https://github.com/mongodb/laravel-mongodb/pull/2738#discussion_r1528659421

src/Query/Builder.php
```diff
@@ -538,9 +596,13 @@ public function generateCacheKey()
         return md5(serialize(array_values($key)));
     }
 
-    /** @inheritdoc */
-    public function aggregate($function, $columns = [])
+    /** @return ($function is null ? AggregationBuilder : mixed) */
+    public function aggregate($function = null, $columns = [])
     {
+        if ($function === null) {
+            return $this->getAggregationBuilder();
```
jmikola: I understand this ignores the `$columns` parameter, but would it be worth asserting that `$columns` is still empty and that a user didn't specify some value(s) there?

src/Query/Builder.php
```diff
@@ -279,6 +284,59 @@ public function dump(mixed ...$args)
...
+            $agg->match(...$wheres);
+        }
+
+        // Distinct query
+        if ($this->distinct) {
+            // Return distinct results directly
+            $column = $this->columns[0] ?? '_id';
+
+            $agg->group(
+                _id: \MongoDB\Builder\Expression::fieldPath($column),
```
jmikola: Should this have a `use` statement?

src/Query/Builder.php
```diff
@@ -279,6 +284,59 @@ public function dump(mixed ...$args)
...
+        }
+
+        if ($this->limit) {
+            $agg->limit($this->limit);
+        }
+
+        // Normal query
+        // Add custom projections.
+        if ($this->projections || $this->columns) {
+            $columns = in_array('*', $this->columns) ? [] : $this->columns;
```
jmikola: What does an asterisk in `$this->columns` indicate? And is it valid for the asterisk to appear alongside other values in the array, or would it make sense to assert that it's the only element if present?

I'm curious as to why an asterisk results in _only_ `$this->projections` being used, instead of selecting _all_ fields (i.e. no projection at all).

src/Query/Builder.php
```diff
@@ -279,6 +284,59 @@ public function dump(mixed ...$args)
...
+
+        $agg = new AggregationBuilder($this->collection, [], $this->options);
+
+        $wheres = $this->compileWheres();
+
+        if (count($wheres)) {
+            $agg->match(...$wheres);
+        }
+
+        // Distinct query
```
jmikola: Are any of these custom operations tested? Basically, anything that would result in `getAggregationBuilder()` prepending something to the pipeline should have a corresponding functional test.

I've having trouble grasping what the "distinct" case here does. It groups by a value (either `_id` or a column) and then reduces each grouped document to the first seen root document containing that value. That seems inconsistent with how the [`distinct` command](https://www.mongodb.com/docs/manual/reference/command/distinct/) works since it emits a list of the distinct values.

src/Query/Builder.php
jmikola: Noted. I would have assumed phpcs might complain about this, but perhaps there's a different configuration here (or I'm mistaken about PHPLIB's own config).

src/Query/Builder.php
GromNaN: It has the SQL meaning: select all columns. In SQL if you write `SELECT *, foo` you get the `foo` column twice in the results.

In Eloquent (SQL or MongoDB), an empty list of columns means a select of all fields (no projection in MongoDB).

src/Query/Builder.php
GromNaN: Good idea. Done.

src/Query/Builder.php
GromNaN: It mimics the way DISTINCT works in SQL. I have added tests.

GromNaN:
After review with @jmikola we decided to remove the feature of initializing the aggregation pipeline from the query builder with `$match`, `$limit`, `$sort` and `$project` stages. Library users are invited to use the Aggregation Builder from the start.

The `AggregationBuilder` class is enhancer with execution methods similar to `Query\Builder` and `Eloquent\Builder`:

- `AggregationBuilder::get($options)` returns the results in a `Laravel\Support\Collection`
- `AggregationBuilder::cursor($options)` returns the results in a `Laravel\Support\LazyCollection` which is a lazy-loading feature.
- `AggregationBuilder::first($options)` returns the first result, after appending the stage `$limit: 1` to the pipeline.

src/Query/AggregationBuilder.php
```diff
@@ -0,0 +1,98 @@
...
+            foreach ($cursor as $item) {
+                yield $item;
+            }
+        });
+    }
+
+    /**
+     * Execute the aggregation pipeline and return the first result.
+     */
+    public function first(array $options = []): mixed
```
GromNaN, line 74: I was concerned of potential conflict with a stage name, but [`$first`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/first/) is an expression operator 😅

composer.json
```diff
@@ -31,6 +31,7 @@
         "mongodb/mongodb": "^1.15"
     },
     "require-dev": {
+        "mongodb/builder": "^0.2",
```
alcaeus, line 34: We could add this to `suggest` to call out that it's necessary to use the agg builder.

src/Query/Builder.php
```diff
@@ -538,9 +540,26 @@ public function generateCacheKey()
         return md5(serialize(array_values($key)));
     }
 
-    /** @inheritdoc */
-    public function aggregate($function, $columns = [])
+    /** @return ($function is null ? AggregationBuilder : mixed) */
+    public function aggregate($function = null, $columns = ['*'])
     {
+        if ($function === null) {
```
alcaeus, line 546: This whole block could use some tests (obviously the missing dependency can't be tested) to protect against regressions.

alcaeus (commented):
LGTM in general, just a couple of minor suggestions.

src/Query/Builder.php
```diff
@@ -538,9 +540,26 @@ public function generateCacheKey()
...
+            if (! trait_exists(FluentFactoryTrait::class)) {
+                // This error will be unreachable when the mongodb/builder package will be merged into mongodb/mongodb
+                throw new BadMethodCallException('Aggregation builder requires package mongodb/builder 0.2+');
+            }
+
+            if ($columns !== [] && $columns !== ['*']) {
```
jmikola: Is there any reason to allow `[]` for `$columns`? IIRC, the previous signature changed the default value to an empty array, but you're preserving `['*']` now. Do you expect users to explicitly call `aggregate(null, [])`? If not, I'd stick to just `['*']` here.

src/Query/AggregationBuilder.php
```diff
@@ -0,0 +1,98 @@
...
+use function collect;
+use function sprintf;
+use function str_starts_with;
+
+class AggregationBuilder
+{
+    use FluentFactoryTrait;
+
+    public function __construct(
+        private MongoDBCollection|LaravelMongoDBCollection $collection,
```
jmikola, line 27: Is there any reason to allow `MongoDB\Collection` here? It's only constructed as such in the test suite, but I presume you could just mock `MongoDB\Laravel\Collection` instead there.

I do realize `MongoDB\Laravel\Collection` just forwards method calls to the PHPLIB class, so there's no technical reason that a `MongoDB\Collection` wouldn't work as well.

OK to leave as-is. I just wanted to bring this up since it makes the API more permissive and I don't expect users would be constructing this on their own anyway.

src/Query/AggregationBuilder.php
```diff
@@ -0,0 +1,98 @@
...
+     * Add a stage without using the builder. Necessary if the stage is built
+     * outside the builder, or it is not yet supported by the library.
+     */
+    public function addRawStage(string $operator, mixed $value): static
+    {
+        if (! str_starts_with($operator, '$')) {
+            throw new InvalidArgumentException(sprintf('The stage name "%s" is invalid. It must start with a "$" sign.', $operator));
+        }
+
+        $this->pipeline[] = [$operator => $value];
```
jmikola, line 42: `FluentFactoryTrait::$pipeline` is annotated as a `list<StageInterface>`. Does this violate that contract?

See: https://github.com/mongodb/mongo-php-builder/blob/aa4f34cde35608b328f23d8fa6270b7cafbe5b18/src/Builder/Stage/FluentFactoryTrait.php#L36C14-L36C34

More generally, I would have expected this kind of escape hatch to exist in the builder itself if at all.

src/Query/AggregationBuilder.php
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    /**
+     * Execute the aggregation pipeline and return the first result.
+     */
+    public function first(array $options = []): mixed
+    {
+        return (clone $this)
+            ->limit(1)
+            ->cursor($options)
```
jmikola: If you're going to apply `limit(1)` I reckon it'd be more performant to just use `get()` here.

src/Query/AggregationBuilder.php
```diff
@@ -0,0 +1,98 @@
...
+            ->first();
+    }
+
+    /**
+     * Execute the aggregation pipeline and return MongoDB cursor.
+     */
+    private function execute(array $options): CursorInterface&Iterator
```
jmikola, line 85: Even though this is internal, I suppose you can't use `MongoDB\Driver\Cursor` here since a CodecCursor might be returned?

Unrelated to this PR, but this made me wonder if we need to keep `CursorInterface&Iterator` in place in PHPLIB 2.0, or if there were plans to integrate the Iterator API into CursorInterface. I asked @alcaeus about this in [PHPLIB-1114](https://jira.mongodb.org/browse/PHPLIB-1114?focusedCommentId=6207158&focusedId=6207158&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-6207158).

Nothing for you here. I merely wanted to cross-reference.

tests/Query/AggregationBuilderTest.php
```diff
@@ -0,0 +1,132 @@
...
+    private static function assertSamePipeline(array $expected, Pipeline $pipeline): void
+    {
+        $expected = Document::fromPHP(['pipeline' => $expected])->toCanonicalExtendedJSON();
+
+        $codec = new BuilderEncoder();
+        $actual = $codec->encode($pipeline);
+        // Normalize with BSON round-trip
+        $actual = Document::fromPHP(['pipeline' => $actual])->toCanonicalExtendedJSON();
+
+        self::assertJsonStringEqualsJsonString($expected, $actual);
```
jmikola, line 145: TIL [`assertJsonStringEqualsJsonString()`](https://docs.phpunit.de/en/9.6/assertions.html#assertjsonstringequalsjsonstring) exists. There's probably a place I could have used this in PHPLIB had I known about it.

src/Query/AggregationBuilder.php
GromNaN: Exact: https://github.com/mongodb/mongo-php-builder/pull/71

GromNaN: Do you think I should add a constructor to `FluentFactoryTrait`? I'm not sure.

src/Query/Builder.php
GromNaN: Tests added.

src/Query/AggregationBuilder.php
jmikola: What do you mean by "add a constructor"? AFAIK, traits cannot have constructors and I don't see why you'd need an initialization method.

Are you suggesting putting `addRawStage()` in the trait?

IMO, you could do without this method entirely and wait until it's requested. That would give you more time to think about the kind of API you'd want when the builder is ultimately used by multiple libraries and not just the Laravel integration.

GromNaN: > What do you mean by "add a constructor"? AFAIK, traits cannot have constructors and I don't see why you'd need an initialization method.

Indeed, the constructor initialization is not necessary.


> Are you suggesting putting `addRawStage()` in the trait?

No, keeping it for Laravel only for now.

tests/Query/AggregationBuilderTest.php
```diff
@@ -0,0 +1,147 @@
...
+
+    public function testAddRawStage(): void
+    {
+        $collection = $this->createMock(MongoDBCollection::class);
+
+        $pipeline = new AggregationBuilder($collection);
+        $pipeline
+            ->addRawStage('$match', ['name' => 'John Doe'])
+            ->addRawStage('$limit', 10)
+            ->addRawStage('$replaceRoot', (object) ['newRoot' => '$$ROOT']);
```
jmikola, line 100: Is there any reason to perform an object cast here?

I noted that `assertSamePipeline()` normalizes the actual and expected values before comparison, but this seems more like a test of _that_ behavior than anything to do with `addRawStage()`.

jmikola (approved):
Please follow up on both open questions, but feel free to merge after doing so. Code changes LGTM.

tests/Query/AggregationBuilderTest.php
GromNaN: This tests that `addRawStage` accepts `stdClass` as 2nd argument.

src/Query/AggregationBuilder.php
GromNaN: I expect `MongoDB\Laravel\Collection` to be deprecated soon, as it's used only for logging commands. We need to switch to driver monitoring features: [PHPORM-56](https://jira.mongodb.org/browse/PHPORM-56). This union type is forward-compatible.

PR #2739, opened 2024-02-28
PHPORM-143 Ensure date are read using local timezone

### LINKED ISSUE #2719: datetime casts not considering timezone
- PHP Version: 8.2
- Database Driver & Version:

### Description:
When using a timezone like "Europe\Amsterdam" the date is stored in UTC but when retrieved is not transformed in local time. 

### Steps to reproduce
1. create a model 
2. dd( $model->created_at , Carbon::now() )

### Expected behaviour
same ( not exactly ) date and time

### Actual behaviour
created_at is 1 hour earlier

GromNaN:
Fix PHPORM-143
Fix #2719

In #2705 ([a5cf5cb](https://github.com/mongodb/laravel-mongodb/commit/a5cf5cba823f5a1d79fb23ce8c9a55e2f27f0ce7#diff-284e164f76aeb13f424d154f9208d91fdbb0112b0a50ad8b1dfa237176f03cf2L144-R140)), the conversion from `MongoDB\BSON\UTCDateTime` to `Carbon\Date` was modified in a way that no longer change use the local timezone. `UTCDateTime::asDateTime` always return a `DateTime` instance with `UTC` timezone, while le previous `Date::createFromTimestampMs($timestampMs)` was using the local timezone.

### Checklist

- [x] ~Update documentation for new features~

### DIFF
--- a/tests/ModelTest.php
+++ b/tests/ModelTest.php
@@ -31,12 +31,15 @@ use MongoDB\Laravel\Tests\Models\User;
 use function abs;
 use function array_keys;
 use function array_merge;
+use function date_default_timezone_set;
 use function get_debug_type;
 use function hex2bin;
 use function sleep;
 use function sort;
 use function strlen;
 use function time;
+
+use const DATE_ATOM;
 
 class ModelTest extends TestCase
 {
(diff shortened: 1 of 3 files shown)


tests/ModelTest.php
```diff
@@ -670,6 +673,30 @@ public function testUnsetDotAttributesAndSet(): void
         $this->assertSame(['note2' => 'DEF', 'note1' => 'ABC'], $user->notes);
     }
 
+    public function testDateUseLocalTimeZone(): void
+    {
+        $tz = 'Australia/Sydney';
+        date_default_timezone_set($tz);
```
jmikola, line 680: This test permanently alters the state of the environment. Would it be preferable to capture the current timezone first and then restore it at the end of the test?

If you're concerned about a failure interrupting things, you can use `tearDown()`, but I don't think that's necessary.

GromNaN: The timezone is reset before each test by this: https://github.com/orchestral/testbench-core/blob/bf3de053bc6566210c0c336635d03e97e270f205/src/Concerns/CreatesApplication.php#L389-L391

But I can explicitly reset the timezone to prevent any regression.

alcaeus (approved):
LGTM pending the timezone things in the tests.

tests/ModelTest.php
jmikola: Ah, I didn't realize the tests in this project used OrchestraTestCase instead of PHPUnit directly. I think a comment explaining that would be sufficient.

jmikola (approved):
LGTM with a comment explaining the timezone being changed isn't a problem.

PR #2741, opened 2024-03-01
PHPORM-148 Fix `null` in `datetime` fields and reset time in `date` field with custom format

### LINKED ISSUE #2729: castAttribute does not handle nullable immutable date times
- Laravel-mongodb Version: 4.1.1
- PHP Version: 8.1
- Database Driver & Version: 6.0.13

### Description:

The `castAttribute` method is overloaded in the base `Model` class. Unlike in the Core Laravel `hasAttributes` trait, it is missing handling for `null` values. See this code snippet from Laravel:

```php
if (is_null($value) && in_array($castType, static::$primitiveCastTypes)) {
    return $value;
}
```

### Steps to reproduce

```php
class MyModel extends \MongoDB\Laravel\Eloquent\Model {
    protected $casts = [
        'my_date' => 'immutable_datetime',
    ];
}

$model = new MyModel();
$model->my_date; // Should be null, instead is current time.
```

### Expected behaviour
`null` should be returned.

### Actual behaviour
The current time is returned.

GromNaN:
Fix PHPORM-148
Fix #2729

- Fix #2729 by removing the method `Model::castAttribute()` that was introduced by #2658 in 4.1.0. Tests added to ensure `null` is allowed in `date`, `datetime`, `immutable_date`, `immutable_datetime` and the variants with custom formats.
- Change the behavior of `immutable_date:j.n.Y H:i` (with custom format), to reset the time. This behave differently than [Laravel 10.46 that treats it like a `immutable_datetime`](https://github.com/laravel/framework/blob/v10.46.0/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L866-L869).

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
-## [4.1.3] - 2024-03-01
+## [4.1.3] - 2024-03-04
 
-* Fix the timezone of `datetime` fields when they are read from the database by @GromNaN in [#2739](https://github.com/mongodb/laravel-mongodb/pull/2739)
-* Fix support for null values in `datetime` fields by @GromNaN in [#2741](https://github.com/mongodb/laravel-mongodb/pull/2741)
+* Fix the timezone of `datetime` fields when they are read from the database. By @GromNaN in [#2739](https://github.com/mongodb/laravel-mongodb/pull/2739)
+* Fix support for null values in `datetime` and reset `date` fields with custom format to the start of the day. By @GromNaN in [#2741](https://github.com/mongodb/laravel-mongodb/pull/2741)
 
 ## [4.1.2] - 2024-02-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
-## [4.1.3] - unreleased
+## [4.1.3] - 2024-03-01
 
 * Fix the timezone of `datetime` fields when they are read from the database by @GromNaN in [#2739](https://github.com/mongodb/laravel-mongodb/pull/2739)
+* Fix support for null values in `datetime` fields by @GromNaN in [#2741](https://github.com/mongodb/laravel-mongodb/pull/2741)
 
 ## [4.1.2] - 2024-02-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
-## [4.1.3] - 2024-03-04
+## [4.1.3] - 2024-03-05
 
 * Fix the timezone of `datetime` fields when they are read from the database. By @GromNaN in [#2739](https://github.com/mongodb/laravel-mongodb/pull/2739)
 * Fix support for null values in `datetime` and reset `date` fields with custom format to the start of the day. By @GromNaN in [#2741](https://github.com/mongodb/laravel-mongodb/pull/2741)
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -316,26 +316,6 @@ abstract class Model extends BaseModel
     }
 
     /** @inheritdoc */
-    protected function castAttribute($key, $value)
-    {
-        $castType = $this->getCastType($key);
-
-        if ($castType === 'immutable_custom_datetime' || $castType === 'immutable_datetime') {
-            if ($value === null) {
-                return null;
-            }
-
-            if (str_starts_with($this->getCasts()[$key], 'immutable_date:')) {
-                return $this->asDate($value)->toImmutable();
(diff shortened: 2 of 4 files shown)


src/Eloquent/Model.php
```diff
@@ -205,9 +205,10 @@ protected function transformModelValue($key, $value)
         if ($this->hasCast($key) && $value instanceof CarbonInterface) {
             $value->settings(array_merge($value->getSettings(), ['toStringFormat' => $this->getDateFormat()]));
 
+            // "date" cast resets the time to 00:00:00.
             $castType = $this->getCasts()[$key];
-            if ($this->isCustomDateTimeCast($castType) && str_starts_with($castType, 'date:')) {
-                $value->startOfDay();
+            if (str_starts_with($castType, 'date:') || str_starts_with($castType, 'immutable_date:')) {
```
GromNaN, line 210: The is no method to check a "date" cast, as opposed to "datetime" cast that keep the time.

src/Eloquent/Model.php
```diff
@@ -315,19 +316,6 @@ protected function fromDecimal($value, $decimals)
...
 
-    /** @inheritdoc */
-    protected function castAttribute($key, $value)
-    {
-        $castType = $this->getCastType($key);
-
-        return match ($castType) {
-            'immutable_custom_datetime','immutable_datetime' => str_starts_with($this->getCasts()[$key], 'immutable_date:') ?
-                $this->asDate($value)->toImmutable() :
-                $this->asDateTime($value)->toImmutable(),
```
GromNaN, line 326: The cast types `immutable_custom_datetime` and `immutable_datetime` are already supported by the parent method: [`HasAttribute::castAttribute()`](https://github.com/laravel/framework/blob/v10.46.0/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L780-L782)

src/Eloquent/Model.php
```diff
@@ -205,9 +205,10 @@ protected function transformModelValue($key, $value)
         if ($this->hasCast($key) && $value instanceof CarbonInterface) {
             $value->settings(array_merge($value->getSettings(), ['toStringFormat' => $this->getDateFormat()]));
 
+            // "date" cast resets the time to 00:00:00.
             $castType = $this->getCasts()[$key];
-            if ($this->isCustomDateTimeCast($castType) && str_starts_with($castType, 'date:')) {
-                $value->startOfDay();
+            if (str_starts_with($castType, 'date:') || str_starts_with($castType, 'immutable_date:')) {
+                $value = $value->startOfDay();
```
GromNaN, line 211: `$value` is reassigned because of `immutable` object.

GromNaN:
~I'm working on this PR, it creates new issues with Laravel 11 (branch 4.2).~ (wrong test)

src/Eloquent/Model.php
jmikola: How does `$this->getCastType($key)` differ from `$this->getCasts()[$key]`? I don't understand why this would match on `immutable_custom_datetime` or `immutable_datetime` but then see a `immutable_date:` prefix from the latter method.

I did note that the function you linked already handles:

```
case 'date':
    return $this->asDate($value);
...
case 'immutable_date':
    return $this->asDate($value)->toImmutable();
```

src/Eloquent/Model.php
```diff
@@ -315,19 +316,6 @@ protected function fromDecimal($value, $decimals)
         return new Decimal128($this->asDecimal($value, $decimals));
     }
 
-    /** @inheritdoc */
-    protected function castAttribute($key, $value)
```
jmikola, line 319: Is the removal of this method what actually fixes the null-handling issue cited in [PHPORM-148](https://jira.mongodb.org/browse/PHPORM-148)?

src/Eloquent/Model.php
GromNaN: - `$this->getCasts()[$key]` is what is defined in the `$casts` property (`date:j.n.Y H:i`).
- `$this->getCastType($key)` is a classification (`custom_datetime`)

src/Eloquent/Model.php
GromNaN: The actual fix is more in `transformModelValue`. It seemed to me that this code was an incomplete attempt to remove the time from `immutable_date` with custom format; which failed with `null` values.

jmikola (approved):
LGTM, although I don't fully grasp the moving parts here.

I'm satisfied with the test cases covering the original reproduction in [PHPORM-148](https://jira.mongodb.org/browse/PHPORM-148).

CHANGELOG.md
```diff
@@ -1,9 +1,10 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
-## [4.1.3] - unreleased
+## [4.1.3] - 2024-03-04
```
alcaeus: This needs changing now:
```suggestion
## [4.1.3] - 2024-03-05
```

alcaeus (approved):

PR #2742, opened 2024-03-04
PHPORM-139 Implement `Model::createOrFirst()` using `findOneAndUpdate` operation

### LINKED ISSUE #2720: Transaction already in progress
- PHP Version: 8.2.15
- Database Driver & Version:  db version v6.0.13 
- PHP MongoDB extension version	**1.17.2**

### Description:
Using firstOrCreate() in a transaction returns an error 

### Steps to reproduce
    $mongodb = DB::connection('app');
    $mongodb->beginTransaction();
    // user does not exist and will be created
    \App\Models\User::firstOrCreate(['username' => 'johncarpenter', 'name' => 'carpenter', 'email' => 'john@example.com']);
    $mongodb->rollBack();

### Expected behaviour
Manage transaction and commit or rollBack 

### Actual behaviour
retuns error :
local.ERROR: MongoDB\Driver\Exception\RuntimeException: Transaction already in progress in vendor/mongodb/mongodb/src/Operation/WithTransaction.php:60

### Notes 
 create() works fine
 firstOrCreate() and updateOrCreate() do work for updates but do not work for creation as they start a transaction in vendor/mongodb/mongodb/src/Operation/WithTransaction.php

### Hack 
We went through adding a test in vendor/mongodb/mongodb/src/Operation/WithTransaction.php

```
 if (!$session->isInTransaction()) {
    $session->startTransaction($this->transactionOptions);
} 
```

GromNaN:
Fix [PHPORM-139](https://jira.mongodb.org/browse/PHPORM-139)
Fix #2720

This may be considered as a new feature and merged into 4.2 instead of 4.1

> [Laravel v10.20](https://laravel-news.com/laravel-10-20-0) came out with a brand-new method called `createOrFirst()`. 
Learn more about this method https://github.com/laravel/framework/discussions/48567: https://laravel-news.com/firstorcreate-vs-createorfirst 

The `createOfFirst` algorithm is:
```
try {
    // insert the new document
} catch (UniqueException) {
    // find the document from criteria
}
```

This is precisely what we can do in a single command with `findOneAndUpdate`.

Unlike Eloquent's implementation, we don't rely on the "unique" index constraint, and just use the `$attribute` array as criteria. 

### Checklist

### DIFF
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -7,9 +7,11 @@ namespace MongoDB\Laravel\Eloquent;
 use Illuminate\Database\ConnectionInterface;
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
 use MongoDB\Driver\Cursor;
+use MongoDB\Laravel\Collection;
 use MongoDB\Laravel\Helpers\QueriesRelationships;
 use MongoDB\Model\BSONDocument;
 
+use function array_intersect_key;
 use function array_key_exists;
 use function array_merge;
 use function collect;
@@ -181,6 +183,22 @@ class Builder extends EloquentBuilder
         }
 
         return $results;
+    }
+
+    public function createOrFirst(array $attributes = [], array $values = [])
+    {
+        // Apply casting and default values to the attributes
+        $instance = $this->newModelInstance($values + $attributes);
+        $values = $instance->getAttributes();
+        $attributes = array_intersect_key($attributes, $values);
+
+        return $this->raw(function (Collection $collection) use ($attributes, $values) {
+            return $collection->findOneAndUpdate(
+                $attributes,
+                ['$setOnInsert' => $values],
+                ['upsert' => true, 'new' => true],
+            );
+        });
     }
 
     /**
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -9,6 +9,7 @@ use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
 use MongoDB\Driver\Cursor;
 use MongoDB\Laravel\Collection;
 use MongoDB\Laravel\Helpers\QueriesRelationships;
+use MongoDB\Laravel\Internal\FindAndModifyCommandSubscriber;
 use MongoDB\Model\BSONDocument;
 
 use function array_intersect_key;
@@ -185,7 +186,14 @@ class Builder extends EloquentBuilder
         return $results;
     }
 
-    public function createOrFirst(array $attributes = [], array $values = [])
+    /**
+     * Attempt to create the record if it does not exist with the matching attributes.
+     * If the record exists, it will be returned.
+     *
+     * @param  array $attributes The attributes to check for duplicate records
+     * @param  array $values     The attributes to insert if no matching record is found
+     */
+    public function createOrFirst(array $attributes = [], array $values = []): Model
     {
         // Apply casting and default values to the attributes
         $instance = $this->newModelInstance($values + $attributes);
@@ -193,11 +201,23 @@ class Builder extends EloquentBuilder
         $attributes = array_intersect_key($attributes, $values);
 
         return $this->raw(function (Collection $collection) use ($attributes, $values) {
-            return $collection->findOneAndUpdate(
-                $attributes,
-                ['$setOnInsert' => $values],
-                ['upsert' => true, 'new' => true],
-            );
+            $listener = new FindAndModifyCommandSubscriber();
+            $collection->getManager()->addSubscriber($listener);
+
+            try {
+                $document = $collection->findOneAndUpdate(
+                    $attributes,
+                    ['$setOnInsert' => $values],
+                    ['upsert' => true, 'new' => true, 'typeMap' => ['root' => 'array', 'document' => 'array']],
+                );
+            } finally {
+                $collection->getManager()->removeSubscriber($listener);
+            }
+
+            $model = $this->model->newFromBuilder($document);
+            $model->wasRecentlyCreated = $listener->created;
+
+            return $model;
         });
     }
 
--- a/tests/ModelTest.php
+++ b/tests/ModelTest.php
@@ -1055,6 +1055,7 @@ class ModelTest extends TestCase
 
         $this->assertSame('taylorotwell@gmail.com', $user1->email);
         $this->assertNull($user1->name);
+        $this->assertTrue($user1->wasRecentlyCreated);
 
         $user2 = User::createOrFirst(
             ['email' => 'taylorotwell@gmail.com'],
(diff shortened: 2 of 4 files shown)


alcaeus (approved):

tests/ModelTest.php
```diff
@@ -1044,4 +1044,43 @@ public function testNumericFieldName(): void
         $this->assertInstanceOf(User::class, $found);
         $this->assertEquals([3 => 'two.three'], $found[2]);
     }
+
+    public function testCreateOrFirst()
+    {
+        $user1 = User::createOrFirst(['email' => 'taylorotwell@gmail.com']);
```
jmikola, line 1050: I'm not sure where these addresses came from but we'd do well to use `@example.com` in email data fixtures.

src/Eloquent/Builder.php
```diff
@@ -183,6 +186,41 @@ public function raw($value = null)
...
+
+        return $this->raw(function (Collection $collection) use ($attributes, $values) {
+            $listener = new FindAndModifyCommandSubscriber();
+            $collection->getManager()->addSubscriber($listener);
+
+            try {
+                $document = $collection->findOneAndUpdate(
+                    $attributes,
+                    ['$setOnInsert' => $values],
+                    ['upsert' => true, 'new' => true, 'typeMap' => ['root' => 'array', 'document' => 'array']],
```
jmikola, line 211: `new` is an option on the internal FindAndModify operation. Since this is using FindOneAndUpdate, the correct option to use would be `returnDocument` with a value of `FindOneAndUpdate::RETURN_DOCUMENT_AFTER`.

src/Eloquent/Builder.php
```diff
@@ -183,6 +186,41 @@ public function raw($value = null)
...
+     * Attempt to create the record if it does not exist with the matching attributes.
+     * If the record exists, it will be returned.
+     *
+     * @param  array $attributes The attributes to check for duplicate records
+     * @param  array $values     The attributes to insert if no matching record is found
+     */
+    public function createOrFirst(array $attributes = [], array $values = []): Model
+    {
+        // Apply casting and default values to the attributes
+        $instance = $this->newModelInstance($values + $attributes);
```
jmikola, line 199: When doing `$values + $attributes`, preference is given to an existing key in `$values`. In that case, it seems possible that actual criteria in `$attributes` might be lost.

Is there a particular reason you didn't do `$attributes + $values`?

src/Eloquent/Builder.php
```diff
@@ -183,6 +186,41 @@ public function raw($value = null)
...
+        $attributes = array_intersect_key($attributes, $values);
+
+        return $this->raw(function (Collection $collection) use ($attributes, $values) {
+            $listener = new FindAndModifyCommandSubscriber();
+            $collection->getManager()->addSubscriber($listener);
+
+            try {
+                $document = $collection->findOneAndUpdate(
+                    $attributes,
+                    ['$setOnInsert' => $values],
```
jmikola, line 210: Since `$values` is over-written above, its value here includes the union of fields from the `$attributes` and `$values` parameters. This is a bit redundant, since upserts already apply the criteria to a new document, but I suppose it makes no functional difference.

Just to confirm, what prevents a user from using query syntax in the `$attributes` parameter? It'd logically work for executing `findOneAndUpdate()`, but I imagine it wouldn't play nice with `Builder::newModelInstance()`.

src/Eloquent/Builder.php
jmikola, line 211: `new` is an option on the internal FindAndModify operation. Since this is using FindOneAndUpdate, the correct option to use would be `returnDocument` with a value of `FindOneAndUpdate::RETURN_DOCUMENT_AFTER`.

src/Eloquent/Builder.php
```diff
@@ -183,6 +186,41 @@ public function raw($value = null)
...
+                    ['upsert' => true, 'new' => true, 'typeMap' => ['root' => 'array', 'document' => 'array']],
+                );
+            } finally {
+                $collection->getManager()->removeSubscriber($listener);
+            }
+
+            $model = $this->model->newFromBuilder($document);
```
jmikola, line 217: How does this method differ from `Builder::newModelInstance()`? It looks like both are used to construct new model instances from "raw" document data.

tests/ModelTest.php
GromNaN: This comes from Laravel's test suite. https://github.com/laravel/framework/blob/a518ef66bebd04c97fa1f04cb3a450795cc3ca11/tests/Database/DatabaseEloquentIntegrationTest.php#L527

src/Eloquent/Builder.php
GromNaN: I'd say the behavior is not well defined if there is a key duplicate between `$attributes` and `$values`. If we look at [the implementation in Laravel](https://github.com/laravel/framework/blob/a518ef66bebd04c97fa1f04cb3a450795cc3ca11/src/Illuminate/Database/Eloquent/Builder.php#L581-L588), it tries to create with an `array_merge($attributes, $values)`, fails if there is a uniqueness constraint in DB and tries to find using `$attributes`.

I used this merge order `$values + $attribute === array_merge($attributes, $values)` to please the unit test that I imported.

src/Eloquent/Builder.php
GromNaN: Oh, I didn't know the `$filter` argument was used to create the document, good to know.

> Just to confirm, what prevents a user from using query syntax in the `$attributes` parameter? It'd logically work for executing `findOneAndUpdate()`, but I imagine it wouldn't play nice with `Builder::newModelInstance()`.

`$parameters` and `$values` are supposed to be used to create a Model instance. Using a query operator would be incorrect.

src/Eloquent/Builder.php
GromNaN: - `Eloquent\Builder::newModelInstance()` applies cast on the attributes. Values comes from the developer.
- `Eloquent\Model::newFromBuilder()` uses data from the database and doesn't apply changes.

src/Eloquent/Builder.php
GromNaN: If `$values` is empty, I get this server error: 
> MongoDB\Driver\Exception\CommandException: Modifiers operate on fields but we found type array instead. For example: {$mod: {<field>: ...}} not {$setOnInsert: []}

If I try to use an empty update, the error is:
> MongoDB\Exception\InvalidArgumentException: Expected update operator(s) or non-empty pipeline for $update

I could add a ternary condition to use `$attribute` as fallback if `$values` is empty, but I do prefer to be explicit and merge all values all the time.

jmikola: > Modifiers operate on fields but we found type array instead.

That could be fixed with an explicit `(object)` cast on the `$setOnInsert` value. Note that server versions prior to 5.0 would still error if that object is empty (this is noted in the [$setOnInsert](https://www.mongodb.com/docs/manual/reference/operator/update/setOnInsert/) docs).

> Expected update operator(s) or non-empty pipeline for $update

Also addressable with an explicit object cast on the value; however, it's secondary to the issue above so not worth looking into.

src/Eloquent/Builder.php
jmikola: This sounds like a possible Laravel bug if you want to pursue that.

PR #2746, opened 2024-03-05
Test Laravel 10 and 11

GromNaN:
Follow-up https://github.com/mongodb/laravel-mongodb/pull/2735

In order to keep testing Laravel 10 when Laravel 11 will be released, I'm using a variable in the CI that specifies the version of `laravel/framework`, which is a meta package containing all `illuminate` packages.

### Checklist

- [ ] ~Add tests and ensure they pass~
- [ ] ~Add an entry to the CHANGELOG.md file~

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -24,16 +24,17 @@ jobs:
                     - "8.2"
                     - "8.3"
                 laravel:
-                  - "10.*"
+                    - "10.*"
+                    - "11.*"
                 include:
                     - php: "8.1"
                       laravel: "10.*"
                       mongodb: "5.0"
                       mode: "low-deps"
-                    # Laravel 11
-                    - php: "8.3"
-                      mongodb: "7.0"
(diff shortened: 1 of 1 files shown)


alcaeus (approved):
LGTM. We should add Laravel 11 to the matrix instead of running two dev builds once it's been released.

GromNaN:
> LGTM. We should add Laravel 11 to the matrix instead of running two dev builds once it's been released.

Yes, but we have to wait for other dependencies to be tagged, like `orchestra` packages.

GromNaN:
@alcaeus updated. We have to update the project config to adapt to the new job names.

alcaeus:
Updated. While doing that I noticed that this will cause issues with pull requests targeting 4.1, as the new build names won't match there. Since branch protection rules won't properly match, should we target 4.1 with this change (excluding Laravel 11) and then add Laravel 11 after merging up?

alcaeus (approved):

PR #2759, opened 2024-03-11
PHPORM-139 Improve `Model::createOrFirst()` tests and error checking

GromNaN:
Applies @jmikola's review on https://github.com/mongodb/laravel-mongodb/pull/2742
Fix PHPORM-139

### DIFF
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -6,6 +6,7 @@ namespace MongoDB\Laravel\Eloquent;
 
 use Illuminate\Database\ConnectionInterface;
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
+use InvalidArgumentException;
 use MongoDB\Driver\Cursor;
 use MongoDB\Laravel\Collection;
 use MongoDB\Laravel\Helpers\QueriesRelationships;
@@ -195,6 +196,10 @@ class Builder extends EloquentBuilder
      */
     public function createOrFirst(array $attributes = [], array $values = []): Model
     {
+        if ($attributes === []) {
+            throw new InvalidArgumentException('You must provide attributes to check for duplicates');
+        }
+
         // Apply casting and default values to the attributes
         $instance = $this->newModelInstance($values + $attributes);
         $values = $instance->getAttributes();
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -12,6 +12,7 @@ use MongoDB\Laravel\Collection;
 use MongoDB\Laravel\Helpers\QueriesRelationships;
 use MongoDB\Laravel\Internal\FindAndModifyCommandSubscriber;
 use MongoDB\Model\BSONDocument;
+use MongoDB\Operation\FindOneAndUpdate;
 
 use function array_intersect_key;
 use function array_key_exists;
@@ -201,6 +202,7 @@ class Builder extends EloquentBuilder
         }
 
         // Apply casting and default values to the attributes
+        // In case of duplicate key between the attributes and the values, the values have priority
         $instance = $this->newModelInstance($values + $attributes);
         $values = $instance->getAttributes();
         $attributes = array_intersect_key($attributes, $values);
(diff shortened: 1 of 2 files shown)


src/Eloquent/Builder.php
```diff
@@ -195,6 +196,10 @@ public function raw($value = null)
      */
     public function createOrFirst(array $attributes = [], array $values = []): Model
     {
+        if ($attributes === []) {
+            throw new InvalidArgumentException('You must provide attributes to check for duplicates');
+        }
```
jmikola, line 202: I'd suggest adding a comment about conflicts between `$attributes` and `$values`, regardless of whether that issue gets reported upstream in Eloquent.

Re: https://github.com/mongodb/laravel-mongodb/pull/2742#discussion_r1520418945

jmikola (commented):
Two outstanding points re: the `findOneAndUpdate()` call:

 * Consider an explicit `(object)` cast when executing the query. Re: https://github.com/mongodb/laravel-mongodb/pull/2742#discussion_r1520408954
 * Use `returnDocument` instead of `new`. Re: https://github.com/mongodb/laravel-mongodb/pull/2742#discussion_r1520021977

src/Eloquent/Builder.php
```diff
@@ -207,8 +214,12 @@ public function createOrFirst(array $attributes = [], array $values = []): Model
             try {
                 $document = $collection->findOneAndUpdate(
                     $attributes,
-                    ['$setOnInsert' => $values],
-                    ['upsert' => true, 'new' => true, 'typeMap' => ['root' => 'array', 'document' => 'array']],
+                    ['$setOnInsert' => (object) $values],
+                    [
+                        'upsert' => true,
+                        'returnDocument' => FindOneAndUpdate::RETURN_DOCUMENT_AFTER,
```
GromNaN, line 222: `new` option was internal. https://github.com/mongodb/laravel-mongodb/pull/2742#discussion_r1520021977

src/Eloquent/Builder.php
```diff
@@ -207,8 +214,12 @@ public function createOrFirst(array $attributes = [], array $values = []): Model
             try {
                 $document = $collection->findOneAndUpdate(
                     $attributes,
-                    ['$setOnInsert' => $values],
-                    ['upsert' => true, 'new' => true, 'typeMap' => ['root' => 'array', 'document' => 'array']],
+                    ['$setOnInsert' => (object) $values],
```
GromNaN, line 219: I added the `(object)` cast, but since we need to support server version 4.4, I need to keep an non-empty object (using filter attributes). https://github.com/mongodb/laravel-mongodb/pull/2742#discussion_r1520408954

jmikola: The object cast certainly doesn't hurt and is arguably correct in the event all field names were numeric (if that's even possible here).

I do think it's worth adding a comment to explain that you're intentionally keeping extra fields in `$values` (instead of filtering down to keys in the original param) to ensure compatibility with pre-5.0 server versions.

jmikola (dismissed):
Suggested a comment above `$setOnInsert` line, but code changes LGTM.

src/Eloquent/Builder.php
```diff
@@ -207,8 +214,14 @@ public function createOrFirst(array $attributes = [], array $values = []): Model
             try {
                 $document = $collection->findOneAndUpdate(
                     $attributes,
-                    ['$setOnInsert' => $values],
-                    ['upsert' => true, 'new' => true, 'typeMap' => ['root' => 'array', 'document' => 'array']],
+                    // Before MongoDB 5.0, the $setOnInsert does support empty document,
+                    // so the filter value are added to avoid an error.
```
jmikola:
```suggestion
                    // Before MongoDB 5.0, $setOnInsert requires a non-empty document.
                    // This is should not be an issue as $values includes the query filter.
```

Optional re-wording.

jmikola (dismissed):

src/Eloquent/Builder.php
GromNaN: Thanks.

jmikola (approved):

PR #2768, opened 2024-03-12
DOCSP-35970: Find one usage example

norareidy:
JIRA - https://jira.mongodb.org/browse/DOCSP-35970
Staging - https://preview-mongodbnorareidy.gatsbyjs.io/laravel/DOCSP-35970-staging/usage-examples/findOne/

### Checklist


docs/usage-examples/findOne.txt
```diff
@@ -0,0 +1,115 @@
...
+===============
+
+You can retrieve a single document from a collection by calling the ``where()`` and
+``first()`` methods on an Eloquent model or a query builder.
+
+Pass a query filter to the ``where()`` method and call the ``first()`` method to return
+one document in the collection that matches the filter. If multiple documents match the
+query filter, ``first()`` returns the first matching document according to the documents'
+:term:`natural order` in the database or according to the sort order specified by the
+``orderBy()`` method.
```
rustagir:
```suggestion
:term:`natural order` in the database or according to the sort order that you can specify by using the
``orderBy()`` method.
```

docs/usage-examples/findOne.txt
```diff
@@ -0,0 +1,115 @@
+.. _laravel-find-one-usage:
+
+===============
+Find a Document
+===============
+
+You can retrieve a single document from a collection by calling the ``where()`` and
+``first()`` methods on an Eloquent model or a query builder.
```
rustagir, line 8:
```suggestion
You can retrieve a single document from a collection by chaining the ``where()`` and
``first()`` methods on an Eloquent model or a query builder.
```

docs/usage-examples/findOne.txt
```diff
@@ -0,0 +1,115 @@
...
+            Marty DeBergi on what proves to be a fateful tour.
+
+To perform the find operation, start your Laravel application by running the following command:
+
+.. code-block:: bash
+
+   php artisan serve
+
+Then, open the URL http://127.0.0.1:8000/browse_movies in your web browser to view the expected
+output.
```
rustagir: S: might be better not to inline the URL to prevent unwanted redirects
```suggestion
Then, open the following URL in your web browser to view the expected output:

.. code-block:: none

   http://127.0.0.1:8000/browse_movies
```

rustagir (changes requested):
a few small things, let me know if you have any questions about my comments!

docs/usage-examples/findOne.txt
```diff
@@ -0,0 +1,118 @@
+.. _laravel-find-one-usage:
+
+===============
+Find a Document
+===============
+
+You can retrieve a single document from a collection by chaining the ``where()`` and
+``first()`` methods on an Eloquent model or a query builder.
+
+Pass a query filter to the ``where()`` method and call the ``first()`` method to return
```
rustagir:
```suggestion
Pass a query filter to the ``where()`` method and then call the ``first()`` method to return
```

docs/usage-examples/findOne.txt
```diff
@@ -0,0 +1,118 @@
...
+-------
+
+This usage example retrieves a document that matches a query filter from the ``movies``
+collection in the ``sample_mflix`` database. The example uses the ``Movie`` Eloquent model
+to represent the ``movies`` collection.
+
+This example performs the following actions:
+
+- Defines a query filter that matches documents in which the value of the ``directors`` field
+  is ``"Rob Reiner"``
```
rustagir: S: I think the `directors` field is an array-valued field

```suggestion
- Defines a query filter that matches documents in which the value of the ``directors`` field
  includes ``"Rob Reiner"``
```

docs/usage-examples/findOne.txt
```diff
@@ -0,0 +1,118 @@
...
+                 {
+                      $movie = Movie::where('directors', 'Rob Reiner')
+                          ->orderBy('_id')
+                          ->first();
+
+                      return view('browse_movies', [
+                          'movie' => $movie
+                      ]);
+                  }
+            }
```
rustagir: S: move into files and include (applies to all code examples)

docs/usage-examples/findOne.txt
```diff
@@ -0,0 +1,118 @@
...
+
+.. code-block:: bash
+
+   php artisan serve
+
+Then, open the following URL in your web browser to view the expected output:
+
+.. code-block:: none
+
+   http://127.0.0.1:8000/browse_movies
```
rustagir: Q: Should we include a link to the full git blob as we do with other driver usage example such as [here](https://www.mongodb.com/docs/drivers/go/current/usage-examples/insertMany/#:~:text=%7D-,View%20a,-fully%20runnable%20example) in Go?
Just a question, we dont do this for some of the other drivers so I am indifferent

rustagir (dismissed):
left some non-blocking comments, Can leave another review if you prefer though!

docs/usage-examples/findOne.txt
norareidy: I'm kind of against it in this case, since we'd have to link not just to one runnable file but to all the necessary Laravel files (Model, controller, view, etc)

docs/includes/usage-examples/FindOneTest.php
```diff
@@ -0,0 +1,34 @@
+<?php
```
norareidy, line 1: I get the following message when running this file:
```
There was 1 PHPUnit test runner warning:

1) No tests found in class "App\Http\Controllers\FindOneTest".

No tests executed!

```

GromNaN: Because there is no assertion.
You can add `$this->expectNotToPerformAssertions();` at the end of tests without assertion.

docs/includes/usage-examples/FindOneTest.php
```diff
@@ -0,0 +1,34 @@
...
+    public function updateOne(): void
+    {
+        require_once __DIR__ . '/Movie.php';
+
+      // begin-find-one
+        $movie = Movie::where('directors', 'Rob Reiner')
+          ->orderBy('_id')
+          ->first();
+
+        print_r($movie->toJson());
```
GromNaN: Do we really want to display the results with `print_r`? That is not very compatible with PHPUnit.

docs/includes/usage-examples/FindOneTest.php
norareidy: I'm still getting the same message after adding `$this->expectNotToPerformAssertions()`.

docs/includes/usage-examples/FindOneTest.php
norareidy: How would you prefer to display the results? Originally I used a view, but I changed it to `print_r` for simplicity

docs/includes/usage-examples/Movie.php
```diff
@@ -0,0 +1,12 @@
...
+
+namespace App\Models;
+
+use MongoDB\Laravel\Eloquent\Model;
+
+class Movie extends Model
+{
+    protected $connection = 'mongodb';
+    protected $collection = 'movies';
+    protected $fillable = ['title', 'year', 'runtime', 'imdb', 'plot'];
```
GromNaN: This is not necessary.

docs/includes/usage-examples/FindOneTest.php
GromNaN: Since it's a simple string, I modified to use `echo`.

docs/includes/usage-examples/FindOneTest.php
```diff
@@ -0,0 +1,33 @@
...
+        $movie = Movie::where('directors', 'Rob Reiner')
+          ->orderBy('_id')
+          ->first();
+
+        echo $movie->toJson();
+        // end-find-one
+
+        $this->assertInstanceOf(Movie::class, $movie);
+
+        $this->expectOutputRegex('^{"_id":"[a-z0-9]{24}","title":"The Shawshank Redemption","directors":["Frank Darabont","Rob Reiner"]}$');
```
GromNaN: Adding assertions on output.

GromNaN (commented):
LGTM. With a simplification of the model. Maybe sharing the same model class for all the find examples would be a good idea.

docs/includes/usage-examples/Movie.php
GromNaN: Seen in other PR that it will be shared with other examples. That's fine.

GromNaN (approved):
LGTM

PR #2747, opened 2024-03-05
DOCSP-37056 eloquent relationships

ccho-mongodb:
JIRA:
https://jira.mongodb.org/browse/DOCSP-37056

Staging:

[Eloquent Models Landing Page](https://preview-mongodbcchomongodb.gatsbyjs.io/laravel/DOCSP-37056-eloquent-relationships/eloquent-models/)
[Eloquent Models Relationships](https://preview-mongodbcchomongodb.gatsbyjs.io/laravel/DOCSP-37056-eloquent-relationships/eloquent-models/relationships/)

Notes for reviewers:

- Most of the content is located in [docs/eloquent-models/relationships.txt](https://github.com/mongodb/laravel-mongodb/pull/2747/files#diff-176abb415112c78673ced6423e881bf068aa6a121a374b58db9c6e2a6449678a). 
- This is the 2nd of 3 PRs on Eloquent Models. The 1st is in [this PR](https://github.com/10gen/docs-laravel/pull/12) and the 3rd (Schema Builder) is covered by https://jira.mongodb.org/browse/DOCSP-37057

### Checklist

- [ ] Add tests and ensure they pass
- [ ] Add an entry to the CHANGELOG.md file
- [ ] Update documentation for new features

PR #2775, opened 2024-03-14
PHPORM-160 Add tests to documentation examples

GromNaN:
Fix PHPORM-160
Based on https://github.com/mongodb/laravel-mongodb/pull/2747

To run the test:
```console
php vendor/bin/phpunit docs/includes/eloquent-models/relationships/RelationshipsExamplesTest.php
```

Preview: https://preview-mongodbcchomongodb.gatsbyjs.io/laravel/GromNaN-PHPORM-160/eloquent-models/

### Checklist


tests/DocumentationExamplesTest.php
```diff
@@ -0,0 +1,156 @@
...
+
+        $controller = new RelationshipController();
+        $controller->oneToMany();
+
+        $planet = Planet::first();
+        $this->assertInstanceOf(Planet::class, $planet);
+        $this->assertCount(2, $planet->moons);
+        $this->assertInstanceOf(Moon::class, $planet->moons->first());
+
+        $controller->planetMoonsDynamic();
```
alcaeus: This and a number of other tests don't run any assertions after the controller call. I assume we only care about the controller not throwing an error. If so, we should note this in a comment here to avoid confusion.

We can also consider using `assert` in the controller outside of the doc comments to at least verify that data was correctly loaded, assuming code outside of the comments doesn't get shown on the docs page.

tests/DocumentationExamplesTest.php
```diff
@@ -0,0 +1,156 @@
...
+
+        $spaceship = SpaceShip::first();
+        $this->assertInstanceOf(SpaceShip::class, $spaceship);
+        $this->assertCount(2, $spaceship->cargo);
+        $this->assertInstanceOf(Cargo::class, $spaceship->cargo->first());
+    }
+
+    /**
+     * @r unInSeparateProcess
+     * @ preserveGlobalState disabled
```
alcaeus: Are the spaces intentional? We should either use the correct annotations or remove the docblock.

```suggestion
     * @runInSeparateProcess
     * @preserveGlobalState disabled
```

alcaeus (changes requested):
I focused my review on the DocumentationExamplesTest file. As for the rest of the commits, are those backports from a newer version? If so, can we backport a squashed commit? Also, is 4.1 the correct target branch here?

tests/DocumentationExamplesTest.php
GromNaN: Good catch. Fixed.

tests/DocumentationExamplesTest.php
GromNaN: I merged the doc examples with the test cases, similar to [library's `DocumentationExamplesTest`](https://github.com/mongodb/mongo-php-library/blob/master/tests/DocumentationExamplesTest.php) but in the `./docs/includes` dir. Adding a new Test Suite in the PHPUnit config.

alcaeus (approved):
Tests LGTM. I'll defer to the docs team to ensure the examples are what they expect.

docs/includes/eloquent-models/relationships/RelationshipsExamplesTest.php
```diff
@@ -0,0 +1,254 @@
...
+use App\Models\SpaceExplorer;
+use App\Models\SpaceShip;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Schema\SQLiteBuilder;
+use Illuminate\Support\Facades\Schema;
+use MongoDB\Laravel\Tests\TestCase;
+
+use function assert;
+
+class RelationshipsExamplesTest extends TestCase
```
GromNaN, line 21: To run this test:

```
php vendor/bin/phpunit docs/includes/eloquent-models/relationships/RelationshipsExamplesTest.php
```

ccho-mongodb (approved):
LGTM

PR #2781, opened 2024-03-15
DOCSP-35974: Update one usage example

norareidy:
JIRA - https://jira.mongodb.org/browse/DOCSP-35974
Staging - https://preview-mongodbnorareidy.gatsbyjs.io/laravel/DOCSP-35974/usage-examples/updateOne/

### Checklist


docs/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,109 @@
...
+   section of the Laravel documentation.
+   
+.. tabs::
+
+   .. tab:: Controller File Code
+      :tabid: find-one-controller
+
+      .. code-block:: php
+
+         class MovieController
```
GromNaN: Could you extract the code example into `.php` files, so we can validate PHP syntax, fix CS and even test them.

GromNaN (changes requested):

docs/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,109 @@
...
+
+You can update a document in a collection by chaining the ``where()``, ``first()``,
+and ``update()`` methods to an Eloquent model or a query builder.
+
+Pass a query filter to the ``where()`` method and call the ``first()`` method to match 
+only one document. Then, pass the document changes to the ``update()`` method to update
+one document that matches the filter.
+
+To learn more about updating data with {+odm-short+}, see the `Updates
+<https://laravel.com/docs/6.x/eloquent#updates>`__ section of the Laravel documentation.
```
GromNaN: You can link to the latest version (currently 11.x) by removing the version from the URL.
```suggestion
<https://laravel.com/docs/eloquent#updates>`__ section of the Laravel documentation.
```

docs/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,109 @@
+.. _laravel-update-one-usage:
+
+=================
+Update a Document
+=================
```
rachel-mack, line 5: I would add the `contents` for the TOC, and the tags while you're working on these pages.
```suggestion
=================

.. contents:: On this page
   :local:
   :backlinks: none
   :depth: 1
   :class: twocols

.. facet::
   :name: genre
   :values: reference
 
.. meta::
   :keywords: <whatever you think>
```

rachel-mack (approved):
One suggestion:

docs/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,77 @@
...
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+You can update a document in a collection by retrieving a single document and calling
+the ``update()`` method on an Eloquent model or a query builder.
+
+Pass a query filter to the ``where()`` method, sort the matching documents, and call the
+``first()`` method to match only one document. Then, update this matching document by passing
```
GromNaN:
```suggestion
``first()`` method to get only the first document. Then, update this matching document by passing
```

docs/includes/usage-examples/UpdateOneTest.php
```diff
@@ -0,0 +1,38 @@
+<?php
```
norareidy, line 1: I get the following error when running this file:
`Could not open input file: vendor/bin/phpunit`

jmikola: That would get installed by running `composer update` from the repository root to install all dependencies in `composer.json` (both `require` and `require-dev`).

Was the error above based on following instructions in [Contributing: Run Tests](https://github.com/mongodb/laravel-mongodb/blob/4.2/CONTRIBUTING.md#run-tests). It's not clear to me how this file would even be picked up by the normal test suite (files under `tests/`), so I'll defer to @GromNaN to follow up here.

docs/includes/usage-examples/UpdateOneTest.php
```diff
@@ -0,0 +1,38 @@
...
+
+declare(strict_types=1);
+
+namespace App\Http\Controllers;
```
jmikola, line 5: @GromNaN: more generally I don't see a clear benefit to putting PHPUnit test files outside of the `tests/` directory like this. This differs from the PHPLIB approach, where we run example/docs scripts from [ExamplesTest](https://github.com/mongodb/mongo-php-library/blob/master/tests/ExamplesTest.php) and assert output.

Based on the current PHPUnit configuration, these files won't be picked up by the test suite (nor tools like phpcs and phpstan). It might make more sense to organize these files within the actual test suite and then have docs refer to them there. And in that case, we can use meaningful namespaces.

docs/includes/usage-examples/UpdateOneTest.php
norareidy: I'm following [these](https://docs.google.com/document/d/12mM87pjVzcTkPCbei2c6Ollb-xNxPQYATHVuavpsRxc/edit#heading=h.hbicvpo3h6hv) instructions, but I was running the command in the wrong folder; here's what I get now:
```
There was 1 PHPUnit test runner warning:

1) No tests found in class "App\Http\Controllers\UpdateOneTest".

No tests executed!
```

docs/includes/usage-examples/UpdateOneTest.php
```diff
@@ -0,0 +1,38 @@
...
+use App\Models\Movie;
+use MongoDB\Laravel\Tests\TestCase;
+
+class UpdateOneTest extends TestCase
+{
+    /**
+     * @runInSeparateProcess
+     * @preserveGlobalState disabled
+     */
+    public function updateOne(): void
```
GromNaN: The method name must start with `test*`.

```suggestion
    public function testUpdateOne(): void
```

docs/includes/usage-examples/UpdateOneTest.php
jmikola: #2795 should address this.

GromNaN (approved):
I updated the example test (formatting and assertions).

docs/includes/usage-examples/UpdateOneTest.php
GromNaN: The namespace is of no particular interest. I think we'll come back to it to clean it up.

PR #2816, opened 2024-04-02
DOCSP-35934: databases and collections

rustagir:
[JIRA](https://jira.mongodb.org/browse/DOCSP-35934)
[Staging](https://preview-mongodbrustagir.gatsbyjs.io/laravel/DOCSP-35934/fundamentals/database-collection/)
[docs PR](https://github.com/10gen/docs-laravel/pull/36)

### Checklist

### DIFF
--- a/docs/fundamentals/database-collection.txt
+++ b/docs/fundamentals/database-collection.txt
@@ -26,18 +26,18 @@ and manage MongoDB databases and collections.
 MongoDB organizes data in a hierarchical structure. A MongoDB
 deployment contains one or more **databases**, and each database
 contains one or more **collections**. In each collection, MongoDB stores
-data as **documents** that contain field-and-value pairs. In Laravel,
-documents are mapped to Eloquent models.
+data as **documents** that contain field-and-value pairs. In
+{+odm-short+}, you can access documents through Eloquent models.
 
 To learn more about the document data format,
 see :manual:`Documents </core/document/>` in the Server manual.
 
 .. _laravel-access-db:
 
-Access a Database
------------------
+Specify the Database in a Connection Configuration
+--------------------------------------------------
 
-You can access a database by specifying a database connection in your
+You specify a database name when configuring a connection in your
 application's ``config/database.php`` file. The ``connections`` property
 in this file stores all of your database connection information, such as
 your connection string, database name, and optionally, authentication
@@ -45,7 +45,7 @@ details. After you specify a database connection, you can perform
 database-level operations and access collections that the database
 contains.
 
-If you set the ``database`` property in a connection to the name of a
+If you set the database name in a connection string to the name of a
 nonexistent database, Laravel still makes a valid connection. When you
 insert any data into a collection in the database, the server creates it
 automatically.
@@ -55,16 +55,15 @@ create a database connection to the ``animals`` database in the
 ``config/database.php`` file:
 
 .. code-block:: php
-   :emphasize-lines: 8
+   :emphasize-lines: 7
    
-   'default' => env('DB_CONNECTION', 'mongodb'),
+   'default' => 'mongodb',
 
    'connections' => [
 
        'mongodb' => [
            'driver' => 'mongodb',
-           'dsn' => env('DB_URI', '<connection string>'),
-           'database' => 'animals',
+           'dsn' => 'mongodb://localhost:27017/animals',
        ], ...
    ]
 
@@ -82,14 +81,12 @@ The following example shows how to specify multiple database connections
 
        'mongodb' => [
            'driver' => 'mongodb',
-           'dsn' => env('DB_URI', '<connection string>'),
-           'database' => 'animals',
+           'dsn' => 'mongodb://localhost:27017/animals',
        ],
 
        'mongodb_alt' => [
            'driver' => 'mongodb',
-           'dsn' => env('DB_URI', '<connection string>'),
-           'database' => 'plants',
+           'dsn' => 'mongodb://localhost:27017/plants',
        ]
 
    ], ...
@@ -101,13 +98,14 @@ The following example shows how to specify multiple database connections
    overhead in using two connections for two distinct databases, so you
    do not need to optimize your connections.
 
-If your application contains multiple
-database connections, you override the ``$connection``
-property when creating a ``Model`` class to apply the model to an
-alternate database.
+If your application contains multiple database connections and you want
+to store your model in a database other than the default, override the
+``$connection`` property when creating a ``Model`` class.
 
 The following example shows how to override the ``$connection`` property
-on the ``Flower`` model class to use the ``mongodb_alt`` connection:
+on the ``Flower`` model class to use the ``mongodb_alt`` connection.
+This directs {+odm-short+} to store the model in the ``flowers``
+collection of the ``plants`` database, instead of in the default database:
 
 .. code-block:: php
 
@@ -121,34 +119,34 @@ on the ``Flower`` model class to use the ``mongodb_alt`` connection:
 Access a Collection
 -------------------
 
-When you create a ``Model`` class in {+odm-short+}, it is automatically
-stored in a collection with a name that is the plural of the model class
-name. For example, if you create a model class called ``Flower``,
+When you create model class that extends
+``MongoDB\Laravel\Eloquent\Model``, {+odm-short+} stores the model data
+in a collection with a name that is the snake case plural form of your
+model class name.
+
+For example, if you create a model class called ``Flower``,
 Laravel applies the model to the ``flowers`` collection in the database.
 
 .. tip::
 
-   You might be storing data in a collection with a different name than the
-   model, or your collection might contain more than one type of data. To
-   learn how to set a collection name when defining a model class, see the
+   To learn how to specify a different collection name when defining a model class, see the
    :ref:`laravel-model-customize-collection-name` section of the Eloquent
    Model Class guide.
 
-For many operations, such as CRUD operations and relationships, we
-recommend that you use the Eloquent ORM to access a collection
-implicitly. The following example specifies a find operation by using
-the ``Flower`` class, so Laravel retrieves results from the ``flowers``
-collection:
+We generally recommend that you use the Eloquent ORM to access a collection
+implicitly for code readability and maintainability. The following
+example specifies a find operation by using the ``Flower`` class, so
+Laravel retrieves results from the ``flowers`` collection:
 
 .. code-block:: php
 
    Flower::where('name', 'Water Lily')->get()
 
-For other collection operations, you can access a collection by using
-a facade to access the query builder class ``DB`` and calling the
-``collection()`` method. The following example shows the same query as
-in the preceding example constructed by using the ``DB::collection()``
-method:
+If you are not able to accomplish your operation by using an Eloquent
+model, you can access a collection by using a facade to access the query
+builder class ``DB`` and calling the ``collection()`` method. The
+following example shows the same query as in the preceding example, but
+the query is constructed by using the ``DB::collection()`` method:
 
 .. code-block:: php
 
@@ -157,42 +155,22 @@ method:
        ->where('name', 'Water Lily')
        ->get()
 
-.. tip::
-   
-   In MongoDB, each document in a collection has a field called ``_id`` which has a value
-   that is a unique identifier. To learn more about customizing this field
-   when you create a model to access a collection, see the :ref:`laravel-model-customize-primary-key` field of the Eloquent
-   Model Class guide.
-   
 List Collections
 ----------------
 
-To see the names of the collections in a database, call the 
+To see information about each of the collections in a database, call the 
 ``listCollections()`` method.
 
-The following example acceses a database connection, then
-calls the ``listCollections()`` method and outputs the collection names:
+The following example accesses a database connection, then
+calls the ``listCollections()`` method to retrieve information about the
+collections in the database:
 
 .. code-block:: php
 
    $collections = DB::connection('mongodb')->getMongoDB()->listCollections();
-   dump($collections);
 
-Drop a Collection
------------------
+Create and Drop Collections
+---------------------------
 
-We recommend performing a **migration** to handle dropping a collection.
-Deleting data from within a migration provides a controlled approach
-that ensures consistency, version control, and reversibility in your
-application.
-
-To learn more about migrations, see the
+To learn how to create and drop collections, see the
 :ref:`laravel-eloquent-migrations` section in the Schema Builder guide.
-
-To drop a collection idiomatically, the Laravel ``Schema`` facade provides the
-``drop()`` and ``dropIfExists()`` methods. The following example
-demonstrates how to drop the ``flowers`` collection:
-
-.. code-block:: php
-
-   Schema::dropIfExists('flowers');
--- a/docs/fundamentals/database-collection.txt
+++ b/docs/fundamentals/database-collection.txt
@@ -1,0 +1,164 @@
+.. _laravel-db-coll:
+
+=========================
+Databases and Collections
+=========================
+
+.. facet::
+   :name: genre
+   :values: reference
+
+.. meta::
+   :keywords: php framework, odm
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+In this guide, you can learn how to use {+odm-short+} to access
+and manage MongoDB databases and collections.
+
+MongoDB organizes data in a hierarchical structure. A MongoDB
+deployment contains one or more **databases**, and each database
+contains one or more **collections**. In each collection, MongoDB stores
+data as **documents** that contain field-and-value pairs.
+
+To learn more about the document data format,
+see :manual:`Documents </core/document/>` in the Server manual.
+
+.. _laravel-access-db:
+
+Access a Database
+-----------------
+
+You can access a database by specifying a database connection in your
+application's ``config/database.php`` file. The ``connections`` property
+in this file stores all of your database connection information, such as
+your connection string, database name, and optionally, authentication
+details. After you specify a database connection, you can perform
+database-level operations and access collections that the database
+contains.
+
+If you set the ``database`` property in a connection to the name of a
+nonexistent database, Laravel still makes a valid connection. When you
+insert any data into a collection in the database, the server creates it
+automatically.
+
+The following example shows how to set a default database connection and
+create a database connection to the ``animals`` database in the
+``config/database.php`` file:
+
+.. code-block:: php
+   :emphasize-lines: 8
+   
+   'default' => env('DB_CONNECTION', 'mongodb'),
+
+   'connections' => [
+
+       'mongodb' => [
+           'driver' => 'mongodb',
+           'dsn' => env('DB_URI', '<connection string>'),
+           'database' => 'animals',
+       ], ...
+   ]
+
+When you set a default database connection, {+odm-short+} uses that
+connection for operations, but you can specify multiple database connections
+in your ``config/database.php`` file.
+
+The following example shows how to specify multiple database connections
+(``mongodb`` and ``mongodb_alt``) to access the ``animals`` and
+``plants`` databases:
+
+.. code-block:: php
+   
+   'connections' => [
+
+       'mongodb' => [
+           'driver' => 'mongodb',
+           'dsn' => env('DB_URI', '<connection string>'),
+           'database' => 'animals',
+       ],
+
+       'mongodb_alt' => [
+           'driver' => 'mongodb',
+           'dsn' => env('DB_URI', '<connection string>'),
+           'database' => 'plants',
+       ]
+
+   ], ...
+
+If your application contains multiple
+database connections, you override the ``$connection``
+property when creating a ``Model`` class to apply the model to an
+alternate database.
+
+The following example shows how to override the ``$connection`` property
+on the ``Flower`` model class to use the ``mongodb_alt`` connection:
+
+.. code-block:: php
+
+   class Flower extends Model
+   {
+       protected $connection = 'mongodb_alt';
+   }
+
+.. _laravel-access-coll:
+
+Access a Collection
+-------------------
+
+When you create a ``Model`` class in {+odm-short+}, it is automatically
+stored in a collection with a name that is the plural of the model class
+name. For example, if you create a model class called ``Flower``,
+Laravel applies the model to the ``flowers`` collection in the database.
+
+.. tip::
+
+   You might be storing data in a collection with a different name than the
+   model, or your collection might contain more than one type of data. To
+   learn how to set a collection name when defining a model class, see the
+   :ref:`laravel-model-customize-collection-name` section of the Eloquent
+   Model Class guide.
+
+For many operations, such as CRUD operations and relationships, we
+recommend that you use the Eloquent ORM to access a collection
+implicitly. The following example specifies a find operation by using
+the ``Flower`` class, so Laravel retrieves results from the ``flowers``
+collection:
+
+.. code-block:: php
+
+   Flower::where('name', 'Water Lily')->get()
+
+For other collection operations, you can access a collection by using
+a facade to access the query builder class ``DB`` and calling the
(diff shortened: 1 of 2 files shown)


docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,164 @@
...
+.. code-block:: php
+   :emphasize-lines: 8
+   
+   'default' => env('DB_CONNECTION', 'mongodb'),
+
+   'connections' => [
+
+       'mongodb' => [
+           'driver' => 'mongodb',
+           'dsn' => env('DB_URI', '<connection string>'),
```
GromNaN: The `config/dabase.php` file is committed in the git repository of each project. The MongoDB DSN usually contains credentials, it should be in the `.env` file.
The 2nd argument of `env()` is the default value. Typically `mongodb://localhost:27017` for the MongoDB DSN.

The application can contain a `.env.example` but in local the developers must create the file `.env`. For production, there is many ways to set the environment variables, from the `.env` file to server or container config.

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,164 @@
...
+           'database' => 'animals',
+       ],
+
+       'mongodb_alt' => [
+           'driver' => 'mongodb',
+           'dsn' => env('DB_URI', '<connection string>'),
```
GromNaN: You can specify that the internal MongoDB driver for PHP reuses the same connection when creating 2 clients with the same DSN. So there's no overhead in using 2 connections for 2 distinct databases, and there's no point in trying to optimize this.

GromNaN (changes requested):

docs/fundamentals/database-collection.txt
rustagir: I believe this is content that will be better suited for the Connection Guide

GromNaN: This placeholder is wrong. It's better removing the fallback value then.
```suggestion
           'dsn' => env('DB_URI'),
```

GromNaN: This is wrong. When a "dsn" is provided, the database name is read from the "dsn", and the  "database" config is ignored.


Your example should look like this, so we don't bother with env vars.
```php

       'mongodb' => [
           'driver' => 'mongodb',
           'dsn' => 'mongodb://localhost:27017/animals,
        ],
```

jmikola: > When a "dsn" is provided, the database name is read from the "dsn", and the "database" config is ignored.

Outside the scope of this PR, but is that something we should consider changing in some future release. The "auth database" component in the DSN should ideally never be specified since it was made obsolete by the `authSource` URI option (as I explained in https://github.com/symfony-cli/symfony-cli/pull/458#issuecomment-2025269218).

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,198 @@
...
+We recommend performing a **migration** to handle dropping a collection.
+Deleting data from within a migration provides a controlled approach
+that ensures consistency, version control, and reversibility in your
+application.
+
+To learn more about migrations, see the
+:ref:`laravel-eloquent-migrations` section in the Schema Builder guide.
+
+To drop a collection idiomatically, the Laravel ``Schema`` facade provides the
+``drop()`` and ``dropIfExists()`` methods. The following example
```
jmikola: @GromNaN: Since dropping a nonexistent collection is a NOP in PHP, is there any practical difference between these two methods? Does it make sense to talk about both of them given the driver behavior?

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,198 @@
...
+   Model Class guide.
+   
+List Collections
+----------------
+
+To see the names of the collections in a database, call the 
+``listCollections()`` method.
+
+The following example acceses a database connection, then
+calls the ``listCollections()`` method and outputs the collection names:
```
jmikola: [`Database::listCollections()`](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBDatabase-listCollections/) is going to yield a list of documents _describing_ the collection. The name is included, but it's not the only field.

In contrast, [`listCollectionNames()`](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBDatabase-listCollectionNames/) returns a list of strings.

Here, it would be more accurate to say that you're outputting information about each collection in the database instead of just the "names".

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,198 @@
...
+       ]
+
+   ], ...
+
+.. note::
+
+   The MongoDB PHP driver reuses the same connection when
+   you create two clients with the same connection string. There is no
+   overhead in using two connections for two distinct databases, so you
+   do not need to optimize your connections.
```
jmikola, line 102: I'll note that if users expect to change the database name for separate connections, that would result in the driver _not_ re-using the same connection since the connection string (i.e. `dsn` option) will technically differ.

This relates to my suggestion in https://github.com/mongodb/laravel-mongodb/pull/2816/files#r1559726362 to encourage users _not_ to specify any "auth database" in the connection string, which the Laravel library just so happens to use as a "default" database.

docs/fundamentals/database-collection.txt
GromNaN: So the doc should tell both methods have the same effect and they exist for compatibility with the SQL Eloquent.

We could remove the collection existence check from this method?
https://github.com/mongodb/laravel-mongodb/blob/95c5b4fac69c2bd5ea01367e62d14dd5278e7513/src/Schema/Builder.php#L87-L92

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,198 @@
...
+--------
+
+In this guide, you can learn how to use {+odm-short+} to access
+and manage MongoDB databases and collections.
+
+MongoDB organizes data in a hierarchical structure. A MongoDB
+deployment contains one or more **databases**, and each database
+contains one or more **collections**. In each collection, MongoDB stores
+data as **documents** that contain field-and-value pairs. In Laravel,
+documents are mapped to Eloquent models.
```
ccho-mongodb: Suggestion:

I think "are mapped" might be ambiguous and potentially be interpreted as something that is performed actively rather than on-demand. 
I think it could be helpful to specify "Laravel MongoDB" rather than "Laravel" since I don't think this is possible without the extension.
Maybe one of the following statements would work better:

"In {+odm-short+}, documents can be accessed through Eloquent models."
"In {+odm-short+}, documents are represented by Eloquent models."

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,198 @@
...
+
+To learn more about the document data format,
+see :manual:`Documents </core/document/>` in the Server manual.
+
+.. _laravel-access-db:
+
+Access a Database
+-----------------
+
+You can access a database by specifying a database connection in your
```
ccho-mongodb: Suggestion:

I think "you can access a database" could be more often interpreted as accessing the data whereas the information provided seems to be focused on specifying the database value in one or more connection configurations.

Maybe "Specify the Database in a Connection Configuration" would be a more accurate description of what we want to communicate to a reader. This could be one way to omit any mention/examples of the connection-related items like "dsn" and "default" which might be more appropriate in the Connection Guide.

I appreciate the thoroughness of the explanation though!

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,198 @@
...
+   overhead in using two connections for two distinct databases, so you
+   do not need to optimize your connections.
+
+If your application contains multiple
+database connections, you override the ``$connection``
+property when creating a ``Model`` class to apply the model to an
+alternate database.
+
+The following example shows how to override the ``$connection`` property
+on the ``Flower`` model class to use the ``mongodb_alt`` connection:
```
ccho-mongodb: Suggestion:

While I think many readers can draw the connection between Flower and "plants", I think it would be helpful to explicitly mention which database the data for this model would be stored in to make sure the reader can draw the connection of the effect.

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,198 @@
...
+
+   The MongoDB PHP driver reuses the same connection when
+   you create two clients with the same connection string. There is no
+   overhead in using two connections for two distinct databases, so you
+   do not need to optimize your connections.
+
+If your application contains multiple
+database connections, you override the ``$connection``
+property when creating a ``Model`` class to apply the model to an
+alternate database.
```
ccho-mongodb: Suggestions:

I think this could be clearer if both of the conditions ("multiple MongoDB connection configs" and "you want the class to apply the model to an alternate database") were listed at the beginning of the sentence.

The current structure is:
If X, do Z to meet the conditions of Y.

The suggested structure is:
If X and Y, do Z.

I think that in this case, X and Y can also be described in a combined manner:

"To make sure your model uses a specific MongoDB database configuration, override the $connection property, setting the value to the key name of the configuration in your Model class."

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,198 @@
...
+       protected $connection = 'mongodb_alt';
+   }
+
+.. _laravel-access-coll:
+
+Access a Collection
+-------------------
+
+When you create a ``Model`` class in {+odm-short+}, it is automatically
+stored in a collection with a name that is the plural of the model class
```
ccho-mongodb: Suggestion:

I like the inclusion of the example, but I think the description could provide some more precise details:

```suggestion
When you create model class that extends ``MongoDB\Laravel\Eloquent\Model``, {+odm-short+} stores the model data in a collection, named by the snake case plural form of your model class.
```

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,198 @@
...
+name. For example, if you create a model class called ``Flower``,
+Laravel applies the model to the ``flowers`` collection in the database.
+
+.. tip::
+
+   You might be storing data in a collection with a different name than the
+   model, or your collection might contain more than one type of data. To
+   learn how to set a collection name when defining a model class, see the
+   :ref:`laravel-model-customize-collection-name` section of the Eloquent
+   Model Class guide.
```
ccho-mongodb: Suggestion:

I think users could have any number of reasons for changing the name, so it could be confusing to provide info of the conditional usage. I'm also not sure whether specifying a collection name solves the problem of "collection might contain more than one type of data" from the current info provided.

"To learn how to specify a different collection name, se the ..."

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,198 @@
...
+recommend that you use the Eloquent ORM to access a collection
+implicitly. The following example specifies a find operation by using
+the ``Flower`` class, so Laravel retrieves results from the ``flowers``
+collection:
+
+.. code-block:: php
+
+   Flower::where('name', 'Water Lily')->get()
+
+For other collection operations, you can access a collection by using
```
ccho-mongodb: Suggestion:

Since the prior recommendation was to use Eloquent models, I think this could be reworded to serve as the exception case.

"If you are not able to accomplish your operation by using the Eloquent models, you can use the DB facade to access the query builder..."

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,198 @@
...
+
+.. tip::
+
+   You might be storing data in a collection with a different name than the
+   model, or your collection might contain more than one type of data. To
+   learn how to set a collection name when defining a model class, see the
+   :ref:`laravel-model-customize-collection-name` section of the Eloquent
+   Model Class guide.
+
+For many operations, such as CRUD operations and relationships, we
```
ccho-mongodb: Suggestion:

I don't think "relationships" should be characterized as operation and the qualifiers ("CRUD operations and relationships") can be omitted. Instead, I think the focus could be on a generic recommendation (rather than trying to explain Eloquent ORM vs Query Builder):

"We recommend that you use Eloquent models to access collections for code readability and maintainability."

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,198 @@
...
+       ->collection('flowers')
+       ->where('name', 'Water Lily')
+       ->get()
+
+.. tip::
+   
+   In MongoDB, each document in a collection has a field called ``_id`` which has a value
+   that is a unique identifier. To learn more about customizing this field
+   when you create a model to access a collection, see the :ref:`laravel-model-customize-primary-key` field of the Eloquent
+   Model Class guide.
```
ccho-mongodb: Suggestion:

I like the idea of including as much related information as possible, but I think this admonition should be omitted from here for the following reasons:

- Keep the section focused on accessing a collection
- None of the examples show what fields are written
- None of the examples show retrieving the data by the primary key / _id field

docs/fundamentals/database-collection.txt
ccho-mongodb: Suggestion:
I think this sentence structure could be simplified and made more concise. E.g.

"The Laravel Schema facade offers the drop() and dropIfExists() methods to drop a collection."

Or, if you end up suggesting just one, maybe something like the following:

"You can drop a collection by calling the "drop()" method of the Laravel Schema facade."

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,198 @@
...
+
+.. code-block:: php
+
+   $collections = DB::connection('mongodb')->getMongoDB()->listCollections();
+   dump($collections);
+
+Drop a Collection
+-----------------
+
+We recommend performing a **migration** to handle dropping a collection.
```
ccho-mongodb: Suggestion 1:

I think it could be easier to find the information if this used the same structure as the previous section on List Collections. E.g.

"To [accomplish some action], call the ... method"

And then describe the recommendations.

Suggestion 2:

If recommending using a migration to drop a collection, I think creating the collection in a migration should also be described on this page since they're often paired together. In that section, you could mention they're automatically created when you save the first document in that collection, but maybe recommend creating the collection with the indexes that you want prior to using the model.
Another solution to this might be to omit this section and instead mention "To 
learn how to create and drop collections..." and link to "Perform Laravel Migrations".

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,198 @@
...
+Deleting data from within a migration provides a controlled approach
+that ensures consistency, version control, and reversibility in your
+application.
```
ccho-mongodb: Suggestion: 
I think this could be great info to add to the "Perform Laravel Migrations" section if you decide to replace this section with a link to there.

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,198 @@
...
+To see the names of the collections in a database, call the 
+``listCollections()`` method.
+
+The following example acceses a database connection, then
+calls the ``listCollections()`` method and outputs the collection names:
+
+.. code-block:: php
+
+   $collections = DB::connection('mongodb')->getMongoDB()->listCollections();
+   dump($collections);
```
ccho-mongodb: Suggestion:
I think that when presenting code examples that include methods to print something out in the success case, it could be helpful to provide some sample output.
Otherwise, it could be good to omit them as they don't necessarily contribute to operation.

ccho-mongodb (approved):
Overall, the content looks good. As mentioned in my comments, I think there could be some structural changes that make the info more focused and easier to locate, but would be ok if you choose not to take them.
I'll also leave it up to you whether you want me to take another look.

docs/fundamentals/database-collection.txt
jmikola: Correct. I see the `Blueprint::drop()` method calls `Collection::drop()`. That NOPs if the namespace doesn't exist.

docs/fundamentals/database-collection.txt
rustagir: Thanks for the information! Since i modified the multiple connections example to specify the default db in the connection strings, should I remove this note then?

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,176 @@
...
+To learn more about the document data format,
+see :manual:`Documents </core/document/>` in the Server manual.
+
+.. _laravel-access-db:
+
+Specify the Database in a Connection Configuration
+--------------------------------------------------
+
+You specify a database name when configuring a connection in your
+application's ``config/database.php`` file. The ``connections`` property
```
ccho-mongodb: Suggestion:

I think this sentence specifies a time ("when"), but I don't think it's connected to  any sequence of instructions we provide. I think the following sentence could avoid associating it with an order:

You can configure the name of the database that a connection uses in your application's config/database.php configuration file.

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,176 @@
...
+.. note::
+
+   The MongoDB PHP driver reuses the same connection when
+   you create two clients with the same connection string. There is no
+   overhead in using two connections for two distinct databases, so you
+   do not need to optimize your connections.
+
+If your application contains multiple database connections and you want
+to store your model in a database other than the default, override the
+``$connection`` property when creating a ``Model`` class.
```
ccho-mongodb: Suggestion:

I think it's only necessary to mention where to make the change.

```suggestion
``$connection`` property in your ``Model`` class.
```

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,176 @@
...
+   }
+
+.. _laravel-access-coll:
+
+Access a Collection
+-------------------
+
+When you create model class that extends
+``MongoDB\Laravel\Eloquent\Model``, {+odm-short+} stores the model data
+in a collection with a name that is the snake case plural form of your
```
ccho-mongodb: Suggestion:

No change necessary, but I just thought this wording might flow better:

```suggestion
in a collection with a name formatted as the snake case plural form of your
```

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,176 @@
...
+We generally recommend that you use the Eloquent ORM to access a collection
+implicitly for code readability and maintainability. The following
+example specifies a find operation by using the ``Flower`` class, so
+Laravel retrieves results from the ``flowers`` collection:
+
+.. code-block:: php
+
+   Flower::where('name', 'Water Lily')->get()
+
+If you are not able to accomplish your operation by using an Eloquent
```
ccho-mongodb: (correcting my prior suggestion)
```suggestion
If you are unable to accomplish your operation by using an Eloquent
```

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,176 @@
...
+model, you can access a collection by using a facade to access the query
+builder class ``DB`` and calling the ``collection()`` method. The
```
ccho-mongodb: Suggestion:

I think it can be more concise to use the "DB facade" nomenclature which the [Laravel docs use](https://laravel.com/docs/11.x/queries#running-database-queries).

```suggestion
model, you can access the query builder by calling the ``collection()`` method on the ``DB`` facade. The
```

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,176 @@
...
+Laravel applies the model to the ``flowers`` collection in the database.
+
+.. tip::
+
+   To learn how to specify a different collection name when defining a model class, see the
+   :ref:`laravel-model-customize-collection-name` section of the Eloquent
+   Model Class guide.
+
+We generally recommend that you use the Eloquent ORM to access a collection
+implicitly for code readability and maintainability. The following
```
ccho-mongodb: Suggestion:
I think there's only one way to reference collections by using Eloquent ORM (through the models), so it could be confusing to include "implicitly":
 
```suggestion
for code readability and maintainability. The following
```

ccho-mongodb (approved):
LGTM, left a couple small suggestions.

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,176 @@
...
+       ], ...
+   ]
+
+When you set a default database connection, {+odm-short+} uses that
+connection for operations, but you can specify multiple database connections
+in your ``config/database.php`` file.
+
+The following example shows how to specify multiple database connections
+(``mongodb`` and ``mongodb_alt``) to access the ``animals`` and
+``plants`` databases:
```
jmikola, line 77: @GromNaN: Do we really not have any alternative way to utilize multiple databases with this library? It's bizarre to me that separate DSNs would be required, and that we'd have to rely on the deprecated auth database component of connection string.

If users need to actually use a DSN with authentication, and omit `authSource`, they'll be in for a bad time.

In Doctrine, we have the ability to use a single connection and just specify different databases as needed for models.

This example is admittedly contrived, since "animals" and "plants" sound like they're used by the same application, and a single application should generally use one database (excluding perhaps something like extra logging).

docs/fundamentals/database-collection.txt
jmikola: I'm asking @GromNaN above, but I very much dislike the example being presented here. If there's any way we can change this to avoid using the "auth database" URI component, I'd like to see it.

I can confirm that this paragraph is no longer relevant given the current state of the PR, but the fact that it's now creating multiple database connections is very disagreeable to me. This doesn't feel like something we should be conveying so simply in docs that some users are bound to copy/paste as reference.

docs/fundamentals/database-collection.txt
GromNaN: Just re-read the class `Laravel\MongoDB\Connection`.

- When the `dsn` config is not set, it is built from the config `host` (string or array) with the auth database that can be read from the config **option** `database` (different from the `database` config in the config). https://github.com/mongodb/laravel-mongodb/blob/bd6aa40bdbd285d9318a20b7dff6b96270115e75/src/Connection.php#L252-L255
Check de examples: https://github.com/mongodb/laravel-mongodb/blob/bd6aa40bdbd285d9318a20b7dff6b96270115e75/tests/ConnectionTest.php#L50-L170
- The database name is read from the `database` config. If not set, the database name is extracted from the dsn, after the `/`. https://github.com/mongodb/laravel-mongodb/blob/bd6aa40bdbd285d9318a20b7dff6b96270115e75/src/Connection.php#L155-L163
- Any additionnal option for the client can be set in `options`: `username`, `password`, `database` for auth database, and all the [driver options](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBClient__construct/#parameters).

docs/fundamentals/database-collection.txt
GromNaN: The correct configuration should be to extract the database into a dedicated config:

```php
   'connections' => [

       'mongodb' => [
           'driver' => 'mongodb',
           'dsn' => 'mongodb://localhost:27017/',
           'database' => 'animals',
       ],

       'mongodb_alt' => [
           'driver' => 'mongodb',
           'dsn' => 'mongodb://localhost:27017/',
           'database' => 'plants',
       ]

   ], ...
```

docs/fundamentals/database-collection.txt
```diff
@@ -0,0 +1,176 @@
...
+.. code-block:: php
+   :emphasize-lines: 7
+   
+   'default' => 'mongodb',
+
+   'connections' => [
+
+       'mongodb' => [
+           'driver' => 'mongodb',
+           'dsn' => 'mongodb://localhost:27017/animals',
```
GromNaN:
```suggestion
           'dsn' => 'mongodb://localhost:27017/animals',
           'database' => 'animals',
```

GromNaN (changes requested):

docs/fundamentals/database-collection.txt
jmikola: @GromNaN: I can't make sense of everything in that comment, but the ["DSN takes precedence over host/port config" test case](https://github.com/mongodb/laravel-mongodb/blob/bd6aa40bdbd285d9318a20b7dff6b96270115e75/tests/ConnectionTest.php#L155) stood out to me. There, a `config.database` option appears to take precedence over the auth database (i.e. path component of the URI) in the `config.dsn` option.

Given that, it seems like there's absolutely no reason we need to specify the "default" database name in a URI, and risk interfering with connection persistence.

GromNaN (approved):

PR #2829, opened 2024-04-08
DOCSP-35951: write operations delete docs

ccho-mongodb:
JIRA:
https://jira.mongodb.org/browse/DOCSP-35951

Staging:
https://preview-mongodbcchomongodb.gatsbyjs.io/laravel/DOCSP-35951-write-operations-delete/fundamentals/write-operations/#delete-documents

Note:
This PR is stacked with all the write operations PRs and is the final section.

Relevant changes:

- [Content](https://github.com/mongodb/laravel-mongodb/pull/2829/files#diff-602d693c77dcd905e0f0c5575aa2b00379ad90872ce40d78e6507d06980e881fR487-R608)
- [Code](https://github.com/mongodb/laravel-mongodb/pull/2829/files#diff-18f5b9e11af53df68556bae083274083009dd56da8ecbbeec6f2b1b216678cedR334-R517)

### Checklist

### DIFF
--- a/docs/fundamentals/write-operations.txt
+++ b/docs/fundamentals/write-operations.txt
@@ -1,0 +1,118 @@
+.. _laravel-fundamentals-write-ops:
+
+================
+Write Operations
+================
+
+.. facet::
+   :name: genre
+   :values: tutorial
+
+.. meta::
+   :keywords: insert, insert one, update, update one, upsert, delete, delete many, code example, mass assignment, eloquent model
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+In this guide, you can learn how to use {+odm-short+} to perform
+**write operations** on your MongoDB collections. Write operations include
+inserting, updating, and deleting data based on criteria that you specify.
+
+This guide shows you how to perform the following tasks:
+
+- Insert Documents
+- Modify Documents
+- Delete Documents
+
+.. _laravel-fundamentals-insert-documents:
+
+Insert Documents
+----------------
+
+In this section, you can learn how to insert documents into MongoDB collections
+from your Laravel application by using the {+odm-long+}.
+
+When you insert the documents, make sure that the data does not violate any
+unique indexes on the collection. When inserting the first document of a
+collection or creating a new collection, MongoDB automatically creates a
+unique index on the ``_id`` field.
+
+For more ionformation creating indexes on MongoDB collections by using the
+Laravel schema builder, see the :ref:`laravel-eloquent-indexes` guide.
+
+This section uses the following example model class to demonstrate how to
+use Eloquent models to perform insert operations:
+
+.. literalinclude:: /includes/fundamentals/write-operations/Concert.php
+   :language: php
+   :dedent:
+   :caption: Concert.php
+
+.. tip::
+
+   The ``$fillable`` attribute lets you use Laravel mass assignment for insert
+   operations. To learn more about mass assignment, see :ref:`laravel-model-mass-assignment`.
+
+To learn more about Eloquent models in {+odm-short+} see the :ref:`laravel-eloquent-models`
+section.
+
+Insert a Document Examples
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The examples in this section show how to use the ``save()`` Eloquent method to
+insert an instance of a ``Concert`` model as a MongoDB document.
+
+When the ``save()`` method succeeds, the instance on which you called the
+method contains the model. If it fails, the instance is assigned a null value.
+
+The example code performs the following actions:
+
+- Creates a new instance of the ``Concert`` model
+- Assigns string values to the ``performer`` and ``venue`` field
+- Assigns a date to the ``performanceDate`` field by using the ``Carbon``
+  package.
+- Inserts the document by calling the ``save()`` method.
+
+.. literalinclude:: /includes/fundamentals/write-operations/WriteOperationsTest.php
+   :language: php
+   :dedent:
+   :start-after: begin model insert one
+   :end-before: end model insert one
+
+If you enable mass assignment by defining either the ``$fillable`` or
+``$guarded`` attributes, you can use the Eloquent model ``create()`` method
+to perform the insert in a single call as shown in the following example:
+
+.. literalinclude:: /includes/fundamentals/write-operations/WriteOperationsTest.php
+   :language: php
+   :dedent:
+   :start-after: begin model insert one mass assign
+   :end-before: end model insert one mass assign
+
+To learn more about the Carbon PHP API extension, see the
+`Carbon <https://github.com/briannesbitt/Carbon>`__ GitHub repository.
+
+Insert Multiple Documents Example
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The examples in this section show how to use the ``insert()`` Eloquent method
+to insert multiple instances of a ``Concert`` model as MongoDB documents.
+
+When the ``insert()`` method succeeds, the method returns the value ``1``.
+If it fails, the method throws an exception.
+
+The example code passes an array containing the data for multiple models
+to the ``insert()`` method:
+
+.. literalinclude:: /includes/fundamentals/write-operations/WriteOperationsTest.php
+   :language: php
+   :dedent:
+   :start-after: begin model insert many
+   :end-before: end model insert many
+
--- a/docs/fundamentals/write-operations.txt
+++ b/docs/fundamentals/write-operations.txt
@@ -326,9 +326,9 @@ Update Arrays in a Document
 In this section, you can see examples of the following operations that
 update array values in a MongoDB document:
 
-- Add values to an array
-- Remove values from an array
-- Update the value of an array element
+- :ref:`Add values to an array <laravel-modify-documents-add-array-values>`
+- :ref:`Remove values from an array <laravel-modify-documents-remove-array-values>`
+- :ref:`Update the value of an array element <laravel-modify-documents-update-array-values>`
 
 These examples modify the sample document created by the following insert
 operation:
@@ -346,7 +346,7 @@ Add Values to an Array Example
 
 This section shows how to use the ``push()`` method to add values to an array
 in a MongoDB document. You can pass one or more values to add and set the
-optional parameter ``unique`` to ``true`` to skip adding any values
+optional parameter ``unique`` to ``true`` to skip adding any duplicate values
 in the array. The following code example shows the structure of a ``push()``
 method call:
 
@@ -359,7 +359,7 @@ method call:
           [<values>], // array or single value to add
           unique: true); // whether to skip existing values
 
-The following example shows how to add the value ``baroque`` to
+The following example shows how to add the value ``"baroque"`` to
 the ``genres`` array field of a matching document. Click the
 :guilabel:`VIEW OUTPUT` button to see the updated document:
 
@@ -381,7 +381,7 @@ the ``genres`` array field of a matching document. Click the
         "genres": [
             "classical",
             "dance-pop",
-            "baroque"
+
         ],
         "updated_at": ...,
         "created_at": ...
@@ -402,7 +402,7 @@ from the array. The following code example shows the structure of a
    :copyable: false
 
    YourModel::where(<match criteria>)
-      ->push(
+      ->pull(
           <field name>,
           [<values>]); // array or single value to remove
 
--- a/docs/fundamentals/write-operations.txt
+++ b/docs/fundamentals/write-operations.txt
@@ -22,7 +22,7 @@ Overview
 
 In this guide, you can learn how to use {+odm-short+} to perform
 **write operations** on your MongoDB collections. Write operations include
-inserting, updating, and deleting data based on criteria that you specify.
+inserting, updating, and deleting data based on specified criteria.
 
 This guide shows you how to perform the following tasks:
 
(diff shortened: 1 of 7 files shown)


docs/fundamentals/write-operations.txt
```diff
@@ -0,0 +1,608 @@
...
+- :ref:`Delete one document <laravel-fundamentals-delete-one>`
+- :ref:`Delete multiple documents <laravel-fundamentals-delete-many>`
+
+To learn about the Laravel features available in {+odm-short+} that modify
+delete behavior, see the following sections:
+
+- :ref:`Soft delete <laravel-model-soft-delete>`, which lets you mark
+  documents as deleted instead of removing them from the database
+- :ref:`Pruning <laravel-model-pruning>`, which lets you define conditions
+  in which a document should be deleted automatically
```
rustagir:
```suggestion
- :ref:`Pruning <laravel-model-pruning>`, which lets you define conditions
  that mark documents for automatic deletion
```

docs/fundamentals/write-operations.txt
```diff
@@ -0,0 +1,608 @@
...
+
+.. _laravel-fundamentals-delete-one:
+
+Delete a Document Examples
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can delete one document in the following ways:
+
+- Call the  ``destroy()`` method on the model to delete a document by the
+  primary key.
```
rustagir:
```suggestion
- Call the  ``destroy()`` method on the model to delete a document by matching the
  primary key.
```

docs/fundamentals/write-operations.txt
```diff
@@ -0,0 +1,608 @@
...
+   :end-before: end model delete by id
+
+When the ``destroy()`` method succeeds, it returns the number of documents
+deleted.
+
+If the primary key value does not match any documents, the ``destroy()`` method
+returns returns  ``0``.
+
+The following example shows how to delete a document by calling ``delete()``
+on an instance of the model:
```
rustagir: S: separate the three examples into subsections or move into bullet points, but as of now stacking them is a bit confusing

docs/fundamentals/write-operations.txt
```diff
@@ -0,0 +1,608 @@
...
+
+If the ``where()`` method does not match any documents, the ``delete()`` method
+returns returns  ``0``.
+
+.. _laravel-fundamentals-delete-many:
+
+Delete Multiple Documents Examples
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can perform deletes on multiple documents in the following ways:
```
rustagir:
```suggestion
You can delete multiple documents in the following ways:
```

rustagir (changes requested):
a few small things!

docs/fundamentals/write-operations.txt
ccho-mongodb: Thanks for pointing out the unclear wording. Since the conditions don't mark documents directly, so maybe I'll use "define conditions that qualify a document for automatic deletion".

docs/fundamentals/write-operations.txt
ccho-mongodb: I think the existing sentence and suggestion might be omitting a detail necessary to understand what to pass it, so I'll figure out how to reword. Will also update the description in the corresponding section for deleting multiple docs.

docs/fundamentals/write-operations.txt
ccho-mongodb: I agree with this and had considered earlier on, but wanted to keep all examples at the same heading level. Both Insert and Update use the same formatting with the same intention.

I think once these sections (Insert, Modify, Delete) are split into individual pages similar to drivers docs are separate pages,  these can be their own sections.

In the meantime, I'll add literalinclude captions (including examples in the prior sections) to identify which method is being demonstrated.

docs/fundamentals/write-operations.txt
```diff
@@ -480,3 +479,133 @@ with ``"contemporary"`` in the ``genres`` array field. Click the
...
+
+.. _laravel-fundamentals-delete-one:
+
+Delete a Document Examples
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can delete one document in the following ways:
+
+- Call the ``destroy()`` method on the model and pass it a primary key to
+  delete the matching document.
```
rustagir: S: revise to avoid the phrasing "pass it"
Ie
```suggestion
- Call the ``destroy()`` method on the model while passing a primary key to
  delete the matching document.
```

docs/fundamentals/write-operations.txt
```diff
@@ -480,3 +479,133 @@ with ``"contemporary"`` in the ``genres`` array field. Click the
...
+
+.. _laravel-fundamentals-delete-many:
+
+Delete Multiple Documents Examples
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can delete multiple documents in the following ways:
+
+- Call the ``destroy()`` method on the model and pass it multiple primary
+  keys to delete the matching documents.
```
rustagir: S: same comment as above re: pass it

rustagir (approved):
lgtm with a few last comments!

docs/fundamentals/write-operations.txt
ccho-mongodb: Discussed on Slack. Replacing with a adverbial phrase to describe the first part of the sentence.

docs/fundamentals/write-operations.txt
```diff
@@ -480,3 +479,134 @@ with ``"contemporary"`` in the ``genres`` array field. Click the
...
+Delete a Document Examples
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can delete one document in the following ways:
+
+- Call the ``destroy()`` method on the model, passing it the primary key of
+  the document to be deleted.
+- Call the ``delete()`` method on an instance of the model.
+- Chain methods to retrieve and delete an instance of a model by calling the
+  ``delete()`` method.
```
GromNaN: `$model->delete()` should be the first in the list as it's probably the more used.

The alternate `Model::destroy($ids)` accepts a single id, a list of ids, or a list of model intances.

```suggestion
- Call the ``$model->delete()`` method on an instance of the model.
- Call the ``Model::destroy($ids)`` method on the model class, passing it the id of
  the document to be deleted.
- Chain methods to retrieve and delete an instance of a model by calling the
  ``delete()`` method.
```

docs/fundamentals/write-operations.txt
```diff
@@ -480,3 +479,134 @@ with ``"contemporary"`` in the ``genres`` array field. Click the
...
+.. _laravel-fundamentals-delete-many:
+
+Delete Multiple Documents Examples
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can delete multiple documents in the following ways:
+
+- Call the ``destroy()`` method on the model, passing the primary keys of
+  the documents to be deleted.
+  keys to delete the matching documents.
```
GromNaN: This sentence seems incomplete. `Model::destroy($ids)` accepts a list of ids, or a list of model instances.

Note that using this method with a large number of ids is unefficient as it sends a distinct command for each id. For performance it's better using `Model::whereIn('id', $ids)->delete()`

docs/includes/fundamentals/write-operations/WriteOperationsTest.php
```diff
@@ -329,4 +330,188 @@ public function testModelPositional(): void
...
+            [
+                'performer' => 'The Rolling Stones',
+                'venue' => 'Soldier Field',
+            ],
+        ];
+        Concert::insert($data);
+
+        // begin model delete one fluent
+        Concert::where('venue', 'Carnegie Hall')
+            ->orderBy('_id')
```
GromNaN: I don't get the point of ordering for deleting. In fact, it's ignored:
https://github.com/mongodb/laravel-mongodb/blob/95c5b4fac69c2bd5ea01367e62d14dd5278e7513/src/Query/Builder.php#L787-L814

docs/includes/fundamentals/write-operations/WriteOperationsTest.php
```diff
@@ -329,4 +330,188 @@ public function testModelPositional(): void
...
+        $concerts = Concert::all();
+
+        $ids = [];
+
+        foreach ($concerts as $concert) {
+            array_push($ids, $concert->id);
+        }
+
+        // begin model delete multiple by id
+        Concert::destroy($ids);
```
GromNaN, line 472: You should show the contents of `$ids` in this example.

docs/includes/fundamentals/write-operations/WriteOperationsTest.php
```diff
@@ -329,4 +330,188 @@ public function testModelPositional(): void
...
+        ];
+        Concert::insert($data);
+
+        $concerts = Concert::all();
+
+        $ids = [];
+
+        foreach ($concerts as $concert) {
+            array_push($ids, $concert->id);
+        }
```
GromNaN: Laravel has [`Collection::pluck()` method](https://laravel.com/docs/11.x/collections#method-pluck) to do that:

```suggestion
        $concerts = Concert::all()->pluck('_id')->all();
```

docs/includes/fundamentals/write-operations/WriteOperationsTest.php
```diff
@@ -329,4 +330,188 @@ public function testModelPositional(): void
...
+            ],
+        ];
+        Concert::insert($data);
+
+        $this->assertEquals(4, Concert::count());
+
+        $id = Concert::first()->id;
+
+        // begin model delete by id
+        Concert::destroy($id);
```
GromNaN, line 380: You should force the value of `_id` in the data an set `$id = '<value>` in the example.

docs/fundamentals/write-operations.txt
ccho-mongodb: Question:
Is the "$model->delete()" a common way to represent a placeholder for an Eloquent model for a Laravel developer audience? 

Otherwise, what would be a common way to explain calling on a method on an instance of a model? I could be wrong, but I suspect the current suggestion might be repetitive (the instance and method are shown, but it's still described as a method).

docs/fundamentals/write-operations.txt
ccho-mongodb: Good catch. I'll add an admonition where the example is currently placed to mention the performance issue.

docs/includes/fundamentals/write-operations/WriteOperationsTest.php
ccho-mongodb: I see, so if the where() query filter matched multiple documents, wouldn't it be important to specify the sort order such that it's known which one will surface when calling limit(1), since that's the one that will be deleted? Does that mean that you can't use indexes with sort orders for the query filter part of the delete operation?

If not possible, I should probably include that information.

docs/fundamentals/write-operations.txt
GromNaN: I believe that code is more explicit than words. This is the convention I use systematically on this project when I want to emphasize the scope of a method.

I've found examples of `Model::method` in [Eloquent doc](https://laravel.com/docs/11.x/eloquent), but not `$model->method`. See https://laravel.com/docs/11.x/eloquent#deleting-models

docs/includes/fundamentals/write-operations/WriteOperationsTest.php
GromNaN: That's currently not possible. That could be a new feature PHPORM-169

docs/includes/fundamentals/write-operations/WriteOperationsTest.php
```diff
@@ -329,4 +329,197 @@ public function testModelPositional(): void
...
+                'venue' => 'Soldier Field',
+            ],
+        ];
+        Concert::insert($data);
+
+        $concerts = Concert::all();
+
+        $ids = [];
+
+        $concerts = Concert::all()->pluck('_id')->all();
```
GromNaN: These variables are no longer used.
```suggestion
```

docs/fundamentals/write-operations.txt
ccho-mongodb: I'm fine with changing conventions, but is this something we could come back to and decide later? I think it might involve updating several of the currently published docs to make them use it consistently. If so, I can ticket that.

docs/fundamentals/write-operations.txt
```diff
@@ -480,3 +479,139 @@ with ``"contemporary"`` in the ``genres`` array field. Click the
...
 in the {+server-docs-name+}.
+
+.. _laravel-fundamentals-delete-documents:
+
+Delete Documents
+----------------
+
+In this section, you can learn how to delete documents from a MongoDB collection
+by using {+odm-short+}. Use delete operations to remove data from your MongoDB
+database permanently.
```
GromNaN: "permanently" is wrong when the `SoftDeletes` trait is used. But you mention it just below, so maybe that's okay.

GromNaN (approved):

docs/includes/fundamentals/write-operations/WriteOperationsTest.php
jmikola: > Does that mean that you can't use indexes with sort orders for the query filter part of the delete operation?

Note: from the server's perspective, there is no way to specify an order with a [delete](https://www.mongodb.com/docs/manual/reference/command/delete/) operation.

Indexes can certainly be used (and [hinted](https://www.mongodb.com/docs/manual/reference/command/delete/#specify-hint-for-delete-operations)). In the case of a [sparse index](https://www.mongodb.com/docs/manual/core/index-sparse/) that might actually have an impact on which document(s) get selected for the deletion.

docs/fundamentals/write-operations.txt
ccho-mongodb: Thanks for providing that perspective, I'll update the text to make that more clear!

PR #2833, opened 2024-04-08
DOCSP-35977: Find multiple usage example

norareidy:
Adds a usage example showing how to retrieve multiple documents.
JIRA - https://jira.mongodb.org/browse/DOCSP-35977
Staging -
https://preview-mongodbcchomongodb.gatsbyjs.io/laravel/DOCSP-35977-ccho/usage-examples/find/

### Checklist

### DIFF
--- a/docs/includes/usage-examples/FindManyTest.php
+++ b/docs/includes/usage-examples/FindManyTest.php
@@ -30,12 +30,11 @@ class FindManyTest extends TestCase
         ]);
 
         // begin-find
-        foreach (
-            Movie::where('runtime', '>', 900)
+        $cursor = Movie::where('runtime', '>', 900)
             ->orderBy('_id')
-            ->cursor() as $movie
-        ) {
-                echo $movie->toJson() . '<br>';
+            ->cursor();
+        foreach ($cursor as $movie) {
+            echo $movie->toJson() . '<br>';
         }
 
         // end-find
--- a/docs/includes/usage-examples/FindManyTest.php
+++ b/docs/includes/usage-examples/FindManyTest.php
@@ -20,9 +20,11 @@ class FindManyTest extends TestCase
         Movie::truncate();
 
         // begin-find
-        foreach (Movie::where('runtime', '>', 900)
+        foreach (
+            Movie::where('runtime', '>', 900)
             ->orderBy('_id')
-            ->cursor() as $movie) {
+            ->cursor() as $movie
+        ) {
                 echo $movie->toJson() . '<br>';
         }
 
--- a/docs/includes/usage-examples/FindManyTest.php
+++ b/docs/includes/usage-examples/FindManyTest.php
@@ -1,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Http\Controllers;
+
+use App\Models\Movie;
+use MongoDB\Laravel\Tests\TestCase;
+
+class FindManyTest extends TestCase
+{
+    /**
+     * @runInSeparateProcess
+     * @preserveGlobalState disabled
+     */
+    public function testFindMany(): void
+    {
+        require_once __DIR__ . '/Movie.php';
+
+        Movie::truncate();
+
+        // begin-find
+        $movies = Movie::where('runtime', '>', 800)
+            ->orderBy('_id')
+            ->get();
+
+        foreach ($movies as $m) {
+            echo $m->toJson() . '<br>';
+        }
+        // end-find
+
+        $this->assertInstanceOf(Movie::class, $movie);
+    }
+}
--- a/docs/includes/usage-examples/FindManyTest.php
+++ b/docs/includes/usage-examples/FindManyTest.php
@@ -39,4 +39,5 @@ class FindManyTest extends TestCase
         }
 
         // end-find
+    }
 }
--- a/docs/includes/usage-examples/FindManyTest.php
+++ b/docs/includes/usage-examples/FindManyTest.php
@@ -27,6 +27,7 @@ class FindManyTest extends TestCase
         foreach ($movies as $m) {
             echo $m->toJson() . '<br>';
         }
+
         // end-find
 
         $this->assertInstanceOf(Movie::class, $movie);
--- a/docs/includes/usage-examples/FindManyTest.php
+++ b/docs/includes/usage-examples/FindManyTest.php
@@ -39,6 +39,4 @@ class FindManyTest extends TestCase
         }
 
         // end-find
-        $this->expectOutputRegex('/^{"_id":"[a-z0-9]{24}","title":"Centennial","runtime":1256}\r\n{"_id":"[a-z0-9]{24}","title":"Baseball","runtime":1140}$/');
-    }
 }
--- a/docs/includes/usage-examples/FindManyTest.php
+++ b/docs/includes/usage-examples/FindManyTest.php
@@ -30,13 +30,11 @@ class FindManyTest extends TestCase
         ]);
 
         // begin-find
-        $cursor = Movie::where('runtime', '>', 900)
+        $movies = Movie::where('runtime', '>', 900)
             ->orderBy('_id')
-            ->cursor();
-        foreach ($cursor as $movie) {
-            echo $movie->toJson() . '<br>';
-        }
-
+            ->get();
         // end-find
+
+        $this->assertEquals(2, $movies->count());
     }
 }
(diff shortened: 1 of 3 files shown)


docs/usage-examples/find.txt
```diff
@@ -0,0 +1,107 @@
...
+      "plot": "The economic and cultural growth of Colorado spanning two centuries from the mid-1700s to the late-1970s.",
+      "genres": [
+          "Action",
+          "Adventure",
+          "Drama"
+      ],
+      "runtime": 1256,
+      ...
+      "title": "Centennial",
+      ...
```
ccho-mongodb: Suggestion:
I think showing the expanded "title", "runtime", and "_id" fields could make the output easier to read and would be sufficient to help the reader understand the query filter and sort order. I think the rest could be hidden with "..."

Perhaps you'd want matches that also aren't easy to mistake for 
another more popular title such as "Taken" since the "runtime" field could be a point of confusion.

Another solution could be to avoid using the "_id" sort and instead using the "runtime" value so that you could omit the "_id" field.

docs/usage-examples/find.txt
```diff
@@ -0,0 +1,107 @@
...
+      }
+
+For instructions on editing your Laravel application to run the usage example, see the
+:ref:`Usage Example landing page <laravel-usage-examples>`.
+
+.. tip::
+
+   To learn more about retrieving documents with {+odm-short+}, see the
+   :ref:`laravel-fundamentals-retrieve` guide
```
ccho-mongodb:
```suggestion
   :ref:`laravel-fundamentals-retrieve` guide.
```

docs/usage-examples/find.txt
```diff
@@ -0,0 +1,107 @@
...
+      "title": "Taken",
+      ...
+      }
+
+For instructions on editing your Laravel application to run the usage example, see the
+:ref:`Usage Example landing page <laravel-usage-examples>`.
+
+.. tip::
+
+   To learn more about retrieving documents with {+odm-short+}, see the
```
ccho-mongodb: I think "where()" is one of many methods you can chain to the model class. Maybe this sentence could be modified to indicate there are other ways of retrieving documents.

"To learn about other ways to retrieve documents..."

docs/usage-examples/find.txt
```diff
@@ -0,0 +1,107 @@
...
+
+Example
+-------
+
+This usage example performs the following actions:
+
+- Uses the ``Movie`` Eloquent model to represent the ``movies`` collection in the
+  ``sample_mflix`` database
+- Retrieves documents from the ``movies`` collection that match a query filter
+
```
ccho-mongodb, line 41: Suggestion:
The example shows code to print out the retrieved documents, so I think it should either be mentioned or the code should be omitted.

docs/usage-examples/find.txt
```diff
@@ -0,0 +1,107 @@
...
+          "Sci-Fi"
+      ],
+      "runtime": 877,
+      ...
+      "title": "Taken",
+      ...
+      }
+
+For instructions on editing your Laravel application to run the usage example, see the
+:ref:`Usage Example landing page <laravel-usage-examples>`.
```
ccho-mongodb: Suggestion:

Since the title of the page is "Usage Examples", I think it would be more compliant with the [style guide](https://www.mongodb.com/docs/meta/style-guide/style/links-and-cross-references/) if they matched.

```suggestion
:ref:`Usage Examples landing page <laravel-usage-examples>`.
```

docs/includes/usage-examples/FindManyTest.php
```diff
@@ -0,0 +1,35 @@
...
+            ->orderBy('_id')
+            ->get();
+
+        foreach ($movies as $m) {
+            echo $m->toJson() . '<br>';
+        }
+
+        // end-find
+
+        $this->assertInstanceOf(Movie::class, $movie);
```
ccho-mongodb: Suggestion:
I think $movie hasn't been declared, so I would omit this assertion, or place it in a loop similar to lines 27-29.

docs/usage-examples/find.txt
```diff
@@ -0,0 +1,107 @@
...
+   :depth: 1
+   :class: singlecol
+
+You can retrieve multiple documents from a collection by calling the ``where()`` method
+on an Eloquent model or a query builder.
+
+Pass a query filter to the ``where()`` method to retrieve documents that meet a
+set of criteria. MongoDB returns the matching documents according to their
+:term:`natural order` in the database or according to the sort order that you can specify
+by using the ``orderBy()`` method.
```
ccho-mongodb: Question:
Can you access the cursor by chaining the "cursor()" method like [this](https://laravel.com/docs/11.x/eloquent#cursors)? If that's the preferred method of retrieving multiple results, I think we should show that.

ccho-mongodb (approved):
LGTM, a few questions and suggestions!

docs/includes/usage-examples/FindManyTest.php
```diff
@@ -0,0 +1,43 @@
...
+        ]);
+
+        // begin-find
+        foreach (
+            Movie::where('runtime', '>', 900)
+            ->orderBy('_id')
+            ->cursor() as $movie
+        ) {
+                echo $movie->toJson() . '<br>';
+        }
```
GromNaN: We don't format the code like this usually. That's better to use a variable, and insist on the fact that we get a `Cursor` here.
```suggestion
        $cursor = Movie::where('runtime', '>', 900)
            ->orderBy('_id')
            ->cursor();
        foreach ($cursor as $movie) {
            echo $movie->toJson() . '<br>';
        }
```

docs/usage-examples/find.txt
```diff
@@ -0,0 +1,82 @@
...
+   :keywords: find many, retrieve, code example
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+You can retrieve multiple documents from a collection by calling the ``where()`` method
+on an Eloquent model or a query builder.
```
GromNaN: `where` doesn't retrieve the documents. It creates a query builder. That the call on `get()` or `cursor()` that runs the query and returns the results.

docs/usage-examples/find.txt
```diff
@@ -0,0 +1,82 @@
...
+
+- Uses the ``Movie`` Eloquent model to represent the ``movies`` collection in the
+  ``sample_mflix`` database
+- Retrieves and prints documents from the ``movies`` collection that match a query filter
+
+The example calls the following methods on the ``Movie`` model:
+
+- ``where()``: matches documents in which the value of the ``runtime`` field is greater than ``900``
+- ``orderBy()``: sorts matched documents by their ascending ``_id`` values
+- ``cursor()``: retrieves the query results by loading each ``Movie`` model into memory one at a time
```
GromNaN: I think most of the time people want to use `->get()` to get a Laravel Collection. It's not a problem if there are few results, and it offers more features than a cursor (count, first, last ...).

ccho-mongodb: Sorry, I had suggested using the [cursor](https://github.com/mongodb/laravel-mongodb/pull/2833/files#r1556598455) since that's how we usually demonstrate this. Will update to use get().

docs/usage-examples/find.txt
```diff
@@ -0,0 +1,82 @@
...
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+You can retrieve multiple documents from a collection by chaining a retrieval
+method such as ``where()`` with the ``get()`` method to retrieve the results.
+You can chain these methods to an Eloquent model or a query builder.
```
GromNaN: To give the reader more perspective, we can tell them that `Model::where()` creates a query builder. And that they can use all the other query builder methods, such as `orWhere`, `whereAll` ...
With a link to the [query builder page](https://github.com/mongodb/laravel-mongodb/blob/4.1/docs/query-builder.txt).

ccho-mongodb: Let me know what you think about [the new description I committed](https://github.com/mongodb/laravel-mongodb/pull/2833/commits/a4cd130549577a119eb1a1c9edf9ec6186ed4797)

If you're ok with this version, I'll create an item in our [cleanup ticket](https://jira.mongodb.org/browse/DOCSP-38344) to update the Usage Example descriptions.

docs/includes/usage-examples/FindManyTest.php
```diff
@@ -0,0 +1,44 @@
...
+                'title' => 'Centennial',
+                'runtime' => 1256,
+            ],
+            [
+                'title' => 'Baseball',
+                'runtime' => 1140,
+            ],
+            [
+                'title' => 'Basketball',
+                'runtime' => 600,
```
GromNaN, line 32: I've added an item with runtime <= 900 for the test to be complete.

GromNaN (approved):
LGTM.

PR #2808, opened 2024-04-01
DOCSP-35947: write operations modify documents

ccho-mongodb:
JIRA: https://jira.mongodb.org/browse/DOCSP-35947

Note for the reviewer:

This contains some updates to the [insert documents](https://preview-mongodbcchomongodb.gatsbyjs.io/laravel/DOCSP-35947-write-operations-modify-documents/fundamentals/write-operations/#insert-documents) section (added new fields to the sample doc).
This is one of three tickets to document the update operations. The other two are:
- [Upsert](https://jira.mongodb.org/browse/DOCSP-35949)
- [Update arrays](https://jira.mongodb.org/browse/DOCSP-35946)

Staging:
https://preview-mongodbcchomongodb.gatsbyjs.io/laravel/DOCSP-35947-write-operations-modify-documents/fundamentals/write-operations/#modify-documents

### Checklist

- [ ] Add tests and ensure they pass
- [ ] Add an entry to the CHANGELOG.md file
- [ ] Update documentation for new features

PR #2836, opened 2024-04-09
DOCSP-35975: Update many usage example

norareidy:
Adds a usage example showing how to update multiple documents
JIRA - https://jira.mongodb.org/browse/DOCSP-35975
Staging - https://preview-mongodbnorareidy.gatsbyjs.io/laravel/DOCSP-35975/usage-examples/updateMany/
### Checklist

### DIFF
--- a/docs/includes/usage-examples/UpdateManyTest.php
+++ b/docs/includes/usage-examples/UpdateManyTest.php
@@ -18,6 +18,22 @@ class UpdateManyTest extends TestCase
         require_once __DIR__ . '/Movie.php';
 
         Movie::truncate();
+        Movie::insert([
+            [
+                'title' => 'Hollywood',
+                'imdb' => [
+                    'rating' => 9.1,
+                    'votes' => 511,
+                ],
+            ],
+            [
+                'title' => 'The Shawshank Redemption',
+                'imdb' => [
+                    'rating' => 9.3,
+                    'votes' => 1513145,
+                ],
+            ],
+        ]);
 
         // begin-update-many
         $updates = Movie::where('imdb.rating', '>', 9.0)
@@ -26,7 +42,7 @@ class UpdateManyTest extends TestCase
         echo 'Updated documents: ' . $updates;
         // end-update-many
 
-        $this->assertTrue($updates);
-        $this->expectOutputString('Updated documents: 20');
+        $this->assertEquals(2, $updates);
+        $this->expectOutputString('Updated documents: 2');
     }
 }
--- a/docs/includes/usage-examples/UpdateManyTest.php
+++ b/docs/includes/usage-examples/UpdateManyTest.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Http\Controllers;
+
+use App\Models\Movie;
+use MongoDB\Laravel\Tests\TestCase;
+
+class UpdateManyTest extends TestCase
+{
+    /**
+     * @runInSeparateProcess
+     * @preserveGlobalState disabled
+     */
+    public function testUpdateMany(): void
+    {
+        require_once __DIR__ . '/Movie.php';
+
+        Movie::truncate();
+
+        // begin-update-many
+        $updates = Movie::where('imdb.rating', '>', 9.0)
+            ->update(['$set' => ['acclaimed' => true]]);
+
+        echo 'Updated documents: ' . $updates;
+        // end-update-many
+
+        $this->assertTrue($updates);
+        $this->expectOutputString('Updated documents: 20');
+    }
+}
--- a/docs/includes/usage-examples/UpdateManyTest.php
+++ b/docs/includes/usage-examples/UpdateManyTest.php
@@ -37,7 +37,7 @@ class UpdateManyTest extends TestCase
 
         // begin-update-many
         $updates = Movie::where('imdb.rating', '>', 9.0)
-            ->update(['$set' => ['acclaimed' => true]]);
+            ->update(['acclaimed' => true]);
 
         echo 'Updated documents: ' . $updates;
         // end-update-many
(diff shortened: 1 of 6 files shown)


docs/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,66 @@
...
+   :keywords: update many, modify, code example
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+You can update multiple documents in a collection by calling the ``update()`` method
+on an Eloquent model or query builder.
```
ccho-mongodb: Question:
Is it possible to call on the model directly? I could be wrong, but I think you need to chain it to an object collection like the result of "where()" for Laravel MongoDB.

I think this description should match what's shown in the example either way.

docs/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,66 @@
...
+document changes to the ``update()`` method.
+
+Example
+-------
+
+This usage example performs the following actions:
+
+- Uses the ``Movie`` Eloquent model to represent the ``movies`` collection in the
+  ``sample_mflix`` database
+- Updates documents from the ``movies`` collection that match a query filter
```
ccho-mongodb, line 34: Similar to [prior feedback](https://github.com/mongodb/laravel-mongodb/pull/2833#discussion_r1556575810), I think printing the output should be mentioned or the code that prints should be hidden.

docs/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,66 @@
...
+      Updated documents: 20
+
+To learn how to edit your Laravel application to run the usage example, see the
+:ref:`Usage Examples landing page <laravel-usage-examples>`.
+
+.. tip::
+
+   To learn more about updating data with {+odm-short+}, see the `Updates
+   <https://laravel.com/docs/{+laravel-docs-version+}/eloquent#updates>`__ section of the
+   Laravel documentation.
```
ccho-mongodb: Question:
Do you know if all the content in this section is available in Laravel MongoDB? For example, do the methods listed in the "Examining Attribute Changes" section work?

If these do not work, perhaps this link should be omitted.

docs/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,66 @@
...
+- Uses the ``Movie`` Eloquent model to represent the ``movies`` collection in the
+  ``sample_mflix`` database
+- Updates documents from the ``movies`` collection that match a query filter
+
+The example calls the following methods on the ``Movie`` model:
+
+- ``where()``: matches documents in which the value of the ``imdb.rating`` nested field
+  is greater than ``9``
+- ``update()``: updates the matching documents by adding an ``acclaimed`` field and setting
+  its value to ``true``
```
ccho-mongodb: Suggestion:
I think it could be helpful to add some information about the value printed even if it seems self-explanatory. E.g.
"The update() method returns the number of documents successfully updated"

ccho-mongodb (approved):
LGTM, a couple questions and suggestions!

docs/usage-examples/updateMany.txt
norareidy: Ah yes, it needs to be on an object collection - fixed

docs/usage-examples/updateMany.txt
norareidy: They don't seem to work, so I'll remove

docs/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,67 @@
...
+      :visible: false
+
+      Updated documents: 20
+
+To learn how to edit your Laravel application to run the usage example, see the
+:ref:`Usage Examples landing page <laravel-usage-examples>`.
+
+.. tip::
+
+   To learn more about updating data with {+odm-short+}, see the :ref:`laravel-fundamentals-modify-documents`
```
norareidy, line 65: Note: since this links to a section of the Write operations page, this PR will be merged after #2808

docs/includes/usage-examples/UpdateManyTest.php
```diff
@@ -0,0 +1,48 @@
...
+                'imdb' => [
+                    'rating' => 9.3,
+                    'votes' => 1513145,
+                ],
+            ],
+        ]);
+
+        // begin-update-many
+        $updates = Movie::where('imdb.rating', '>', 9.0)
+            ->update(['$set' => ['acclaimed' => true]]);
```
GromNaN: `$set` is optional, you can pass the values directly for compatibility with SQL Eloquent.

```suggestion
            ->update(['acclaimed' => true]);
```

The operator `$set` is [added automatically](https://github.com/mongodb/laravel-mongodb/blob/95c5b4fac69c2bd5ea01367e62d14dd5278e7513/src/Query/Builder.php#L711-L721) to the command when the values doesn't contain any $-prefixed name. Example: https://github.com/mongodb/laravel-mongodb/blob/95c5b4fac69c2bd5ea01367e62d14dd5278e7513/tests/QueryTest.php#L578

docs/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,67 @@
...
+   :keywords: update many, modify, code example
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+You can update multiple documents in a collection by calling the ``update()`` method
+on an object collection or query builder.
```
GromNaN: There is no `update()` method on Laravel Collections: https://laravel.com/api/11.x/Illuminate/Support/Collection.html

Update can be called on a single Model instance (updateOne doc) or on a query builder. I think there is only one way to update many documents, it's using the query builder. Otherwise you have to loop on the collection of model instance to update one-by-one.

GromNaN (changes requested):

docs/includes/usage-examples/UpdateManyTest.php
norareidy: Good to know, thanks!

GromNaN (approved):
LGTM

PR #2838, opened 2024-04-09
PHPORM-167 Fix and test MongoDB failed queue provider

GromNaN:
Fix PHPORM-167

In Laravel 11.35, the `DatabaseFailedJobProvider::ids` was added in order to optimize loading this ids only instead of the full documents. https://github.com/laravel/framework/pull/49186

This implementation is not compatible with MongoDB since it references the field `id` instead of `_id`.

Also, the method [`DatabaseFailedJobProvider::prune`](https://github.com/laravel/framework/blob/d63d2b1f50b6bbf8beb88b7b4ad5a4d230244443/src/Illuminate/Queue/Failed/DatabaseFailedJobProvider.php#L140) is not compatible with MongoDB because it tries to delete document by bulks of 1000, which is not supported. MongoDB can only delete one or all documents matching a criteria.

I'm also adding tests to the methods inherited from the parent `DatabaseFailedJobProvider` to ensure they work with a MongoDB connection.

### Checklist

### DIFF
--- a/src/Queue/Failed/MongoFailedJobProvider.php
+++ b/src/Queue/Failed/MongoFailedJobProvider.php
@@ -90,7 +90,7 @@ class MongoFailedJobProvider extends DatabaseFailedJobProvider
      *
      * @param  string|null $queue
      *
-     * @return array
+     * @return list<mixed>
      */
     public function ids($queue = null)
     {
--- a/src/Queue/Failed/MongoFailedJobProvider.php
+++ b/src/Queue/Failed/MongoFailedJobProvider.php
@@ -8,6 +8,7 @@ use Carbon\Carbon;
 use DateTimeInterface;
 use Exception;
 use Illuminate\Queue\Failed\DatabaseFailedJobProvider;
+use MongoDB\BSON\ObjectId;
 use MongoDB\BSON\UTCDateTime;
 
 use function array_map;
@@ -86,11 +87,11 @@ class MongoFailedJobProvider extends DatabaseFailedJobProvider
     }
 
     /**
-     * Get the IDs of all of the failed jobs.
+     * Get the IDs of all the failed jobs.
      *
      * @param  string|null $queue
      *
-     * @return list<mixed>
+     * @return list<ObjectId>
      */
     public function ids($queue = null)
     {
@@ -102,7 +103,7 @@ class MongoFailedJobProvider extends DatabaseFailedJobProvider
     }
 
     /**
-     * Prune all entries older than the given date.
+     * Prune all failed jobs older than the given date.
      *
      * @param  DateTimeInterface $before
      *
--- a/src/Queue/Failed/MongoFailedJobProvider.php
+++ b/src/Queue/Failed/MongoFailedJobProvider.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace MongoDB\Laravel\Queue\Failed;
 
 use Carbon\Carbon;
+use DateTimeInterface;
 use Exception;
 use Illuminate\Queue\Failed\DatabaseFailedJobProvider;
 use MongoDB\BSON\UTCDateTime;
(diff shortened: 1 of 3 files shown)


src/Queue/Failed/MongoFailedJobProvider.php
```diff
@@ -75,12 +76,43 @@ public function find($id)
...
     {
         return $this->getTable()->where('_id', $id)->delete() > 0;
     }
+
+    /**
+     * Get the IDs of all of the failed jobs.
+     *
+     * @param  string|null $queue
+     *
+     * @return array
```
jmikola: Since there are no type requirements on the `_id` field, this is essentially a list of mixed values, correct? Any reason to use not to use the [list type](https://phpstan.org/writing-php-code/phpdoc-types#lists)?

src/Queue/Failed/MongoFailedJobProvider.php
```diff
@@ -75,12 +76,43 @@ public function find($id)
...
+     */
+    public function ids($queue = null)
+    {
+        return $this->getTable()
```
jmikola, line 98: Does `getTable()` need different typing information for static analysis?

I see [`DatabaseFailedJobProvider::getTable()`](https://laravel.com/api/11.x/Illuminate/Queue/Failed/DatabaseFailedJobProvider.html#method_getTable) returns [Illuminate\Database\Query\Builder](https://laravel.com/api/11.x/Illuminate/Database/Query/Builder.html). Is that a parent class shared by the MongoDB library's builder, and thus still statisfied?

src/Queue/Failed/MongoFailedJobProvider.php
GromNaN: Correct.
```suggestion
     * @return list<mixed>
```

src/Queue/Failed/MongoFailedJobProvider.php
GromNaN: We are using Eloquent methods here. Nothing specific to MongoDB. The type `Illuminate\Database\Query\Builder` from [`DatabaseFailedJobProvider::getTable()`](https://github.com/laravel/framework/blob/48246da2320c95a17bfae922d36264105a917906/src/Illuminate/Queue/Failed/DatabaseFailedJobProvider.php#L168-L171) is fine.

src/Queue/Failed/MongoFailedJobProvider.php
```diff
@@ -75,12 +76,43 @@ public function find($id)
...
+            ->when($queue !== null, static fn ($query) => $query->where('queue', $queue))
+            ->orderBy('_id', 'desc')
+            ->pluck('_id')
+            ->all();
+    }
+
+    /**
+     * Prune all entries older than the given date.
```
jmikola: Should this refer to "failed entries"?

tests/Queue/Failed/MongoFailedJobProviderTest.php
```diff
@@ -0,0 +1,150 @@
...
+{
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        DB::connection('mongodb')
+            ->collection('failed_jobs')
+            ->raw()
+            ->insertMany(array_map(static fn ($i) => [
+                '_id' => new ObjectId(sprintf('%024d', $i)),
```
jmikola, line 27: Are `_id` fields within jobs _always_ an ObjectId? I saw that MongoFailedJobProvider made some assumptions about that by using `string` for type hints and casting.

jmikola (approved):
Suggested comment change.

Also a question about `_id` types. If the assumption is correct, then just resolve my comment there.

tests/Queue/Failed/MongoFailedJobProviderTest.php
GromNaN: Documents are added by the `log` method which doesn't specify `_id`. So we get an automatic ObjectId.

PR #2868, opened 2024-04-15
DOCSP-37641: GridFS page

ccho-mongodb:
JIRA: https://jira.mongodb.org/browse/DOCSP-37641

Staging: https://preview-mongodbcchomongodb.gatsbyjs.io/laravel/DOCSP-37641-gridfs-page/fundamentals/gridfs/

Note:
I was unable to figure out the best way to test the code examples. I added methods that allowed me to check that the code worked as described, but differ by import aliasing and such.

### Checklist

### DIFF
--- a/docs/fundamentals/gridfs.txt
+++ b/docs/fundamentals/gridfs.txt
@@ -98,7 +98,6 @@ The following example shows how to select a bucket named ``grid`` on the
 ``MongoDB\GridFS\Bucket``:
 
 .. code-block:: php
-   :dedent:
 
    $manager = new MongoDB\Driver\Manager(/* connection string */);
    $bucket = new MongoDB\GridFS\Bucket(
@@ -115,7 +114,6 @@ The following example shows how to select a bucket named ``grid`` on the
 ``selectGridFSBucket()`` method:
 
 .. code-block:: php
-   :dedent:
 
    $db = (new MongoDB\Client(/* connection string */))->filesDb;
    $bucket = $db->selectGridFSBucket([
@@ -148,7 +146,6 @@ The following example shows how to open an upload stream and write to it
 directly:
 
 .. code-block:: php
-   :dedent:
 
    $stream = $bucket->openUploadStream('new-file.txt');
    $contents = file_get_contents('/path/to/source-file.txt');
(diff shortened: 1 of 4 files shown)


GromNaN (commented):
This doc doesn't tell how to use gridfs in a Laravel application. Since we don't have created a facade or a flysystem bridge, there is no easy way to configure the bucket and access it in the application.

ccho-mongodb:
> This doc doesn't tell how to use gridfs in a Laravel application. Since we don't have created a facade or a flysystem bridge, there is no easy way to configure the bucket and access it in the application.

@GromNaN thanks for letting me know.
I created the page since Rishabh mentioned we needed it (see the JIRA ticket for details). If there is a way that you recommend that we configure it for a Laravel application, can you let me know so I can add that content? Otherwise, can you ask Rishabh to close the ticket?

docs/fundamentals/gridfs.txt
```diff
@@ -0,0 +1,194 @@
...
+Select a GridFS Bucket
+----------------------
+
+To store or retrieve files from GridFS, select a new or existing bucket on
+a MongoDB database. MongoDB creates a bucket only when you upload the first
+file to it.
+
+You can select a GridFS bucket by using one of the following methods:
+
+- Instantiate a ``MongoDB\GridFS\Bucket`` by using the PHP extension
```
jmikola, line 86: The Bucket exists in the library, not the extension.

Also, I don't think we should suggest manual construction since users would need to provide a `MongoDB\Driver\Manager` instance (see: [docs](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBGridFSBucket__construct/)).

The examples you have below where you're constructing a Manager or Client directly should not be demonstrated, as it's bypassing the Laravel library and its configuration entirely. There should be a single method in Laravel to obtain the underlying Client object, and that's the only thing we should be presenting. @GromNaN should be able to advise about that.

docs/includes/fundamentals/gridfs/GridFSTest.php
```diff
@@ -0,0 +1,132 @@
...
+use MongoDB\GridFS\Bucket;
+use MongoDB\Laravel\Tests\TestCase;
+
+use function fclose;
+use function file_get_contents;
+use function fopen;
+use function fwrite;
+use function stream_get_contents;
+
+class GridFSTest extends TestCase
```
jmikola, line 19: I don't see any reference to this file from the docs, so why is this being introduced?

The lines constructing Manager and Client instances don't use the test suites connection string, so I expect this may not run in all CI environments.

ccho-mongodb:
Thanks everyone for the input! Closing this PR for now until there's a better way to access GridFS from Laravel.

PR #2877, opened 2024-04-16
PHPORM-99 Add optimized cache and lock drivers

### LINKED ISSUE #2609: Can't store cache lock
- Laravel-mongodb Version: #.#.#
- PHP Version: #.#.#
- Database Driver & Version:

### Description:

I have issue when store data to cache lock with MongoDB driver. error also occurs when dispatch a job implements ShouldBeUnique.

Method acquire Illuminate\Cache\DatabaseLock catch QueryException to update record cache lock table. But with mongodb driver connection throw MongoDB\Driver\Exception\BulkWriteException so can not excute snippet code in catch block

Illuminate\Cache\DatabaseLock

```
public function acquire()
    {
        $acquired = false;

        try {
            $this->connection->table($this->table)->insert([
                'key' => $this->name,
                'owner' => $this->owner,
                'expiration' => $this->expiresAt(),
            ]);

            $acquired = true;
        } catch (QueryException $e) {
            $updated = $this->connection->table($this->table)
                ->where('key', $this->name)
                ->where(function ($query) {
                    return $query->where('owner', $this->owner)->orWhere('expiration', '<=', time());
                })->update([
                    'owner' => $this->owner,
                    'expiration' => $this->expiresAt(),
                ]);

            $acquired = $updated >= 1;
        }

        if (random_int(1, $this->lottery[1]) <= $this->lottery[0]) {
            $this->connection->table($this->table)->where('expiration', '<=', time())->delete();
        }

        return $acquired;
    }
```

### Steps to reproduce
$lock = Cache::lock('foo', 10);
$lock->block(5);

Job:
```
class ExampleJob implements ShouldQueue, ShouldBeUnique {
        use Dispatchable, InteractsWithQueue, Queueable, SerializesModels, UsesBackoffStrategy;

        public RequestQueue $request;
        public $tries = 2;

        /**
         * The number of seconds the job can run before timing out.
         *
         * @var int
         */
        public $timeout = 600;

        /**
         * Indicate if the job should be marked as failed on timeout.
         *
         * @var bool
         */
        public $failOnTimeout = true;

        /**
         * Create a new job instance.
         *
         * @return void
         */
        public function __construct() {
            //
        }

        /**
         * Execute the job.
         *
         * @return void
         */
        public function handle() {
       }
```
dispatch job:
`ExampleJob ::dispatch(); `

Exception message like:

`local.ERROR: E11000 duplicate key error collection: mes.cache_locks index: key_1 dup key: { key: "_cachefoo" } {"userId":"admin","exception":"[object] (MongoDB\\Driver\\Exception\\BulkWriteException(code: 11000): E11000 duplicate key error collection: mes.cache_locks index: key_1 dup key: { key: \"_cachefoo\" } at /var/www/vendor/mongodb/mongodb/src/Operation/InsertMany.php:157)`

</details>

Could you suggest any ways solve this issue ?

### LINKED ISSUE #2718: Need to support upsert function in Laravel 10
- PHP Version: 8.3
- Database Driver & Version: 4

### Description:
The implementation of Cache's put function is modified to db's upsert function in Laravel10.
image

### Steps to reproduce
When the driver of cache is changed to mongodb，This will lead to an error. Such as this:
```php
 \Cache::store('mongdb-conn')->forever($key, $val);

// or 
 \Cache::store('mongdb-conn')->put($key, $val);

```

The error is: This database engine does not support upserts.

GromNaN:
Fix fo PHPORM-99

In theory, we can use [`DatabaseStore`](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Cache/DatabaseStore.php) and [`DatabaseLock`](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Cache/DatabaseLock.php). But various issues prove the changing nature of their implementation, based on new features in the query builder, make this feature unstable for MongoDB users. fix #2718, fix #2609 

By introducing dedicated drivers, we can optimize the implementation to use the mongodb library directly instead of the subset of features provided by laravel query builder.

Usage:

```php
# config/cache.php
return [

    'stores' => [

        'mongodb' => [
            'driver' => 'mongodb',
            'connection' => 'mongodb',
            'collection' => 'cache',
            'lock_connection' => 'mongodb',
            'lock_collection' => 'cache_locks',
            'lock_lottery' => [2, 100],
            'lock_timeout' => '86400',
        ]
    ]

]
```

Cache:
```php
// Store any value into the cache. The value is serialized in MongoDB
Cache::set('foo', [1, 2, 3]);

// Read the value
dump(Cache::get('foo'));

// Clear the cache
Cache::flush();

```

Lock:

```php

// Get an unique lock. It's very important to keep this object in memory
// so that the lock can be released.
$lock = Cache::lock('foo');
$lock->block(10); // Wait 10 seconds before throwing an exception if the lock isn't released

// Any time-consuming task
sleep(5);

// Release the lock
$lock->release();
```

### Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,10 +25,11 @@
         "php": "^8.1",
         "ext-mongodb": "^1.15",
         "composer-runtime-api": "^2.0.0",
-        "illuminate/support": "^10.0|^11",
+        "illuminate/cache": "^10.36|^11",
         "illuminate/container": "^10.0|^11",
         "illuminate/database": "^10.30|^11",
         "illuminate/events": "^10.0|^11",
+        "illuminate/support": "^10.0|^11",
         "mongodb/mongodb": "^1.15"
     },
     "require-dev": {
--- a/src/Cache/MongoLock.php
+++ b/src/Cache/MongoLock.php
@@ -1,0 +1,134 @@
+<?php
+
+namespace MongoDB\Laravel\Cache;
+
+use Illuminate\Cache\Lock;
+use MongoDB\Laravel\Collection;
+use MongoDB\Operation\FindOneAndUpdate;
+use Override;
+
+use function random_int;
+
+final class MongoLock extends Lock
+{
+    /**
+     * Create a new lock instance.
+     *
+     * @param Collection  $collection              The MongoDB collection
+     * @param string      $name                    Name of the lock
+     * @param int         $seconds                 Time-to-live of the lock in seconds
+     * @param string|null $owner                   A unique string that identifies the owner. Random if not set
+     * @param array       $lottery                 The prune probability odds
+     * @param int         $defaultTimeoutInSeconds The default number of seconds that a lock should be held
+     */
+    public function __construct(
+        private readonly Collection $collection,
+        string $name,
+        int $seconds,
+        ?string $owner = null,
+        private readonly array $lottery = [2, 100],
+        private readonly int $defaultTimeoutInSeconds = 86400,
+    ) {
+        parent::__construct($name, $seconds, $owner);
+    }
+
+    /**
+     * Attempt to acquire the lock.
+     */
+    public function acquire(): bool
+    {
+        // The lock can be acquired if: it doesn't exist, it has expired,
+        // or it is already owned by the same lock instance.
+        $isExpiredOrAlreadyOwned = [
+            '$or' => [
+                ['$lte' => ['$expiration', $this->currentTime()]],
+                ['$eq' => ['$owner', $this->owner]],
+            ],
+        ];
+        $result = $this->collection->findOneAndUpdate(
+            ['_id' => $this->name],
+            [
+                [
+                    '$set' => [
+                        'owner' => [
+                            '$cond' => [
+                                'if' => $isExpiredOrAlreadyOwned,
+                                'then' => $this->owner,
+                                'else' => '$owner',
+                            ],
+                        ],
+                        'expiration' => [
+                            '$cond' => [
+                                'if' => $isExpiredOrAlreadyOwned,
+                                'then' => $this->expiresAt(),
+                                'else' => '$expiration',
+                            ],
+                        ],
+                    ],
+                ],
+            ],
+            [
+                'upsert' => true,
+                'returnDocument' => FindOneAndUpdate::RETURN_DOCUMENT_AFTER,
+                'projection' => ['owner' => 1],
+            ],
+        );
+
+        if (random_int(1, $this->lottery[1]) <= $this->lottery[0]) {
+            $this->collection->deleteMany(['expiration' => ['$lte' => $this->currentTime()]]);
+        }
+
+        return $result['owner'] === $this->owner;
+    }
+
+    /**
+     * Release the lock.
+     */
+    #[Override]
+    public function release(): bool
+    {
+        $result = $this->collection
+            ->deleteOne([
+                '_id' => $this->name,
+                'owner' => $this->owner,
+            ]);
+
+        return $result->getDeletedCount() > 0;
+    }
+
+    /**
+     * Releases this lock in disregard of ownership.
+     */
+    #[Override]
+    public function forceRelease(): void
+    {
+        $this->collection->deleteOne([
+            '_id' => $this->name,
+        ]);
+    }
+
+    /**
+     * Returns the owner value written into the driver for this lock.
+     */
+    #[Override]
+    protected function getCurrentOwner(): ?string
+    {
+        return $this->collection->findOne(
+            [
+                '_id' => $this->name,
+                'expiration' => ['$gte' => $this->currentTime()],
+            ],
+            ['projection' => ['owner' => 1]],
+        )['owner'] ?? null;
+    }
+
+    /**
+     * Get the UNIX timestamp indicating when the lock should expire.
+     */
+    private function expiresAt(): int
+    {
+        $lockTimeout = $this->seconds > 0 ? $this->seconds : $this->defaultTimeoutInSeconds;
+
+        return $this->currentTime() + $lockTimeout;
+    }
+}
--- a/src/Cache/MongoStore.php
+++ b/src/Cache/MongoStore.php
@@ -1,0 +1,296 @@
+<?php
+
+namespace MongoDB\Laravel\Cache;
+
+use Illuminate\Cache\RetrievesMultipleKeys;
+use Illuminate\Contracts\Cache\LockProvider;
+use Illuminate\Contracts\Cache\Store;
+use Illuminate\Support\InteractsWithTime;
+use MongoDB\Laravel\Collection;
+use MongoDB\Laravel\Connection;
+use MongoDB\Operation\FindOneAndUpdate;
+use Override;
+
+use function is_float;
+use function is_int;
+use function is_string;
+use function serialize;
+use function str_contains;
+use function unserialize;
+
+final class MongoStore implements LockProvider, Store
+{
+    use InteractsWithTime;
+    // Provides "many" and "putMany" in a non-optimized way
+    use RetrievesMultipleKeys;
+
+    private const TEN_YEARS_IN_SECONDS = 315360000;
+
+    private Collection $collection;
+
+    /**
+     * @param Connection      $connection                  The MongoDB connection to use for the cache
+     * @param string          $collectionName              Name of the collection where cache items are stored
+     * @param string          $prefix                      Prefix for the name of cache items
+     * @param Connection|null $lockConnection              The MongoDB connection to use for the lock, if different from the cache connection
+     * @param string          $lockCollectionName          Name of the collection where locks are stored
+     * @param array{int, int} $lockLottery                 Probability [chance, total] of pruning expired cache items
+     * @param int             $defaultLockTimeoutInSeconds Time-to-live of the locks in seconds
+     */
+    public function __construct(
+        private readonly Connection $connection,
+        private readonly string $collectionName = 'cache',
+        private readonly string $prefix = '',
+        private readonly ?Connection $lockConnection = null,
+        private readonly string $lockCollectionName = 'cache_locks',
+        private readonly array $lockLottery = [2, 100],
+        private readonly int $defaultLockTimeoutInSeconds = 86400,
+    ) {
+        $this->collection = $this->connection->getCollection($this->collectionName);
+    }
+
+    /**
+     * Get a lock instance.
+     *
+     * @param string      $name
+     * @param int         $seconds
+     * @param string|null $owner
+     */
+    #[Override]
+    public function lock($name, $seconds = 0, $owner = null): MongoLock
+    {
+        return new MongoLock(
+            ($this->lockConnection ?? $this->connection)->getCollection($this->lockCollectionName),
+            $this->prefix . $name,
+            $seconds,
+            $owner,
+            $this->lockLottery,
+            $this->defaultLockTimeoutInSeconds,
+        );
+    }
+
+    /**
+     * Restore a lock instance using the owner identifier.
+     */
+    #[Override]
+    public function restoreLock($name, $owner): MongoLock
+    {
+        return $this->lock($name, 0, $owner);
+    }
+
+    /**
+     * Store an item in the cache for a given number of seconds.
+     *
+     * @param string $key
+     * @param mixed  $value
+     * @param int    $seconds
+     */
+    #[Override]
+    public function put($key, $value, $seconds): bool
+    {
+        $result = $this->collection->updateOne(
+            [
+                '_id' => $this->prefix . $key,
+            ],
+            [
+                '$set' => [
+                    'value' => $this->serialize($value),
+                    'expiration' => $this->currentTime() + $seconds,
+                ],
+            ],
+            [
+                'upsert' => true,
+
+            ],
+        );
+
+        return $result->getUpsertedCount() > 0 || $result->getModifiedCount() > 0;
+    }
+
+    /**
+     * Store an item in the cache if the key doesn't exist.
+     *
+     * @param string $key
+     * @param mixed  $value
+     * @param int    $seconds
+     */
+    public function add($key, $value, $seconds): bool
+    {
+        $result = $this->collection->updateOne(
+            [
+                '_id' => $this->prefix . $key,
+            ],
+            [
+                [
+                    '$set' => [
+                        'value' => [
+                            '$cond' => [
+                                'if' => ['$lte' => ['$expiration', $this->currentTime()]],
+                                'then' => $this->serialize($value),
+                                'else' => '$value',
+                            ],
+                        ],
+                        'expiration' => [
+                            '$cond' => [
+                                'if' => ['$lte' => ['$expiration', $this->currentTime()]],
+                                'then' => $this->currentTime() + $seconds,
+                                'else' => '$expiration',
+                            ],
+                        ],
+                    ],
+                ],
+            ],
+            ['upsert' => true],
+        );
+
+        return $result->getUpsertedCount() > 0 || $result->getModifiedCount() > 0;
+    }
+
+    /**
+     * Retrieve an item from the cache by key.
+     *
+     * @param string $key
+     */
+    #[Override]
+    public function get($key): mixed
+    {
+        $result = $this->collection->findOne(
+            ['_id' => $this->prefix . $key],
+            ['projection' => ['value' => 1, 'expiration' => 1]],
+        );
+
+        if (! $result) {
+            return null;
+        }
+
+        if ($result['expiration'] <= $this->currentTime()) {
+            $this->forgetIfExpired($key);
+
+            return null;
+        }
+
+        return $this->unserialize($result['value']);
+    }
+
+    /**
+     * Increment the value of an item in the cache.
+     *
+     * @param string    $key
+     * @param int|float $value
+     */
+    #[Override]
+    public function increment($key, $value = 1): int|float|false
+    {
+        $this->forgetIfExpired($key);
+
+        $result = $this->collection->findOneAndUpdate(
+            [
+                '_id' => $this->prefix . $key,
+                'expiration' => ['$gte' => $this->currentTime()],
+            ],
+            [
+                '$inc' => ['value' => $value],
+            ],
+            [
+                'returnDocument' => FindOneAndUpdate::RETURN_DOCUMENT_AFTER,
+            ],
+        );
+
+        if (! $result) {
+            return false;
+        }
+
+        if ($result['expiration'] <= $this->currentTime()) {
+            $this->forgetIfExpired($key);
+
+            return false;
+        }
+
+        return $result['value'];
+    }
+
+    /**
+     * Decrement the value of an item in the cache.
+     *
+     * @param string    $key
+     * @param int|float $value
+     */
+    #[Override]
+    public function decrement($key, $value = 1): int|float|false
+    {
+        return $this->increment($key, -1 * $value);
+    }
+
+    /**
+     * Store an item in the cache indefinitely.
+     *
+     * @param string $key
+     * @param mixed  $value
+     */
+    #[Override]
+    public function forever($key, $value): bool
+    {
+        return $this->put($key, $value, self::TEN_YEARS_IN_SECONDS);
+    }
+
+    /**
+     * Remove an item from the cache.
+     *
+     * @param string $key
+     */
+    #[Override]
+    public function forget($key): bool
+    {
+        $result = $this->collection->deleteOne([
+            '_id' => $this->prefix . $key,
+        ]);
+
+        return $result->getDeletedCount() > 0;
+    }
+
+    /**
+     * Remove an item from the cache if it is expired.
+     *
+     * @param string $key
+     */
+    public function forgetIfExpired($key): bool
+    {
+        $result = $this->collection->deleteOne([
+            '_id' => $this->prefix . $key,
+            'expiration' => ['$lte' => $this->currentTime()],
+        ]);
+
+        return $result->getDeletedCount() > 0;
+    }
+
+    public function flush(): bool
+    {
+        $this->collection->deleteMany([]);
+
+        return true;
+    }
+
+    public function getPrefix(): string
+    {
+        return $this->prefix;
+    }
+
+    private function serialize($value): string|int|float
+    {
+        // Don't serialize numbers, so they can be incremented
+        if (is_int($value) || is_float($value)) {
+            return $value;
+        }
+
+        return serialize($value);
+    }
+
+    private function unserialize($value): mixed
+    {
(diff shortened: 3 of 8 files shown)


src/Cache/MongoLock.php
```diff
@@ -0,0 +1,145 @@
...
+                'upsert' => true,
+                'returnDocument' => FindOneAndUpdate::RETURN_DOCUMENT_AFTER,
+            ],
+        );
+
+        if (random_int(1, $this->lottery[1]) <= $this->lottery[0]) {
+            $this->collection->deleteMany(['expiration' => ['$lte' => $this->currentTime()]]);
+        }
+
+        return $result->owner === $this->owner;
```
GromNaN: Here I'm using `findOneAndUpdate` and a comparison of the value to return `true` when the lock is acquired twice by the same instance during the same second. If I use `updateOne` and check `$result->getUpsertedCount() > 0 || $result->getModifiedCount() > 0;`, the result is `false`.

src/MongoDBServiceProvider.php
```diff
@@ -40,5 +45,28 @@ public function register()
...
         });
+
+        // Add cache store.
+        $this->app->resolving('cache', function (CacheManager $cache) {
+            $cache->extend('mongodb', function (Application $app, array $config): Repository {
+                $connection = $app['db']->connection($config['connection'] ?? null);
+
+                $store = new MongoStore(
+                    $connection,
+                    $config['collection'] ?? $config['table'] ?? throw new InvalidArgumentException('Missing "collection" name for MongoDB cache'),
```
GromNaN: Both `collection` and `table` configs are accepted for compatibility with the `database` config. But I may remove the `table` support to be cleaner.

This needs to be consistent with `lock_table` below.

src/Query/Builder.php
```diff
@@ -659,6 +660,26 @@ public function whereBetween($column, iterable $values, $boolean = 'and', $not =
...
+    }
+
+    /** @inheritdoc */
+    public function insertOrIgnore(array $values)
+    {
+        try {
+            return $this->performInsert($values, ['ordered' => false]);
+        } catch (BulkWriteException $exception) {
+            // Ignore "duplicate key error"
+            if ($exception->getCode() !== 11000) {
```
GromNaN: I'm not sure if I should ignore only "duplicate key errors". [Laravel docs](https://laravel.com/docs/11.x/queries#insert-statements)
> The `insertOrIgnore` method will ignore errors while inserting records into the database. When using this method, you should be aware that duplicate record errors will be ignored and other types of errors may also be ignored depending on the database engine.

src/Query/Builder.php
```diff
@@ -659,6 +660,26 @@ public function whereBetween($column, iterable $values, $boolean = 'and', $not =
 
     /** @inheritdoc */
     public function insert(array $values)
+    {
+        return $this->performInsert($values);
+    }
+
+    /** @inheritdoc */
+    public function insertOrIgnore(array $values)
```
GromNaN: I'll open a distinct PR for this method that is no longer required. It was used in [`DatabaseStore::add`](https://github.com/laravel/framework/blob/ed718c7c7de17cad775e0eabbd441238fa7d32ec/src/Illuminate/Cache/DatabaseStore.php#L164), but I replaced the implementation for MongoDB.

src/Cache/MongoStore.php
```diff
@@ -0,0 +1,90 @@
...
+    public function put($key, $value, $seconds): bool
+    {
+        $key = $this->prefix . $key;
+        $value = $this->serialize($value);
+        $expiration = $this->getTime() + $seconds;
+        $collection = $this->table()->raw(null);
+        assert($collection instanceof Collection);
+
+        $result = $collection->updateOne(
+            ['key' => ['$eq' => $key]],
```
alcaeus: Why the explicit use of `$eq` instead of the shorter variant?
```suggestion
            ['key' => $key],
```

src/Cache/MongoStore.php
```diff
@@ -0,0 +1,90 @@
...
+    public function add($key, $value, $seconds): bool
+    {
+        $key = $this->prefix . $key;
+        $value = $this->serialize($value);
+        $expiration = $this->getTime() + $seconds;
+        $collection = $this->table()->raw(null);
+        assert($collection instanceof Collection);
+
+        $result = $collection->updateOne(
+            ['key' => ['$eq' => $key]],
```
alcaeus: Same as above, this could use the shorter variant.

src/Cache/MongoStore.php
```diff
@@ -0,0 +1,90 @@
...
+            ],
+            ['upsert' => true],
+        );
+
+        return $result->getUpsertedCount() > 0 || $result->getModifiedCount() > 0;
+    }
+
+    public function lock($name, $seconds = 0, $owner = null)
+    {
+        assert($this->connection instanceof Connection);
```
alcaeus: Wouldn't this always be true since this is set through the constructor, or is it technically possible for the value to change in the parent class?

src/Query/Builder.php
alcaeus: I dislike the ambiguity that is built into the documentation - you can't really be sure what's ignored. One option is to ignore any errors that purely come from the data (e.g. unique key violations, document constraint violations) but still report more serious problems (like server selection errors). IMO, users should be more aware of what could go wrong and explicitly ignore those errors if they don't care for them.

For the purpose of the cache/lock work, this seems sufficient so I'm fine with only ignoring unique key errors and returning to this if people have more errors they think should be ignored.

src/Cache/MongoStore.php
GromNaN: The property is defined in the parent class. I'm removing the inheritance on `DatabaseStore`, so the code will be more optimized with less type mess.

src/Cache/MongoStore.php
GromNaN: I assumed that the verbose version would be more explicit to avoid edge cases.

src/Query/Builder.php
GromNaN: This method is no longer used by the cache. I'm reverting this change, and creating a dedicated ticket PHPORM-170

src/Query/Builder.php
GromNaN: See PHPORM-170

src/Cache/MongoStore.php
```diff
@@ -0,0 +1,282 @@
...
+use function is_string;
+use function serialize;
+use function str_contains;
+use function unserialize;
+
+final class MongoStore implements LockProvider, Store
+{
+    use InteractsWithTime;
+    // Provides "many" and "putMany" in a non-optimized way
+    use RetrievesMultipleKeys;
```
GromNaN, line 25: This can be improved later in a new PR.

src/MongoDBServiceProvider.php
GromNaN: Moved to `collection` and `lock_collection` with default values `cache` and `cache_locks`.

composer.json
```diff
@@ -25,13 +25,15 @@
         "php": "^8.1",
         "ext-mongodb": "^1.15",
         "composer-runtime-api": "^2.0.0",
-        "illuminate/support": "^10.0|^11",
+        "illuminate/cache": "^10.36|^11",
```
GromNaN, line 28: Required to access `Cache::isOwnedCurrentProcess`
https://github.com/laravel/framework/commit/234444e8fb17daf16a703877e3fb8ccb4b84cad4

src/Cache/MongoLock.php
```diff
@@ -0,0 +1,138 @@
...
+{
+    /**
+     * Create a new lock instance.
+     *
+     * @param  Collection  $collection              The MongoDB collection
+     * @param  string      $name
+     * @param  int         $seconds
+     * @param  string|null $owner
+     * @param  array       $lottery                 The prune probability odds
+     * @param int         $defaultTimeoutInSeconds The default number of seconds that a lock should be held
```
jmikola, line 22:
```suggestion
     * @param int          $defaultTimeoutInSeconds The default number of seconds that a lock should be held
```

Are all of the `@param` annotations necessary? Elsewhere, I've only seen them used when a description of the parameter is necessary. In that case, you'd only need to do it for `$lottery` and `$defaultTimeoutInSeconds` (the `$collection` description doesn't seem necessary).

src/Cache/MongoLock.php
```diff
@@ -0,0 +1,138 @@
...
+        // The lock can be acquired if: it doesn't exist, it has expired,
+        // or it is already owned by the same lock instance.
+        $condition = [
+            '$or' => [
+                ['$lte' => ['$expiration', $this->currentTime()]],
+                ['$eq' => ['$owner', $this->owner]],
+            ],
+        ];
+        $result = $this->collection->findOneAndUpdate(
+            ['key' => ['$eq' => $this->name]],
```
jmikola:
```suggestion
            ['key' => $this->name],
```

If `name` is a string, I don't see any reason to use the verbose syntax here. According to the [server docs](https://www.mongodb.com/docs/manual/reference/operator/query/eq/#syntax), the only edge case is when the matched value is a regex.

src/Cache/MongoLock.php
```diff
@@ -0,0 +1,138 @@
...
+                    ],
+                ],
+            ],
+            [
+                'upsert' => true,
+                'returnDocument' => FindOneAndUpdate::RETURN_DOCUMENT_AFTER,
+            ],
+        );
+
+        if (random_int(1, $this->lottery[1]) <= $this->lottery[0]) {
```
jmikola, line 77: I realize this is beyond the PR, but is there a particular reason that `[2, 100]` is used instead of `0.02`? This just seems like an odd API.

src/Cache/MongoLock.php
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Releases this lock in disregard of ownership.
+     *
+     * @return void
+     */
+    #[Override]
+    public function forceRelease(): void
+    {
+        $this->collection->deleteMany([
```
jmikola: Why does this use `deleteMany()`? I realize there's no index here, but aren't locks uniquely identified by their `key` field?

Also, did you consider using the key string as the `_id` to enforce that? Doing so would be an easy way to ensure that an index is used for the queries.

src/Cache/MongoLock.php
```diff
@@ -0,0 +1,138 @@
...
+
+        return $result->owner === $this->owner;
+    }
+
+    /**
+     * Get the UNIX timestamp indicating when the lock should expire.
+     *
+     * @return int
+     */
+    protected function expiresAt()
```
jmikola: Any reason not to make this `private`? I expect it was protected in [Illuminate/Cache/DatabaseLock](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Cache/DatabaseLock.php) because other classes might extend that, but that doesn't seem necessary here.

src/Cache/MongoLock.php
jmikola: > when the lock is acquired twice by the same instance during the same second

What is the significance of "same instance during the same second"?

If `updateOne` reports that no document was upserted or modified, wouldn't that imply that the lock could not be acquired (i.e. has another owner and has not expired)?

I was testing this locally:

```php
$coll = $client->selectCollection('test', 'coll');
$coll->drop();

$key = 'foo';
$owner = 'bar';
$currentTime = time();
$expiresAt = $currentTime + 10;

$isExpiredOrAlreadyOwned = ['$or' => [
    ['$lte' => ['$expiration', $currentTime]],
    ['$eq' => ['$owner', $owner]],
]];

$result = $coll->updateOne(
    ['key' => $key],
    [
        ['$set' => [
            'owner' => ['$cond' => [$isExpiredOrAlreadyOwned, $owner, '$owner']],
            'expiration' => ['$cond' => [$isExpiredOrAlreadyOwned, $expiresAt, '$expiration']],
        ]],
    ],
    ['upsert' => true],
);

printf("\n\nupdateOne: matched(%d) modified(%d) upserted(%d)\n\n",
    $result->getMatchedCount(),
    $result->getModifiedCount(),
    $result->getUpsertedCount(),
);

if ($result->getMatchedCount()) {
    echo "lock '$key' already existed\n";
}

if ($result->getModifiedCount()) {
    echo "took over an expired lock for '$owner', which will now expire at $expiresAt\n";
}

if ($result->getUpsertedCount()) {
    echo "created the lock for '$owner', which will now expire at $expiresAt\n";
}
```

src/Cache/MongoStore.php
```diff
@@ -0,0 +1,282 @@
...
+            $this->prefix . $name,
+            $seconds,
+            $owner,
+            $this->lockLottery,
+            $this->defaultLockTimeoutInSeconds,
+        );
+    }
+
+    /**
+     * Restore a lock instance using the owner identifier.
```
jmikola, line 73: I realize this was copied from [DatabaseStore](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Cache/DatabaseStore.php#L295), but how is this method supposed to be used? I don't see it in the test suite.

"Restore a lock instance using the owner identifier" seems like an inaccurate description for a method that constructs and returns a new MongoLock instance.

I assume this has something to do with the `$owner` being a required parameter, which will differ from `lock()` generating a random string by default.

src/Cache/MongoStore.php
```diff
@@ -0,0 +1,282 @@
...
+     * Retrieve an item from the cache by key.
+     *
+     * @param  string $key
+     */
+    #[Override]
+    public function get($key): mixed
+    {
+        $result = $this->collection->findOne(
+            [
+                'key' => $this->prefix . $key,
```
jmikola: Did you consider using `_id` for the cache key field? I assume there shouldn't be duplicates, and this would ensure that the queries are indexed.

src/Cache/MongoStore.php
```diff
@@ -0,0 +1,282 @@
...
+    /**
+     * Store an item in the cache indefinitely.
+     *
+     * @param  string $key
+     * @param  mixed  $value
+     */
+    #[Override]
+    public function forever($key, $value): bool
+    {
+        return $this->put($key, $value, 315360000);
```
jmikola: Should `315360000` be a constant? I'll note that this also isn't an indefinite value. Would it make more sense to have a sentinel value for the expiration timestamp? For instance, zero could be used if you made adjustments to the expiration check (e.g. `forgetIfExpired()`).

src/Cache/MongoLock.php
GromNaN: In fact, here I've been able to add the types to the parameters so I can remove the `@param`. Elsewhere, I'm implementing an interface, so I can't add PHP types on parameters.

src/Cache/MongoLock.php
GromNaN: This comes from Laravel's Lottery: https://laravel.com/api/11.x/Illuminate/Support/Lottery.html

Used in DatabaseLock config: https://github.com/laravel/framework/blob/9b3bb15ab4d8e7a80bda82ffbe92613f15f69fee/src/Illuminate/Cache/DatabaseLock.php#L49

src/Cache/MongoLock.php
GromNaN: Just forgot to change it, thanks for noticing
```suggestion
    private function expiresAt()
```

src/Cache/MongoLock.php
GromNaN: Good idea, I'll change `key` to `_id` everywhere. It won't change anything in the way it works.

src/Cache/MongoLock.php
jmikola: I see see the same structure used in DatabaseLock, but not the Lottery class itself. Consistency is probably more important here. It may be worth elaborating on the structure of that two-element array in the doc block comment, though. As is, it's not entirely clear how the two values relate to one another.

src/Cache/MongoStore.php
GromNaN: Apparently, that's useful for cross-process locks: https://laravel.com/docs/11.x/cache#managing-locks-across-processes

It's part of the `LockProvider` interface, so I must implement it. But all the implementations are identical.

src/Cache/MongoLock.php
GromNaN: I'm talking about this test:
https://github.com/mongodb/laravel-mongodb/blob/3fa18f3553e0907f8ac1d19d1560a84afddb9ff8/tests/Cache/MongoLockTest.php#L30-L34

`updateOne` reports no document modified if the values in `$set` are not different from the values already stored.

In your example, if you run `updateOne` 2 times, the 2nd time you get "lock 'foo' already existed" and nothing else. In Laravel, `true` is expected because the lock is acquired and owned by the current instance.

src/Cache/MongoStore.php
GromNaN: This is 10 years. I've added a private constant.
Changing to a sentinel value would add much complexity that I don't think is necessary at this point.

src/Cache/MongoLock.php
GromNaN: I've added a description to each parameter.

src/Cache/MongoLock.php
jmikola: Now I understand the significance of "same second" in your first comment. The timestamp field wouldn't change on subsequent updates within the same clock second, and you rely on checking the `owner` field (not possible with `updateOne`) to determine if the lock is acquired.

src/Cache/MongoStore.php
```diff
@@ -0,0 +1,296 @@
...
+            '_id' => $this->prefix . $key,
+            'expiration' => ['$lte' => $this->currentTime()],
+        ]);
+
+        return $result->getDeletedCount() > 0;
+    }
+
+    public function flush(): bool
+    {
+        $this->collection->deleteMany([]);
```
jmikola, line 268: Did you consider dropping the collection? My storage engine knowledge is outdated, but I still reckon that'd be a faster operation.

And if there's no concern about preserving indexes or other metadata (e.g. validation rules), there isn't much benefit to keeping the original collection in place.

tests/Cache/MongoCacheStoreTest.php
```diff
@@ -0,0 +1,230 @@
...
+
+    public function testValueCanUpdateExistCacheInTransaction()
+    {
+        $store = $this->getStore();
+
+        $store->put('foo', 'bar', 60);
+
+        DB::beginTransaction();
+        $store->put('foo', 'new-bar', 60);
+        DB::commit();
```
jmikola, line 69: Is this actually testing anything? The underlying `updateOne` operation doesn't pass a session, so don't imagine it's actually going to execute within a transaction.

src/MongoDBServiceProvider.php
```diff
@@ -34,6 +40,26 @@ public function register()
...
+            $cache->extend('mongodb', function (Application $app, array $config): Repository {
+                // The closure is bound to the CacheManager
+                assert($this instanceof CacheManager);
+
+                $store = new MongoStore(
+                    $app['db']->connection($config['connection'] ?? null),
+                    $config['collection'] ?? 'cache',
+                    $this->getPrefix($config),
+                    $app['db']->connection($config['lock_connection'] ?? $config['connection'] ?? null),
+                    ($config['lock_collection'] ?? $config['collection'] ?? 'cache') . '_locks',
```
GromNaN:
```suggestion
                    $config['lock_collection'] ?? (($config['collection'] ?? 'cache') . '_locks'),
```

src/Cache/MongoStore.php
GromNaN: Why not, if the engine supports it well. What is the risk of race-conditioning if there is an insert at the same time?

tests/Cache/MongoCacheStoreTest.php
GromNaN: At least we can make sure it doesn't crash.

src/Cache/MongoStore.php
jmikola: Dropping [locks the collection](https://www.mongodb.com/docs/manual/reference/method/db.collection.drop/#resource-locking), and thus I expect it'd avoid any potential race conditions. In contrast, I expect `deleteMany()` would allow other operations to interleave.

To clarify: I don't consider this a real risk for the PR. The drop just seemed like a more straightforward operation.

tests/Cache/MongoCacheStoreTest.php
jmikola: My main concern is that "InTransaction" in this test (and at least one other below) suggests that the operation actually used a transaction. To ensure that no one gets the wrong idea by reading this test case, I'd suggest adding a comment explaining that transactions aren't used by the Cache adapters at all.

jmikola (approved):
I've said my peace, so I'll defer to you to incorporate any feedback you like.

src/Cache/MongoLock.php
```diff
@@ -0,0 +1,134 @@
...
+                'returnDocument' => FindOneAndUpdate::RETURN_DOCUMENT_AFTER,
+                'projection' => ['owner' => 1],
+            ],
+        );
+
+        if (random_int(1, $this->lottery[1]) <= $this->lottery[0]) {
+            $this->collection->deleteMany(['expiration' => ['$lte' => $this->currentTime()]]);
+        }
+
+        return $result['owner'] === $this->owner;
```
GromNaN, line 81: Reverted to use array-access to please static analysis tools that doesn't understand that [`MongoDB\MongoDB\BSONDocument`](https://github.com/mongodb/mongo-php-library/blob/e3a588407120d0ac9f3cf7b23fb739fc79d95340/src/Model/BSONDocument.php#L58) has magic property-access.

https://phpstan.org/r/1e7b5d92-8b98-4e53-863f-b0c8652820c1
https://psalm.dev/r/ab76084ecf

```php
$doc = new ArrayObject(['foo' => 'bar'], ArrayObject::ARRAY_AS_PROPS);
$doc->bar;
```

It's also more robust in case `array` is used as default typemap.

tests/Cache/MongoCacheStoreTest.php
GromNaN: Comment added.

src/Cache/MongoStore.php
GromNaN: I don't want to drop the collection. You could legitimately want to create a TTL index on the `expiration` field, and dropping the collection to purge the cache would have the side-effect of deleting the index.

GromNaN: See additional PR that enable TTL index: https://github.com/mongodb/laravel-mongodb/pull/2891

src/Cache/MongoLock.php
GromNaN: I reverted to `updateOne` in this [additional PR](https://github.com/mongodb/laravel-mongodb/pull/2891), because `UTCDateTime` is more precise and acquiring a lock with the same owner in the same microsecond will not happen.

alcaeus (approved):

PR #2879, opened 2024-04-18
DOCSP-35962: sorts

rustagir:
https://jira.mongodb.org/browse/DOCSP-35962

[STAGING](https://preview-mongodbrustagir.gatsbyjs.io/laravel/DOCSP-35962/fundamentals/read-operations/#sort-query-results)
[DOCS PR](https://github.com/10gen/docs-laravel/pull/64)

### Checklist

### DIFF
--- a/docs/fundamentals/read-operations.txt
+++ b/docs/fundamentals/read-operations.txt
@@ -94,11 +94,11 @@ retrieve documents that meet the following criteria:
 
       Use the following syntax to specify the query:
 
-      .. code-block:: php
-
-         $movies = Movie::where('year', 2010)
-             ->where('imdb.rating', '>', 8.5)
-             ->get();
+      .. literalinclude:: /includes/fundamentals/read-operations/ReadOperationsTest.php
+         :language: php
+         :dedent:
+         :start-after: start-query
+         :end-before: end-query
 
    .. tab:: Controller Method
       :tabid: controller
@@ -209,12 +209,11 @@ documents.
 
       Use the following syntax to specify the query:
 
-      .. code-block:: php
-
-         $movies = Movie::where('year', 1999)
-             ->skip(2)
-             ->take(3)
-             ->get();
+      .. literalinclude:: /includes/fundamentals/read-operations/ReadOperationsTest.php
+         :language: php
+         :dedent:
+         :start-after: start-skip-limit
+         :end-before: end-skip-limit
 
    .. tab:: Controller Method
       :tabid: controller
@@ -300,12 +299,11 @@ This example queries for documents in which the value of the ``countries`` field
 
       Use the following syntax to specify the query:
 
-      .. code-block:: php
-
-         $movies = Movie::where('countries', 'Indonesia')
-             ->orderBy('year')
-             ->orderBy('title', 'desc')
-             ->get();
+      .. literalinclude:: /includes/fundamentals/read-operations/ReadOperationsTest.php
+         :language: php
+         :dedent:
+         :start-after: start-sort
+         :end-before: end-sort
 
    .. tab:: Controller Method
       :tabid: controller
@@ -368,7 +366,7 @@ This example queries for documents in which the value of the ``countries`` field
 
    - :manual:`Natural order </reference/glossary/#std-term-natural-order>`
      in the Server manual glossary
-   - `Ordering, Grouping, Limit and Offset <https://laravel.com/docs/10.x/queries#ordering-grouping-limit-and-offset>`__
+   - `Ordering, Grouping, Limit and Offset <https://laravel.com/docs/11.x/queries#ordering-grouping-limit-and-offset>`__
      in the Laravel documentation
 
 .. _laravel-retrieve-one:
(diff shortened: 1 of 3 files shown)


docs/fundamentals/read-operations.txt
```diff
@@ -269,6 +271,103 @@ documents.
...
+.. _laravel-sort:
+
+Sort Query Results
+~~~~~~~~~~~~~~~~~~
+
+To order query results based on the values of specified fields, use the ``where()`` method
+followed by the ``orderBy()`` method. You can set an **ascending** or
+**descending** sort direction on results. By default, the ``orderBy()``
+method sets an ascending sort on the supplied field name. To specify a
+descending sort, pass ``'desc'`` as the second parameter.
```
ccho-mongodb: Suggestion:
I think the orderBy() method also accepts `asc' for ascending sort. It could be helpful to let the know the user exists (for completeness) and to mention that it's not necessary since it's the default value. I think keeping the asc examples as-is is fine though.

ccho-mongodb (approved):
LGTM! Left one suggestion.

docs/fundamentals/read-operations.txt
```diff
@@ -269,6 +271,106 @@ documents.
...
+            
+            ...
+
+.. tip::
+
+   To learn more about sorting, see the following resources:
+
+   - :manual:`Natural order </reference/glossary/#std-term-natural-order>`
+     in the Server manual glossary
+   - `Ordering, Grouping, Limit and Offset <https://laravel.com/docs/10.x/queries#ordering-grouping-limit-and-offset>`__
```
GromNaN: The Laravel version in the doc url needs to be updated to 11.x for laravel-mongodb 4.2

docs/fundamentals/read-operations.txt
```diff
@@ -269,6 +271,106 @@ documents.
...
+.. tabs::
+
+   .. tab:: Query Syntax
+      :tabid: query-syntax
+
+      Use the following syntax to specify the query:
+
+      .. code-block:: php
+
+         $movies = Movie::where('countries', 'Indonesia')
```
GromNaN: Could you move the examples to a test?

rustagir: I will move the query syntax examples into a test, but I'm not sure how to move the Controller method snippets into include files.

docs/fundamentals/read-operations.txt
```diff
@@ -269,6 +270,105 @@ documents.
...
+            
+            ...
+
+.. tip::
+
+   To learn more about sorting, see the following resources:
+
+   - :manual:`Natural order </reference/glossary/#std-term-natural-order>`
+     in the Server manual glossary
+   - `Ordering, Grouping, Limit and Offset <https://laravel.com/docs/11.x/queries#ordering-grouping-limit-and-offset>`__
```
GromNaN: That may be a subject to be dealt with separately, but we can't freeze the version like that. How to update the links when a new Laravel version is released? Perhaps the easiest way is to link to the latest version. But you run the risk of having a broken link.

```suggestion
   - `Ordering, Grouping, Limit and Offset <https://laravel.com/docs/queries#ordering-grouping-limit-and-offset>`__
```

docs/includes/fundamentals/read-operations/ReadOperationsTest.php
```diff
@@ -0,0 +1,97 @@
...
+    {
+        // start-sort
+        $movies = Movie::where('countries', 'Indonesia')
+            ->orderBy('year')
+            ->orderBy('title', 'desc')
+            ->get();
+        // end-sort
+
+        $this->assertNotNull($movies);
+        $this->assertCount(4, $movies);
```
GromNaN, line 79: I've added fixtures to make the tests work. But I had to update the expected number of results, so we don't have too much data to load.

docs/includes/fundamentals/read-operations/ReadOperationsTest.php
```diff
@@ -0,0 +1,97 @@
...
+    public function testFirst(): void
+    {
+        // start-first
+        $movie = Movie::where('runtime', 30)
+            ->orderBy('_id')
+            ->first();
+        // end-first
+
+        $this->assertNotNull($movie);
+        $this->assertInstanceOf(Movie::class, $movie);
```
GromNaN, line 95: I've renamed `$movies` into `$movie`, as a single object is expected. Not a list of 1.

docs/fundamentals/read-operations.txt
GromNaN: Ok, let's keep the controller example here for now.

GromNaN (approved):
LGTM after fixing the Laravel doc link.

docs/fundamentals/read-operations.txt
rustagir: We can link to the latest version, or we can store the laravel version as a source constant. Either way, the risk of linking to a broken link is there, so I'll update to just use latest for now.

docs/includes/fundamentals/read-operations/ReadOperationsTest.php
rustagir: Thanks for fixing the tests!

GromNaN (approved):

PR #2888, opened 2024-04-19
DOCSP-36792: Add default database connection admonition

norareidy:
JIRA - https://jira.mongodb.org/browse/DOCSP-36792
Staging - 
https://preview-mongodbnorareidy.gatsbyjs.io/laravel/DOCSP-36792/quick-start/configure-mongodb/
https://preview-mongodbnorareidy.gatsbyjs.io/laravel/DOCSP-36792/query-builder/
### Checklist

### DIFF
--- a/docs/query-builder.txt
+++ b/docs/query-builder.txt
@@ -36,12 +36,6 @@ lets you perform database operations. Facades, which are static interfaces to
 classes, make the syntax more concise, avoid runtime errors, and improve
 testability.
 
-.. tip::
-
-   Before using the ``DB`` facade, ensure that you specify MongoDB as your application's
-   default database connection. For instructions on setting the database connection,
-   see the :ref:`laravel-quick-start-connect-to-mongodb` step in the Quick Start.
-
 {+odm-short+} aliases the ``DB`` method ``table()`` as the ``collection()``
 method. Chain methods to specify commands and any constraints. Then, chain
 the ``get()`` method at the end to run the methods and retrieve the results.
@@ -52,6 +46,14 @@ The following example shows the syntax of a query builder call:
    DB::collection('<collection name>')
        // chain methods by using the "->" object operator
        ->get();
+.. tip::
+
+   Before using the ``DB::collection()`` method, ensure that you specify MongoDB as your application's
+   default database connection. For instructions on setting the database connection,
+   see the :ref:`laravel-quick-start-connect-to-mongodb` step in the Quick Start.
+
+   If MongoDB is not your application's default database, you can use the ``DB::connection()`` method
+   to specify a MongoDB connection. Pass a value of ``"mongodb"`` to the ``connection()`` method.
 
 This guide provides examples of the following types of query builder operations:
 
--- a/docs/query-builder.txt
+++ b/docs/query-builder.txt
@@ -53,7 +53,12 @@ The following example shows the syntax of a query builder call:
    see the :ref:`laravel-quick-start-connect-to-mongodb` step in the Quick Start.
 
    If MongoDB is not your application's default database, you can use the ``DB::connection()`` method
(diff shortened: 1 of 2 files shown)


rustagir (approved):
lgtm

docs/quick-start/configure-mongodb.txt
```diff
@@ -22,7 +22,7 @@ Configure Your MongoDB Connection
 
       .. code-block:: php
 
-         'default' => env('DB_CONNECTION', 'mongodb'),
+         'default' => 'mongodb',
```
GromNaN: If people want to update their default connection, it might be in the env var `DB_CONNECTION` in the `.env` file.

GromNaN (changes requested):
This is wrong, the `DB` facade can be used for any connection. Each model class specifies the connection it uses (for Eloquent builder), and the connection name can be specified before creating a query builder:

```php
DB::connection('mongodb')->collection('users')->...
```

docs/quick-start/configure-mongodb.txt
norareidy: This line should set the default connection to MongoDB even if the `DB_CONNECTION` variable is set to something else in `.env`. But I updated the admonition (in query-builder.txt) to give two options on how to set MongoDB as the connection, if that helps clarify

GromNaN: I remain convinced that this is the not the right way to change the default connection.

GromNaN: I remain convinced that this is the not the right way to change the default connection. @alcaeus any opinion?

alcaeus: Laravel's default configuration uses env variables, and I agree that we shouldn't deviate too far from this.

docs/quick-start/configure-mongodb.txt
```diff
@@ -35,7 +35,7 @@ Configure Your MongoDB Connection
          'connections' => [
            'mongodb' => [
              'driver' => 'mongodb',
-             'dsn' => env('DB_URI', '<connection string>'),
+             'dsn' => '<connection string>',
```
alcaeus: As the connection string is likely to differ between environments (e.g. dev, staging, prod), we should stick with the Laravel default of providing this in an environment variable.
```suggestion
             'dsn' => env('DB_URI', '<connection string>'),
```

docs/query-builder.txt
```diff
@@ -46,6 +46,14 @@ The following example shows the syntax of a query builder call:
...
        // chain methods by using the "->" object operator
        ->get();
+.. tip::
+
+   Before using the ``DB::collection()`` method, ensure that you specify MongoDB as your application's
+   default database connection. For instructions on setting the database connection,
+   see the :ref:`laravel-quick-start-connect-to-mongodb` step in the Quick Start.
+
+   If MongoDB is not your application's default database, you can use the ``DB::connection()`` method
+   to specify a MongoDB connection. Pass a value of ``"mongodb"`` to the ``connection()`` method.
```
alcaeus: This should mention that one should pass the connection name (for which we use "mongodb" in our examples). Also, I suggest using a code snippet for this purpose:
```suggestion
   If MongoDB is not your application's default database, you can use the ``DB::connection()`` method
   to specify a MongoDB connection. Pass a the name of the connection to the ``connection()`` method:
   
   .. code-block:: php

      $connection = DB::connection('mongodb');
```

alcaeus (dismissed):

docs/quick-start/configure-mongodb.txt
norareidy: Okay, I changed these code snippets back to using env variables. Since we don't explain the .env file or its variables in the quick start, I made a [ticket](https://jira.mongodb.org/browse/DOCSP-38890) to add this info.

GromNaN (approved):

PR #2889, opened 2024-04-19
DOCSP-35959: search text

rustagir:
[JIRA](https://jira.mongodb.org/browse/DOCSP-35959)

[STAGING](https://preview-mongodbrustagir.gatsbyjs.io/laravel/DOCSP-35959/fundamentals/read-operations/#search-text-fields)

[docs PR](https://github.com/10gen/docs-laravel/pull/67)

### Checklist


docs/fundamentals/read-operations.txt
```diff
@@ -175,6 +176,124 @@ Use the following syntax to run a find operation that matches all documents:
    more information about ``take()``, see the :ref:`laravel-modify-find` section of this
    guide.
 
+.. _laravel-retrieve-text-search:
+
+Search Text Fields
+------------------
+
+You can perform a text search by using the :manual:`$text
```
ccho-mongodb: Suggestion:

I think it could be better to swap this paragraph (describes how to perform one) with the one following it (describes what a text search is).  The first paragraph also seems more connected to the third paragraph ("After building your query...").

docs/fundamentals/read-operations.txt
```diff
@@ -175,6 +176,124 @@ Use the following syntax to run a find operation that matches all documents:
...
+   the text-valued field. To learn more about creating
+   indexes, see the :ref:`laravel-eloquent-indexes` section of the
+   Schema Builder guide.
+
+After building your query with the ``where()`` method, chain the ``get()``
+method to retrieve the query results.
+
+This example calls the ``where()`` method on the ``Movie`` Eloquent model to
+retrieve documents in which the ``plot`` field contains the phrase
+``"love story"``:
```
ccho-mongodb: Issue:

I think the creation of the index needs to be mentioned or shown to understand that "plot" is the indexed field.

Suggestion:
I think either mentioning that the text index was created on the "plot" field or showing an example of creating the index similar to [this Java documentation](https://www.mongodb.com/docs/drivers/java/sync/current/fundamentals/crud/read-operations/text/#text-index) would help a reader understand the connection.

docs/fundamentals/read-operations.txt
```diff
@@ -175,6 +176,124 @@ Use the following syntax to run a find operation that matches all documents:
...
+      in the ``MovieController.php`` file to resemble the following code:
+
+      .. io-code-block::
+         :copyable: true
+
+         .. input::
+            :language: php
+
+            class MovieController
+            {
```
ccho-mongodb, line 237: Suggestion:
If it makes it easier to test, I think you can omit the class and function declarations since it's already described as being inside the show() function of the MovieController.php file.
I would not recommend spending much effort on this though since it shouldn't impact the user's understanding.

docs/fundamentals/read-operations.txt
```diff
@@ -175,6 +176,124 @@ Use the following syntax to run a find operation that matches all documents:
...
+
+   To specify a phrase as the text search criteria, you must include it
+   with escaped quotes in the query filter.
+
+A text search assigns a numerical text score to indicate how closely
+each result matches the string in your query filter. You can sort the
+results by relevance by using the ``orderBy()`` method to sort on the
+``textScore`` metadata:
+
+.. literalinclude:: /includes/fundamentals/read-operations/ReadOperationsTest.php
```
ccho-mongodb, line 282: Suggestion:

I think it could be good to link to a resource that explains $meta and the output. From a quick search, I wasn't able to find a direct explanation, but maybe this [link](https://www.mongodb.com/docs/upcoming/core/indexes/index-types/index-text/control-text-search-results/#assign-weights-to-text-search-results) can help.

docs/fundamentals/read-operations.txt
```diff
@@ -175,6 +176,124 @@ Use the following syntax to run a find operation that matches all documents:
...
+``"love story"``:
+
+.. tabs::
+
+   .. tab:: Query Syntax
+      :tabid: query-syntax
+
+      Use the following syntax to specify the query:
+
+      .. literalinclude:: /includes/fundamentals/read-operations/ReadOperationsTest.php
```
ccho-mongodb, line 218: Great example!

docs/fundamentals/read-operations.txt
```diff
@@ -175,6 +176,124 @@ Use the following syntax to run a find operation that matches all documents:
...
+            IMDB Rating: 6.1
+            IMDB Votes: 223
+            Plot: A girl. A boy. A love story ...
+
+            ...
+            
+.. note::
+
+   To specify a phrase as the text search criteria, you must include it
+   with escaped quotes in the query filter.
```
ccho-mongodb: Question:
Perhaps a question for the technical reviewer, but is it possible to use single quotes instead of double quotes to represent the value of the $search key (which would hopefully avoid having to escape the quotes and concatenate strings)?

ccho-mongodb (approved):
LGTM. A couple suggestions and one issue, but I think it can be resolved without another review. Feel free to reach out if you want me to take another look!

docs/fundamentals/read-operations.txt
GromNaN: I use [single quote](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.single) everywhere, unless I want to use variable injection with double quotes.

docs/fundamentals/read-operations.txt
```diff
@@ -175,6 +176,124 @@ Use the following syntax to run a find operation that matches all documents:
...
+         :copyable: true
+
+         .. input::
+            :language: php
+
+            class MovieController
+            {
+                public function show()
+                {
+                     $movies = Movie::where('$text', ['$search' => "\"" . 'love story' . "\""])
```
GromNaN: Sure it's better like this. The doubles quotes will probably come from user input.
```suggestion
                     $movies = Movie::where('$text', ['$search' => '"love story"'])
```

docs/fundamentals/read-operations.txt
rustagir: I mentioned this in the note, but I can also add a sentence that explains that the example assumes a text index on the `plot` field

docs/fundamentals/read-operations.txt
rustagir: Leaving as is, since the other examples on this page use the same format for the code examples.

docs/fundamentals/read-operations.txt
rustagir: thanks for this improvement! Was struggling to make this work, didnt realize the solution was so simple.

docs/fundamentals/read-operations.txt
rustagir: tested this improvement and it works! removing the admonition

docs/includes/fundamentals/read-operations/ReadOperationsTest.php
```diff
@@ -94,4 +101,35 @@ public function testFirst(): void
...
     }
+
+    /**
+     * @runInSeparateProcess
+     * @preserveGlobalState disabled
+     */
+    public function testText(): void
+    {
+        // start-text
+        $movies = Movie::where('$text', ['$search' => '"love story"'])
```
GromNaN, line 112: Test failing, the result is 1: only the movie with the title "this is a love story" is matched. We get 2 results by removing the quotes.

```suggestion
        $movies = Movie::where('$text', ['$search' => 'love story'])
```

docs/includes/fundamentals/read-operations/ReadOperationsTest.php
```diff
@@ -94,4 +101,35 @@ public function testFirst(): void
...
+    }
+
+    /**
+     * @runInSeparateProcess
+     * @preserveGlobalState disabled
+     */
+    public function testTextRelevance(): void
+    {
+        // start-text-relevance
+        $movies = Movie::where('$text', ['$search' => '"love story"'])
```
GromNaN, line 127: Same as the other test:

```suggestion
        $movies = Movie::where('$text', ['$search' => 'love story'])
```

GromNaN (approved):
LGTM. I updated the test and examples.

docs/includes/fundamentals/read-operations/ReadOperationsTest.php
rustagir: Ah okay! Since the example is meant to show text search for a **phrase**, I think it is better to change the test to assert one matched document instead of changing the query filter. Let me know if you will make this change or if I should make the correcting commit.

GromNaN: I let you make the change that best illustrate your purpose. You have to update the inserted documents.

PR #2891, opened 2024-04-22
PHPORM-99 Enable TTL index to auto-purge of expired cache and lock items

GromNaN:
Fix PHPORM-99
Complement to #2877

By using an `UTCDateTime` for the expiration field of cache and lock collections, we can create a [TTL index](https://www.mongodb.com/docs/manual/core/index-ttl/).

```php
// Create TTL index on cache collection
Cache::driver('mongodb')->createTTLIndex();

// Create TTL index on lock collection
Cache::driver('mongodb')->lock('')->createTTLIndex();
```

Once TTL index have been created (using a migration), the `lock_lottery` value can be set to `[0, 0]` in order to disable random pruning of expired locks.

Also, using a `UTCDateTime` is more precise than the int timestamp in seconds. So we can use `updateOne` as [discussed here](https://github.com/mongodb/laravel-mongodb/pull/2877#discussion_r1569031485).

### DIFF
--- a/src/Cache/MongoLock.php
+++ b/src/Cache/MongoLock.php
@@ -41,7 +41,7 @@ final class MongoLock extends Lock
         // or it is already owned by the same lock instance.
         $isExpiredOrAlreadyOwned = [
             '$or' => [
-                ['$lte' => ['$expiration', $this->getUTCDateTime()]],
+                ['$lte' => ['$expires_at', $this->getUTCDateTime()]],
                 ['$eq' => ['$owner', $this->owner]],
             ],
         ];
@@ -57,11 +57,11 @@ final class MongoLock extends Lock
                                 'else' => '$owner',
                             ],
                         ],
-                        'expiration' => [
+                        'expires_at' => [
                             '$cond' => [
                                 'if' => $isExpiredOrAlreadyOwned,
                                 'then' => $this->getUTCDateTime($this->seconds),
-                                'else' => '$expiration',
+                                'else' => '$expires_at',
                             ],
                         ],
                     ],
@@ -74,8 +74,8 @@ final class MongoLock extends Lock
             ],
         );
 
-        if (! empty($this->lottery[0]) && random_int(1, $this->lottery[1]) <= $this->lottery[0]) {
-            $this->collection->deleteMany(['expiration' => ['$lte' => $this->getUTCDateTime()]]);
+        if ($this->lottery[0] <= 0 && random_int(1, $this->lottery[1]) <= $this->lottery[0]) {
+            $this->collection->deleteMany(['expires_at' => ['$lte' => $this->getUTCDateTime()]]);
         }
 
         return $result->getModifiedCount() > 0 || $result->getUpsertedCount() > 0;
@@ -112,7 +112,7 @@ final class MongoLock extends Lock
     {
         $this->collection->createIndex(
             // UTCDateTime field that holds the expiration date
-            ['expiration' => 1],
+            ['expires_at' => 1],
             // Delay to remove items after expiration
             ['expireAfterSeconds' => 0],
         );
@@ -127,7 +127,7 @@ final class MongoLock extends Lock
         return $this->collection->findOne(
             [
                 '_id' => $this->name,
-                'expiration' => ['$gte' => $this->getUTCDateTime()],
+                'expires_at' => ['$gte' => $this->getUTCDateTime()],
             ],
             ['projection' => ['owner' => 1]],
         )['owner'] ?? null;
--- a/src/Cache/MongoLock.php
+++ b/src/Cache/MongoLock.php
@@ -3,6 +3,8 @@
 namespace MongoDB\Laravel\Cache;
 
 use Illuminate\Cache\Lock;
+use Illuminate\Support\Carbon;
+use MongoDB\BSON\UTCDateTime;
 use MongoDB\Laravel\Collection;
 use MongoDB\Operation\FindOneAndUpdate;
 use Override;
@@ -41,11 +43,11 @@ final class MongoLock extends Lock
         // or it is already owned by the same lock instance.
         $isExpiredOrAlreadyOwned = [
             '$or' => [
-                ['$lte' => ['$expiration', $this->currentTime()]],
+                ['$lte' => ['$expiration', $this->getUTCDateTime()]],
                 ['$eq' => ['$owner', $this->owner]],
             ],
         ];
-        $result = $this->collection->findOneAndUpdate(
+        $result = $this->collection->updateOne(
             ['_id' => $this->name],
             [
                 [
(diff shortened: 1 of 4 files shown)


alcaeus (approved):
LGTM. Not sure about what prompted the change for the return value of the `acquire` method, but it looks like it achieves the same result.

GromNaN:
> LGTM. Not sure about what prompted the change for the return value of the `acquire` method, but it looks like it achieves the same result.

I think it saves the server from having to return the modified document, which means fewer operations. It's probably negligible, but in any case unnecessary.

alcaeus:
> I think it saves the server from having to return the modified document, which means fewer operations. It's probably negligible, but in any case unnecessary.

I'm not an expert, but IIRC `update` does have a few advantages over `findAndModify`, good call on that!

src/Cache/MongoLock.php
```diff
@@ -107,6 +107,17 @@ public function forceRelease(): void
         ]);
     }
 
+    /** Creates a TTL index that automatically deletes expired objects. */
+    public function createTTLIndex(): void
```
jmikola, line 119: I intentionally didn't suggest using a TTL index in https://github.com/mongodb/laravel-mongodb/pull/2877 because it seemed redundant with the `$lottery` constructor argument that remained in place.

Given that these are just public methods on the MongoLock and MongoStore classes, how do you expect users to interact with them? My understanding is that both classes would typically be used while handling a request, and this method seems more like schema setup. It makes sense for MongoDB, of course, but doesn't fit very well into whatever common API Laravel intended for these classes.

GromNaN: This is an additional feature that should be documented as I have described. The methods will probably be called in a migration. I think that for applications with a significant volume of traffic, using TTL indexes will have a positive impact on performance without having to search for the correct lottery value.

src/Cache/MongoLock.php
```diff
@@ -107,6 +107,17 @@ public function forceRelease(): void
...
+    {
+        $this->collection->createIndex(
+            // UTCDateTime field that holds the expiration date
+            ['expiration' => 1],
```
jmikola: If you have the flexibility to customize the field name, I'd suggest something like `expiresAt` for consistency with `createdAt` and `updatedAt`.

I didn't bother to raise this in the previous PR but it seems more relevant if there are also indexes involved and people may have eyes on the schema.

src/Cache/MongoLock.php
```diff
@@ -116,19 +127,14 @@ protected function getCurrentOwner(): ?string
...
-    /**
-     * Get the UNIX timestamp indicating when the lock should expire.
-     */
-    private function expiresAt(): int
+    private function getUTCDateTime(int $additionalSeconds = 0): UTCDateTime
     {
-        $lockTimeout = $this->seconds > 0 ? $this->seconds : $this->defaultTimeoutInSeconds;
-
-        return $this->currentTime() + $lockTimeout;
+        return new UTCDateTime(Carbon::now()->addSeconds($additionalSeconds));
```
jmikola, line 146: What allows UTCDateTime to be constructed from a Carbon instance? Is this relying on Carbon's string representation?

I looked at the [Addition and Subtraction](https://carbon.nesbot.com/docs/#api-addsub) docs for Carbon and the strings in the examples don't look correct.

src/Cache/MongoStore.php
```diff
@@ -181,12 +182,9 @@ public function get($key): mixed
     #[Override]
     public function increment($key, $value = 1): int|float|false
     {
-        $this->forgetIfExpired($key);
-
         $result = $this->collection->findOneAndUpdate(
             [
                 '_id' => $this->prefix . $key,
-                'expiration' => ['$gte' => $this->currentTime()],
```
jmikola, line 189: By removing this criteria and the preceding call to `forgetIfExpired()`, you're potentially incrementing an expired document. Is that a concern?

Since the TTL index is optional, I wouldn't have expected this to change.

src/Cache/MongoLock.php
```diff
@@ -74,11 +74,11 @@ public function acquire(): bool
             ],
         );
 
-        if (random_int(1, $this->lottery[1]) <= $this->lottery[0]) {
-            $this->collection->deleteMany(['expiration' => ['$lte' => $this->currentTime()]]);
+        if (! empty($this->lottery[0]) && random_int(1, $this->lottery[1]) <= $this->lottery[0]) {
```
jmikola: Does anything enforce that `$this->lottery` is a tuple? This only checks that the first element exists and is non-empty, but there is no check for a second element or both being integers.

Perhaps that should be done in the constructor.

I assume you intentionally decided not to make `$lottery` nullable as a means to disable it.

src/Cache/MongoLock.php
```diff
@@ -74,11 +74,11 @@ public function acquire(): bool
...
+            $this->collection->deleteMany(['expiration' => ['$lte' => $this->getUTCDateTime()]]);
         }
 
-        return $result['owner'] === $this->owner;
+        return $result->getModifiedCount() > 0 || $result->getUpsertedCount() > 0;
```
jmikola: Wasn't this logic problematic in the original PR? Are you now relying on the UTCDateTime always changing due to sub-second precision?

src/Cache/MongoLock.php
```diff
@@ -14,20 +16,18 @@ final class MongoLock extends Lock
...
+     * @param string|null     $owner      A unique string that identifies the owner. Random if not set
+     * @param array{int, int} $lottery    Probability [chance, total] of pruning expired cache items. Set to [0, 0] to disable
      */
     public function __construct(
         private readonly Collection $collection,
         string $name,
         int $seconds,
         ?string $owner = null,
         private readonly array $lottery = [2, 100],
-        private readonly int $defaultTimeoutInSeconds = 86400,
```
jmikola, line 30: Why was this removed entirely? It looks like it previously served a purpose if `$seconds` was zero. Now, specifying zero for `$seconds` would cause locks to immediately expire.

src/Cache/MongoLock.php
GromNaN: The [`Carbon` class extends `DateTime`](https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Carbon.php#L836), and the `UTCDateTime` [constructor](https://www.php.net/manual/en/mongodb-bson-utcdatetime.construct.php) accepts any `DateTimeInterface`. There is no string transformation.

src/Cache/MongoLock.php
GromNaN: [Laravel uses underscores](https://laravel.com/docs/11.x/eloquent#timestamps) for `created_at` and `updated_at`, I've updated to `expires_at`.

src/Cache/MongoStore.php
GromNaN: I've removed the previous `forgetIfExpired` in order to avoid useless command. A nominal usage must be a single command.

Even if the expired item is incremented here, it is deleted just after.

src/Cache/MongoLock.php
GromNaN: It seemed to me that the "**zero** chances on **whatever**" value was explicit enough to say that we wanted to disable this mechanism.

src/Cache/MongoLock.php
GromNaN: Since `$defaultTimeoutInSeconds` was only a default value, I moved the processing to `MongoStore::lock`. It is not possible to disable lock expiration.

src/Cache/MongoLock.php
GromNaN: No problem with the previous logic. I said to myself that it was impossible to have the same lock with the same owner acquired in the same microsecond. But I've reverted to that, so you don't have to worry.

src/Cache/MongoLock.php
jmikola: Ah, I missed Carbon's inheritance. SGTM!

jmikola (approved):
I trust there will be a separate PR for documentation (since this was a follow-up on an original PR).

PR #2894, opened 2024-04-22
DOCSP-35939: tls

rustagir:
https://jira.mongodb.org/browse/DOCSP-35939

[STAGING](https://preview-mongodbrustagir.gatsbyjs.io/laravel/DOCSP-35939/fundamentals/connection/tls/)

Docs PR - https://github.com/10gen/docs-laravel/pull/69

### Checklist


docs/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,192 @@
...
+========================
+Enable and Configure TLS
+========================
+
+.. facet::
+   :name: genre
+   :values: reference
+
+.. meta::
+   :keywords: code example, security, connection options
```
ccho-mongodb: Suggestion:
I think adding "SSL" to the keywords could be helpful since some people may look for the outdated SSL feature.

docs/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,192 @@
...
+   :class: singlecol
+
+Overview
+--------
+
+In this guide, you can learn how to use the TLS protocol to secure 
+your connection to a MongoDB deployment. To configure your connection to
+use TLS, enable the TLS option and provide your certificates for
+validation when configuring a connection in your application's
+``config/database.php`` file.
```
ccho-mongodb: Suggestion:

I think the "when" could be omitted since it doesn't necessarily need to be a specific point in time.

```suggestion
validation in your application's
``config/database.php`` file.
```

docs/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,192 @@
...
+----------
+
+In your application's ``config/database.php`` file, you can enable TLS
+on a connection to your MongoDB deployment in one of the following ways:
+
+- Setting the ``tls`` option to ``true`` in your connection string
+- Setting the ``tls`` option to ``true`` in the ``options`` property
+
+Select from the following :guilabel:`Connection String` and
+:guilabel:`URI Options` tabs to see a corresponding code sample:
```
ccho-mongodb: Suggestion:
I think URI Options might be valid term, but since these are represented as an "options" array in the config file, maybe it could be described as "connection options", which is similar to how they're categorized on the [Connection Options](https://github.com/mongodb/laravel-mongodb/blob/4.3/docs/fundamentals/connection/connection-options.txt#L49-L61) page.

docs/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,192 @@
...
+      :tabid: connection string tls true
+
+      .. code-block:: php
+         :emphasize-lines: 5
+
+         'connections' => [
+      
+             'mongodb' => [
+                 'driver' => 'mongodb',
+                 'dsn' => 'mongodb://<hostname>:<port>?tls=true',
```
ccho-mongodb: Suggestion:

I think there's usually a "/" before the "?" character, just [as convention](https://github.com/mongodb/laravel-mongodb/blob/4.3/docs/includes/figures/connection_uri_parts.png), but probably not necessary.

docs/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,192 @@
...
+                 'database' => 'myDB',
+             ], ...
```
ccho-mongodb: Suggestion:

I think it would be ok to omit the "..." ellipses in this case, since it's focusing on a single connection configuration.

Applies to all instances.

docs/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,192 @@
...
+   * - Client Certificate
+     - A digital certificate that allows the server to verify the identity
+       of your application to establish an encrypted network connection.
+
+   * - Certificate Key
+     - The client certificate private key file. This key is often
+       included within the certificate file itself.
+
+   * - Passphrase
+     - The password to decrypt the private client key if it is encrypted.
```
ccho-mongodb: Suggestion:

I think this is great info, but would definitely be helpful if there was explicit mention of which settings or files that they correspond to.

docs/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,192 @@
...
+   ]
+
+.. note::
+
+   You can alternatively reference your certificates by setting options
+   in the ``driverOptions`` property in your connection configuration.
+   To learn more about these options, see the
+   `MongoDB\Driver\Manager::__construct()
+   <https://www.php.net/manual/en/mongodb-driver-manager.construct.php>`__
+   API documentation.
```
ccho-mongodb: Suggestion:

In the referenced link, it's not clear to me which settings are safe to use. The driverOptions settings, such as such as ca_file, pem_file, and pem_pwd, on the page are described as "deprecated alias for the [option name] URI option" . 

I think it could be helpful to mention which settings to recommend since it's not clear on the linked page. Maybe @GromNaN can provide more guidance here.

docs/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,192 @@
...
+   
+       'mongodb' => [
+           'driver' => 'mongodb',
+           'dsn' => '<connection string>',
+           'database' => 'myDB',
+           'options'  => [
+               'tls' => true,
+               'tlsCAFile' => '<path to CA certificate>',
+               'tlsCertificateKeyFile' => '<path to public client certificate>',
+           ],
```
ccho-mongodb: Suggestion:
I think it could be helpful to include  `tlsCertificateKeyFilePassword` since it's mentioned as something they should set in "Set the following options...".

docs/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,192 @@
...
+flexibility in your application.
+
+Set the following options in the ``options`` property to reference your
+certificates:
+
+- ``tlsCAFile``
+- ``tlsCertificateKeyFile``
+- ``tlsCertificateKeyFilePassword``
+
+You can specify additional options to configure TLS on your connection.
```
ccho-mongodb: Suggestion:
If this refers to the "for testing purposes" settings, I think this sentence can be omitted. I think including this sentence might be duplicative and may raise a question about whether the additional options are the ones presented for testing purposes.

docs/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,192 @@
...
+Reference Certificates
+----------------------
+
+You must reference your certificates when configuring your ``mongodb``
+connection so that the server can validate them before the client connects.
+
+We recommend that you reference your certificates and set other TLS
+options in the ``options`` property of your connection configuration
+instead of in the connection string. This improves readability and
+flexibility in your application.
```
ccho-mongodb: Suggestion:

I wasn't able to think of a way that this could improve flexibility of an application, but if does improve flexibility, feel free to ignore the suggestion.

```suggestion
instead of in the connection string. This improves the readability of your configuration file.```

File: docs/fundamentals/connection/tls.txt [hunk 11]
```diff
@@ -0,0 +1,192 @@
...
+   
+.. warning::
+
+   Setting these options to ``true`` disables
+   certificate and hostname validation.
+   
+   Specifying these options in a production environment makes
+   your application insecure and potentially
+   vulnerable to expired certificates and foreign processes posing
+   as valid client instances.
```
ccho-mongodb: Suggestion:

I think the content on https://www.mongodb.com/docs/manual/reference/connection-string/#tls-options could be a better resource for providing specific information and recommendations about these settings.

Perhaps this info could be replaced with:

"For testing purposes, you can disable validation by setting the following options:

- ...

To learn more, see the Connection Options settings in the Server manual. Select PHP from the Select your language dropdown menu on the right side of the page.
"

ccho-mongodb (approved):
Looks great -- I think a few straightforward tweaks could improve the focus of the content. I don't think I need to review again, but happy to take another look if you want.

docs/fundamentals/connection/tls.txt
GromNaN: I'd ask @jmikola. I don't know this option. I learned about this feature by reading your doc.

docs/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,202 @@
+.. _laravel-tls:
+
+========================
+Enable and Configure TLS
```
jmikola, line 4: I'm beating a dead horse here, but it's bizarre to me that we're documenting this stuff on the Laravel library instead of in PHPLIB itself. Even then, this functionality is entirely contained within the extension (and libmongoc), but I understand that PHPLIB needs to fulfill the role of being the main docs for the PHP driver.

The only thing I would expect Laravel docs to cover would be how to specify these options in its own configuration (i.e. the mapping between Laravel config and the underlying driver options).

Is this the same content that's going to ultimately end up in PHPLIB as well? And is the official plan to have it duplicated?

docs/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,202 @@
...
+   
+   If your connection string uses a DNS SRV record by including
+   the ``mongodb+srv`` prefix, TLS is enabled on your connection by
+   default.
+
+Configure Certificates
+----------------------
+
+To successfully initiate a TLS request, your application must present 
+cryptographic certificates to prove its identity. Your application's
```
jmikola, line 94: Why is a client-side certificate necessary to initiate TLS connections to a cluster? I only expect this might be needed for [X509 authentication](https://www.mongodb.com/docs/manual/tutorial/configure-x509-client-authentication/).

docs/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,202 @@
...
+.. note::
+
+   For **testing purposes**, you can set the following options to
+   disable validation:
+
+   - ``tlsAllowInvalidCertificates``
+   - ``tlsAllowInvalidHostnames``
+   - ``tlsInsecure`` 
+   
+   Specifying these options in a production environment might
```
jmikola:
```suggestion
   Specifying these options in a production environment might make
```

I think you accidentally a word here.

docs/fundamentals/connection/tls.txt
jmikola: The URI options prefixed with `tls` supersede the `driverOptions`. There are very few options that are _only_ supported through `driverOptions` (e.g. `crl_file`), and I don't think any of those are relevant to these docs.

I think the right call is to remove any discussion of `driverOptions`. You can still link to the [Manager constructor](https://www.php.net/manual/en/mongodb-driver-manager.construct.php) as canonical documentation for URI options, though. In that case, you can drop a link to it in the "Additional Information" section below (probably _above_ the server docs, since it will be more relevant to PHP).

----

> ``driverOptions`` property

Is this actually a property in the Laravel configuration? If you're referring to the Manager constructor, then the proper terminology would be "`driverOptions` _parameter_" (not "property").

But this may be a moot point since the entire note can be removed.

docs/fundamentals/connection/tls.txt
rustagir: There is some content on setting SSL options in the existing PHP lib documentation - https://www.mongodb.com/docs/php-library/current/tutorial/connecting/#connecting-with-ssl-and-authentication

I guess the root issue here is that we are standardizing and creating content for the Laravel docs _before_ doing the same process for the PHPLIB docs. Since there is no comprehensive TLS documentation in the PHPLIB docs at the moment, we can cover this feature in this docs set, at least until we have a plan to standardize the PHP docs - maybe at that point, we can duplicate the content from here to there and replace the content in Laravel with a link to the PHP docs.

rustagir: @caitlindavey @ccho-mongodb If you have more context, feel free to comment!

docs/fundamentals/connection/tls.txt
rustagir: For any TLS content that I or another writer has written, we have used information from the Server docs to guide the driver documentation - https://www.mongodb.com/docs/manual/core/security-transport-encryption/#certificates

docs/fundamentals/connection/tls.txt
jmikola: Side note: [Connecting with SSL and Authentication](https://www.mongodb.com/docs/php-library/current/tutorial/connecting/) is outdated, since it still uses the old `ssl` URI option instead of `tls`.

> maybe at that point, we can duplicate the content from here to there and replace the content in Laravel with a link to the PHP docs.

This is likely the best possible outcome I can hope for.

docs/fundamentals/connection/tls.txt
jmikola: I don't see anything in [Certificates](https://www.mongodb.com/docs/manual/core/security-transport-encryption/#certificates) that talks about clients supplying their own certificates. The introductory paragraph of that section links to [TLS/SSL Configuration for Clients](https://www.mongodb.com/docs/manual/tutorial/configure-ssl-clients/), which mainly discusses how to configure a client to validate the servers' certificates.

If this kind of content is ending up in various drivers' documentation outside of a discussion about X509, I think this needs to be revisited across the board. In any event, this does not seem relevant to PHP users as written.

GromNaN:
I leave it to @jmikola to approve.

docs/fundamentals/connection/tls.txt
jmikola: I'd still appreciate a follow-up on where this content came from, since I didn't find it on the [Certificates](https://www.mongodb.com/docs/manual/core/security-transport-encryption/#certificates) page you linked. If this is truly coming from some other docs page, I expect _that_ should also be revised.

----

I just asked about this in the drivers channel:

> Are the `tlsCertificateKeyFile` and `tlsCertificateKeyFilePassword` URI options only utilized for [X509 auth](https://www.mongodb.com/docs/manual/tutorial/configure-x509-client-authentication/#authenticate-with-a-x-509-certificate)? The [URI option spec](https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst) makes no mention of X509, and the [X509 section in the auth spec](https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#mongodb-x509) makes no mention of the URI options.

I'll follow up once I have a response, but in the event that `tlsCertificateKeyFile` and `tlsCertificateKeyFilePassword` are only relevant to X509 auth I'd propose removing them in favor of discussing them in some page covering auth mechanisms and not TLS connections. If anything, OCSP seems far more relevant to a docs page on TLS connections than anything pertaining to auth.

Note: the `tlsCAFile` file _is_ generally useful if an application's system is poorly configured (we've run into that before with users connecting to Atlas) and is _not_ specific to X509 auth. That can remain.

jmikola: @rcsanchez97 examined the C driver and confirmed that these URI options are only used for X509 authentication, so I reckon that supports my suggestion to remove them from general TLS docs.

jmikola: @kevinAlbs just followed up with a demonstration that these options might still be used irrespective of X509 auth: https://gist.github.com/kevinAlbs/fcdfe12af21c2cbfbd77f72aaa78a1f7.

I think the relevant server docs for this is [Set Up mongod and mongos with Client Certificate Validation](https://www.mongodb.com/docs/manual/tutorial/configure-ssl/#set-up-mongod-and-mongos-with-client-certificate-validation); however, it's not clear to me exactly what server configuration option is responsible for requiring clients to present a certificate.

I did come across the following for [`net.tls.allowConnectionsWithoutCertificates`](https://www.mongodb.com/docs/manual/reference/configuration-options/#mongodb-setting-net.tls.allowConnectionsWithoutCertificates):

> By default, the server bypasses client certificate validation unless the server is configured to use a CA file. If a CA file is provided, the following rules apply:
>
> * For clients that don't provide certificates, [mongod](https://www.mongodb.com/docs/manual/reference/program/mongod/#mongodb-binary-bin.mongod) or [mongos](https://www.mongodb.com/docs/manual/reference/program/mongos/#mongodb-binary-bin.mongos) encrypts the TLS/SSL connection, assuming the connection is successfully made.
> * For clients that present a certificate, [mongos](https://www.mongodb.com/docs/manual/reference/program/mongos/#mongodb-binary-bin.mongos) or [mongod](https://www.mongodb.com/docs/manual/reference/program/mongod/#mongodb-binary-bin.mongod) performs certificate validation using the root certificate chain specified by [CAFile](https://www.mongodb.com/docs/manual/reference/configuration-options/#mongodb-setting-net.tls.CAFile) and reject clients with invalid certificates.

That suggests [`net.tls.CAFile`](https://www.mongodb.com/docs/manual/reference/configuration-options/#mongodb-setting-net.tls.CAFile) is responsible, although the first bullet is unclear as it suggests a connection might still be made.

This likely needs clarification from someone on the server team.

But just coming back to this PR, even if we leave the options documented in this section I take issue with the following lines that suggest clients must always present certificates to connect via TLS:

> To successfully initiate a TLS request, your application must present 
cryptographic certificates to prove its identity.

and

> The following list describes the components that your client must
present to establish a TLS-enabled connection:

rustagir: While we wait for clarification from the Server team, would it be sufficient to change the phraseage in those sentences/potentially other places to be less affirmative? Something like:

```
To successfully initiate a TLS request, your application might need to present
cryptographic certificates to prove its identity.
```

jmikola: Replacing "must" works for me.

docs/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,197 @@
...
+- ``tlsCertificateKeyFilePassword``
+   
+.. note::
+
+   For **testing purposes**, you can set the following options to
+   disable validation:
+
+   - ``tlsAllowInvalidCertificates``
+   - ``tlsAllowInvalidHostnames``
+   - ``tlsInsecure``
```
jmikola: Quoting https://www.php.net/manual/en/mongodb-driver-manager.construct.php:

> Specifying [true](https://www.php.net/manual/en/reserved.constants.php#constant.true) for this option has the same effect as specifying [true](https://www.php.net/manual/en/reserved.constants.php#constant.true) for both the "tlsAllowInvalidCertificates" and "tlsAllowInvalidHostnames" URI options. Defaults to [false](https://www.php.net/manual/en/reserved.constants.php#constant.false).

There should be no reason for the user to specify all three options. The improvement here would be to note that `tlsInsecure` implies the other two options.

Sorry for missing this earlier. This change should probably be applied to whatever content this was copied/adapted from (assuming other driver docs would be affected).

jmikola (approved):
LGTM. I'll leave you to address the `tlsInsecure` documentation on your own and don't need to review that a second time.

norareidy (approved):
lgtm

PR #2904, opened 2024-04-23
PHPORM-81 implement `mongodb` driver for batch

GromNaN:
Fix PHPORM-81

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 * Support collection name prefix by @GromNaN in [#2930](https://github.com/mongodb/laravel-mongodb/pull/2930)
 * Add `mongodb` driver for Batching by @GromNaN in [#2904](https://github.com/mongodb/laravel-mongodb/pull/2904)
 * Rename queue option `table` to `collection`
+* Replace queue option `expire` with `retry_after`
 
 ## [4.3.0] - 2024-04-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 ## [4.4.0] - unreleased
 
 * Support collection name prefix by @GromNaN in [#2930](https://github.com/mongodb/laravel-mongodb/pull/2930)
+* Add `mongodb` driver for Batching by @GromNaN in [#2904](https://github.com/mongodb/laravel-mongodb/pull/2904)
+* Rename queue option `table` to `collection`
 
 ## [4.3.0] - 2024-04-26
 
--- a/docs/queues.txt
+++ b/docs/queues.txt
@@ -11,7 +11,7 @@ Queues
 .. meta::
    :keywords: php framework, odm, code example
 
-If you want to use MongoDB as your database backend for Laravel Queue, change 
+If you want to use MongoDB as your database backend for Laravel Queue, change
 the driver in ``config/queue.php``:
 
 .. code-block:: php
@@ -20,27 +20,107 @@ the driver in ``config/queue.php``:
        'database' => [
            'driver' => 'mongodb',
            // You can also specify your jobs specific database created on config/database.php
-           'connection' => 'mongodb-job',
-           'table' => 'jobs',
+           'connection' => 'mongodb',
+           'collection' => 'jobs',
            'queue' => 'default',
            'expire' => 60,
        ],
    ],
 
-If you want to use MongoDB to handle failed jobs, change the database in 
+.. list-table::
+   :header-rows: 1
+   :widths: 25 75
+
+   * - Setting
+     - Description
+
+   * - ``driver``
+     - **Required**. Specifies the queue driver to use. Must be ``mongodb``.
+
+   * - ``connection``
+     - Uses the default connection by default. The database connection used to store jobs. It must be a ``mongodb`` connection.
+
+   * - ``collection``
+     - **Required**. Name of the MongoDB collection to store jobs to process.
+
+   * - ``queue``
+     - **Required**. Name of the queue.
+
+   * - ``retry_after``
+     - Default ``60``. Specify how many seconds the queue connection should wait before retrying a job that is being processed
+
+If you want to use MongoDB to handle failed jobs, change the database in
 ``config/queue.php``:
 
 .. code-block:: php
 
    'failed' => [
        'driver' => 'mongodb',
-       // You can also specify your jobs specific database created on config/database.php
-       'database' => 'mongodb-job',
-       'table' => 'failed_jobs',
+       'database' => 'mongodb',
+       'collection' => 'failed_jobs',
    ],
+
+.. list-table::
+   :header-rows: 1
+   :widths: 25 75
+
+   * - Setting
+     - Description
+
+   * - ``driver``
+     - **Required**. Specifies the queue driver to use. Must be ``mongodb``.
+
+   * - ``connection``
+     - Uses the default connection by default. The database connection used to store jobs. It must be a ``mongodb`` connection.
+
+   * - ``collection``
+     - Default ``failed_jobs``. Name of the MongoDB collection to store failed jobs.
+
 
 Add the service provider in ``config/app.php``:
 
 .. code-block:: php
 
    MongoDB\Laravel\MongoDBQueueServiceProvider::class,
+
+
+Job Batching
+------------
+
+`Laravel's job batching https://laravel.com/docs/{+laravel-docs-version+}/queues#job-batching>`__
+feature allows you to easily execute a batch of jobs and then perform some action
+when the batch of jobs has completed executing.
+
+With MongoDB, you don't have to create any collection before using this feature.
+The collection ``job_batches`` will be created automatically to store meta
+information about your job batches, such as their completion percentage
+
+.. code-block:: php
+
+    'batching' => [
+       'driver' => 'mongodb',
+       'database' => 'mongodb',
+       'collection' => 'job_batches',
+   ],
+
+.. list-table::
+   :header-rows: 1
+   :widths: 25 75
+
+   * - Setting
+     - Description
+
+   * - ``driver``
+     - **Required**. Specifies the queue driver to use. Must be ``mongodb``.
+
+   * - ``connection``
+     - Uses the default connection by default. The database connection used to store jobs. It must be a ``mongodb`` connection.
+
+   * - ``collection``
+     - Default ``job_batches``. Name of the MongoDB collection to store job batches.
+
+Add the service provider in ``config/app.php``:
+
+.. code-block:: php
+
+   MongoDB\Laravel\MongoDBBusServiceProvider::class,
--- a/docs/queues.txt
+++ b/docs/queues.txt
@@ -38,7 +38,7 @@ the driver in ``config/queue.php``:
      - **Required**. Specifies the queue driver to use. Must be ``mongodb``.
 
    * - ``connection``
-     - Uses the default connection by default. The database connection used to store jobs. It must be a ``mongodb`` connection.
+     - The database connection used to store jobs. It must be a ``mongodb`` connection. The driver uses the default connection if a connection is not specified.
 
    * - ``collection``
      - **Required**. Name of the MongoDB collection to store jobs to process.
@@ -47,7 +47,7 @@ the driver in ``config/queue.php``:
      - **Required**. Name of the queue.
 
    * - ``retry_after``
-     - Default ``60``. Specify how many seconds the queue connection should wait before retrying a job that is being processed
+     - Specifies how many seconds the queue connection should wait before retrying a job that is being processed. Defaults to ``60``. 
 
 If you want to use MongoDB to handle failed jobs, change the database in
 ``config/queue.php``:
@@ -71,10 +71,10 @@ If you want to use MongoDB to handle failed jobs, change the database in
      - **Required**. Specifies the queue driver to use. Must be ``mongodb``.
 
    * - ``connection``
-     - Uses the default connection by default. The database connection used to store jobs. It must be a ``mongodb`` connection.
+     - The database connection used to store jobs. It must be a ``mongodb`` connection. The driver uses the default connection if a connection is not specified.
 
    * - ``collection``
-     - Default ``failed_jobs``. Name of the MongoDB collection to store failed jobs.
+     - Name of the MongoDB collection to store failed jobs. Defaults to ``failed_jobs``. 
 
 
 Add the service provider in ``config/app.php``:
@@ -87,12 +87,12 @@ Add the service provider in ``config/app.php``:
 Job Batching
 ------------
 
-`Job batching https://laravel.com/docs/{+laravel-docs-version+}/queues#job-batching>`__
-is a Laravel feature to execute batch of jobs and subsequent actions before,
+`Job batching <https://laravel.com/docs/{+laravel-docs-version+}/queues#job-batching>`__
+is a Laravel feature to execute a batch of jobs and subsequent actions before,
 after, and during the execution of the jobs from the queue.
 
-With MongoDB, you don't have to create any collection before using this feature.
-The collection ``job_batches`` will be created automatically to store meta
+With MongoDB, you don't have to create any collection before using job batching.
+The ``job_batches`` collection is created automatically to store meta
 information about your job batches, such as their completion percentage.
 
 .. code-block:: php
@@ -114,10 +114,10 @@ information about your job batches, such as their completion percentage.
      - **Required**. Specifies the queue driver to use. Must be ``mongodb``.
 
    * - ``connection``
-     - Uses the default connection by default. The database connection used to store jobs. It must be a ``mongodb`` connection.
+     - The database connection used to store jobs. It must be a ``mongodb`` connection. The driver uses the default connection if a connection is not specified.
 
    * - ``collection``
-     - Default ``job_batches``. Name of the MongoDB collection to store job batches.
+     - Name of the MongoDB collection to store job batches. Defaults to ``job_batches``. 
 
 Add the service provider in ``config/app.php``:
 
--- a/docs/queues.txt
+++ b/docs/queues.txt
@@ -23,7 +23,7 @@ the driver in ``config/queue.php``:
            'connection' => 'mongodb',
            'collection' => 'jobs',
            'queue' => 'default',
-           'expire' => 60,
+           'retry_after' => 60,
        ],
    ],
 
--- a/docs/queues.txt
+++ b/docs/queues.txt
@@ -87,13 +87,13 @@ Add the service provider in ``config/app.php``:
 Job Batching
 ------------
 
-`Laravel's job batching https://laravel.com/docs/{+laravel-docs-version+}/queues#job-batching>`__
-feature allows you to easily execute a batch of jobs and then perform some action
-when the batch of jobs has completed executing.
+`Job batching https://laravel.com/docs/{+laravel-docs-version+}/queues#job-batching>`__
+is a Laravel feature to execute batch of jobs and subsequent actions before,
+after and during the execution of the jobs from the queue.
 
 With MongoDB, you don't have to create any collection before using this feature.
 The collection ``job_batches`` will be created automatically to store meta
-information about your job batches, such as their completion percentage
+information about your job batches, such as their completion percentage.
 
 .. code-block:: php
 
--- a/docs/queues.txt
+++ b/docs/queues.txt
@@ -89,7 +89,7 @@ Job Batching
 
 `Job batching https://laravel.com/docs/{+laravel-docs-version+}/queues#job-batching>`__
 is a Laravel feature to execute batch of jobs and subsequent actions before,
-after and during the execution of the jobs from the queue.
+after, and during the execution of the jobs from the queue.
 
 With MongoDB, you don't have to create any collection before using this feature.
 The collection ``job_batches`` will be created automatically to store meta
--- a/src/Bus/MongoBatchRepository.php
+++ b/src/Bus/MongoBatchRepository.php
@@ -20,7 +20,6 @@ use MongoDB\Laravel\Connection;
 use MongoDB\Operation\FindOneAndUpdate;
 use Override;
 
-use function date_default_timezone_get;
 use function is_string;
 use function serialize;
 use function unserialize;
@@ -66,6 +65,9 @@ class MongoBatchRepository extends DatabaseBatchRepository implements PrunableBa
         $batch = $this->collection->findOne(
             ['_id' => $batchId],
             [
+                // If the select query is executed faster than the database replication takes place,
+                // then no batch is found. In that case an exception is thrown because jobs are added
+                // to a null batch.
                 'readPreference' => new ReadPreference(ReadPreference::PRIMARY),
                 'typeMap' => ['root' => 'array', 'array' => 'array', 'document' => 'array'],
             ],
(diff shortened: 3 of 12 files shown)


tests/Bus/MongoBatchRepositoryTest.php
```diff
@@ -0,0 +1,364 @@
...
+
+        $repository = new MongoBatchRepository(new BatchFactory($queue), $connection, 'job_batches');
+
+        $pendingBatch = (new PendingBatch(new Container(), collect()))
+            ->progress(function (Batch $batch) {
+                $_SERVER['__progress.batch'] = $batch;
+                $_SERVER['__progress.count']++;
+            })
+            ->then(function (Batch $batch) {
+                $_SERVER['__then.batch'] = $batch;
```
GromNaN, line 346: We can't use `$this` because the closure is serialized.

src/Bus/MongoBatchRepository.php
```diff
@@ -0,0 +1,270 @@
...
+        $this->collection = $connection->getCollection($collection);
+
+        parent::__construct($factory, $connection, $collection);
+    }
+
+    #[Override]
+    public function get($limit = 50, $before = null): array
+    {
+        if (is_string($before)) {
+            $before = new ObjectId($before);
```
alcaeus, line 47: Beware that this may throw an exception if the string is not a well-formed `ObjectId`. Not sure if that's something we want to consider, but I figured I'd point it out.

src/Bus/MongoBatchRepository.php
```diff
@@ -0,0 +1,270 @@
...
+                'sort' => ['_id' => -1],
+                'typeMap' => ['root' => 'array', 'document' => 'array', 'array' => 'array'],
+            ],
+        )->toArray();
+    }
+
+    #[Override]
+    public function find(string $batchId): ?Batch
+    {
+        $batchId = new ObjectId($batchId);
```
alcaeus, line 63: Can we guarantee that `batchId` will always be a valid `ObjectId` in string representation? If not, it shouldn't be problematic to use a string as primary key.

src/MongoDBQueueServiceProvider.php
```diff
@@ -51,6 +54,15 @@ protected function registerFailedJobServices()
      */
     protected function mongoFailedJobProvider(array $config): MongoFailedJobProvider
     {
-        return new MongoFailedJobProvider($this->app['db'], $config['database'], $config['table']);
+        if (! isset($config['collection']) && isset($config['table'])) {
+            trigger_error('Since mongodb/laravel-mongodb 4.4: Using "table" option for the queue is deprecated. Use "collection" instead.', E_USER_DEPRECATED);
+            $config['collection'] = $config['table'];
+        }
```
alcaeus, line 60: 👍

alcaeus (commented):
I'm not sure about assuming that any string passed will be a valid `ObjectId`. A possible workaround would be to try and create an `ObjectId` instance, but leave the value as-is if it isn't valid. That way, the operation will either succeed (if such a record exists) or be a no-op (which would be the assumed behaviour for any batch that doesn't exist). As is, passing a string that isn't `ObjectId`-like would result in an exception, which IMO is not desirable.

src/Bus/MongoBatchRepository.php
GromNaN: In the `store` method, the `_id` is not set, so only `ObjectId` can be inserted.

alcaeus: I wanted to point this out as theoretically users could call the method with any string. No objection to assuming valid ObjectIDs are passed and leaving this as-is.

docs/queues.txt
```diff
@@ -20,27 +20,107 @@ the driver in ``config/queue.php``:
...
 .. code-block:: php
 
    MongoDB\Laravel\MongoDBQueueServiceProvider::class,
+
+
+Job Batching
+------------
+
+`Laravel's job batching https://laravel.com/docs/{+laravel-docs-version+}/queues#job-batching>`__
+feature allows you to easily execute a batch of jobs and then perform some action
```
jmikola: I dunno where the MongoDB style guide is for docs, but I think we generally want to avoid subjective terms like "easily" or "simply". I remember this being relevant for server docs.

docs/queues.txt
```diff
@@ -20,27 +20,107 @@ the driver in ``config/queue.php``:
...
+when the batch of jobs has completed executing.
+
+With MongoDB, you don't have to create any collection before using this feature.
+The collection ``job_batches`` will be created automatically to store meta
+information about your job batches, such as their completion percentage
```
jmikola:
```suggestion
information about your job batches, such as their completion percentage.
```

src/Bus/MongoBatchRepository.php
```diff
@@ -0,0 +1,270 @@
...
+use Override;
+
+use function date_default_timezone_get;
+use function is_string;
+use function serialize;
+use function unserialize;
+
+// Extending DatabaseBatchRepository is necessary so methods pruneUnfinished and pruneCancelled
+// are called by PruneBatchesCommand
+class MongoBatchRepository extends DatabaseBatchRepository implements PrunableBatchRepository
```
jmikola, line 29: Is there any reason for inconsistent use of "Mongo" and "MongoDB" as class prefixes? This isn't the first time I've noticed it, and I understand that there may be historical inconsistencies in the project, but this PR introduces two new classes with inconsistent names.

src/Bus/MongoBatchRepository.php
```diff
@@ -0,0 +1,270 @@
...
+        );
+    }
+
+    #[Override]
+    public function markAsFinished(string $batchId): void
+    {
+        $batchId = new ObjectId($batchId);
+        $this->collection->updateOne(
+            ['_id' => $batchId],
+            ['$set' => ['finished_at' => new UTCDateTime(Carbon::now())]],
```
jmikola: Is `Carbon::now()` mainly used to support mocking of time in the test suite? If so, is this worthy of a comment for future readers?

src/Bus/MongoBatchRepository.php
```diff
@@ -0,0 +1,270 @@
...
+    /**
+     * Rollback the last database transaction for the connection.
+     */
+    #[Override]
+    public function rollBack(): void
+    {
+        $this->connection->rollBack();
+    }
+
+    /** Mark the batch that has the given ID as finished. */
```
jmikola: Is this comment accurate? The `deleteMany()` operation in this method doesn't refer to any "given ID".

src/Bus/MongoBatchRepository.php
```diff
@@ -0,0 +1,270 @@
...
+
+    #[Override]
+    public function find(string $batchId): ?Batch
+    {
+        $batchId = new ObjectId($batchId);
+
+        $batch = $this->collection->findOne(
+            ['_id' => $batchId],
+            [
+                'readPreference' => new ReadPreference(ReadPreference::PRIMARY),
```
jmikola, line 71: Why does `find()` use an explicit read preference, while `get()` does not? This may warrant a comment.

src/Bus/MongoBatchRepository.php
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    /** @return ($date is null ? null : CarbonImmutable) */
+    private function toCarbon(?UTCDateTime $date): ?CarbonImmutable
+    {
+        if ($date === null) {
+            return null;
+        }
+
+        return CarbonImmutable::createFromTimestamp((string) $date, date_default_timezone_get());
```
jmikola: AFAIK, `createFromTimestamp()` and `createFromTimestampUTC()` both expect the argument to denote the timestamp in seconds. [`UTCDateTime::__toString()`](https://www.php.net/manual/en/mongodb-bson-utcdatetime.tostring.php) returns a numeric string denoting milliseconds.

Would `createFromTimestampMsUTC()` be the correct method to use here? I'm also concerned about _not_ using the "UTC" variant for the Carbon method here, as that might blindly apply the current system timezone to the timestamp. The Carbon docs aren't clear about _how_ the optional timezone gets applied to the timestamp (ideally, it would always assume a timestamp is UTC and apply a conversion).

src/MongoDBQueueServiceProvider.php
```diff
@@ -51,6 +54,15 @@ protected function registerFailedJobServices()
      */
     protected function mongoFailedJobProvider(array $config): MongoFailedJobProvider
     {
-        return new MongoFailedJobProvider($this->app['db'], $config['database'], $config['table']);
+        if (! isset($config['collection']) && isset($config['table'])) {
```
jmikola, line 57: Should it be an outright exception to specify _both_ `table` and `collection`? As-is, I assume `table` would just be ignored.

tests/Bus/MongoBatchRepositoryTest.php
jmikola: How would `$this` be used here even if it _was_ supported?

Offhand, is [laravel/serializable-closure](https://github.com/laravel/serializable-closure) used internally for serialization? AFAIK, the stock `serialize()` function doesn't support closures on its own.

jmikola (commented):
Forgive me for any ignorant questions.

docs/queues.txt
GromNaN: I merely copied the [Laravel doc](https://laravel.com/docs/11.x/queues#job-batching), but I agree the working doesn't follow best practices. I'll reword.

src/Bus/MongoBatchRepository.php
GromNaN: I also named `MongoLock` and `MongoStore` with the same unsaid naming rule, following the naming examples of `MongoConnector`, `MongoQueue` ...

src/Bus/MongoBatchRepository.php
GromNaN: Because there is an explicit `useWritePdo` [in `DatabaseBatchRepository`](https://github.com/laravel/framework/blob/279c444b70a28c3a2187899ba420364cac9b702a/src/Illuminate/Bus/DatabaseBatchRepository.php#L80). Introduced by https://github.com/laravel/framework/pull/43737

I suspect that `get` is not used in Laravel, which would explain why it hasn't been fixed too.

src/MongoDBQueueServiceProvider.php
GromNaN: Yes, `table` will be ignored if `collection` is set. Allowing both make migration easier as you can specify both during transition.

tests/Bus/MongoBatchRepositoryTest.php
GromNaN: All the `$_SERVER['something']` could have been replaced by `$this->data['something']`. But that's not allowed by the package you mentioned.

src/Bus/MongoBatchRepository.php
jmikola: That definitely seems worth of a comment that mentions the SQL version using a writable connection.

src/Bus/MongoBatchRepository.php
GromNaN: Good point, I hadn't looked closely enough at `Carbon` methods.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 ## [4.4.0] - unreleased
 
 * Support collection name prefix by @GromNaN in [#2930](https://github.com/mongodb/laravel-mongodb/pull/2930)
+* Add `mongodb` driver for Batching by @GromNaN in [#2904](https://github.com/mongodb/laravel-mongodb/pull/2904)
+* Rename queue option `table` to `collection`
```
jmikola, line 8: Do you also need to note deprecating `expire` for `retry_after`?

src/Bus/MongoBatchRepository.php
```diff
@@ -0,0 +1,276 @@
...
+    {
+        $batchId = new ObjectId($batchId);
+        $this->collection->deleteOne(['_id' => $batchId]);
+    }
+
+    /** Execute the given Closure within a storage specific transaction. */
+    #[Override]
+    public function transaction(Closure $callback): mixed
+    {
+        return $this->connection->transaction(fn () => $callback());
```
jmikola: Is there any significance to wrapping `$callback` in an additional closure?

GromNaN: Removed. Also in Laravel from where I copied the code: https://github.com/laravel/framework/pull/51513

docs/queues.txt
```diff
@@ -20,27 +20,107 @@ the driver in ``config/queue.php``:
...
 
    MongoDB\Laravel\MongoDBQueueServiceProvider::class,
+
+
+Job Batching
+------------
+
+`Job batching https://laravel.com/docs/{+laravel-docs-version+}/queues#job-batching>`__
+is a Laravel feature to execute batch of jobs and subsequent actions before,
+after and during the execution of the jobs from the queue.
```
jmikola:
```suggestion
after, and during the execution of the jobs from the queue.
```

Oxford comma.

src/Bus/MongoBatchRepository.php
```diff
@@ -0,0 +1,276 @@
...
+class MongoBatchRepository extends DatabaseBatchRepository implements PrunableBatchRepository
+{
+    private Collection $collection;
+
+    public function __construct(
+        BatchFactory $factory,
+        Connection $connection,
+        string $collection,
+    ) {
+        $this->collection = $connection->getCollection($collection);
```
jmikola, line 38: Apologies for not noticing this before, but is it typical for the database name to not be configurable across components? I do see that [Connection](https://github.com/mongodb/laravel-mongodb/blob/4.3.0/src/Connection.php#L66) relies on `getDefaultDatabaseName()`, which suggests every component within this library will end up using the same database.

src/Bus/MongoBatchRepository.php
```diff
@@ -0,0 +1,276 @@
...
+            'failed_jobs' => 0,
+            'failed_job_ids' => [],
+            // Serialization is required for Closures
+            'options' => serialize($batch->options),
+            'created_at' => $this->getUTCDateTime(),
+            'cancelled_at' => null,
+            'finished_at' => null,
+        ]);
+
+        return $this->find($result->getInsertedId());
```
jmikola: Is there a particular reason to use separate `insertOne` and `findOne` (via `$this->find()`) operations here? This looks like a candidate for `findOneAndUpdate()` with an upsert option and `RETURN_DOCUMENT_AFTER`.

src/MongoDBBusServiceProvider.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    /**
+     * Register the service provider.
+     */
+    public function register()
+    {
+        $this->app->singleton(MongoBatchRepository::class, function (Container $app) {
+            return new MongoBatchRepository(
+                $app->make(BatchFactory::class),
+                $app->make('db')->connection($app->config->get('queue.batching.database')),
```
jmikola, line 23: I realize this isn't your fault, but it's bizarre to me that the config option is `queue.batching.database` when we're actually selecting a connection. I expect this only affects the MongoDB integration, though.

jmikola (approved):
You can decide which of my comments are worth addressing.

src/Bus/MongoBatchRepository.php
GromNaN: The database is per connection. To use an other database, they need to configure a distinct connection. I don't think this is an issue.

src/Bus/MongoBatchRepository.php
GromNaN: Since `_id` is the only unknown value, I'm updating this method to hydrate the `Batch` object directly from inserted data, instead of running a query.

src/MongoDBBusServiceProvider.php
GromNaN: That's how it's designed in Laravel: https://github.com/laravel/framework/blob/93658ff0ca839352a557bee2599cb5ec8d295d88/src/Illuminate/Bus/BusServiceProvider.php#L57

CHANGELOG.md
GromNaN: Added.

jmikola (approved):

docs/queues.txt
```diff
@@ -20,27 +20,107 @@ the driver in ``config/queue.php``:
...
+   :widths: 25 75
+
+   * - Setting
+     - Description
+
+   * - ``driver``
+     - **Required**. Specifies the queue driver to use. Must be ``mongodb``.
+
+   * - ``connection``
+     - Uses the default connection by default. The database connection used to store jobs. It must be a ``mongodb`` connection.
```
jordan-smith721: I think this would be more clear if the default is specified at the end, so the reader has more context. (also reworded a bit to avoid repeating "default")

```suggestion
     - The database connection used to store jobs. It must be a ``mongodb`` connection. The driver uses the default connection if a connection is not specified.
```

docs/queues.txt
```diff
@@ -20,27 +20,107 @@ the driver in ``config/queue.php``:
...
+
+   * - ``collection``
+     - **Required**. Name of the MongoDB collection to store jobs to process.
+
+   * - ``queue``
+     - **Required**. Name of the queue.
+
+   * - ``retry_after``
+     - Default ``60``. Specify how many seconds the queue connection should wait before retrying a job that is being processed
```
jordan-smith721: Change for consistency. Also reorganizing to match my earlier suggestion of showing the default last:
```suggestion
     - Specifies how many seconds the queue connection should wait before retrying a job that is being processed. Defaults to ``60``. 
```

docs/queues.txt
```diff
@@ -20,27 +20,107 @@ the driver in ``config/queue.php``:
...
+   :widths: 25 75
+
+   * - Setting
+     - Description
+
+   * - ``driver``
+     - **Required**. Specifies the queue driver to use. Must be ``mongodb``.
+
+   * - ``connection``
+     - Uses the default connection by default. The database connection used to store jobs. It must be a ``mongodb`` connection.
```
jordan-smith721:
```suggestion
     - The database connection used to store jobs. It must be a ``mongodb`` connection. The driver uses the default connection if a connection is not specified.
```

docs/queues.txt
```diff
@@ -20,27 +20,107 @@ the driver in ``config/queue.php``:
...
+
+   * - ``collection``
+     - Default ``failed_jobs``. Name of the MongoDB collection to store failed jobs.
```
jordan-smith721:
```suggestion
     - Name of the MongoDB collection to store failed jobs. Defaults to ``failed_jobs``. 
```

docs/queues.txt
```diff
@@ -20,27 +20,107 @@ the driver in ``config/queue.php``:
...
 
 .. code-block:: php
 
    MongoDB\Laravel\MongoDBQueueServiceProvider::class,
+
+
+Job Batching
+------------
+
+`Job batching https://laravel.com/docs/{+laravel-docs-version+}/queues#job-batching>`__
```
jordan-smith721:
```suggestion
`Job batching <https://laravel.com/docs/{+laravel-docs-version+}/queues#job-batching>`__
```

docs/queues.txt
```diff
@@ -20,27 +20,107 @@ the driver in ``config/queue.php``:
...
+is a Laravel feature to execute batch of jobs and subsequent actions before,
```
jordan-smith721:
```suggestion
is a Laravel feature to execute a batch of jobs and subsequent actions before,
```

docs/queues.txt
```diff
@@ -20,27 +20,107 @@ the driver in ``config/queue.php``:
...
+after, and during the execution of the jobs from the queue.
+
+With MongoDB, you don't have to create any collection before using this feature.
```
jordan-smith721: Suggestion: Clarify what "this feature" is since we're in a new paragraph
```suggestion
With MongoDB, you don't have to create any collection before using job batching.
```

docs/queues.txt
```diff
@@ -20,27 +20,107 @@ the driver in ``config/queue.php``:
...
+The collection ``job_batches`` will be created automatically to store meta
```
jordan-smith721: Change passive to active voice

```suggestion
The ``job_batches`` collection is created automatically to store meta
```

docs/queues.txt
```diff
@@ -20,27 +20,107 @@ the driver in ``config/queue.php``:
...
+   :widths: 25 75
+
+   * - Setting
+     - Description
+
+   * - ``driver``
+     - **Required**. Specifies the queue driver to use. Must be ``mongodb``.
+
+   * - ``connection``
+     - Uses the default connection by default. The database connection used to store jobs. It must be a ``mongodb`` connection.
```
jordan-smith721:
```suggestion
     - The database connection used to store jobs. It must be a ``mongodb`` connection. The driver uses the default connection if a connection is not specified.
```

docs/queues.txt
```diff
@@ -20,27 +20,107 @@ the driver in ``config/queue.php``:
...
+
+   * - ``collection``
+     - Default ``job_batches``. Name of the MongoDB collection to store job batches.
```
jordan-smith721:
```suggestion
     - Name of the MongoDB collection to store job batches. Defaults to ``job_batches``. 
```

jordan-smith721 (changes requested):
Just a few suggestions for the docs portion

docs/queues.txt
```diff
@@ -38,7 +38,7 @@ the driver in ``config/queue.php``:
      - **Required**. Specifies the queue driver to use. Must be ``mongodb``.
 
    * - ``connection``
-     - Uses the default connection by default. The database connection used to store jobs. It must be a ``mongodb`` connection.
+     - The database connection used to store jobs. It must be a ``mongodb`` connection. The driver uses the default connection if a connection is not specified.
```
rustagir, line 41:
```suggestion
     - Specifies the database connection used to store jobs. It must be a ``mongodb`` connection. The driver uses the default connection if a connection is not specified.
```

docs/queues.txt
```diff
@@ -20,27 +20,107 @@ the driver in ``config/queue.php``:
...
            // You can also specify your jobs specific database created on config/database.php
-           'connection' => 'mongodb-job',
-           'table' => 'jobs',
+           'connection' => 'mongodb',
+           'collection' => 'jobs',
            'queue' => 'default',
-           'expire' => 60,
+           'retry_after' => 60,
        ],
    ],
```
rustagir, line 29: S: add an introductory sentence before the table, such as:
"The following table describes settings that you can specify to configure the behavior of the queue:"

docs/queues.txt
```diff
@@ -20,27 +20,107 @@ the driver in ``config/queue.php``:
...
+     - Description
+
+   * - ``driver``
+     - **Required**. Specifies the queue driver to use. Must be ``mongodb``.
+
+   * - ``connection``
+     - The database connection used to store jobs. It must be a ``mongodb`` connection. The driver uses the default connection if a connection is not specified.
+
+   * - ``collection``
+     - **Required**. Name of the MongoDB collection to store jobs to process.
```
rustagir, line 44:
```suggestion
     - **Required**. Specifies the name of the MongoDB collection to store jobs to process.
```

rustagir: Applies to all entries

docs/queues.txt
```diff
@@ -20,27 +20,107 @@ the driver in ``config/queue.php``:
...
 .. code-block:: php
 
    'failed' => [
        'driver' => 'mongodb',
-       // You can also specify your jobs specific database created on config/database.php
-       'database' => 'mongodb-job',
-       'table' => 'failed_jobs',
+       'database' => 'mongodb',
+       'collection' => 'failed_jobs',
    ],
```
rustagir, line 62: S: same as earlier, add an introductory sentence to introduce this table

rustagir (approved):
approving with comments, when this PR is merged I will make an edits PR!

jordan-smith721 (approved):

PR #2905, opened 2024-04-23
PHPORM-171 Set timestamps when using `createOrFirst`

GromNaN:
Fix PHPORM-171
Issue reported by @nicohell https://github.com/mongodb/laravel-mongodb/issues/2720#issuecomment-2071709454

### Checklist

### DIFF
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -204,6 +204,12 @@ class Builder extends EloquentBuilder
         // Apply casting and default values to the attributes
         // In case of duplicate key between the attributes and the values, the values have priority
         $instance = $this->newModelInstance($values + $attributes);
+
+        /* @see \Illuminate\Database\Eloquent\Model::performInsert */
+        if ($instance->usesTimestamps()) {
+            $instance->updateTimestamps();
+        }
+
         $values = $instance->getAttributes();
         $attributes = array_intersect_key($attributes, $values);
 
--- a/tests/ModelTest.php
+++ b/tests/ModelTest.php
@@ -1048,12 +1048,17 @@ class ModelTest extends TestCase
 
     public function testCreateOrFirst()
     {
+        Carbon::setTestNow('2010-06-22');
+        $createdAt = Carbon::now()->getTimestamp();
         $user1 = User::createOrFirst(['email' => 'john.doe@example.com']);
 
         $this->assertSame('john.doe@example.com', $user1->email);
         $this->assertNull($user1->name);
         $this->assertTrue($user1->wasRecentlyCreated);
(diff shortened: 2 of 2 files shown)


src/Eloquent/Builder.php
```diff
@@ -204,6 +204,12 @@ public function createOrFirst(array $attributes = [], array $values = []): Model
         // Apply casting and default values to the attributes
         // In case of duplicate key between the attributes and the values, the values have priority
         $instance = $this->newModelInstance($values + $attributes);
+
+        /* @see \Illuminate\Database\Eloquent\Model::performInsert */
+        if ($instance->usesTimestamps()) {
+            $instance->updateTimestamps();
+        }
```
GromNaN, line 211: https://github.com/laravel/framework/blob/d1f17cff0285e8aa6a69a2f18d0ad7a2a0fd5ea8/src/Illuminate/Database/Eloquent/Model.php#L1290-L1295

tests/ModelTest.php
```diff
@@ -1048,12 +1048,17 @@ public function testNumericFieldName(): void
 
     public function testCreateOrFirst()
     {
+        Carbon::setTestNow('2010-06-22');
+        $createdAt = Carbon::now()->getTimestamp();
```
jmikola, line 1052: Consider `$currentTime` (or something more generic), since this is used for both assertions.

tests/ModelTest.php
```diff
@@ -1048,12 +1048,17 @@ public function testNumericFieldName(): void
...
         $user1 = User::createOrFirst(['email' => 'john.doe@example.com']);
 
         $this->assertSame('john.doe@example.com', $user1->email);
         $this->assertNull($user1->name);
         $this->assertTrue($user1->wasRecentlyCreated);
+        $this->assertEquals($createdAt, $user1->created_at->getTimestamp());
```
jmikola, line 1058: Should this use a comparison to assert that the `created_at` and `updated_at` fields are greater than or equal to the previously calculated current time?

I don't see anything in the PR that would guarantee the entire test completes within a single timestamp.

GromNaN: `Carbon::setTestNow` freezes the clock, so that the date point that is used to populate `create_at` is the one that is set.

tests/ModelTest.php
GromNaN: This is to be distinct from the `$updatedAt` that will be used to test update: https://github.com/mongodb/laravel-mongodb/pull/2906/files#diff-f1edeec321592c2aa6369488adcfa51ba57065b21063b4ef615c84b0f4fe2d2eR1115

jmikola (approved):

Issue #2922, opened 2024-05-01
PHPStan errors after 4.3.0 upgrade


promatik:
- Laravel-mongodb Version: 4.3.0
- PHP Version: 8.3.6
- Database Driver & Version: 6.0

### Description:

After upgrade to 4.3.0 there are dozens of [phpstan](https://github.com/phpstan/phpstan) errors, those errors are all similar;

```
Line   app\Http\Controllers\Admin\MailCrudController.php
:58    Parameter #2 $room expects App\Models\Talk\Room, Illuminate\Database\Eloquent\Model|null given.
```
```
Line   app\Listeners\Talk\CallEndedListener.php
:26    Access to an undefined property Illuminate\Database\Eloquent\Collection<int, Illuminate\Database\Eloquent\Model>|Illuminate\Database\Eloquent\Model::$id.
```

I don't know exactly how phpstan works, but before (v4.2) the models were being correctly interpreted. Now with v4.3.0 all Mongo models are being considered `Illuminate\Database\Eloquent\Model|null` so we have these errors like `Access to an undefined property ID`.

### Expected behaviour
PHP Stan still working, correctly inferring the model.

### Actual behaviour
There are now dozens of phpstan errors. Mongo models are always interpreted as `Illuminate\Database\Eloquent\Model`.

promatik:
I'll close this one since this is most probably an issue in Larastan.

PR #2923, opened 2024-05-01
Removed `@mixin Builder` comment from Model.php

promatik:
Fix for https://github.com/mongodb/laravel-mongodb/issues/2922.

This change https://github.com/mongodb/laravel-mongodb/commit/fdfb5e5027bf46744c28862d042b3063c7d7a782#diff-284e164f76aeb13f424d154f9208d91fdbb0112b0a50ad8b1dfa237176f03cf2 introduced a breaking change for phpstan, it now wrongly infers models as `Illuminate\Database\Eloquent\Model`, instead of it's main class.

Salut @GromNaN!
As the author of that change, is it necessary? Can we go back to what it was?

### Checklist

- [ ] ~~Add tests and ensure they pass~~
- [ ] ~~Update documentation for new features~~

### DIFF
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -49,6 +49,7 @@ use function strcmp;
 use function uniqid;
 use function var_export;
 
+/** @mixin QueryBuilder */
 abstract class Model extends BaseModel
 {
     use HybridRelations;
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -49,7 +49,6 @@ use function strcmp;
 use function uniqid;
 use function var_export;
 
-/** @mixin Builder */
 abstract class Model extends BaseModel
 {
     use HybridRelations;


GromNaN:
Thanks for the feedback. Could you provide an example of static analysis that is broken with this change, that was not before. Thanks.

GromNaN:
From my POV, this annotation enables autocompletion of builder's methods in PHPStorm.
If that creates an issue, the solution will be to declare all the methods with `@method` annotation.

promatik:
> Thanks for the feedback. Could you provide an example of static analysis that is broken with this change, that was not before. Thanks.

Hi @GromNaN, yes, I added a few examples in the issue; https://github.com/mongodb/laravel-mongodb/issues/2922.

Basically all my Models that used to extend the Mongo Model, now "have become" `Illuminate\Database\Eloquent\Model|null`.

For instance I had a `Room` model that extends `MongoDB\Laravel\Eloquent\Model`, after this change anywhere I used that model I have errors like;

```
Parameter #2 $room expects App\Models\Talk\Room, Illuminate\Database\Eloquent\Model|null given.
```

or:

```
Access to an undefined property Illuminate\Database\Eloquent\Collection<int, Illuminate\Database\Eloquent\Model>|Illuminate\Database\Eloquent\Model::$id.
```

promatik:
**UPDATE:**

@GromNaN I think I found the issue;

```php
use ...
use MongoDB\Laravel\Query\Builder as QueryBuilder;
use ...
```

Probably PHPStorm understands that Builder is actually the QueryBuilder, but PHPStand doesn't.

I changed `Builder` to `QueryBuilder` and PHPStan is not throwing me any errors 🎉

Also updated the PR.
Can you please test this on PHPStorn?

GromNaN:
That's a different class, the exposed methods are wrong with this change.
Example, `MongoDB\Laravel\Eloquent\Builder::aggregates` returns `AggregationBuilder | mixed` while `MongoDB\Laravel\Query\Builder::aggregates` returns `AggregationBuilder | self`.

I tried to reproduce with PHPStan but didn't find the bug.
https://phpstan.org/r/10adb986-2c43-4b77-9a80-6f31d562441f

Try removing or changing the `@mixin` annotation to see the error.

promatik:
Hey @GromNaN!

Yeah, sorry for the mistake, now I see it's a different class.

Anyway, I can see mixin works as expected in your demo, but I can't fix my issue 🫤
With the mixing there, my model wrongly becomes a `Illuminate\Database\Eloquent\Model`.

```
 ------ -------------------------------------------------------------------------------------- 
  Line   2024_04_26_225522_add_share_slug_to_calls_table.php
 ------ --------------------------------------------------------------------------------------
  :16    Access to an undefined property Illuminate\Database\Eloquent\Model::$share_code.
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property
 ------ --------------------------------------------------------------------------------------
```

By the way, I'm using `larastan/larastan` and I think the problem comes from there, because I just tried phpstan alone, and everything works 🤷‍♂️

Let's move the issue to Larastan 👌
https://github.com/larastan/larastan/issues/1940

promatik:
Hey @GromNaN, sorry for bothering again.
Based on this comment; https://github.com/larastan/larastan/issues/1940#issuecomment-2102532494

Might this be an issue on this implementation? I have no idea about these inner ways of work of stan.

I'm trying to upgrade to v4.3, but the 30+ phpstan errors are keeping me away form it.

GromNaN:
Fixed by #2981

promatik:
Great!
Thank you @GromNaN 🎉

PR #2931, opened 2024-05-03
Implement `Builder::toSql` using JSON representation of the MQL

### LINKED ISSUE #2797: toSql() Alias from toMql()
### Is your feature request related to an issue?
In some plugins such as rappasoft/laravel-liveware-tables the debug uses the toSql() function and when I used it at that time it gave an error

### Describe the solution you would like
It could be that they aliased the toMql() function to use it as toSql() and then that exception would not arise.

### Additional context
In my case, what I did was make a fork of rappasoft/laravel-liveware-tables to be able to use it with mongodb, because it also makes several somewhat strange functions and I changed the function from toSql() to toMql()

GromNaN:
Fix #2797

The method `toSql` is used by third party packages like [`rappasoft/laravel-liveware-tables`](https://github.com/rappasoft/laravel-livewire-tables/blob/2bd477ec1d5b10fede53cfa0cbc8ffbb815b3a74/src/DataTransferObjects/DebuggableData.php#L19). 

### Checklist

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -18,6 +18,7 @@ use Illuminate\Support\LazyCollection;
 use InvalidArgumentException;
 use LogicException;
 use MongoDB\BSON\Binary;
+use MongoDB\BSON\Document;
 use MongoDB\BSON\ObjectID;
 use MongoDB\BSON\Regex;
 use MongoDB\BSON\UTCDateTime;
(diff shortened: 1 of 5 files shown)


src/Query/Builder.php
```diff
@@ -1535,4 +1546,10 @@ public function orWhereIntegerNotInRaw($column, $values, $boolean = 'and')
     {
         throw new BadMethodCallException('This method is not supported by MongoDB');
     }
+
+    /** @internal This method is not supported by MongoDB. */
+    public function insertUsing(array $columns, $query)
```
GromNaN, line 1551: We relied on the exception thrown by the parent method that uses `toSql`.

src/Query/Builder.php
```diff
@@ -1446,16 +1447,26 @@ public function __call($method, $parameters)
         return parent::__call($method, $parameters);
     }
 
-    /** @internal This method is not supported by MongoDB. */
+    /**
+     * Return the JSON representation of the MongoDB Query.
+     * Naming is inherited and used for compatibility with third-party packages.
+     *
+     * @internal
```
jmikola, line 1454: Should this still be `@internal`?

I don't have enough context to understand how/why [rappasoft/laravel-livewire-tables](https://github.com/rappasoft/laravel-livewire-tables) needs to support MongoDB, but isn't there some concern that returning JSON strings here might break _other_ packages that expect these methods to return actual SQL statements?

My thinking is that packages wishing to support this library should be the ones to change and conditionally call `toMql()`.

GromNaN: It's used for debugging: https://tables.laravel-boilerplate.com/tailwind
But [Laravel Debugbar](https://github.com/barryvdh/laravel-debugbar) would provide the same information. 

And I agree this can lead to other problems if packages expect SQL and use the returned value as such.

PR #2937, opened 2024-05-08
PHPORM-181: Add SBOM lite

alcaeus:
PHPORM-181

### DIFF
--- a/sbom.json
+++ b/sbom.json
@@ -1,0 +1,85 @@
+{
+    "$schema": "http://cyclonedx.org/schema/bom-1.5.schema.json",
+    "bomFormat": "CycloneDX",
+    "specVersion": "1.5",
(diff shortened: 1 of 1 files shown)


sbom.json
```diff
@@ -0,0 +1,85 @@
+{
```
alcaeus, line 1: The file was generated using the cyclonedx/cyclonedx-php-composer library to generate an SBOM from composer.lock, then manually removing all components and dependencies as we only report bundled dependencies:
```shell
composer require --dev cyclonedx/cyclonedx-php-composer
composer CycloneDX:make-sbom --output-file=sbom.json --output-format=JSON --omit=dev --omit=plugin
```

jmikola (approved):
Noted this is the same as https://github.com/mongodb/mongo-php-library/pull/1292.

PR #2965, opened 2024-05-22
PHPORM-179: Upload PHPStan report to GitHub Code Scanning

alcaeus:
PHPORM-179

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -108,7 +108,13 @@ jobs:
             phpstan-result-cache-
 
       - name: Run PHPStan
-        run: ./vendor/bin/phpstan analyse --no-interaction --no-progress --ansi
+        run: ./vendor/bin/phpstan analyse --no-interaction --no-progress --ansi --error-format=sarif > phpstan.sarif
+
+      - name: "Upload SARIF report"
+        if: always()
+        uses: "github/codeql-action/upload-sarif@v3"
+        with:
+          sarif_file: phpstan.sarif
 
       - name: Save cache PHPStan results
         id: phpstan-cache-save
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -14,3 +14,11 @@ parameters:
     ignoreErrors:
         - '#Unsafe usage of new static#'
         - '#Call to an undefined method [a-zA-Z0-9\\_\<\>]+::[a-zA-Z]+\(\)#'
+
+services:
+    errorFormatter.sarif:
+        class: MongoDB\Laravel\Tests\PHPStan\SarifErrorFormatter
+        arguments:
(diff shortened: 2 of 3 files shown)


src/PHPStan/SarifErrorFormatter.php
```diff
@@ -0,0 +1,129 @@
...
+use PHPStan\Command\AnalysisResult;
+use PHPStan\Command\ErrorFormatter\ErrorFormatter;
+use PHPStan\Command\Output;
+use PHPStan\File\RelativePathHelper;
+use PHPStan\Internal\ComposerHelper;
+
+use function array_values;
+
+/** @internal */
+class SarifErrorFormatter implements ErrorFormatter
```
GromNaN: I don't like that we have this class in the package. Maybe create a 3rd party package that we add as dev dependency?

src/Connection.php
```diff
@@ -202,7 +202,7 @@
      */
     public function ping(): void
     {
-        $this->getMongoClient()->getManager()->selectServer(new ReadPreference(ReadPreference::PRIMARY_PREFERRED));
+        $this->getMongoClient()->getManager()->selectServer(new ReadPreference(ReadPreference::PRIMARY_PREFERRED), true);
```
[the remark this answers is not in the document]
GromNaN: We can't dismiss outdated messages like this?

src/PHPStan/SarifErrorFormatter.php
GromNaN: Or maybe you'll do a PR to integrate it into PHPStan directly?

GromNaN: Or use the suggested package: https://github.com/phpstan/phpstan/issues/5973#issuecomment-1407681159

alcaeus: I'll point you to the comment at the top of the file; I started out with the 3rd party package but wanted to add a rule. Given that the 3rd-party package has little to no activity and little to no installs, there's also no point in relying on an external tool.

src/Connection.php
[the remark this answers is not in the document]
alcaeus: Not sure if it was there before or only because you commented, but I see a "Resolve conversation" button here. In the code scanning results (see the link in the first comment) the error is marked as "fixed".

src/PHPStan/SarifErrorFormatter.php
GromNaN: Ok, so let's move the class into the `tests` directory, so that it's not in the package artefact.

alcaeus: It's not a test, so I've gone ahead and excluded it via the .gitattributes file

alcaeus: Moved.

GromNaN (approved):

PR #2967, opened 2024-05-22
[fix] Prevent exception when connection is closed

SanderMuller:
When the MongoDB connection disconnects, it currently unsets the connection property altogether. This causes the following exception when trying to use a MongoDB connection afterwards: `Prevent Undefined property: MongoDB\Laravel\Connection::$connection`

This PR sets the connection (client) to `null` instead of using `unset()`.

Unfortunately there doesn't seem to be a reconnector configured by default, otherwise `$connection->reconnect()` could be used, but this has no effect right now. Could be a good next step for someone to implement.

It's possible that this issue only occurs when using Database Transactions (that's my use case. Didn't get any exceptions locally despite long periods of heavy load, but on our Atlas instance it is getting this exception a real lot)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 ## [4.3.1]
 
 * Fix memory leak when filling nested fields using dot notation by @GromNaN in [#2962](https://github.com/mongodb/laravel-mongodb/pull/2962)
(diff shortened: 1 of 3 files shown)


GromNaN (approved):
LGTM. I'll change the target branch to 4.3 and upmerge into 4.4.

GromNaN:
Thank you @SanderMuller

Issue #2945, opened 2024-05-14
updateOrCreate -> createOrFirst doesn't work as expected


Boevsson:
- Laravel-mongodb Version: 4.2
- PHP Version: 8.3.4
- Database Driver & Version: mongod 5.0 / 7.0

### Steps to reproduce
1. $id = $request->get('id') ?? null; // in the case id is null
2. Do `Model::updateOrCreate(['_id' => $id], $data);`

### Expected behaviour

A new record should be created with the given data and a generated _id.

### Actual behaviour

It queries the first record from the collection and updates it.

This works fine when using the package 4.0 because the createOrFirst is not overriden there in the Mongo Builder class

GromNaN:
Thanks for the report.
Related to #2935
~Tracked in Jira PHPORM-183~

GromNaN:
After investigation, `null` is a valid value for MongoDB `_id`. That's why the `Model::updateOrCreate(['_id' => null], $data)` creates or updates a new document with `_id: null`.

This is inconsistent with how SQL works because SQL databases will generate a value with a request like this: `INSERT INTO table (id, name) VALUES (null, 'foo')`.

Can you modify your code like this:

```php
$id = $request->get('id'); // null is the default value
if ($id) {
    $model = Model::updateOrCreate(['_id' => $id], $data);
} else {
    $model = Model::create($data);
}
```

GromNaN:
Closing as this is the expected behavior for MongoDB.

PR #2969, opened 2024-05-22
Unset `_id: null` to let it be autogenerated

GromNaN:
Tentative fix for #2945

### Checklist

### DIFF
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -234,8 +234,8 @@ class Builder extends EloquentBuilder
             try {
                 $document = $collection->findOneAndUpdate(
                     $attributes,
-                    // Before MongoDB 5.0, $setOnInsert requires a non-empty document,
-                    // this should not be an issue as $values include the query filter.
+                    // Before MongoDB 5.0, $setOnInsert requires a non-empty document.
+                    // This should not be an issue as $values includes the query filter.
                     ['$setOnInsert' => (object) $values],
                     [
                         'upsert' => true,
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -21,6 +21,7 @@ use function array_merge;
 use function collect;
 use function is_array;
 use function iterator_to_array;
+use function json_encode;
 
 /** @method \MongoDB\Laravel\Query\Builder toBase() */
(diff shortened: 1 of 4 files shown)


src/Eloquent/Builder.php
```diff
@@ -233,8 +234,8 @@ public function createOrFirst(array $attributes = [], array $values = []): Model
             try {
                 $document = $collection->findOneAndUpdate(
                     $attributes,
-                    // Before MongoDB 5.0, $setOnInsert requires a non-empty document.
-                    // This should not be an issue as $values includes the query filter.
+                    // Before MongoDB 5.0, $setOnInsert requires a non-empty document,
+                    // this should not be an issue as $values include the query filter.
```
jmikola: Grammar was correct as-is. `$values` is a singular variable, so the verb would be plural.

tests/ModelTest.php
```diff
@@ -1151,4 +1151,21 @@ public function testUpdateOrCreate(array $criteria)
...
+    {
+        $user = User::create(['_id' => null, 'email' => 'foo@bar']);
+        $this->assertNotNull(ObjectId::class, $user->id);
+        $this->assertSame(1, User::count());
+    }
+
+    public function testUpdateOrCreateWithNullId()
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('You must provide attributes to check for duplicates');
```
jmikola, line 1165: Noted that `expectExceptionMessage()` is a substring match.

tests/ModelTest.php
```diff
@@ -1151,4 +1151,21 @@ public function testUpdateOrCreate(array $criteria)
...
+        User::updateOrCreate(
+            ['_id' => null],
```
jmikola, line 1167: Do users typically pass an explicit `null` value to this method, or would this argument come from another source such as an empty model object?

GromNaN: From the [original issue](https://github.com/mongodb/laravel-mongodb/issues/2945), `null` is a fallback value.

```php
$id = $request->get('id') ?? null;
Model::updateOrCreate(['_id' => $id], $data);
```

jmikola (approved):

PR #2980, opened 2024-05-27
PHPORM-180 Trigger events in Model::createOrFirst

### LINKED ISSUE #2935: `createOrFirst` is not triggering `creating` nor `created` model events
- Laravel-mongodb Version: 4.3.0
- Laravel version: 11.6.0
- PHP Version: 8.2.2

### Description:

There seem to be a problem, where using the `createOrFirst` builder method, that it does not trigger the `creating` or `created` model events when the model is being created (not retrieved, as there is no matching model found). 

This seems to be caused by the fact that the `createOrFirst` method is overwritten in MongoDB package, but it is not calling the `$model->save()` method after an attempt to find the model, which would trigger those two events. 

This is how it behaved in the past (I just upgraded from v3 to v4) and it is how it behaves in Laravel without MongoDB.

### Steps to reproduce
1. set up a simple Model class
2. call `Model::query()->createOrFirst()` on it 
3. see that it won't trigger the `creating` and `created` event if it recently created the model (there was no matching model for the attributes) 

This will however trigger the `retrieving` event instead - even if the model was recently created, but not actually "retrieved"

### Expected behaviour

When a model is being created as a new one with use of `createOrFirst` method, it should trigger the `creating` and `created` model events. 

### Actual behaviour
When a model is being created as a new one with use of `createOrFirst` method, there is no `creating` nor `created` events dispatched. 

### Workaround 

At the moment the only workaround I could think of is to manually try to fetch the model first with use of find/where query, and if its not found, create it with use of `Model::query()->create()` method, or `$model->save()` - which both will dispatch the `creating` and `created` events.

GromNaN:
Fix PHPORM-180 
Fix #2935

In order to trigger the same events as Eloquent's `createOrFirst`, the process must be in the `Model` class. So I created a new `saveOrFirst` method that needs to be public to be called by the `Eloquent\Query` class, but I want to keep it internal. I tried to add an option to `Model::save`, but this method returns an boolean and I need the object if found.

### Checklist

### DIFF
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -8,16 +8,13 @@ use Illuminate\Database\ConnectionInterface;
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
 use InvalidArgumentException;
 use MongoDB\Driver\Cursor;
-use MongoDB\Laravel\Collection;
 use MongoDB\Laravel\Helpers\QueriesRelationships;
-use MongoDB\Laravel\Internal\FindAndModifyCommandSubscriber;
 use MongoDB\Laravel\Query\AggregationBuilder;
 use MongoDB\Model\BSONDocument;
-use MongoDB\Operation\FindOneAndUpdate;
 
-use function array_intersect_key;
 use function array_key_exists;
 use function array_merge;
+use function assert;
 use function collect;
 use function is_array;
 use function iterator_to_array;
@@ -218,40 +215,9 @@ class Builder extends EloquentBuilder
         // Apply casting and default values to the attributes
         // In case of duplicate key between the attributes and the values, the values have priority
         $instance = $this->newModelInstance($values + $attributes);
+        assert($instance instanceof Model);
 
-        /* @see \Illuminate\Database\Eloquent\Model::performInsert */
-        if ($instance->usesTimestamps()) {
-            $instance->updateTimestamps();
-        }
-
-        $values = $instance->getAttributes();
-        $attributes = array_intersect_key($attributes, $values);
-
-        return $this->raw(function (Collection $collection) use ($attributes, $values) {
-            $listener = new FindAndModifyCommandSubscriber();
-            $collection->getManager()->addSubscriber($listener);
-
-            try {
-                $document = $collection->findOneAndUpdate(
-                    $attributes,
-                    // Before MongoDB 5.0, $setOnInsert requires a non-empty document.
-                    // This should not be an issue as $values includes the query filter.
-                    ['$setOnInsert' => (object) $values],
-                    [
-                        'upsert' => true,
-                        'returnDocument' => FindOneAndUpdate::RETURN_DOCUMENT_AFTER,
-                        'typeMap' => ['root' => 'array', 'document' => 'array'],
-                    ],
-                );
-            } finally {
-                $collection->getManager()->removeSubscriber($listener);
-            }
-
-            $model = $this->model->newFromBuilder($document);
-            $model->wasRecentlyCreated = $listener->created;
-
-            return $model;
-        });
+        return $instance->saveOrFirst($attributes);
     }
 
     /**
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace MongoDB\Laravel\Eloquent;
 
 use BackedEnum;
+use BadMethodCallException;
 use Carbon\CarbonInterface;
 use DateTimeInterface;
 use DateTimeZone;
@@ -31,6 +32,7 @@ use function array_keys;
 use function array_merge;
 use function array_unique;
 use function array_values;
+use function assert;
 use function class_basename;
 use function count;
 use function date_default_timezone_get;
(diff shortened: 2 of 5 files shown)


src/Eloquent/Model.php
```diff
@@ -759,6 +761,71 @@ public function save(array $options = [])
...
+        // update timestamps on this model, which are maintained by us for developer
+        // convenience. After, we will just continue saving these model instances.
+        if ($this->usesTimestamps()) {
+            $this->updateTimestamps();
+        }
+
+        // If the model has an incrementing key, we can use the "insertGetId" method on
+        // the query builder, which will give us back the final inserted ID for this
+        // table from the database. Not all tables have to be incrementing though.
+        $attributes = $this->getAttributesForInsert();
```
jmikola, line 801: Why does the comment above refer to `insertGetId()`? It's not clear to me how that's getting used here, unless it's through `getAttributesForInsert()`.

Does `getAttributesForInsert` respect modifications from earlier model methods such as `setUniqueIds()` and `updateTimestamps()`? I assume this all pulls from the same state.

src/Eloquent/Builder.php
```diff
@@ -218,40 +215,9 @@ public function createOrFirst(array $attributes = [], array $values = []): Model
...
-            } finally {
-                $collection->getManager()->removeSubscriber($listener);
-            }
-
-            $model = $this->model->newFromBuilder($document);
-            $model->wasRecentlyCreated = $listener->created;
-
-            return $model;
-        });
+        return $instance->saveOrFirst($attributes);
```
jmikola, line 220: I don't recall the original reason for introducing `createOrFirst()` in https://github.com/mongodb/laravel-mongodb/pull/2742 ([PHPORM-139](https://jira.mongodb.org/browse/PHPORM-139)). It looks like this was necessary to avoid some issue with transactions.

Just wanted to confirm that a custom implementation is still required at all, given that we'll no longer be using `findAndModify`.

src/Eloquent/Model.php
```diff
@@ -759,6 +761,71 @@ public function save(array $options = [])
         return $saved;
     }
 
+    /** @internal Not part of Laravel Eloquent API. Use raw findOneAndModify if necessary */
+    public function saveOrFirst(array $criteria): ?static
+    {
+        $this->mergeAttributesFromCachedCasts();
```
jmikola, line 767: I assume much of the contents for this method were lifted from an existing method in Eloquent. If so, can we reference that in the method doc block?

I'm a little worried that if anything changes upstream we'll probably need to find out the hard way that this needs to be updated, but that may be unavoidable.

src/Eloquent/Model.php
```diff
@@ -759,6 +761,71 @@ public function save(array $options = [])
...
+        // table from the database. Not all tables have to be incrementing though.
+        $attributes = $this->getAttributesForInsert();
+
+        $document = $query->getQuery()
+            ->where($criteria)
+            ->insertOrFirst($attributes);
+
+        $connection = $query->getConnection();
+        if (! $this->getConnectionName() && $connection) {
+            $this->setConnection($connection->getName());
```
jmikola, line 809: What's the reason for inheriting the query's connection? IS this necessary for dispatching the `created` event?

src/Eloquent/Model.php
```diff
@@ -759,6 +761,71 @@ public function save(array $options = [])
...
+        }
+
+        // If a document matching the criteria was found, it is returned. Nothing was saved.
+        if (! $document['wasRecentlyCreated']) {
+            return $this->newInstance($document);
+        }
+
+        // If the model is successfully saved, we need to do a few more things once
+        // that is done. We will call the "saved" method here to run any actions
```
jmikola, line 818: I don't see any reference to a `saved` method. Is this comment accurate?

tests/ModelTest.php
```diff
@@ -1073,7 +1078,9 @@ public function testCreateOrFirst()
         $this->assertFalse($user2->wasRecentlyCreated);
         $this->assertEquals($createdAt, $user1->created_at->getTimestamp());
         $this->assertEquals($createdAt, $user1->updated_at->getTimestamp());
+        $this->assertEquals(['saving', 'creating'], $events);
```
jmikola, line 1081: Does the preceding call to `User::createOrFirst()` return via the `wasRecentlyCreated` check? I'm trying to understand how only these two events are dispatched.

Some comments within the test method might be helpful, similar to how you had `// Update` within `testUpdateOrCreate()`.

src/Eloquent/Builder.php
GromNaN: We are still using `findAndModify`, but in `Query\Builder::findAndModify`.

The call chain is:
- `Model::createOrFirst`
- `Eloquent\Query::createOrFirst`
- `Model::saveOrFirst`
- `Query\Builder::findAndModify`
- `Collection::findAndModify`

src/Eloquent/Model.php
GromNaN: I'm also concerned that we have a lot of duplicated code that is hard to understand and we are not sure we can maintain in the long-term.
Creating a simple alternative PR: https://github.com/mongodb/laravel-mongodb/pull/2984

tests/ModelTest.php
GromNaN: This events are dispatched before the DB call. And the others are dispatched only when the document was created by this call. This is consistent with Laravel's behavior.

GromNaN:
Replaced by #2984

PR #2981, opened 2024-05-27
Remove `@mixin` on Model class

GromNaN:
Fix #2923

It seems that `@mixin` is too naive, doesn't support all dynamic methods and doesn't translate return types correctly.

### Checklist

### DIFF
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -48,7 +48,6 @@ use function str_starts_with;
 use function strcmp;
 use function var_export;
 
-/** @mixin Builder */
 abstract class Model extends BaseModel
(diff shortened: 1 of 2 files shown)


src/Eloquent/Model.php
```diff
@@ -48,7 +48,6 @@
 use function strcmp;
 use function var_export;
 
-/** @mixin Builder */
```
jmikola, line 51: I noted there were some other solutions proposed in https://github.com/larastan/larastan/issues/1940#issuecomment-2125174053 but I assume they're not something we can fix independently within this package.

jmikola (approved):

PR #2984, opened 2024-05-29
PHPORM-180 Keep createOrFirst in 2 commands to simplify implementation

GromNaN:
Fix PHPORM-180
Alternative to #2980

Revert optimisation from #2742 in order to keep all the processes from Eloquent and simplify maintenance.

The main feature difference is that it's requiring a unique index whereas my previous implementation was only using the filter "attributes". This new implementation is closer to [Laravel's implementation](https://github.com/laravel/framework/blob/33340933acfa47b64cc1f87fa61e03b7d7dd1325/src/Illuminate/Database/Eloquent/Builder.php#L589-L596), it only removes the nested transaction and catch the MongoDB exception instead of the PDO exception.

### Checklist


GromNaN:
Unfortunately, that doesn't solve the nested transaction issue reported initially #2720.

```
MongoDB\Driver\Exception\CommandException: Transaction with { txnNumber: 1 } has been aborted.
```

src/Eloquent/Builder.php
```diff
@@ -202,56 +198,17 @@ public function raw($value = null)
...
-                        'returnDocument' => FindOneAndUpdate::RETURN_DOCUMENT_AFTER,
-                        'typeMap' => ['root' => 'array', 'document' => 'array'],
-                    ],
-                );
-            } finally {
-                $collection->getManager()->removeSubscriber($listener);
+        try {
+            return $this->create(array_merge($attributes, $values));
+        } catch (WriteException $e) {
+            if ($e->getCode() === self::DUPLICATE_KEY_ERROR) {
```
jmikola, line 226: This approach seems untenable if any DuplicateKeyError will cause the transaction to be aborted.

Do I assume correctly that the library isn't creating its own transaction here, and it's only an issue when the application calls this _within_ its own transaction (as demonstrated in https://github.com/mongodb/laravel-mongodb/issues/2720)?

I think that in order to avoid DuplicateKeyError potentially aborting, we'd need to flip this to the original design of `firstOrCreate()` and only insert if the initial query found no match. That wouldn't eliminate the possibility of a write conflict between _two_ transactions, though.

jmikola: Alternatively, maybe you can check if you're already in a transaction and only then resort to an initial `find()` to try and avoid a DuplicateKeyError.

You could argue that two transactions conflicting with one another is a more general situation that users should anticipate.

GromNaN: Some events (`saving` and `creating`) are triggered by `createOrFirst` but not by `firstOrCreate` when the document is found.

GromNaN: I implemented your suggestion. When in transaction, `createOrFirst` fallback to `firstOrCreate`.

src/Eloquent/Builder.php
```diff
@@ -203,55 +199,42 @@ public function raw($value = null)
...
-     * If the record exists, it will be returned.
+     * Get the first record matching the attributes. If the record is not found, create it.
      *
-     * @param  array $attributes The attributes to check for duplicate records
-     * @param  array $values     The attributes to insert if no matching record is found
+     * @param  array $attributes
+     * @param  array $values
      */
-    public function createOrFirst(array $attributes = [], array $values = []): Model
+    public function firstOrCreate(array $attributes = [], array $values = []): Model|Builder
```
jmikola: When would this return a Builder instance instead of a Model?

src/Eloquent/Builder.php
```diff
@@ -203,55 +199,42 @@ public function raw($value = null)
...
+        if ($instance !== null) {
+            return $instance;
         }
 
-        // Apply casting and default values to the attributes
-        // In case of duplicate key between the attributes and the values, the values have priority
-        $instance = $this->newModelInstance($values + $attributes);
+        // createOrFirst is not supported in transaction.
+        if ($this->getConnection()->getSession()?->isInTransaction()) {
+            return $this->create(array_merge($attributes, $values));
```
jmikola, line 210: Should `+` be used instead of `array_merge` to ensure that string _and_ numeric keys overwrite each other instead of possibly appending numerics?

tests/ModelTest.php
```diff
@@ -1086,21 +1107,21 @@ public function testCreateOrFirst()
...
+        $events = [];
         $user4 = User::createOrFirst(
             ['name' => 'Robert Doe'],
             ['name' => 'Maria Doe', 'email' => 'maria.doe@example.com'],
         );
 
         $this->assertSame('Maria Doe', $user4->name);
         $this->assertTrue($user4->wasRecentlyCreated);
-    }
+        $this->assertEquals(['saving', 'creating', 'created', 'saved'], $events);
```
jmikola, line 1124: Noted that these events are asserted _before_ the transaction commits. I assume that means they're not reliable for inferring that changes were actually persisted to the database.

This reminded me of a PR @alcaeus recently worked on in Doctrine ODM re: lifecycle events and transactions.

tests/ModelTest.php
```diff
@@ -1048,19 +1048,34 @@ public function testNumericFieldName(): void
...
 
         $this->assertSame('john.doe@example.com', $user1->email);
         $this->assertNull($user1->name);
         $this->assertTrue($user1->wasRecentlyCreated);
         $this->assertEquals($createdAt, $user1->created_at->getTimestamp());
         $this->assertEquals($createdAt, $user1->updated_at->getTimestamp());
+        $this->assertEquals(['saving', 'creating', 'created', 'saved'], $events);
 
         Carbon::setTestNow('2020-12-28');
+        $events = [];
```
jmikola, line 1078: What's the point of calling `Carbon::setTestNow()` above? Is it related asserting `created_at` and `updated_at` still have the same value (i.e. the model was not modified)?

I suppose those assertions are consistent with expecting `$user2->wasRecentlyCreated` to still be false.

tests/ModelTest.php
```diff
@@ -1073,7 +1088,13 @@ public function testCreateOrFirst()
         $this->assertFalse($user2->wasRecentlyCreated);
         $this->assertEquals($createdAt, $user1->created_at->getTimestamp());
         $this->assertEquals($createdAt, $user1->updated_at->getTimestamp());
+        if ($transaction) {
+            $this->assertEquals([], $events);
+        } else {
+            $this->assertEquals(['saving', 'creating'], $events);
+        }
```
jmikola, line 1099: Why would there be no events when executing within a transaction? Considering both code paths:

 * When we're in a transaction, `User::createOrFirst` will delegate to `firstOrCreate` and return `(clone $this)->where($attributes)->first()`.
 * When we're not in a transaction, `createOrFirst()` will return `$this->create(array_merge($attributes, $values))`. I assume that always dispatches these two events regardless of whether the document exists server-side.

Is there any concern with these event expectations differing from Eloquent?

Either way, I think a comment explaining different assertions would be helpful.

tests/ModelTest.php
GromNaN: Yes, that ensures the timestamps are not modified.

tests/ModelTest.php
GromNaN: This 2 events are dispatched in the `Eloquent\Query::create` callchain:
- "save" in [`Model::save`](https://github.com/laravel/framework/blob/97238d27e271d6f8b36b64680756f788130f19c5/src/Illuminate/Database/Eloquent/Model.php#L1126)
- "creating" in [`Model::performInsert`](https://github.com/laravel/framework/blob/97238d27e271d6f8b36b64680756f788130f19c5/src/Illuminate/Database/Eloquent/Model.php#L1286)

Then the "created" and "saved" events are not dispatched because of the server exception.

None of this events are dispatched when using `firstOrCreate` because `find` does not trigger this events and return a value.

Comment added.

tests/ModelTest.php
GromNaN: There is no commit or UoW in Eloquent. Every operation is sent to the DB directly, and events are dispatched in the process, even if the transaction is rollback later.

src/Eloquent/Builder.php
GromNaN: Copied from https://github.com/laravel/framework/blob/97238d27e271d6f8b36b64680756f788130f19c5/src/Illuminate/Database/Eloquent/Builder.php#L592

src/Eloquent/Builder.php
GromNaN: Never, I was fooled by the `@return Model|static` [in the parent class](https://github.com/laravel/framework/blob/97238d27e271d6f8b36b64680756f788130f19c5/src/Illuminate/Database/Eloquent/Builder.php#L571).

src/Eloquent/Builder.php
```diff
@@ -202,56 +198,37 @@ public function raw($value = null)
...
 
-    /**
-     * Attempt to create the record if it does not exist with the matching attributes.
-     * If the record exists, it will be returned.
-     *
-     * @param  array $attributes The attributes to check for duplicate records
-     * @param  array $values     The attributes to insert if no matching record is found
-     */
-    public function createOrFirst(array $attributes = [], array $values = []): Model
+    public function firstOrCreate(array $attributes = [], array $values = [])
```
jmikola, line 201: I'm not actually suggesting this, but it's amusing to me that this entire method body could be reduced to:

```php
return (clone $this)->where($attributes)->first() ??
    $this->getConnection()->getSession()?->isInTransaction()
        ? $this->create(array_merge($attributes, $values))
        : $this->createOrFirst($attributes, $values);
```

jmikola (approved):

src/Eloquent/Builder.php
GromNaN: Sure, but it won't help me to use xdebug's step-by-step debugger.

PR #2985, opened 2024-05-30
PHPORM-186 GridFS adapter for Filesystem

GromNaN:
Fix PHPORM-186

Integration of [Flysystem GridFS Adapter](https://flysystem.thephpleague.com/docs/adapter/gridfs/) for [Laravel File Storage](https://laravel.com/docs/11.x/filesystem).

The configuration is similar to the [Symfony Flysystem Bundle](https://github.com/thephpleague/flysystem-bundle/pull/162) as it provides various way to configure the bucket:
- using the database connection (default connection by default)
- setting a service name for the `bucket` instance
- setting a factory function for the `bucket` instance

Unlike the Symfony Bundle, I do not provide instantiation of a new `MongoDB\Client` from the configuration as we always have a database connection with Eloquent.

### Checklist


docs/filesystems.txt
```diff
@@ -0,0 +1,144 @@
...
+.. meta::
+   :keywords: php framework, gridfs, code example
+
+Overview
+--------
+
+Using MongoDB to store files can be done using the
+`GridFS Adapter for Flysystem <https://flysystem.thephpleague.com/docs/adapter/gridfs/>`__.
+GridFS lets you store files of unlimited size in the same database as your data.
+This ensures the integrity of transactions and backups.
```
jmikola: Driver implementations of GridFS do not support transactions ([DRIVERS-2062](https://jira.mongodb.org/browse/DRIVERS-2062)).

docs/filesystems.txt
```diff
@@ -0,0 +1,144 @@
...
+
+   'disks' => [
+       'gridfs' => [
+           'driver' => 'gridfs',
+           'connection' => 'mongodb',
+           'database' => 'files',
+           'bucket' => 'fs',
+           'prefix' => '',
+           'read-only' => false,
+           'throw' => false,
```
jmikola: Is this the complete list of supported options, with default values repeated just so users know what they can set? Since you have a table below with all options, I'd consider reducing this code example to just the necessary options.

docs/filesystems.txt
```diff
@@ -0,0 +1,144 @@
...
+     - Specifies a prefix for the name of the files that are stored in the bucket. Using a distinct bucket is recommended
+       in order to store the files in a different collection.
+
+   * - ``read-only``
+     - If ``true``, writing to the GridFS bucket is disabled. Write operations will return ``false`` or throw exceptions
+       depending on the configuration of ``throw``. Defaults to ``false``.
+
+   * - ``throw``
+     - If ``true``, exceptions are thrown when an operation cannot be performed. Defaults to ``false``, the operations
+       return ``true`` in case of succes, ``false`` in case of error.
```
jmikola: The second sentence looks like a run-on. Consider:

```suggestion
     - If ``true``, exceptions are thrown when an operation cannot be performed. If ``false``,
       operations return ``true`` on success and ``false`` on error. Defaults to ``false``.
```

docs/filesystems.txt
```diff
@@ -0,0 +1,144 @@
...
+     - **Required**. Specifies the filesystem driver to use. Must be ``gridfs`` for MongoDB.
+
+   * - ``connection``
+     - The database connection used to store jobs. It must be a ``mongodb`` connection. The driver uses the default connection if a connection is not specified.
+
+   * - ``database``
+     - Name of the MongoDB database fot the GridFS bucket. Use the database of the connection if not specified.
+
+   * - ``bucket``
+     - Name of the GridFS bucket. A database can contain multiple buckets identified by their name. Defaults to ``fs``.
```
jmikola: Based on the example below, shouldn't this description also mention that this option accepts a `MongoDB\GridFS\Bucket` instance?

docs/filesystems.txt
```diff
@@ -0,0 +1,144 @@
...
+                   ]);
+           },
+       ],
+   ],
+
+Usage
+-----
+
+The benefits of using Laravel File Storage facade, is that it provides a common
+interface for all the supported file systems. Use the ``gridfs`` disk in the
```
jmikola: I don't think the comma serves a purpose here. Consider instead:

```suggestion
A benefit of using Laravel's File Storage facade is that it provides a common
interface for all the supported file systems. Use the ``gridfs`` disk in the
```

Alternatively "the Laravel's File Storage". Also, the facade is just the API using `Storage::`, no? Would it be more correct to refer to this as a component? I'll defer to you.

I assume this is what we're referring to: https://laravel.com/docs/11.x/filesystem

docs/filesystems.txt
```diff
@@ -0,0 +1,144 @@
...
+   echo $disk->get('hello.txt'); // Hello World!
+
+To learn more Laravel File Storage, see
+`Laravel File Storage <https://laravel.com/docs/{+laravel-docs-version+}/filesystem>`__.
+
+Versioning
+----------
+
+In GridFS, file names are metadata to file objects identified by unique MongoDB ObjectID.
+There may be more than one file with the same name, they are called "revisions":
```
jmikola:
```suggestion
In GridFS, file names are metadata in file documents identified by unique MongoDB ObjectID.
If multiple documents share the same file name, they are considered "revisions" and
further distinguished by creation timestamps.
```

I'm using the word "timestamps" here, but if you think that might cause some confusion (re: BSON timestamp type), you may want to use "date" instead.

docs/filesystems.txt
```diff
@@ -0,0 +1,144 @@
...
+
+- Reading a file reads the last revision of this file name
+- Writing to a file name creates a new revision for this file name
+- Renaming a file renames all the revisions of this file name
+- Deleting a file deletes all the revisions of this file name
+
+The GridFS Adapter for Flysystem does not provide access to a specific revision
+of a filename, you must use the :manual:`GridFS API </tutorial/gridfs/>` if you
```
jmikola: Does `:manual:` link to the PHPLIB docs? I assume this is the intended target: https://www.mongodb.com/docs/php-library/current/tutorial/gridfs/

src/MongoDBServiceProvider.php
```diff
@@ -66,5 +79,61 @@ public function register()
...
+            $filesystemManager->extend('gridfs', static function (Application $app, array $config) {
+                if (! class_exists(GridFSAdapter::class)) {
+                    throw new RuntimeException('GridFS adapter for Flysystem is missing. Try running "composer require league/flysystem-gridfs"');
+                }
+
+                $bucket = $config['bucket'] ?? null;
+
+                // Get the bucket from a factory function
+                if ($bucket instanceof Closure) {
+                    $bucket = $bucket($app, $config);
```
jmikola, line 99: Given the way these `if` conditions are stacked, is it possible that the Closure returns a string that is then resolved from a service, and finally resolved from a database connection? If that's not intentional, consider rewriting this to avoid this chain of resolution.

Based on the test below, I expect you may want to assert that the Closure returns a MongoDB\GridFS\Bucket instance.

tests/FilesystemsTest.php
```diff
@@ -0,0 +1,150 @@
...
+
+        $this->app['config']->set('filesystems.disks.' . $this->dataName(), $options);
+
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage($message);
+
+        Storage::disk($this->dataName());
+    }
+
+    public function testReadOnlyAndThrowOption()
```
jmikola:
```suggestion
    public function testReadOnlyAndThrowOptions()
```

tests/FilesystemsTest.php
```diff
@@ -0,0 +1,150 @@
...
+    public function testPrefix()
+    {
+        $this->app['config']->set('filesystems.disks.gridfs-prefix', [
+            'driver' => 'gridfs',
+            'connection' => 'mongodb',
+            'prefix' => 'foo/bar/',
+        ]);
+
+        $disk = Storage::disk('gridfs-prefix');
+        $disk->put('hello/world.txt', 'Hello World!');
```
jmikola, line 141: I may be mistaken and/or misunderstanding something, but aren't all files supposed to be prefixed with a leading directory separator?

composer.json
```diff
@@ -34,6 +34,8 @@
     },
     "require-dev": {
         "mongodb/builder": "^0.2",
+        "league/flysystem-gridfs": "^3.28",
+        "league/flysystem-read-only": "^3.0",
```
jmikola, line 38: Should one or both of these go into the `suggest` list?

docs/filesystems.txt
```diff
@@ -0,0 +1,144 @@
...
+   :values: tutorial
+
+.. meta::
+   :keywords: php framework, gridfs, code example
+
+Overview
+--------
+
+Using MongoDB to store files can be done using the
+`GridFS Adapter for Flysystem <https://flysystem.thephpleague.com/docs/adapter/gridfs/>`__.
```
norareidy: S: to avoid the repetition of "using", you could reword this part a little. I'd also clarify which types of files you can store with GridFS.

```suggestion
You can use the
`GridFS Adapter for Flysystem <https://flysystem.thephpleague.com/docs/adapter/gridfs/>`__ to store large files in MongoDB.
```

docs/filesystems.txt
```diff
@@ -0,0 +1,144 @@
...
+GridFS lets you store files of unlimited size in the same database as your data.
+This ensures the integrity of transactions and backups.
+
+
+Configuration
+-------------
+
+Before using the GridFS driver, you will need to install the Flysystem GridFS package via the Composer package manager:
```
norareidy: S: we try to avoid the [future tense](https://www.mongodb.com/docs/meta/style-guide/style/voice-and-tone/#best-practices) and ["via"](https://www.mongodb.com/docs/meta/style-guide/terminology/general-term-guidelines/avoid-obscure-words/) in the docs. Also, I'd edit this sentence slightly so it introduces the code example:

```suggestion
Before using the GridFS driver, install the Flysystem GridFS package through the Composer package manager by running the following command:
```

docs/filesystems.txt
```diff
@@ -0,0 +1,144 @@
...
+
+.. code-block:: bash
+
+   composer require league/flysystem-gridfs
+
+Configure `Laravel File Storage <https://laravel.com/docs/{+laravel-docs-version+}/filesystem>`__,
```
norareidy: S: you can remove this comma
```suggestion
Configure `Laravel File Storage <https://laravel.com/docs/{+laravel-docs-version+}/filesystem>`__
```

docs/filesystems.txt
```diff
@@ -0,0 +1,144 @@
...
+           'driver' => 'gridfs',
+           'connection' => 'mongodb',
+           'database' => 'files',
+           'bucket' => 'fs',
+           'prefix' => '',
+           'read-only' => false,
+           'throw' => false,
+       ],
+   ],
+
```
norareidy, line 49: S: Add a sentence that introduces this table, like:
```suggestion
You can configure the following ``gridfs`` settings in ``config/filesystems.php``: 
```

docs/filesystems.txt
```diff
@@ -0,0 +1,144 @@
...
+     - Description
+
+   * - ``driver``
+     - **Required**. Specifies the filesystem driver to use. Must be ``gridfs`` for MongoDB.
+
+   * - ``connection``
+     - The database connection used to store jobs. It must be a ``mongodb`` connection. The driver uses the default connection if a connection is not specified.
+
+   * - ``database``
+     - Name of the MongoDB database fot the GridFS bucket. Use the database of the connection if not specified.
```
norareidy: S: small typo, and I think you can reword the second sentence so it's more complete:

```suggestion
     - Name of the MongoDB database for the GridFS bucket. The driver uses the database of the connection if a database is not specified.
```

docs/filesystems.txt
```diff
@@ -0,0 +1,144 @@
...
+
+   * - ``read-only``
+     - If ``true``, writing to the GridFS bucket is disabled. Write operations will return ``false`` or throw exceptions
+       depending on the configuration of ``throw``. Defaults to ``false``.
+
+   * - ``throw``
+     - If ``true``, exceptions are thrown when an operation cannot be performed. Defaults to ``false``, the operations
+       return ``true`` in case of succes, ``false`` in case of error.
+
+If you have specific requirements, you can use a factory or a service name to define the bucket. The options
```
norareidy: Q: what do you mean by "specific requirements"? I think it would be helpful to clarify them here

docs/filesystems.txt
```diff
@@ -0,0 +1,144 @@
...
+   $disk = Storage::disk('gridfs');
+
+   // Write the file "hello.txt" into GridFS
+   $disk->put('hello.txt', 'Hello World!');
+
+   // Read the file
+   echo $disk->get('hello.txt'); // Hello World!
+
+To learn more Laravel File Storage, see
+`Laravel File Storage <https://laravel.com/docs/{+laravel-docs-version+}/filesystem>`__.
```
norareidy: S: clarify where this link comes from
```suggestion
`Laravel File Storage <https://laravel.com/docs/{+laravel-docs-version+}/filesystem>`__ in the Laravel documentation.
```

docs/filesystems.txt
```diff
@@ -0,0 +1,144 @@
...
+
+Versioning
+----------
+
+In GridFS, file names are metadata to file objects identified by unique MongoDB ObjectID.
+There may be more than one file with the same name, they are called "revisions":
+
```
norareidy, line 136: S: introduce this list with something like:
```suggestion
GridFS operations interact with file revisions in the following ways:
```

docs/filesystems.txt
norareidy: I: break these two independent clauses into different sentences
```suggestion
of a filename. You must use the :manual:`GridFS API </tutorial/gridfs/>` if you
```

norareidy: `:manual:` links to the Server manual, so I'd remove it here and just use the full https://www.mongodb.com/docs/php-library/current/tutorial/gridfs/ link

docs/filesystems.txt
```diff
@@ -0,0 +1,144 @@
...
+- Reading a file reads the last revision of this file name
+- Writing to a file name creates a new revision for this file name
+- Renaming a file renames all the revisions of this file name
+- Deleting a file deletes all the revisions of this file name
+
+The GridFS Adapter for Flysystem does not provide access to a specific revision
+of a filename, you must use the :manual:`GridFS API </tutorial/gridfs/>` if you
+need to work with revisions.
```
norareidy: S: add to this sentence so it introduces the code example
```suggestion
need to work with revisions, as shown in the following code:
```

docs/filesystems.txt
```diff
@@ -0,0 +1,144 @@
+.. _laravel-queues:
+
+======
+Queues
+======
```
norareidy: I: Edit the page title and ref anchor to reflect the content:
```suggestion
.. _laravel-gridfs:

======
GridFS
======
```

norareidy (changes requested):
I added some comments, mostly about wording and introducing the code examples in the text. Let me know if you have any questions!

tests/FilesystemsTest.php
GromNaN: The directory separator is always added between the prefix and the file name by the [PathPrefixer](https://github.com/thephpleague/flysystem/blob/3.x/src/PathPrefixer.php#L26).

docs/filesystems.txt
GromNaN: Noted, that's too bad.

docs/filesystems.txt
GromNaN: Why not, but personally I find the complete example more synthetic and quicker to read, or even copy and paste.
I understand that we want developers to use the default minimal config, so that's what I'm going to keep.

src/MongoDBServiceProvider.php
GromNaN: Code modified to prevent this sequence of resolution.

composer.json
GromNaN: Added `league/flysystem-gridfs` to the suggestions.

docs/filesystems.txt
jmikola: > I find the complete example more synthetic and quicker to read

Same. If this example was annotated with comments to call out default values (and thus assist user's with reducing their own config) I think it'd stand well on its own and we wouldn't even need the table that follows.

tests/FilesystemsTest.php
jmikola: I recall that the PathPrefixer adds it, but I thought the docs might present the canonical form. I just noted the absence of a leading `/` is consistent with examples in the [Laravel Filesystem docs](https://laravel.com/docs/11.x/filesystem#deleting-files).

docs/filesystems.txt
```diff
@@ -0,0 +1,148 @@
...
+     - If ``true``, exceptions are thrown when an operation cannot be performed. If ``false``,
+	   operations return ``true`` on success and ``false`` on error. Defaults to ``false``.
+
+You can also use a factory or a service name to create an instance of ``MongoDB\GridFS\Bucket``.
+In this case, the options ``connection`` and ``database`` are ignored:
+
+.. code-block:: php
+
+    use Illuminate\Foundation\Application;
+    use MongoDB\GridFS\Bucket;
```
jmikola:
```suggestion
   use Illuminate\Foundation\Application;
   use MongoDB\GridFS\Bucket;
```

Consistent indentation with `'disks'` below. Code blocks use a three-space indent.

docs/filesystems.txt
```diff
@@ -0,0 +1,148 @@
...
+
+To learn more Laravel File Storage, see
+`Laravel File Storage <https://laravel.com/docs/{+laravel-docs-version+}/filesystem>`__
+in the Laravel documentation.
+
+Versioning
+----------
+
+In GridFS, file names are metadata in file documents identified by unique
+MongoDB ObjectID. If multiple documents share the same file name, they are
```
jmikola:
```suggestion
File names in GridFS are metadata in file documents, which are identified by
unique ObjectIDs. If multiple documents share the same file name, they are
```

Moving the comma around to make this more readable. Also dropped the MongoDB prefix for ObjectID (seemed redundant) and fixed subject/verb agreement.

docs/filesystems.txt
```diff
@@ -0,0 +1,148 @@
...
+if you need to work with revisions, as shown in the following code:
+
+.. code-block:: php
+
+   // Create a bucket service from the MongoDB connection
+   /** @var \MongoDB\GridFS\Bucket $bucket */
+   $bucket = $app['db']->connection('mongodb')->getMongoDB()->selectGridFSBucket();
+
+   // Download the last but one version of a file
+   $bucket->openDownloadStreamByName('hello.txt', ['revision' => -2])
```
jmikola, line 154: Since PathPrefixer doesn't apply here, would users need to write `/hello.txt`, or do the prefixes only kick in when a path is used? I expect this could be a pain point for users, that may warrant a note.

docs/filesystems.txt
GromNaN: Added optional settings in comments. Keeping the table for details.

tests/FilesystemsTest.php
GromNaN: When the prefix is empty, the file name in the `fs.files` collection does not have the leading `/`. The only situation you can get a leading `/` is by having it in the prefix.

docs/filesystems.txt
GromNaN: We don't use the leading `/` in the [PHP GridFS documentation](https://www.mongodb.com/docs/php-library/v1.17/tutorial/gridfs/#uploading-files-with-writable-streams). I don't know what to write in a note.

jmikola: There's no prefixing done in PHPLIB, so the filename is used exactly as provided.

My question here is: would `hello.txt` here refer to the same file written by Flysystem, or would you need to use `/hello.txt`? It's not clear to me if Flysystem adds a prefix to every file name or just those that appear to use paths (i.e. already contain a `/` somewhere in the middle of the string).

jmikola: Perhaps this is answered by https://github.com/mongodb/laravel-mongodb/pull/2985#discussion_r1625525675. In that case, I suppose the note would be to remind users that prefixing is entirely handled by Flysystem. So if they're using a non-empty `prefix` option they'll need to ensure that filenames are prefixed manually when using PHPLIB directly.

GromNaN: Flysystem always adds the prefix to the file name stored in MongoDB.

jmikola (approved):

docs/filesystems.txt
```diff
@@ -0,0 +1,159 @@
...
+           'connection' => 'mongodb',
+           // 'database' => null,
+           // 'bucket' => 'fs',
+           // 'prefix' => '',
+           // 'read-only' => false,
+           // 'throw' => false,
+       ],
+   ],
+
+You can configure the disk the following settings in ``config/filesystems.php``:
```
norareidy, line 50: I: I think you can delete "the disk"

```suggestion
You can configure the following settings in ``config/filesystems.php``:
```

docs/filesystems.txt
```diff
@@ -0,0 +1,159 @@
...
+   :widths: 25 75
+
+   * - Setting
+     - Description
+
+   * - ``driver``
+     - **Required**. Specifies the filesystem driver to use. Must be ``gridfs`` for MongoDB.
+
+   * - ``connection``
+     - The database connection used to store jobs. It must be a ``mongodb`` connection. The driver uses the default connection if a connection is not specified.
```
norareidy, line 63: S: remove the article to match the other descriptions
```suggestion
     - Database connection used to store jobs. It must be a ``mongodb`` connection. The driver uses the default connection if a connection is not specified.
```

docs/filesystems.txt
```diff
@@ -0,0 +1,159 @@
...
+   * - ``database``
+     - Name of the MongoDB database for the GridFS bucket. The driver uses the database of the connection if a database is not specified.
+
+   * - ``bucket``
+     - Name or instance of the GridFS bucket. A database can contain multiple buckets identified by their name. Defaults to ``fs``.
+
+   * - ``prefix``
+     - Specifies a prefix for the name of the files that are stored in the bucket. Using a distinct bucket is recommended
+       in order to store the files in a different collection, instead of using a prefix.
+       The prefix should not start with a leading slash ``/``.
```
norareidy, line 74: S: Add parentheses around the slash
```suggestion
       The prefix should not start with a leading slash (``/``).
```

docs/filesystems.txt
```diff
@@ -0,0 +1,159 @@
...
+                       'chunkSizeBytes' => 261120,
+                   ]);
+           },
+       ],
+   ],
+
+Usage
+-----
+
+A benefit of using Laravel Filesystem is that it provides a common interface
```
norareidy, line 109: S: Reword so this sentence is a little less subjective

```suggestion
Laravel Filesystem provides a common interface
```

docs/filesystems.txt
```diff
@@ -0,0 +1,159 @@
...
+for all the supported file systems. You can use the ``gridfs`` disk in the same
+way as the ``local`` disk.
+
```
norareidy, line 112: S: introduce the code

```suggestion
The following example writes a file to the ``gridfs`` disk, then reads the file:
```

docs/filesystems.txt
```diff
@@ -0,0 +1,159 @@
...
+To learn more Laravel File Storage, see
+`Laravel File Storage <https://laravel.com/docs/{+laravel-docs-version+}/filesystem>`__
+in the Laravel documentation.
+
+Versioning
+----------
+
+File names in GridFS are metadata in file documents, which are identified by
+unique ObjectIDs. If multiple documents share the same file name, they are
```
norareidy, line 131: S: since this is the first time that file metadata documents are mentioned, I think it would be helpful to describe them 

```suggestion
GridFS creates file documents for each uploaded file. These documents contain metadata, including the file name and a
unique ObjectId. If multiple documents share the same file name, they are
```

docs/filesystems.txt
```diff
@@ -0,0 +1,159 @@
...
+The Laravel MongoDB integration uses the GridFS Flysystem adapter. It interacts
+with file revisions in the following ways:
+
+- Reading a file reads the last revision of this file name
+- Writing a file creates a new revision for this file name
+- Renaming a file renames all the revisions of this file name
+- Deleting a file deletes all the revisions of this file name
+
+The GridFS Adapter for Flysystem does not provide access to a specific revision
+of a filename. You must use the
```
norareidy, line 143: S: I'd stay consistent with using either "filename" or "file name"

```suggestion
of a file name. You must use the
```

docs/filesystems.txt
```diff
@@ -0,0 +1,159 @@
...
+   /** @var \MongoDB\GridFS\Bucket $bucket */
+   $bucket = $app['db']->connection('mongodb')->getMongoDB()->selectGridFSBucket();
+
+   // Download the last but one version of a file
+   $bucket->openDownloadStreamByName('hello.txt', ['revision' => -2])
+
+.. note::
+
+   If you use a prefix the Filesystem component, you will have to handle it
+   by yourself when using the GridFS API directly.
```
norareidy, line 159: S: clarify the first part of this sentence ("If you use a prefix the Filesystem component" part)

Q: what do you mean by "handle it by yourself" - could you clarify that in the note?

```suggestion
   If you specify the ``prefix`` GridFS setting, you will have to handle it
   by yourself when using the GridFS API directly.
```

norareidy (approved):
LGTM with some comments!

GromNaN:
> LGTM with some comments!

I applied your suggested changes in a new PR: https://github.com/mongodb/laravel-mongodb/pull/2993

PR #2995, opened 2024-06-04
Update compatibility tables for v4.4

norareidy:
JIRA - https://jira.mongodb.org/browse/DOCSP-40161
Staging - https://preview-mongodbnorareidy.gatsbyjs.io/laravel/DOCSP-40161-4.4/compatibility/

### Checklist

- [ ] Add tests and ensure they pass
- [ ] Add an entry to the CHANGELOG.md file
- [ ] Update documentation for new features

PR #2997, opened 2024-06-04
v4.5 compat table

GromNaN:
Same as https://github.com/mongodb/laravel-mongodb/pull/2995

But avoids having to many identical versions.


rustagir (approved):
lgtm

PR #3004, opened 2024-06-13
PHPORM-185, PHPORM-191, PHPORM-192: Publish SSDLC assets on release

alcaeus:
This PR fixes the following issues:

- PHPORM-185
- PHPORM-191
- PHPORM-192

In order to have all resources available when we publish SSDLC assets, it is necessary to run the static analysis workflow as part of the release workflow. Through this we can ensure that up-to-date code scanning results are available for export.

Along with the necessary changes to publish all assets, I've also made a couple of improvements related to the release process:

- The token generation and checkout now uses the `secure-checkout` action from drivers-github-tools
- Creating the release tag uses the `tag-version` action from drivers-github-tools


.github/workflows/coding-standards.yml
```diff
@@ -67,59 +66,3 @@ jobs:
         uses: stefanzweifel/git-auto-commit-action@v5
         with:
           commit_message: "apply phpcbf formatting"
-
-  analysis:
```
alcaeus, line 71: This job was extracted to its own workflow

.github/workflows/release.yml
```diff
@@ -84,14 +74,77 @@ jobs:
       # 3. push next branch
       # 4. switch back to release branch, then push
 
-      - name: "Push changes from release branch"
-        run: git push
```
alcaeus, line 88: This step was never necessary

.github/workflows/release.yml
```diff
@@ -84,14 +74,77 @@ jobs:
...
-
-      # Pushing the release tag starts build processes that then produce artifacts for the release
-      - name: "Push release tag"
-        run: git push origin ${{ inputs.version }}
```
alcaeus, line 92: This step is now handled by the `tag-version` action

.github/workflows/release.yml
```diff
@@ -84,14 +74,77 @@ jobs:
...
+        with:
+          product_name: "MongoDB Laravel Integration"
+          release_version: ${{ inputs.version }}
+          filenames: ""
+          token: ${{ env.GH_TOKEN }}
+
+      - name: "Download SBOM file from Silk"
+        uses: mongodb-labs/drivers-github-tools/sbom@v2
+        with:
+          silk_asset_group: mongodb-laravel-integration
```
jmikola, line 129: Since this projects exists within the scope of DBX and its PHP team, wouldn't it make sense for these terms to share common prefixes (e.g. `mongodb-dbx-php-laravel`)?

Enforcing some taxonomy early on should help us down the road, but only if we can get everyone on board with this. To that end, maybe this is a better subject for the SSDLC channel.

.github/workflows/release.yml
```diff
@@ -84,14 +74,77 @@ jobs:
...
+      - name: "Generate compliance report"
+        uses: mongodb-labs/drivers-github-tools/compliance-report@v2
+        with:
+          token: ${{ env.GH_TOKEN }}
+
+      - name: Upload S3 assets
+        uses: mongodb-labs/drivers-github-tools/upload-s3-assets@v2
+        with:
+          version: ${{ inputs.version }}
+          product_name: laravel-mongodb
```
jmikola, line 150: I suppose this needs to match the repository name, so enforcing some taxonomy is out of the question. From https://github.com/mongodb/mongo-php-driver/pull/1584, I noted that PHPC is using `mongo-php-driver`.

jmikola (approved):

.github/workflows/release.yml
alcaeus: Yeah, these names were auto-assigned by DEVPROD, but your suggestion makes sense. Maybe they can rename them or something

.github/workflows/release.yml
alcaeus: The assets are uploaded to a separate S3 bucket for the PHP team only, so I opted for the repository name since it makes most sense to us. Depending on who consumes this the names might change soemtime down the road.

PR #3006, opened 2024-06-17
Upload code scanning results to correct ref when releasing

alcaeus:
Same issue as with https://github.com/mongodb/mongo-php-library/pull/1346.

When running the static analysis workflow from the release workflow, the `github.ref` and `github.sha` point to the release branch instead of the release tag. We have to account for this when uploading the sarif. Using `inputs.ref` works, but the `sarif-upload` action also expects a `sha` input when `ref` is given. To account for this, we get the ref for `HEAD` on `workflow_dispatch` events so the code scanning results are correctly attached to the release tag.

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -29,6 +29,11 @@ jobs:
         with:
           ref: ${{ github.event_name == 'workflow_dispatch' && inputs.ref || github.ref }}
 
+      - name: "Get SHA hash of checked out ref"
+        if: ${{ github.event_name == 'workflow_dispatch' }}
+        run: |
+          echo CHECKED_OUT_SHA=$(git rev-parse HEAD) >> $GITHUB_ENV
+
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/static-analysis.yml
```diff
@@ -58,12 +63,21 @@ jobs:
 
       - name: Run PHPStan
         run: ./vendor/bin/phpstan analyse --no-interaction --no-progress --ansi --error-format=sarif > phpstan.sarif
+        continue-on-error: true
```
GromNaN, line 66: If there's an error, the file won't be generated and subsequent steps will crash, isn't it?

alcaeus: It depends. IIRC, if PHPStan detects errors it will exit with a non-zero exit code but still write the file. Default behaviour in GHA would be to stop the job, but we want to at least continue to the next step to upload a SARIF file if it exists. If PHPStan crashed and didn't produce a file, I'm fine with the upload-sarif step failing as well.

alcaeus:
Note: fixed a wrong filename due to copy/paste. The PHPStan workflow should now pass again.

GromNaN (approved):

PR #3021, opened 2024-07-01
PHPORM-206 Add model schema version feature

florianJacques:
Added a feature for managing mongoDB document versioning.
The default key is "schema_version"
If the version is not specified, t is automatically set to 1

Usage

```php
use MongoDB\Laravel\Eloquent\HasDocumentVersion;
use MongoDB\Laravel\Eloquent\Model as Eloquent;

/** @property int $schema_version */
class SchemaVersionned extends Eloquent
{
    use HasSchemaVersion;

    public const int SCHEMA_VERSION = 1;

    protected $connection       = 'mongodb';
    protected $collection       = 'documentVersion';
    protected static $unguarded = true;
}
```

For overwrite version key:

```php
/** @property int $__v */
class DocumentVersion extends Eloquent
{
    use HasSchemaVersion;

    protected $connection       = 'mongodb';
    protected $collection       = 'documentVersion';
    protected static $unguarded = true;

    public static function getSchemaVersionKey(): string
    {
       return '__v';
    }
}
```

For update document version:

```php
/** @property int $schema_version */
class DocumentVersion extends Eloquent
{
    use HasSchemaVersion;

    public const int SCHEMA_VERSION = 2;

    protected $connection       = 'mongodb';
    protected $collection       = 'documentVersion';
    protected static $unguarded = true;
}
```


tests/ModelTest.php
```diff
@@ -1188,6 +1190,12 @@ public function testCreateWithNullId()
         $this->assertSame(1, User::count());
     }
 
+    public function testDocumentVersion()
```
GromNaN: This test class is already very long. You can create a new `DocumentVersionTest` class.

src/Eloquent/HasDocumentVersion.php
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Get document version key
+     *
+     * @return string
+     */
+    protected static function getDocumentVersionKey(): string
+    {
+        return defined(static::class.'::DOCUMENT_VERSION_KEY')
+            ? (string) static::DOCUMENT_VERSION_KEY
+            : '__v';
```
GromNaN: In [MongoDB's tutorial on Schema Versioning](https://www.mongodb.com/docs/manual/tutorial/model-data-for-schema-versioning/), this field name is named `schema_version`.

tests/Models/DocumentVersion.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Laravel\Tests\Models;
+
+use MongoDB\Laravel\Eloquent\HasDocumentVersion;
+use MongoDB\Laravel\Eloquent\Model as Eloquent;
+
+/** @property int __v */
```
GromNaN: This trait should contain a method `getDocumentVersion`. So that you don't have to access the property and worry about the name configuration.

GromNaN (commented):
Thanks for proposing this feature.

The ability to store documents with different schemas in the same collection is undoubtedly one of the great strengths of the MongoDB model. We have a [tutorial on this subject](https://www.mongodb.com/docs/manual/tutorial/model-data-for-schema-versioning/).

I think there's a piece missing: the code to automatically update documents when they're read from the DB. This would probably be done in a listener of the `retrieved` event.

In the trait:
```php
        static::retrieved(function (self $model) {
            $model->upgradeDocumentVersion($model, $model->{static::getDocumentVersionKey()} ?? 0);
        });
```

In the code would look like this (example for the docs):

```php
    public function upgradeDocumentVersion(Model $document, int $fromVersion): void
    {
        switch (true) {
            case $fromVersion < 2:
                // Field renamed
                $document->newField = $document->oldField;
            // No break to apply all changes
            case $fromVersion < 3:
                // Field type updated
                $document->price = (float) $document->price;
            // ...
        }
    }
```

tests/ModelTest.php
florianJacques: Ok, i create a new class

src/Eloquent/HasDocumentVersion.php
florianJacques: It's renamed.

tests/Models/DocumentVersion.php
florianJacques: I added this, and made a version setter

src/Eloquent/HasDocumentVersion.php
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * Get document version key
+     *
+     * @return string
+     */
+    protected static function getDocumentVersionKey(): string
+    {
+        return defined(static::class.'::DOCUMENT_VERSION_KEY')
+            ? (string) static::DOCUMENT_VERSION_KEY
+            : 'schema_version';
```
GromNaN: Changing the name of this attribute should be exceptional. I suggest deleting the constant and keeping only the method, which can be overridden.
```suggestion
        return 'schema_version';
```

src/Eloquent/HasDocumentVersion.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Laravel\Eloquent;
+
+trait HasDocumentVersion
```
GromNaN: "Document Version" sounds like we are versioning the document with revisions, with an history. Naming everything "Schema Version" would be more appropriate.

src/Eloquent/HasDocumentVersion.php
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * @return int
+     */
+    protected static function getCurrentVersion(): int
+    {
+        return defined(static::class.'::DOCUMENT_VERSION')
+            ? (int) static::DOCUMENT_VERSION
+            : 1;
+    }
```
GromNaN: The `DOCUMENT_VERSION` should be initialized to `1` in the trait, and overridden in the model class.
```
public const int DOCUMENT_VERSION = 1;
```

src/Eloquent/HasDocumentVersion.php
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * migrate model document version schema
+     *
+     * @param int|null $fromVersion
+     * @return void
+     */
+    public function migrateDocumentVersion($fromVersion): void
+    {
+    }
```
GromNaN: When the phpdoc duplicate the native types, you can omit them, same for other methods.

Also updating the comment.
```suggestion
    /**
     * Migrate document to the current schema version.
     */
    public function migrateDocumentVersion(int $fromVersion): void
    {
    }
```

src/Eloquent/HasDocumentVersion.php
```diff
@@ -0,0 +1,79 @@
...
+     * @return void
+     */
+    public static function bootHasDocumentVersion(): void
+    {
+        static::saving(function ($model) {
+          $version = $model->getDocumentVersion();
+
+          if (!$version) {
+              $model->setDocumentVersion(static::getCurrentVersion());
+          }
```
GromNaN: `$version` is used only once, not necessary.
```suggestion
          if (! $model->getDocumentVersion()) {
              $model->setDocumentVersion(static::getCurrentVersion());
          }
```

What's the behavior when the stored document doesn't contain the `schema_version` field.  It should be considered `0` and `$model->migrateDocumentVersion` should be called.

src/Eloquent/HasDocumentVersion.php
```diff
@@ -0,0 +1,79 @@
...
+        });
+
+        static::retrieved(function ($model) {
+            $model->migrateDocumentVersion($model->getDocumentVersion());
```
GromNaN: Skip the call if `$this->getDocumentVersion() >= static::getCurrentVersion()`

src/Eloquent/HasDocumentVersion.php
florianJacques: I wanted to reproduce the philosophy of eloquent models with timestamp keys

src/Eloquent/HasDocumentVersion.php
florianJacques: Yes but impossible to overwrite a const

src/Eloquent/HasDocumentVersion.php
florianJacques: I agree with you, I will make the changes

src/Eloquent/HasDocumentVersion.php
florianJacques: Understood

src/Eloquent/HasDocumentVersion.php
florianJacques: Of course, I don't know why I didn't think of that.

src/Eloquent/HasSchemaVersion.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Laravel\Eloquent;
+
+trait HasSchemaVersion
+{
+    public $currentSchemaVersion = 1;
```
florianJacques: I don't think this is the best solution, do you have any ideas?

src/Eloquent/HasSchemaVersion.php
```diff
@@ -0,0 +1,78 @@
...
+    public static function bootHasSchemaVersion(): void
+    {
+        static::saving(function ($model) {
+          if (!$model->getSchemaVersion()) {
+              $model->setSchemaVersion($model->currentSchemaVersion);
+          }
+        });
+
+        static::retrieved(function ($model) {
+            $model->upgradeSchemaVersion();
```
florianJacques: I think we need to find a way to give control over the call to this function

src/Eloquent/HasSchemaVersion.php
GromNaN:
```suggestion
    public int $currentSchemaVersion = 1;
```

src/Eloquent/HasSchemaVersion.php
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public function migrateSchema(int $fromVersion): void
+    {
+    }
+
+    /**
+     * migrate schema and set current model version
+     * @return void
+     */
+    public function upgradeSchemaVersion(): void
```
GromNaN: Why do you need this public method? All this code can be safely in the "retrieved" closure.

tests/SchemaVersionTest.php
```diff
@@ -0,0 +1,45 @@
...
+    public function testWithBasicDocument()
+    {
+        $document = new SchemaVersion(['name' => 'Luc']);
+        $this->assertEmpty($document->getSchemaVersion());
+        $document->save();
+
+        $this->assertEquals(1, $document->getSchemaVersion());
+        $this->assertNull($document->age);
+
+        $document->currentSchemaVersion = 2;
```
GromNaN: You should not update this value directly. That's the responsability of `migrateSchema`.
```suggestion
```

src/Eloquent/HasSchemaVersion.php
florianJacques: I think it's a good idea to let the user choose which method to use, between migrating + upgrading the version or migrating and upgrading the version later.

tests/SchemaVersionTest.php
florianJacques: I think the choice of method/parameter name is wrong, because you haven't understood its role.
Here, it's about forcing the model to believe that it should play a migration.
Maybe remplace the naming by "latestSchemaVersion"?

src/Eloquent/HasSchemaVersion.php
```diff
@@ -0,0 +1,82 @@
...
+
+    protected static function getSchemaVersionKey(): string
+    {
+        return 'schema_version';
+    }
+
+    protected function getModelSchemaVersion(): int
+    {
+        try {
+            return $this::SCHEMA_VERSION;
```
GromNaN, line 77: Update: The schema version is linked to the code. Using a constant feels more appropriate.

tests/SchemaVersionTest.php
```diff
@@ -0,0 +1,58 @@
...
+        $this->assertEmpty($document->getSchemaVersion());
+        $document->save();
+
+        // The current schema version of the model is stored by default
+        $this->assertEquals(2, $document->getSchemaVersion());
+
+        // Test automatic migration
+        SchemaVersion::insert([
+            ['name' => 'Vador', 'schema_version' => 1],
+        ]);
```
GromNaN, line 32: Update: inserting raw document to test the migration when the document is read.

GromNaN (approved):
@florianJacques I updated the implementation so that
- the `SCHEMA_VERSION` constant is required
- the tests ensure the documents are updated when retrieved
- adding documentation in the trait description

florianJacques:
Thanks for your work !

GromNaN:
Thank you @florianJacques

PR #3024, opened 2024-07-02
fix attributeToArray method

florianJacques:
- Fix model serialization, when model has embed relation or nested fields.

When dealing with embedded models with relationships (EmbedOne and/or EmbedMany), the serialization methods do not serialize Mongo BSON types in nested fields, as the function only converts top-level attributes.

In the exemple, prices is embedMany Relationship

```php
 public function prices(): EmbedsMany
 {
      return $this->embedsMany(Price::class, 'prices', 'product');
 }
```

Before fix =>

```php
 [
    "_id" => "667be7212ea8e4cc8b0dd636",
    "isActive" => true,
    "name" => "CREDIT_PACKAGE_3",
    "category" => "CREDIT_BALANCE",
    "quantity" => 3.0,
    "prices" => [
      [
        "currency" => "EUR",
        "price" => 5.99,
        "_id" => MongoDB\BSON\ObjectId {#4456
          +"oid": "667be7212ea8e4cc8b0dd634",
        },
      ],
    ],
]
```
After fix =>

```php
 [
    "_id" => "667be7212ea8e4cc8b0dd636",
    "isActive" => true,
    "name" => "CREDIT_PACKAGE_3",
    "category" => "CREDIT_BALANCE",
    "quantity" => 3.0,
    "prices" => [
      [
        "currency" => "EUR",
        "price" => 5.99,
        "_id" => "667be7212ea8e4cc8b0dd634",
      ],
    ],
]
```

### DIFF
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -328,12 +328,20 @@ abstract class Model extends BaseModel
         // MongoDB related objects to a string representation. This kind
         // of mimics the SQL behaviour so that dates are formatted
         // nicely when your models are converted to JSON.
-        foreach ($attributes as $key => &$value) {
+        $convertMongoObjects = function (&$value) use (&$convertMongoObjects) {
             if ($value instanceof ObjectID) {
                 $value = (string) $value;
             } elseif ($value instanceof Binary) {
                 $value = (string) $value->getData();
+            } elseif (is_array($value)) {
+                foreach ($value as &$embedValue) {
+                    $convertMongoObjects($embedValue);
+                }
             }
+        };
+
+        foreach ($attributes as $key => &$value) {
+            $convertMongoObjects($value);
         }
 
         return $attributes;
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -333,6 +333,8 @@ abstract class Model extends BaseModel
                 $value = (string) $value;
             } elseif ($value instanceof Binary) {
                 $value = (string) $value->getData();
+            } elseif ($value instanceof UTCDateTime) {
+                $value = $this->serializeDate($value->toDateTime());
             } elseif (is_array($value)) {
                 foreach ($value as &$embedValue) {
                     $convertMongoObjects($embedValue);
--- a/src/Eloquent/Model.php
+++ b/src/Eloquent/Model.php
@@ -319,31 +319,44 @@ abstract class Model extends BaseModel
         return new Decimal128($this->asDecimal($value, $decimals));
     }
 
+    /**
+     * Convert MongoDB objects to their string representations.
+     *
+     * This method converts MongoDB related objects (ObjectID, Binary, UTCDateTime)
+     * to their serialized representations, ensuring the values can be correctly
+     * serialized when the model is converted to JSON.
+     *
+     * @param mixed $value The value to convert
+     *
+     * @return mixed
+     */
+    protected function convertMongoObjects(mixed $value): mixed
+    {
+        if ($value instanceof ObjectID) {
+            $value = (string) $value;
+        } elseif ($value instanceof Binary) {
+            $value = (string) $value->getData();
+        } elseif ($value instanceof UTCDateTime) {
+            $value = $this->serializeDate($value->toDateTime());
+        } elseif (is_array($value)) {
+            foreach ($value as &$nestedValue) {
+                $nestedValue = $this->convertMongoObjects($nestedValue);
+            }
+        }
+
+        return $value;
+    }
+
     /** @inheritdoc */
     public function attributesToArray()
     {
         $attributes = parent::attributesToArray();
 
-        // Because the original Eloquent never returns objects, we convert
         // MongoDB related objects to a string representation. This kind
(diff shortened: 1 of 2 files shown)


GromNaN:
Thank you @florianJacques for contributing. To review your PR correctly, I need that you describe the issue, and if possible add tests validating the fix.

florianJacques:
> Thank you @florianJacques for contributing. To review your PR correctly, I need that you describe the issue, and if possible add tests validating the fix.

Test ok

tests/EmbeddedRelationsTest.php
```diff
@@ -945,4 +946,13 @@ public function testGetQueueableRelationsEmbedsOne()
...
         $this->assertEquals([], $user->father->getQueueableRelations());
     }
+
+    public function testEmbedManySerialization()
+    {
+        $user = User::create(['name' => 'John Doe']);
+        $user->addresses()->save(new Address(['city' => 'New York']));
+
+        $results = $user->toArray();
+        assertIsString($results['addresses'][0]['_id']);
```
GromNaN: We use the `TestCase` method.
```suggestion
        $this->assertIsString($results['addresses'][0]['_id']);
```

src/Eloquent/Model.php
```diff
@@ -328,12 +328,22 @@ public function attributesToArray()
         // MongoDB related objects to a string representation. This kind
         // of mimics the SQL behaviour so that dates are formatted
         // nicely when your models are converted to JSON.
-        foreach ($attributes as $key => &$value) {
+        $convertMongoObjects = function (&$value) use (&$convertMongoObjects) {
```
GromNaN: This should be a private method.

GromNaN (changes requested):
The new test does not cover the issue, it works without your fix.

In order to convert date objects nested into embedded object (not embedded model), you can define a regular cast using a dot for the field path.

Example: https://github.com/mongodb/laravel-mongodb/blob/ffacc6b48b7f8e08006a27e323e587976dfa67ea/tests/Models/User.php#L42-L46

Validated by this test:
https://github.com/mongodb/laravel-mongodb/blob/ffacc6b48b7f8e08006a27e323e587976dfa67ea/tests/ModelTest.php#L774-L776

src/Eloquent/Model.php
florianJacques: Ok now is split

florianJacques: For the test, I simply reused existing model. Would you like me to create a new one?
Yes for nested fields (with no embed relationship), but will it work without my patch for more deep level fields?

GromNaN: Yes, you can create a new model class for the test.

src/Eloquent/Model.php
```diff
@@ -319,21 +319,44 @@ protected function fromDecimal($value, $decimals)
...
+        return $value;
+    }
+
     /** @inheritdoc */
     public function attributesToArray()
     {
         $attributes = parent::attributesToArray();
 
-        // Because the original Eloquent never returns objects, we convert
         // MongoDB related objects to a string representation. This kind
```
GromNaN: Please revert this.
```suggestion
        // Because the original Eloquent never returns objects, we convert
        // MongoDB related objects to a string representation. This kind
```

tests/EmbeddedRelationsTest.php
```diff
@@ -945,4 +946,16 @@ public function testGetQueueableRelationsEmbedsOne()
...
+    public function testEmbedManySerialization()
+    {
+        $address = new Address(['country' => 'France']);
+        $address->save();
+
+        $address->addresses()->create(['city' => 'Paris']);
+        $address->addresses()->create(['city' => 'Nice']);
+
+        $results = $address->toArray();
+        $this->assertIsString($results['addresses'][0]['_id']);
```
GromNaN, line 959: Unfortunately, this test does not cover the corrected bug. I've just checked, even if your fix is removed, the test still passes.

src/Eloquent/Model.php
```diff
@@ -319,6 +319,34 @@ protected function fromDecimal($value, $decimals)
...
+     *
+     * @return mixed
+     */
+    protected function convertMongoObjects(mixed $value): mixed
+    {
+        if ($value instanceof ObjectID) {
+            $value = (string) $value;
+        } elseif ($value instanceof Binary) {
+            $value = (string) $value->getData();
+        } elseif ($value instanceof UTCDateTime) {
```
GromNaN, line 339: I don't think dates should be handled here. They were not in the current code.

GromNaN (changes requested):
I'm thinking of rejecting your proposal.

Embeds should not be used in this way. Instead of calling `$results['addresses']`, you need to call `$addresses->addresses()->get()` which returns a collection of `Address` objects by applying the castings defined in the Model.

You can also work with objects directly, in which case they are not embedded and don't automatically have an id, but you can add one.

src/Eloquent/Model.php
florianJacques: I understand, indeed if relationships are called, casts are automatic. Sorry for the misunderstanding.
You can reject the fix proposal. Looking forward to contributing on something else.

PR #3040, opened 2024-07-10
PHPORM-147 Make `id` an alias for `_id`

### LINKED ISSUE #2489: Querying subdocuments using whereIn by "_id" field does not work, yet it does by "id" key
### Description:
Querying using whereIn("subarray._id", [...]) does not work, when subarray has an array of items with '_id' field. If you had subarray of items with 'id' instead of '_id' and then queried for "subarray.id" it would work.
However querying using whereRaw(['subarray._id' => ['$in' => ['id_value'']]]) works as expected. I tried transforming the array of _id strings into an array of ObjectIds but it didn't help.

### Steps to reproduce
1. Create a document with a subarray of items where each item will have an _id field
2. Try a query using whereIn("subarray._id", [...]) = will not return results
3. Now do the same but create a document with a subarray of items where each item will have an id field
4. Try a query using whereIn("subarray.id", [...]) = will return results

### Expected behaviour
It should work as normal, as in step 4

### Actual behaviour
It returned no results while there definitely were matches

### LINKED ISSUE #3022: [Feature Request] SessionServiceProvider for database
Install laravel breeze react example, setup SESSION_DRIVER=database, the first error is duplicate key for id, deleting unique index solve this but in any post request, it will always cause 419 | PAGE EXPIRED

Temp solution just using any driver like file. I would like guidance on how to make Session driver database work for mongodb thanks

GromNaN:
Fix PHPORM-147
Fix https://github.com/mongodb/laravel-mongodb/issues/3022 (PHPORM-201)
Fix https://github.com/mongodb/laravel-mongodb/issues/2489

Laravel and Eloquent uses `id` as conventional field name for models and various queries. MongoDB's primary key is called `_id`. The name is source of various issues with Laravel packages:

- `DatabaseSessionHandler` https://github.com/mongodb/laravel-mongodb/issues/3022#issuecomment-2220068550
- User email validation https://github.com/laravel/breeze/pull/357

Breaking change: it is no longer possible to set a different value for `id` and `_id`. If you need a distinct ID field, use an other name.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [5.0.0] - next
+
+* **BREAKING CHANGE** Use `id` as an alias for `_id` in commands and queries for compatibility with Eloquent packages by @GromNaN in [#3040](https://github.com/mongodb/laravel-mongodb/pull/3040)
+
 ## [4.8.0] - next
 
 * Add `Query\Builder::incrementEach()` and `decrementEach()` methods by @SmallRuralDog in [#2550](https://github.com/mongodb/laravel-mongodb/pull/2550)
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -375,7 +375,7 @@ class Builder extends BaseBuilder
 
             // Apply order and limit
             if ($this->orders) {
-                $pipeline[] = ['$sort' => $this->orders];
+                $pipeline[] = ['$sort' => $this->aliasIdForQuery($this->orders)];
             }
 
             if ($this->offset) {
@@ -416,7 +416,7 @@ class Builder extends BaseBuilder
 
         // Normal query
         // Convert select columns to simple projections.
-        $projection = array_fill_keys($columns, true);
+        $projection = $this->aliasIdForQuery(array_fill_keys($columns, true));
 
         // Add custom projections.
         if ($this->projections) {
@@ -431,7 +431,7 @@ class Builder extends BaseBuilder
         }
 
         if ($this->orders) {
-            $options['sort'] = $this->orders;
+            $options['sort'] = $this->aliasIdForQuery($this->orders);
         }
 
         if ($this->offset) {
@@ -506,13 +506,17 @@ class Builder extends BaseBuilder
         if ($returnLazy) {
             return LazyCollection::make(function () use ($result) {
                 foreach ($result as $item) {
-                    yield $item;
+                    yield $this->aliasIdForResult($item);
                 }
             });
         }
 
         if ($result instanceof Cursor) {
             $result = $result->toArray();
+        }
+
+        foreach ($result as &$document) {
+            $document = $this->aliasIdForResult($document);
         }
 
         return new Collection($result);
@@ -593,7 +597,7 @@ class Builder extends BaseBuilder
     /** @inheritdoc */
     public function exists()
     {
-        return $this->first(['_id']) !== null;
+        return $this->first(['id']) !== null;
     }
 
     /** @inheritdoc */
@@ -682,6 +686,18 @@ class Builder extends BaseBuilder
             $values = [$values];
         }
 
+        // Compatibility with Eloquent queries that uses "id" instead of MongoDB's _id
+        foreach ($values as &$document) {
+            if (isset($document['id'])) {
+                if (isset($document['_id']) && $document['_id'] !== $document['id']) {
+                    throw new InvalidArgumentException('Cannot insert document with different "id" and "_id" values');
+                }
+
+                $document['_id'] = $document['id'];
+                unset($document['id']);
+            }
+        }
+
         $options = $this->inheritConnectionOptions();
 
         $result = $this->collection->insertMany($values, $options);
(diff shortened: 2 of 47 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+# [4.7.0] - upcoming
+
+* **BREAKING CHANGE** Use `id` and an alias for `_id` in commands and queries for compatibility with Eloquent packages by @GromNaN in [#3040](https://github.com/mongodb/laravel-mongodb/pull/3040)
```
alcaeus:
```suggestion
* **BREAKING CHANGE** Use `id` as an alias for `_id` in commands and queries for compatibility with Eloquent packages by @GromNaN in [#3040](https://github.com/mongodb/laravel-mongodb/pull/3040)
```

bisht2050:
Should this target a major release?

masterbater:
Seems like a minor breaking change, since id is already correlated with _id. Hopefully this get merged soon. This is a must update that can also unblock more laravel packages together with DocumentModel trait.
@GromNaN 🫡🙏🏻thanks for being thoughtful of the community

alcaeus (approved):
I'll admit, I'm not entirely convinced we want to break BC in a minor release, but given the state of the whole `_id` vs. `id` situation, it might be better to just get this over with.

Correct me if I'm wrong, but as I understand the current situation is:
* Models always have an `_id` field
* The value of `_id` is returned when fetching the `id` field
* Using `id` in the query builder is already changed to use `_id`

With this change, we're using the value of `id` for `_id` when saving models, but throw if there are different values for `id` and `_id`. This ensures that other libraries are able to query MongoDB models the same way they query relational models. Did I understand all of this correctly?

GromNaN:
>  Using `id` in the query builder is already changed to use `_id`

`_id` was already used by the `Query\Builder::find` method. But a `where('id', <value>)` was not converted to `{ "_id": <value> }`. With Eloquent models, there is already an alias when the attribute `id` is read
https://github.com/mongodb/laravel-mongodb/blob/f654b833d31ec3cac96887e6db61c5fef4f68b4a/src/Eloquent/DocumentModel.php#L75-L81

src/Query/Builder.php
```diff
@@ -979,21 +987,26 @@ public function runPaginationCountQuery($columns = ['*'])
...
      */
-    protected function performUpdate($query, array $options = [])
+    protected function performUpdate(array $update, array $options = [])
     {
         // Update multiple items by default.
         if (! array_key_exists('multiple', $options)) {
             $options['multiple'] = true;
         }
 
+        // Since "id" is an alias for "_id", we prevent updating it
```
GromNaN, line 1073: TODO: support alias for nested "id" field.
what if `$update` is a pipeline? That becomes too complex.

rustagir (approved):
lgtm

masterbater:
I dont know if this an issue can be improved for now I made adjustment on my side by unsetting the _id field
Example

Address model
User with address field

```php
$address = Address::find(objectIdhere);

$users = User::insert([
                        'name' => $user->fullname,
                        'username' => $user->username,
                        'address' => $address->toArray()])
```
Cannot have both "id" and "_id" fields. {"exception":"[object] (InvalidArgumentException(code: 0): Cannot have both \"id\" and \"_id\" fields. at /app/vendor/mongodb/laravel-mongodb/src/Query/Builder.php:1620)
It seems good that it can detect also nested documents that has both id and _id

GromNaN:
I was on the fence about unsetting `_id` or not. That's why I left this comment:
https://github.com/mongodb/laravel-mongodb/blob/7551f76b41e70e4883c3005cb557288a1f2a2ff4/src/Query/Builder.php#L1662

It makes it impossible to work with `_id` in result as it would be natively, while `_id` or `id` can both be used in a query.

Thanks for your feedback. It seems that we should unset `_id` indead.

masterbater:
I dont know if this an issue or anything that can be improved for now I made adjustment on my side by unsetting the _id field
Example

Address model
User with address field

```php
$address = Address::find(objectIdhere);
  // unset($address ['_id']); Solution
$users = User::insert([
                        'name' => $user->fullname,
                        'username' => $user->username,
                        'address' => $address->toArray()])
```
Cannot have both "id" and "_id" fields. {"exception":"[object] (InvalidArgumentException(code: 0): Cannot have both "id" and "_id" fields. at /app/vendor/mongodb/laravel-mongodb/src/Query/Builder.php:1620)
It seems good that it can detect also nested documents that has both id and _id, but is there a way it only returns id or do we just need to keep unsettting the _id field

GromNaN:
We can also detect when `_id === id` in a query and `unset(id)`. That would solve your issue.

masterbater:
> I was on the fence about unsetting `_id` or not. That's why I left this comment:
> 
> https://github.com/mongodb/laravel-mongodb/blob/7551f76b41e70e4883c3005cb557288a1f2a2ff4/src/Query/Builder.php#L1662
> 
> It makes it impossible to work with `_id` in result as it would be natively, while `_id` or `id` can both be used in a query.
> 
> Thanks for your feedback. It seems that we should unset `_id` indeed.

Thanks for replying so fast, I edited my question now its below to the answer hahaha.

So the aliasIdForResult will just return the id field, but we cant access the native _id Objectid. This would be in line to Laravel behavior but probably needs to be documented. This is acceptable though just like prisma uses id for mongodb too.

masterbater:
> We can also detect when `_id === id` in a query and `unset(id)`. That would solve your issue.

When using v5 I always assume that _id is non existent so I always use id both for saving data and retrieving it.

PR #3041, opened 2024-07-11
V5

GromNaN:
- Query Builder return stdClass: https://github.com/laravel/framework/blob/213a370b703592587bafcd52d38a0ad772ff7442/src/Illuminate/Database/Connection.php#L118C16-L118C25 
- Alias id for _id everywhere.
- Don’t expose _id field: No _id attribute in Eloquent.
- Always convert datetime to UTCDateTime: https://github.com/laravel/framework/blob/213a370b703592587bafcd52d38a0ad772ff7442/src/Illuminate/Database/Connection.php#L733-L734
- Timezone : using Carbon instead of system timezone.

### Checklist


src/Eloquent/Model.php
```diff
@@ -16,13 +16,6 @@ abstract class Model extends BaseModel
 {
     use DocumentModel;
 
-    /**
-     * The primary key for the model.
-     *
-     * @var string
-     */
-    protected $primaryKey = '_id';
```
GromNaN, line 24: The primary key can keep the default value `id`, which is converted to `_id` by the query builder.

src/Query/Builder.php
```diff
@@ -451,7 +456,7 @@ public function toMql(): array
         }
 
         // Fix for legacy support, converts the results to arrays instead of objects.
-        $options['typeMap'] = ['root' => 'array', 'document' => 'array'];
+        $options['typeMap'] = ['root' => 'object', 'document' => 'array'];
```
GromNaN, line 457: Major breaking change to meet Laravel's behavior: the query builder [returns a `stdClass` object](https://github.com/laravel/framework/blob/213a370b703592587bafcd52d38a0ad772ff7442/src/Illuminate/Database/Connection.php#L118C16-L118C25).

src/Queue/Failed/MongoFailedJobProvider.php
```diff
@@ -4,116 +4,9 @@
...
-use DateTimeInterface;
-use Exception;
 use Illuminate\Queue\Failed\DatabaseFailedJobProvider;
-use MongoDB\BSON\ObjectId;
-use MongoDB\BSON\UTCDateTime;
-
-use function array_map;
 
+/** @deprecated Use {@see DatabaseFailedJobProvider} */
 class MongoFailedJobProvider extends DatabaseFailedJobProvider
```
GromNaN: This class is no longer necessary as the Laravel class works out of the box thanks to the change of `id` field, automatic transformation of dates, and results returned as objects.

PR #3043, opened 2024-07-15
Implement `Connection::getServerVersion`

GromNaN:
Fix [PHPORM-213](https://jira.mongodb.org/browse/PHPORM-213)

This method is used by the [command `db:show`](https://github.com/laravel/framework/blob/60d780f3aa96e28a4a1cbeb75c6514eb6bc58871/src/Illuminate/Database/Console/ShowCommand.php#L49). It was added in Laravel 11 by https://github.com/laravel/framework/pull/48864

The value looks like `7.0.11`, similar to what is returned for SQL databases.

### Checklist

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -327,6 +327,11 @@ class Connection extends BaseConnection
         return $this->db->$method(...$parameters);
     }
 
+    public function getServerVersion(): string
+    {
+        return $this->db->command(['buildInfo' => 1])->toArray()[0]['version'];
+    }
+
     private static function getVersion(): string
     {
         return self::$version ?? self::lookupVersion();
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -327,6 +327,12 @@ class Connection extends BaseConnection
         return $this->db->$method(...$parameters);
     }
 
+    /**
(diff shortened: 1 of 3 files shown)


src/Connection.php
```diff
@@ -327,6 +327,11 @@ public function __call($method, $parameters)
         return $this->db->$method(...$parameters);
     }
 
+    public function getServerVersion(): string
+    {
+        return $this->db->command(['buildInfo' => 1])->toArray()[0]['version'];
```
alcaeus, line 338: A couple of notes here:
* `buildInfo` is not part of the [stable API](https://www.mongodb.com/docs/manual/reference/stable-api/), so if the client was configured to use a stable API version this command will fail
* In a replica set, this will only return the version of the primary server; it may be worth noting that other servers in the replica set may be of a different version
* In a sharded deployment, the command is run on an arbitrary `mongos` from the discovered topology. This means that two subsequent runs of the `buildInfo` command could yield different results in a mixed deployment.

It's not necessarily critical, but it may be worth noting this in a comment. We can also consider marking the method as internal given the limitations.

GromNaN: Added a comment to the method.

alcaeus (approved):

PR #3045, opened 2024-07-15
PHPORM-215 Implement Schema::getColumns and getIndexes

GromNaN:
Fix [PHPORM-215](https://jira.mongodb.org/browse/PHPORM-215)

This method is used by the command `db:table`:

Example of output:
```
php artisan db:table --database=mongodb users

  users .............................................................................................................  
  Columns ......................................................................................................... 8  
  Size ..................................................................................................... 60.00 KB  

  Column ....................................................................................................... Type  
  _id objectId ............................................................................................. objectId  
  created_at date, nullable .................................................................................... date  
  email string, nullable ..................................................................................... string  
  email_verified_at date, nullable ............................................................................. date  
  name string, nullable ...................................................................................... string  
  password string, nullable .................................................................................. string  
  remember_token string, nullable ............................................................................ string  
  updated_at date, nullable .................................................................................... date  

  Index .............................................................................................................  
  _id_ _id ......................................................................................... default, primary  
  email_1 email ..................................................................................... default, unique  


```

### Checklist

### DIFF
--- a/src/Schema/Builder.php
+++ b/src/Schema/Builder.php
@@ -154,8 +154,8 @@ class Builder extends \Illuminate\Database\Schema\Builder
     public function getColumns($table)
     {
         $stats = $this->connection->getMongoDB()->selectCollection($table)->aggregate([
-            // Sample 10,000 documents to get a representative sample of the collection
-            ['$sample' => ['size' => 10_000]],
+            // Sample 1,000 documents to get a representative sample of the collection
+            ['$sample' => ['size' => 1_000]],
             // Convert each document to an array of fields
             ['$project' => ['fields' => ['$objectToArray' => '$$ROOT']]],
             // Unwind to get one document per field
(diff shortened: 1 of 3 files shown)


src/Schema/Builder.php
```diff
@@ -146,6 +151,70 @@ public function getTableListing()
         return $collections;
     }
 
+    public function getColumns($table)
```
alcaeus, line 154: This method could get slow for larger collections, so it may be worth adding a [`$sample`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/sample/) pipeline stage down the road.

Depending on how much the feature is used, we can also consider expanding its functionality in a separate project to be more similar to [Compass' schema analysis feature](https://www.mongodb.com/docs/compass/current/schema/). For now, the current solution provides a good starting point.

alcaeus (approved):

src/Schema/Builder.php
GromNaN: I don't think we should try to replicate what is done by Compass. I added `$sample` and `$limit` stages.

alcaeus (approved):

src/Schema/Builder.php
GromNaN: I tried on atlas sample dataset `sample_airbnb.listingsAndReviews`. With `$sample`, I get this error:
```
  PlanExecutor error during aggregation :: caused by :: Sort exceeded memory limit of 33554432 bytes, but did not opt in to external sorting. Aborting operation. Pass allowDiskUse:true to opt in.
```

It works without it. There might be an optimization on the server for this use-case.

PR #3053, opened 2024-07-18
PHPORM-114 Implement `Builder::upsert`

GromNaN:
Fix PHPORM-114

Laravel documentation for this method: https://laravel.com/docs/11.x/eloquent#upserts

Unlike `createOrFirst` and `firstOrCreate`, there is no event triggered on the model by this operation. So we can use MongoDB's bulk write operation to perform multiple upserts in a single command.

### Checklist

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -735,7 +735,6 @@ class Builder extends BaseBuilder
         $this->applyBeforeQueryCallbacks();
 
         $options = $this->inheritConnectionOptions();
-        $options['upsert'] = true;
         $uniqueBy = array_fill_keys((array) $uniqueBy, 1);
 
         // If no update fields are specified, all fields are updated
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -752,7 +752,7 @@ class Builder extends BaseBuilder
                     $filter[$key] = $val;
                 }
 
-                if ($update === null || isset($update[$key])) {
+                if ($update === null || array_key_exists($key, $update)) {
                     $operation['$set'][$key] = $val;
                 } else {
                     $operation['$setOnInsert'][$key] = $val;
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -726,6 +726,48 @@ class Builder extends BaseBuilder
     }
 
     /** @inheritdoc */
+    public function upsert(array $values, $uniqueBy, $update = null): int
+    {
+        if ($values === []) {
+            return 0;
+        }
+
+        $this->applyBeforeQueryCallbacks();
+
+        $options = $this->inheritConnectionOptions();
+        $options['upsert'] = true;
+        $uniqueBy = array_fill_keys((array) $uniqueBy, 1);
+
+        // If no update fields are specified, all fields are updated
+        if ($update !== null) {
+            $update = array_fill_keys((array) $update, 1);
+        }
+
+        $bulk = [];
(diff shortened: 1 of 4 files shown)


src/Query/Builder.php
```diff
@@ -725,6 +725,48 @@ public function update(array $values, array $options = [])
...
+    public function upsert(array $values, $uniqueBy, $update = null): int
+    {
+        if ($values === []) {
+            return 0;
+        }
+
+        $this->applyBeforeQueryCallbacks();
+
+        $options = $this->inheritConnectionOptions();
+        $options['upsert'] = true;
```
alcaeus: This shouldn't be necessary as the `upsert` option is set on each bulk operation, not the entire bulk write.

src/Query/Builder.php
```diff
@@ -725,6 +725,48 @@ public function update(array $values, array $options = [])
...
+        $bulk = [];
+
+        foreach ($values as $value) {
+            $filter = $operation = [];
+            foreach ($value as $key => $val) {
+                if (isset($uniqueBy[$key])) {
+                    $filter[$key] = $val;
+                }
+
+                if ($update === null || isset($update[$key])) {
```
alcaeus: Note that `array_key_exists` is pretty much as fast as `isset` with the optimisation done in https://github.com/php/php-src/commit/f5044a12ddb6f79d59f72c9c12a6cad36d5f6492.

As a further point, instead of looping over all fields, you could theoretically use `array_intersect_key` and `array_diff_key` to assemble the filter, update, and setOnInsert fields:
```php
// All unique fields serve as filter
$filter = array_intersect_key($value, $uniqueBy);

if ($update === null) {
  // Update all fields, nothing will be set on insert only
  $update = $value;
  $setOnInsert = [];
} else {
  // Only update fields marked as updatable
  $update = array_intersect_key($value, $update);
  // Other fields, excluding uniques, are only set on insert. The filter is implicitly contained in $setOnInsert as part of an upsert
  $setOnInsert = array_diff_key($value, $update, $uniqueBy);
}
```

I'll let you decide whether you want to keep iteration or use the suggestion above.

tests/ModelTest.php
```diff
@@ -143,6 +143,42 @@ public function testUpdate(): void
         $this->assertEquals('Hans Thomas', $check->fullname);
     }
 
+    public function testUpsert()
+    {
+        //$user = User::upsert(['name' => 'John Doe'], ['name' => 'John Doe', 'title' => 'admin', 'age' => 35]);
```
alcaeus: This should probably be removed.

src/Query/Builder.php
GromNaN: Ok, switched to `array_key_exists` as it's more safe in case of typo in the variable name.
```suggestion
                if ($update === null || array_key_exists($key, $update)) {
```

For the other array functions, I thought about it, but I find in this version it's easier to understand that the value is in one of the 2 arrays.

tests/ModelTest.php
GromNaN:
```suggestion
```

src/Query/Builder.php
GromNaN:
```suggestion
```

src/Query/Builder.php
alcaeus: Sounds good!

alcaeus (approved):

PR #3071, opened 2024-07-23
PHPORM-222 Register the `BusServiceProvider` when `BatchRepository` is built

### LINKED ISSUE #3060: Conflict with horizon
- Laravel-mongodb Version: #.#.#
- PHP Version: 8.3
- Database Driver & Version: latest horizon, latest laravel-mongodb, latest laravel

### Description:

Your service provider conflicts with horizon: 
https://github.com/laravel/horizon/issues/1476

Works only if I comment the BatchRepository::class in MongoDBBusServiceProvider.php
photo_2024-07-22_14-55-29

GromNaN:
Fix PHPORM-222
Fix #3060

`MongoDBBusServiceProvider` is not the provider of `BatchRepository`.
I could not reproduce the issue described in #3060, but it looks like this change should fix it.

### Checklist

### DIFF
--- a/src/MongoDBBusServiceProvider.php
+++ b/src/MongoDBBusServiceProvider.php
@@ -8,7 +8,10 @@ use Illuminate\Bus\BusServiceProvider;
 use Illuminate\Container\Container;
 use Illuminate\Contracts\Support\DeferrableProvider;
 use Illuminate\Support\ServiceProvider;
+use InvalidArgumentException;
 use MongoDB\Laravel\Bus\MongoBatchRepository;
+
+use function sprintf;
 
 class MongoDBBusServiceProvider extends ServiceProvider implements DeferrableProvider
 {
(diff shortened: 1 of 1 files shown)


alcaeus (dismissed):

degecko:
Correct me if I'm wrong, but doesn't this fix render the code from the register method useless?

If I'm correct in assuming that the idea was to extend the BatchRepository service provider for the case where the queue batching driver was set to mongodb, then the proposed solution will invalidate that solution.

It's also true that the way the BatchRepository resolver has been extended never actually worked. I think that's the actual bug that needs fixing.

GromNaN:
Ok, I think I've found how this should be done. 

`Illuminate\Bus\BusServiceProvider` is a `DeferrableProvider`. It's registered only when one of the services it provides is requested. But by extending this service in my provider, I replace the service provider that is registered. I with Laravel would call every providers that provide a given service, but there is only one. 
I need to explicitly register the `Illuminate\Bus\BusServiceProvider` when my provider is registered.

GromNaN:
Ideally that could be solved in Laravel: https://github.com/laravel/framework/discussions/52251

alcaeus (approved):

PR #3074, opened 2024-07-25
Fix wrong name for driver options in docs

alcaeus:
The docs use `driverOptions` as key in the connection config, but the connection class actually checks the `driver_options` key. Since the latter name has been in use for a while, I figured that the docs were wrong and updated the documentation so that the configuration is properly accepted.


GromNaN (approved):
Good catch. https://github.com/mongodb/laravel-mongodb/blob/046b92ab109368a5faad7d0848d11af7d4dd5edc/src/Connection.php#L183-L185

friday-deployer (approved):
Good to go

norareidy (approved):
LGTM

PR #3093, opened 2024-08-01
Added ability to use owner key if provided to the morph relation

alejojperez:
In Laravel Polymorphic relations, you can specify the owner key to the relation

```php
protected function getResultsByType($type)
{
    $instance = $this->createModelByType($type);

    $key = $instance->getKeyName(); // <--- Before the change
    $ownerKey = $this->ownerKey ?? $instance->getKeyName(); // <---- After the change

    $query = $instance->newQuery();

    return $query->whereIn($ownerKey, $this->gatherKeysByType($type, $instance->getKeyType()))->get();
}
``` 

### Checklist

### DIFF
--- a/src/Relations/MorphTo.php
+++ b/src/Relations/MorphTo.php
@@ -29,11 +29,11 @@ class MorphTo extends EloquentMorphTo
     {
         $instance = $this->createModelByType($type);
 
-        $key = $instance->getKeyName();
+        $ownerKey = $this->ownerKey ?? $instance->getKeyName();
 
         $query = $instance->newQuery();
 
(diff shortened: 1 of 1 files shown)


GromNaN:
That sounds correct to me. Could you add a test?

Tracking in [PHPORM-226](https://jira.mongodb.org/browse/PHPORM-226)

GromNaN:
I've tried to reproduce the error, but I can't find the condition in which the code you've fixed is called.

It seems to be with eager loading of relations, but I don't find how. Could you show an example of your model definition and your query?

GromNaN:
Fixed by #3162

PR #3100, opened 2024-08-13
PHPORM-227 Fix single document upsert

GromNaN:
Fix PHPORM-227

Caused by #3053

### Checklist

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -732,6 +732,11 @@ class Builder extends BaseBuilder
             return 0;
         }
 
+        // Single document provided
+        if (! array_is_list($values)) {
+            $values = [$values];
+        }
+
         $this->applyBeforeQueryCallbacks();
 
         $options = $this->inheritConnectionOptions();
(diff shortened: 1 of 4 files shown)


src/Query/Builder.php
```diff
@@ -732,6 +733,11 @@ public function upsert(array $values, $uniqueBy, $update = null): int
             return 0;
         }
 
+        // Single document provided
+        if (! is_array(reset($values))) {
```
alcaeus: Just double-checking that we never expect this method to be called with an array of objects? Technically the driver would be able to store those.

GromNaN: This was copied from Laravel: https://github.com/laravel/framework/blob/d16ce251beb53f46d602f5f523ba82900ae4be91/src/Illuminate/Database/Eloquent/Builder.php#L1132-L1135

But I could use [`array_is_list`](https://www.php.net/manual/en/function.array-is-list.php) to check the array instead of the values.

```suggestion
        if (! array_is_list($values)) {
```

alcaeus: I think that's a better option in general, regardless of whether or not the method should accept objects.

alcaeus (approved):
LGTM. Feel free to add the `array_is_list` improvement before merging.

PR #3107, opened 2024-08-21
PHPORM-229 Make Query\Builder return objects instead of array to match Laravel behavior

GromNaN:
Fix PHPORM-229

Laravel Query Builder always returns objects ([source](https://github.com/laravel/framework/blob/213a370b703592587bafcd52d38a0ad772ff7442/src/Illuminate/Database/Connection.php#L118)).

This is a major breaking change that requires changing all array access into property access. Returning a `MongoDB\Model\BSONDocument` and `MongoDB\Model\BSONArray` is not possible as the data is casted with native PHP cast feature in Laravel code: `(object) $result` and `(array) $result`.

### Checklist

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -25,6 +25,7 @@ use MongoDB\Builder\Stage\FluentFactoryTrait;
 use MongoDB\Driver\Cursor;
 use Override;
 use RuntimeException;
+use stdClass;
 
 use function array_fill_keys;
 use function array_is_list;
@@ -45,6 +46,7 @@ use function explode;
 use function func_get_args;
 use function func_num_args;
 use function get_debug_type;
+use function get_object_vars;
 use function implode;
 use function in_array;
 use function is_array;
@@ -52,11 +54,13 @@ use function is_bool;
 use function is_callable;
 use function is_float;
 use function is_int;
+use function is_object;
 use function is_string;
 use function md5;
 use function preg_match;
 use function preg_quote;
 use function preg_replace;
+use function property_exists;
 use function serialize;
 use function sprintf;
 use function str_ends_with;
@@ -391,7 +395,7 @@ class Builder extends BaseBuilder
             }
 
             $options = [
-                'typeMap' => ['root' => 'array', 'document' => 'array'],
+                'typeMap' => ['root' => 'object', 'document' => 'array'],
             ];
 
             // Add custom query options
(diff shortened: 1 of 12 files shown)


src/Query/Builder.php
```diff
@@ -451,8 +455,7 @@ public function toMql(): array
             $options['projection'] = $projection;
         }
 
-        // Fix for legacy support, converts the results to arrays instead of objects.
-        $options['typeMap'] = ['root' => 'array', 'document' => 'array'];
+        $options['typeMap'] = ['root' => 'object', 'document' => 'array'];
```
GromNaN, line 457: Shall we update the type of document to `object` also, to get the same type for root and embedded documents?

rustagir (approved):

masterbater:
Is it possible that all v5 pr be merge this week? I just have an expectation that v5 will be release this week. But if not, no pressure just want those big changes to be merge so we can evaluate if we need to add rdbms or just this package can do fine itself.

alcaeus:
> data is casted with native PHP cast feature in Laravel code: `(object) $result` and `(array) $result`.

Note that we could leverage the [`cast_object` object handler](https://www.phpinternalsbook.com/php7/classes_objects/object_handlers.html#c.cast_object) in extension classes to facilitate converting a `PackedArray` into an array.

src/Query/Builder.php
alcaeus: If the root document is returned as an object, it would make sense to return all embedded documents as objects as well.

src/Query/Builder.php
```diff
@@ -1651,16 +1654,31 @@ private function aliasIdForQuery(array $values): array
         return $values;
     }
 
-    private function aliasIdForResult(array $values): array
+    private function aliasIdForResult(array|object $values): array|object
```
alcaeus, line 1654: You could use an `@template` annotation to indicate that the return type will be the same as the parameter type.

GromNaN: Yes, but it's a private function and doesn't pose any downstream problems.

GromNaN:
> > data is casted with native PHP cast feature in Laravel code: `(object) $result` and `(array) $result`.
> 
> Note that we could leverage the [`cast_object` object handler](https://www.phpinternalsbook.com/php7/classes_objects/object_handlers.html#c.cast_object) in extension classes to facilitate converting a `PackedArray` into an array.

Tracking in [PHPC-2426](https://jira.mongodb.org/browse/PHPC-2426)

src/Query/Builder.php
GromNaN: Fixed.

src/Query/Builder.php
GromNaN: Yes, butLaravel have sort of embedded document with "JSON" fields. They are returned as array...

alcaeus (approved):

src/Query/Builder.php
alcaeus: Ah, I see. In that case, let's leave it as-is for consistency with Laravel, even though it seems strange in general.

PR #3105, opened 2024-08-21
PHPORM-230 Convert DateTimeInterface to UTCDateTime in queries

GromNaN:
Fix PHPORM-230

By default, `DateTimeInterface` objects are converted to empty objects by the PHP driver. In this package, they are already explicitly converted to `MongoDB\BSON\UTCDateTime` objects in Eloquent models (depending on cast) and DB queries.

This PR generalize this transformation by applying the conversion to every command: including `DB::insert` and `DB::update` values.

### Checklist

- [x] Add tests and ensure they pass
- [ ] Add an entry to the CHANGELOG.md file
- [ ] Update documentation for new features

PR #3119, opened 2024-08-27
PHPORM-234 Convert dates in DB Query results

GromNaN:
Fix [PHPORM-234](https://jira.mongodb.org/browse/PHPORM-234)
Related to https://github.com/mongodb/laravel-mongodb/pull/3105#issuecomment-2302221218

~Blocked by [PHPC-2429](https://jira.mongodb.org/browse/PHPC-2429)~

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 * Remove support for Laravel 10 by @GromNaN in [#3123](https://github.com/mongodb/laravel-mongodb/pull/3123)
 * **BREAKING CHANGE** Use `id` as an alias for `_id` in commands and queries for compatibility with Eloquent packages by @GromNaN in [#3040](https://github.com/mongodb/laravel-mongodb/pull/3040)
 * **BREAKING CHANGE** Make Query\Builder return objects instead of array to match Laravel behavior by @GromNaN in [#3107](https://github.com/mongodb/laravel-mongodb/pull/3107)
+* **BREAKING CHANGE** In DB query results, convert BSON `UTCDateTime` objects into `Carbon` date with the default timezone by @GromNaN in [#3119](https://github.com/mongodb/laravel-mongodb/pull/3119)
 * Remove `MongoFailedJobProvider`, replaced by Laravel `DatabaseFailedJobProvider` by @GromNaN in [#3122](https://github.com/mongodb/laravel-mongodb/pull/3122)
 
 ## [4.8.0] - 2024-08-27
--- a/src/Eloquent/DocumentModel.php
+++ b/src/Eloquent/DocumentModel.php
@@ -5,12 +5,14 @@ declare(strict_types=1);
 namespace MongoDB\Laravel\Eloquent;
 
 use BackedEnum;
+use Carbon\Carbon;
 use Carbon\CarbonInterface;
 use DateTimeInterface;
 use DateTimeZone;
 use Illuminate\Contracts\Queue\QueueableCollection;
 use Illuminate\Contracts\Queue\QueueableEntity;
 use Illuminate\Contracts\Support\Arrayable;
+use Illuminate\Database\Eloquent\Concerns\HasAttributes;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Support\Arr;
@@ -97,8 +99,14 @@ trait DocumentModel
         return $this->getKeyName();
     }
 
-    /** @inheritdoc */
-    public function fromDateTime($value)
+    /**
+     * Convert a DateTime to a storable UTCDateTime.
+     *
+     * @see HasAttributes::fromDateTime()
+     *
+     * @param  mixed $value
+     */
+    public function fromDateTime($value): UTCDateTime
     {
         // If the value is already a UTCDateTime instance, we don't need to parse it.
         if ($value instanceof UTCDateTime) {
@@ -113,8 +121,14 @@ trait DocumentModel
         return new UTCDateTime($value);
     }
 
-    /** @inheritdoc */
-    protected function asDateTime($value)
+    /**
+     * Return a timestamp as DateTime object.
+     *
+     * @see HasAttributes::asDateTime()
+     *
+     * @param  mixed $value
+     */
+    protected function asDateTime($value): Carbon
     {
         // Convert UTCDateTime instances to Carbon.
         if ($value instanceof UTCDateTime) {
--- a/src/Eloquent/DocumentModel.php
+++ b/src/Eloquent/DocumentModel.php
@@ -100,7 +100,7 @@ trait DocumentModel
     }
 
     /**
-     * Convert a DateTime to a storable UTCDateTime.
+     * Convert a DateTimeInterface (including Carbon) to a storable UTCDateTime.
      *
      * @see HasAttributes::fromDateTime()
      *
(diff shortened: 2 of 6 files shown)


src/Eloquent/DocumentModel.php
```diff
@@ -113,18 +118,6 @@ public function fromDateTime($value)
...
-    protected function asDateTime($value)
-    {
-        // Convert UTCDateTime instances to Carbon.
-        if ($value instanceof UTCDateTime) {
-            return Date::instance($value->toDateTime())
-                ->setTimezone(new DateTimeZone(date_default_timezone_get()));
-        }
-
-        return parent::asDateTime($value);
-    }
```
GromNaN: This is not necessary at model level because the conversion is done in query result.

tests/QueryBuilderTest.php
```diff
@@ -676,27 +676,32 @@ public function testUpdateSubdocument()
...
     {
         DB::table('users')->insert([
-            ['name' => 'John Doe', 'birthday' => new UTCDateTime(Date::parse('1980-01-01 00:00:00'))],
-            ['name' => 'Robert Roe', 'birthday' => new UTCDateTime(Date::parse('1982-01-01 00:00:00'))],
-            ['name' => 'Mark Moe', 'birthday' => new UTCDateTime(Date::parse('1983-01-01 00:00:00.1'))],
-            ['name' => 'Frank White', 'birthday' => new UTCDateTime(Date::parse('1960-01-01 12:12:12.1'))],
+            ['name' => 'John Doe', 'birthday' => Date::parse('1980-01-01 00:00:00')],
+            ['name' => 'Robert Roe', 'birthday' => Date::parse('1982-01-01 00:00:00')],
+            ['name' => 'Mark Moe', 'birthday' => Date::parse('1983-01-01 00:00:00.1')],
+            ['name' => 'Frank White', 'birthday' => Date::parse('1975-01-01 12:12:12.1')],
```
GromNaN, line 682: Changed the date to avoid the issue reported in [PHPC-2429](https://jira.mongodb.org/browse/PHPC-2429)

masterbater:
Neat, is this the last feature for mongodb 5.0, excited the for the release thanks for the hardwork everyone @GromNaN  @alcaeus @norareidy for continuing this open source package and making better laravel compatibility.  I know my comment is useless to you guys and unnecessary, I just want to cheer you guys. I cant pr but I can report issue and other integration issue if I found one. Have a good day!

tests/QueryBuilderTest.php
```diff
@@ -676,27 +676,32 @@ public function testUpdateSubdocument()
...
         ]);
 
         $user = DB::table('users')
-            ->where('birthday', new UTCDateTime(Date::parse('1980-01-01 00:00:00')))
+            ->where('birthday', Date::parse('1980-01-01 00:00:00'))
```
jmikola, line 686: Noted that [`Date::parse()`](https://laravel.com/api/11.x/Illuminate/Support/Facades/Date.html#method_parse) returns a Carbon instance

src/Query/Builder.php
```diff
@@ -1660,7 +1663,10 @@ private function aliasIdForResult(array|object $values): array|object
             }
 
             foreach ($values as $key => $value) {
-                if (is_array($value) || is_object($value)) {
+                if ($value instanceof UTCDateTime) {
+                    $values[$key] = Date::instance($value->toDateTime())
+                        ->setTimezone(new DateTimeZone(date_default_timezone_get()));
+                } elseif (is_array($value) || $value instanceof stdClass) {
```
jmikola: This changes `is_object()` to a stdClass check. Was that intentional?

I'm not familiar with how `aliasIdForResult()` is used, but this looks like it may be an optimization to avoid unnecessary recursion. `aliasIdForResult()` only handled array and stdClass instances and returned `$values` as-is otherwise.

src/Query/Builder.php
```diff
@@ -1660,7 +1663,10 @@ private function aliasIdForResult(array|object $values): array|object
             }
 
             foreach ($values as $key => $value) {
-                if (is_array($value) || is_object($value)) {
+                if ($value instanceof UTCDateTime) {
+                    $values[$key] = Date::instance($value->toDateTime())
+                        ->setTimezone(new DateTimeZone(date_default_timezone_get()));
```
jmikola, line 1668: I'll note that ODM also applies the default timezone in its [DateType class](https://github.com/doctrine/mongodb-odm/blob/2.8.1/lib/Doctrine/ODM/MongoDB/Types/DateType.php#L50).

src/Eloquent/DocumentModel.php
```diff
@@ -97,8 +99,14 @@ public function getQualifiedKeyName()
         return $this->getKeyName();
     }
 
-    /** @inheritdoc */
-    public function fromDateTime($value)
+    /**
+     * Convert a DateTime to a storable UTCDateTime.
```
jmikola: Should this refer to DateTimeInterface? You can also write "DateTimeInterface (including Carbon)" for more clarity.

Related thoughts (no action needed) follow.

Although UTCDateTime accepts a DateTimeInterface, it doesn't actually rely on any of its methods. That interface is effectively used in place of a union on DateTime and DateTimeImmutable. The [UTCDateTime constructor](https://github.com/mongodb/mongo-php-driver/blob/1.19.3/src/BSON/UTCDateTime.c#L151) accesses the internal `php_date_obj` struct directly.

Thankfully, this is safe because [DateTimeInterface](https://www.php.net/manual/en/class.datetimeinterface.php) is not able to be implemented by userland classes. So any instance will be either a DateTime or DateTimeImmutable object with the internal struct.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 
 * **BREAKING CHANGE** Use `id` as an alias for `_id` in commands and queries for compatibility with Eloquent packages by @GromNaN in [#3040](https://github.com/mongodb/laravel-mongodb/pull/3040)
 * **BREAKING CHANGE** Make Query\Builder return objects instead of array to match Laravel behavior by @GromNaN in [#3107](https://github.com/mongodb/laravel-mongodb/pull/3107)
+* **BREAKING CHANGE** Convert BSON `UTCDateTime` objects into `Carbon` date in query results by @GromNaN in [#3119](https://github.com/mongodb/laravel-mongodb/pull/3119)
```
jmikola: Would it be worth noting that the converted object gets the default timezone applied?

If Carbon objects were previously converted to UTCDateTime and left as-is, I suppose users would already be accustomed to having any original time zone dropped for UTC. But seeing that Carbon types are now returned might lead them to believe that the original Carbon object's timezone is restored, which isn't necessarily the case.

Doing so would require preserving the original timezone alongside the UTCDateTime in some kind of nested object (see POC in [PHPC-640](https://jira.mongodb.org/browse/PHPC-640)). I'm not advocating for such an implementation -- just pointing out that it was a consideration during the early designs of PHPC's BSON APIs.

jmikola (approved):

jmikola (approved):
Two questions but nothing to hold up merging.

src/Query/Builder.php
GromNaN: Yes, but this method may accept any class defined in the typeMap. Not something supported for now as the typeMap cannot be modified. Tracked in [PHPORM-233](https://jira.mongodb.org/browse/PHPORM-233)

src/Query/Builder.php
GromNaN: Good to know, thanks for the investigation.

CHANGELOG.md
GromNaN: Storing the timezone is something we could consider with custom date cast (need a [mutator](https://laravel.com/docs/11.x/eloquent-mutators))

PR #3136, opened 2024-09-05
PHPORM-236 Remove _id from query results

GromNaN:
Fix PHPORM-236
related to https://github.com/mongodb/laravel-mongodb/pull/3040

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 ## [5.0.0] - next
 
 * Remove support for Laravel 10 by @GromNaN in [#3123](https://github.com/mongodb/laravel-mongodb/pull/3123)
-* **BREAKING CHANGE** Use `id` as an alias for `_id` in commands and queries for compatibility with Eloquent packages by @GromNaN in [#3040](https://github.com/mongodb/laravel-mongodb/pull/3040)
+* **BREAKING CHANGE** Use `id` as an alias for `_id` in commands and queries for compatibility with Eloquent packages by @GromNaN in [#3040](https://github.com/mongodb/laravel-mongodb/pull/3040) and [#3136](https://github.com/mongodb/laravel-mongodb/pull/3136)
 * **BREAKING CHANGE** Make Query\Builder return objects instead of array to match Laravel behavior by @GromNaN in [#3107](https://github.com/mongodb/laravel-mongodb/pull/3107)
 * **BREAKING CHANGE** In DB query results, convert BSON `UTCDateTime` objects into `Carbon` date with the default timezone by @GromNaN in [#3119](https://github.com/mongodb/laravel-mongodb/pull/3119)
 * Remove `MongoFailedJobProvider`, replaced by Laravel `DatabaseFailedJobProvider` by @GromNaN in [#3122](https://github.com/mongodb/laravel-mongodb/pull/3122)
--- a/docs/includes/usage-examples/FindOneTest.php
+++ b/docs/includes/usage-examples/FindOneTest.php
@@ -31,6 +31,6 @@ class FindOneTest extends TestCase
         // end-find-one
 
         $this->assertInstanceOf(Movie::class, $movie);
-        $this->expectOutputRegex('/^{"_id":"[a-z0-9]{24}","title":"The Shawshank Redemption","directors":\["Frank Darabont","Rob Reiner"\],"id":"[a-z0-9]{24}"}$/');
+        $this->expectOutputRegex('/^{"title":"The Shawshank Redemption","directors":\["Frank Darabont","Rob Reiner"\],"id":"[a-z0-9]{24}"}$/');
     }
 }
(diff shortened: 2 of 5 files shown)


docs/includes/usage-examples/FindOneTest.php
```diff
@@ -31,6 +31,6 @@ public function testFindOne(): void
         // end-find-one
 
         $this->assertInstanceOf(Movie::class, $movie);
-        $this->expectOutputRegex('/^{"_id":"[a-z0-9]{24}","title":"The Shawshank Redemption","directors":\["Frank Darabont","Rob Reiner"\],"id":"[a-z0-9]{24}"}$/');
+        $this->expectOutputRegex('/^{"title":"The Shawshank Redemption","directors":\["Frank Darabont","Rob Reiner"\],"id":"[a-z0-9]{24}"}$/');
```
GromNaN, line 34: We can't preserve the key/property order [without creating a new object/array](https://luigicavalieri.com/blog/renaming-keys-of-multidimensional-associative-array-while-preserving-elements-order/). This overhead and complexity is not justified here. "id" will always be last.

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 ## [5.0.0] - next
 
 * Remove support for Laravel 10 by @GromNaN in [#3123](https://github.com/mongodb/laravel-mongodb/pull/3123)
-* **BREAKING CHANGE** Use `id` as an alias for `_id` in commands and queries for compatibility with Eloquent packages by @GromNaN in [#3040](https://github.com/mongodb/laravel-mongodb/pull/3040)
+* **BREAKING CHANGE** Use `id` as an alias for `_id` in commands and queries for compatibility with Eloquent packages by @GromNaN in [#3040](https://github.com/mongodb/laravel-mongodb/pull/3040) and [#3136](https://github.com/mongodb/laravel-mongodb/pull/3136)
```
jmikola, line 7: I added an association between [PHPORM-147](https://jira.mongodb.org/browse/PHPORM-147) and [PHPORM-236](https://jira.mongodb.org/browse/PHPORM-236) since these appear to be closely related.

docs/includes/usage-examples/FindOneTest.php
jmikola: I realize this was in the original pattern, so it's not a change from this PR, but what causes the ObjectId to be printed as a plain string? It doesn't look like this library overrides `Model::toJson()`, so is the ObjectId typing dropped entirely for the model's in-memory representation?

tests/QueryBuilderTest.php
```diff
@@ -449,25 +449,34 @@ public function testDistinct()
...
         $item = DB::table('items')->find('knife');
         $this->assertEquals('knife', $item->id);
+        $this->assertObjectNotHasProperty('_id', $item);
+        $this->assertEquals('sharp', $item->tags[0]['id']);
+        $this->assertArrayNotHasKey('_id', $item->tags[0]);
 
         $item = DB::table('items')->where('id', 'fork')->first();
         $this->assertEquals('fork', $item->id);
 
+        // tags.id is translated into tags._id in query
```
jmikola, line 471: There are plenty of assertions in this test for `_id` _not_ existing on the model object(s), but what actually asserts that `_id` is stored in the database? Presumably, the query below would work even if `id` was never translated to `_id` on the way to the server.

If it's not the intention to test that here, feel free to resolve. This comment made me think otherwise.

tests/Queue/Failed/DatabaseFailedJobProviderTest.php
```diff
@@ -77,7 +77,7 @@ public function testAll(): void
         $all = $this->getProvider()->all();
 
         $this->assertCount(5, $all);
-        $this->assertEquals(new ObjectId(sprintf('%024d', 5)), $all[0]->_id);
+        $this->assertEquals(new ObjectId(sprintf('%024d', 5)), $all[0]->id);
         $this->assertEquals(sprintf('%024d', 5), $all[0]->id, 'id field is added for compatibility with DatabaseFailedJobProvider');
```
jmikola: Unrelated to the diff, but would this benefit from using `assertSame` to also check that a string type is used?

jmikola (approved):
A few questions, but I don't think any changes are required.

rustagir (approved):

docs/includes/usage-examples/FindOneTest.php
GromNaN: By default, `id` is casted into string when `Model::toArray()` is called.

tests/QueryBuilderTest.php
GromNaN: I added a query on `_id` field in this test.

tests/Queue/Failed/DatabaseFailedJobProviderTest.php
GromNaN: They are not the same. 2 distinct objects line 80.
I add assertions on the type of the "id" property.

mshamaseen:
This change results in omitting `_id` from `toArray()` and `toJson()`. Thus, introducing a high-impact breaking change.

GromNaN:
See https://github.com/mongodb/laravel-mongodb/issues/3184 for the feature request to request the `_id` field.

PR #3137, opened 2024-09-05
PHPORM-56 Replace Collection proxy class with Driver monitoring

GromNaN:
Fix PHPORM-56

Use [MongoDB driver monitoring](https://www.php.net/manual/en/mongodb.tutorial.apm.php) feature to log all commands.

Benefit: any 3rd party class that accepts a `MongoDB\Collection` instance can be used and still having query logs in Laravel.

Upgrade guide:
- Change type hints from `MongoDB\Laravel\Collection` to `MongoDB\Collection`.
```diff
- use MongoDB\Laravel\Collection;
+ use MongoDB\Collection;
```

### Checklist


composer.json
```diff
@@ -30,7 +30,7 @@
         "illuminate/database": "^11",
         "illuminate/events": "^11",
         "illuminate/support": "^11",
-        "mongodb/mongodb": "^1.15"
+        "mongodb/mongodb": "^1.18"
```
GromNaN, line 33: Requires https://github.com/mongodb/mongo-php-library/pull/1195

norareidy (approved):

tests/ConnectionTest.php
```diff
@@ -264,6 +265,16 @@ public function testQueryLog()
...
         $this->assertCount(5, DB::getQueryLog());
+
+        // Error
+        try {
+            DB::table('items')->where('name', 'test')->update(
+                ['$set' => ['embed' => ['foo' => 'bar']], '$unset' => ['embed' => ['foo']]],
+            );
+            self::fail('Expected BulkWriteException');
+        } catch (BulkWriteException) {
+            $this->assertCount(6, $logs = DB::getQueryLog());
```
alcaeus: `$logs` isn't used anywhere
```suggestion
            $this->assertCount(6, DB::getQueryLog());
```

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 * Remove `MongoFailedJobProvider`, replaced by Laravel `DatabaseFailedJobProvider` by @GromNaN in [#3122](https://github.com/mongodb/laravel-mongodb/pull/3122)
 * Remove custom `PasswordResetServiceProvider`, use the default `DatabaseTokenRepository` by @GromNaN in [#3124](https://github.com/mongodb/laravel-mongodb/pull/3124)
 * Remove `Blueprint::background()` method by @GromNaN in [#3132](https://github.com/mongodb/laravel-mongodb/pull/3132)
+* Replace `Collection` proxy class with Driver monitoring by @GromNaN in [#3137]((https://github.com/mongodb/laravel-mongodb/pull/3137)
```
alcaeus, line 13: Might apply to other BC breaks as well, but do we want to leverage deprecations in 1.x to announce the break?

src/CommandSubscriber.php
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    private function logQuery(CommandSucceededEvent|CommandFailedEvent $event): void
+    {
+        $startedEvent = $this->commands[$event->getOperationId()];
+        unset($this->commands[$event->getOperationId()]);
+
+        $command = [];
+        foreach (get_object_vars($startedEvent->getCommand()) as $key => $value) {
+            if ($key[0] !== '$' && ! in_array($key, ['lsid', 'txnNumber'])) {
```
alcaeus, line 46: I never checked the performance of this, but why not use `str_starts_with`?

CHANGELOG.md
GromNaN: We can mark the class as deprecated in 1.x. Unless we add an option to switch the monitoring feature from the proxy class to the monitoring events, there will be no way to disable the deprecation.

src/CommandSubscriber.php
GromNaN: The array key is never empty, and I need to check only 1 char. This skips 1 method call.

```php
public function benchStrStartsWith()
{
    $key = '$db';
    return str_starts_with($key, '$');
}

public function benchStrictCmp()
{
    $key = '$db';
    return $key[0] === '$';
}
```

```
benchStrStartsWith......................I2 - Mo0.185μs (±0.26%)
benchStrictCmp..........................I2 - Mo0.135μs (±0.60%)
```

CHANGELOG.md
alcaeus: Got it - not sure we need to do this, just wanted to check 👍

alcaeus (approved):

src/Connection.php
```diff
@@ -198,6 +203,8 @@ public function ping(): void
     /** @inheritdoc */
     public function disconnect()
     {
+        $this->connection?->removeSubscriber($this->commandSubscriber);
+        $this->commandSubscriber = null;
```
jmikola, line 207: Unregistering the subscriber seems correct, but I'm more generally curious how the `disconnect()` method is used in Laravel. Wouldn't it be misleading to users given that PHPC persists the connections by default?

I assume Laravel never uses the `disableClientPersistence` driver option.

GromNaN: I'm just making sure that the objects are deleted without requiring a garbage collector cycle.

jmikola: I understand why you made the additions to this method. It was a more general question about what the purpose of `disconnect()` was for the Laravel MongoDB library. Is it something we're obligated to implement for Eloquent?

Wirone:
FYI: [5.0.0 release notes](https://github.com/mongodb/laravel-mongodb/releases/tag/5.0.0) contain crucial mistake, it says `MongoDB\Laravel\Connection` was replaced by `MongoDB\Connection`, but in fact it was Co**ll**ection. It leads to broken code if such change is applied.

image

alcaeus:
Thank you for pointing this out @Wirone! I've fixed the wrong class names in the release notes.

PR #3140, opened 2024-09-06
DOCSP-42957: DateTimeInterface in queries

norareidy:
JIRA - https://jira.mongodb.org/browse/DOCSP-42957
Staging - https://preview-mongodbnorareidy.gatsbyjs.io/laravel/DOCSP-42957-stage/upgrade/#version-5.x-breaking-changes

### Checklist

### DIFF
--- a/docs/includes/query-builder/QueryBuilderTest.php
+++ b/docs/includes/query-builder/QueryBuilderTest.php
@@ -11,6 +11,7 @@ use MongoDB\BSON\ObjectId;
 use MongoDB\BSON\Regex;
 use MongoDB\Collection;
 use MongoDB\Laravel\Tests\TestCase;
+use Carbon\Carbon;
 
 use function file_get_contents;
 use function json_decode;
--- a/docs/includes/query-builder/QueryBuilderTest.php
+++ b/docs/includes/query-builder/QueryBuilderTest.php
@@ -148,14 +148,26 @@ class QueryBuilderTest extends TestCase
         $this->assertInstanceOf(\Illuminate\Support\Collection::class, $result);
     }
 
-    public function testWhereDate(): void
+    public function testWhereCarbon(): void
     {
         // begin query where date
         $result = DB::connection('mongodb')
             ->table('movies')
-            ->where('released', Carbon::create(2010, 1, 15, 0, 0, 0, 'UTC'))
+            ->where('released', Carbon::create(2010, 1, 15))
             ->get();
         // end query where date
+
+        $this->assertInstanceOf(\Illuminate\Support\Collection::class, $result);
+    }
+
+    public function testWhereDate(): void
+    {
+        // begin query whereDate string
+        $result = DB::connection('mongodb')
+            ->table('movies')
+            ->whereDate('released', '2010-1-15')
+            ->get();
+        // end query whereDate string
 
         $this->assertInstanceOf(\Illuminate\Support\Collection::class, $result);
     }
--- a/docs/includes/query-builder/QueryBuilderTest.php
+++ b/docs/includes/query-builder/QueryBuilderTest.php
@@ -150,12 +150,13 @@ class QueryBuilderTest extends TestCase
 
     public function testWhereDate(): void
     {
-        // begin query whereDate
+        // begin query where date
         $result = DB::connection('mongodb')
             ->table('movies')
-            ->whereDate('released', '2010-1-15')
-            ->get();
-        // end query whereDate
+            ->where('released', new UTCDateTime(
+                Carbon::create(2010, 1, 15, 0, 0, 0, 'UTC')
+            ))->get();
+        // end query where date
 
         $this->assertInstanceOf(\Illuminate\Support\Collection::class, $result);
     }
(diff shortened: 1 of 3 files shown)


docs/upgrade.txt
```diff
@@ -89,6 +89,9 @@ This library version introduces the following breaking changes:
   of this behavior, you cannot have two separate ``id`` and ``_id`` fields in your
   documents.
 
+- When using a ``DateTimeInterface`` object in a query, the library converts
+  the ``DateTimeInterface`` to a ``MongoDB\BSON\UTCDateTime`` object.
```
GromNaN:
```suggestion
- When using a ``DateTimeInterface`` object (including ``Carbon``) in a query, the library converts
  the ``DateTimeInterface`` to a ``MongoDB\BSON\UTCDateTime`` object.
```

GromNaN (commented):
The same applies in the other direction. UTCDateTime objects received as results are converted to `Carbon` date with the default timezone. https://github.com/mongodb/laravel-mongodb/pull/3119

I think this should be mentioned in the query builder docs, with an example.

docs/upgrade.txt
rustagir: S: It's a bit unclear what it means to use an object "in a query". Perhaps you could provide a short example (code or descriptive)

rustagir (changes requested):
I left a suggestion that might change the scope of the PR, so requesting another look!

docs/upgrade.txt
GromNaN: Query refers to the filter passed to `Query\Builder::where()`, or the data to `insert()` or `update()`.

docs/includes/query-builder/QueryBuilderTest.php
```diff
@@ -150,12 +150,13 @@ public function testWhereIn(): void
...
-        // begin query whereDate
+        // begin query where date
         $result = DB::connection('mongodb')
             ->table('movies')
-            ->whereDate('released', '2010-1-15')
-            ->get();
-        // end query whereDate
+            ->where('released', new UTCDateTime(
+                Carbon::create(2010, 1, 15, 0, 0, 0, 'UTC')
+            ))->get();
```
GromNaN: The previous example was a lot better as it used `whereDate` method that convert the string to a date object. I don't think it makes sense to change it.

The purpose is to show that users can work with `Carbon` object without worring with `UTCDateTime`. This example shows the opposite of what should be done.

```php
->where('released', Carbon::create(2010, 1, 15));
```

docs/query-builder.txt
```diff
@@ -348,23 +348,26 @@ query builder method to retrieve documents from the
 Match Dates Example
 ^^^^^^^^^^^^^^^^^^^
 
-The following example shows how to use the ``whereDate()``
+The following example shows how to use the ``where()``
 query builder method to retrieve documents from the
-``movies`` collection that match the specified date of
-``2010-1-15`` in the ``released`` field:
+``movies`` collection in which the ``released`` value
+is ``2010, 1, 15, 0, 0, 0``, specified in a ``Carbon`` object:
```
GromNaN: The value is quoted without explaining its meanings. It's pointless. I wouldn't even put it there. The previous example was more explicit.

docs/upgrade.txt
norareidy: I added a brief description and linked to an example on the query builder page

GromNaN (changes requested):
Adding a 2nd example would make more sense that changing the existing one.

docs/upgrade.txt
```diff
@@ -81,6 +81,16 @@ This library version introduces the following breaking changes:
...
     
+- When using a ``DateTimeInterface`` object, including ``Carbon``, in a query,
+  the library converts the ``DateTimeInterface`` to a ``MongoDB\BSON\UTCDateTime``
+  object. This conversion applies to ``DateTimeInterface`` objects passed as query
+  filters to the ``where()`` method or as data passed to the ``insert()`` and
+  ``update()`` methods. 
+  
+  For an example that passes a ``Carbon`` object to the
+  ``DB::where()`` method, see the :ref:`laravel-query-builder-wheredate`
+  section of the Query Builder guide.
```
rustagir:
```suggestion
  To view an example that passes a ``Carbon`` object to the
  ``DB::where()`` method, see the :ref:`laravel-query-builder-wheredate`
  section of the Query Builder guide.
```

docs/query-builder.txt
rustagir: S: describe what the plain date meaning of this is

rustagir (approved):
your changes really cleared things up! left some small suggestions

docs/includes/query-builder/QueryBuilderTest.php
norareidy: I added back the old example and kept this one - lmk what you think

GromNaN (approved):
LGTM

docs/includes/query-builder/QueryBuilderTest.php
GromNaN: Thanks, that's perfect.

PR #3143, opened 2024-09-12
Update merge-up config for new branch pattern

GromNaN:
This PR updates the configuration for the merge-up workflow. The new `devBranchNamePattern` set to `<major>.x` to keep a single working branch `5.x`.

Related to https://github.com/alcaeus/automatic-merge-up-action/pull/41

### DIFF
--- a/.github/workflows/merge-up.yml
+++ b/.github/workflows/merge-up.yml
@@ -3,7 +3,7 @@ name: Merge up
 on:
   push:
     branches:
-      - "[0-9]+.[0-9]+"
+      - "[0-9]+.[0-9x]+"
 
 env:
   GH_TOKEN: ${{ secrets.MERGE_UP_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/merge-up.yml
```diff
@@ -27,5 +27,5 @@ jobs:
         uses: alcaeus/automatic-merge-up-action@main
         with:
           ref: ${{ github.ref_name }}
-          branchNamePattern: '<major>.<minor>'
+          branchNamePattern: '<major>.x'
```
alcaeus: If we want to have a similar pattern as we do in PHPLIB and PHPC, this won't work.

I'd assume we continue to have branches for stable versions (e.g. 4.8, 5.0, etc.) and development branches for major verisons (e.g. 4.x, 5.x). In that case, the config should be:

```suggestion
          branchNamePattern: '<major>.<minor>'
          devBranchNamePattern: '<major>.x'
```

alcaeus (changes requested):

.github/workflows/merge-up.yml
GromNaN: Oh, I reread 2 times the description of your other PRs but did not realize the property was renamed. My bad.

alcaeus (approved):

PR #3167, opened 2024-10-03
PHPORM-248 register command subscriber only when logs are enabled

### LINKED ISSUE #3166: Performance Regression 5.x
- Laravel-mongodb Version: 5.0.2
- PHP Version: 8.3.12
- Database Driver & Version: pecl/mongodb-1.20.0

### Description:

Hello,

We are experiencing a significant performance regression after upgrading to the latest version 5.0.2.

Our test suite, comprising approximately 400 tests, 2,000 assertions, and 15,000 queries, used to complete in about 15 seconds with version 4.8.0. After the upgrade, it now takes around 50 seconds to complete. (on my local dev box)

Individual HTTP requests or single tests do not exhibit any issues. In fact, performance for our average web request has slightly improved.

The issue only shows up in long-running processes, such as a test suite.

I have created a demo Laravel application that demonstrates the problem in a simple, isolated manner: [Demp-App](https://github.com/jensk/lara-mongo-debug)
We were able to reproduce the issue in 3 different environments: 
* WSL2-Ubuntu/Mongo-7 via docker, 
* Archlinux/Mongo-7 via docker, 
* Gentoo/Mongo 4.4 without docker

### Steps to reproduce
1. git clone  https://github.com/jensk/lara-mongo-debug
2. composer install
3. vendor/bin/phpunit

### Expected behaviour
Execution time of the tests should remain stable. ([Example results with 4.8](https://github.com/jensk/lara-mongo-debug/blob/main/phpunit-output-v4.8.0.txt))

### Actual behaviour
Execution time increases slightly and linearly with each call to the various methods of laravel-mongodb. (e.g. truncate, delete, create ...) ([Example results with 5.0.2](https://github.com/jensk/lara-mongo-debug/blob/main/phpunit-output-v5.0.2.txt))

GromNaN:
Fix PHPORM-248
Fix https://github.com/mongodb/laravel-mongodb/issues/3166

The command subscriber has significant impact on command execution. It can be enabled only when necessary.

### Checklist


src/Connection.php
```diff
@@ -141,6 +139,36 @@ public function getDatabaseName()
...
+    {
+        parent::enableQueryLog();
+
+        $this->commandSubscriber = new CommandSubscriber($this);
+        $this->connection->addSubscriber($this->commandSubscriber);
+    }
+
+    public function disableQueryLog()
+    {
+        parent::disableQueryLog(); // TODO: Change the autogenerated stub
```
alcaeus: There's a todo comment left 😉

src/Connection.php
```diff
@@ -141,6 +139,36 @@ public function getDatabaseName()
...
+
+        $this->connection->removeSubscriber($this->commandSubscriber);
```
alcaeus: Needs a safe guard in case no subscriber is registered

alcaeus (changes requested):

src/Connection.php
GromNaN: Added a condition to never add or remove the subscriber twice.

alcaeus (approved):

PR #3168, opened 2024-10-04
fixbug in pluck function

fuyuki0511:
Fix bug in pluck function.
Cannot use object of type stdClass as array

### Checklist

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -854,7 +854,7 @@ class Builder extends BaseBuilder
         // Convert ObjectID's to strings
         if (((string) $key) === '_id') {
             $results = $results->map(function ($item) {
-                $item['_id'] = (string) $item['_id'];
+                $item->_id = (string) $item->id;
 
                 return $item;
             });


src/Query/Builder.php
```diff
@@ -854,7 +854,7 @@ public function pluck($column, $key = null)
         // Convert ObjectID's to strings
         if (((string) $key) === '_id') {
             $results = $results->map(function ($item) {
-                $item['_id'] = (string) $item['_id'];
+                $item->_id = (string) $item->id;
```
GromNaN, line 857: I think this piece of code can be removed as `Arr::pluck` already convert stringable key value into string. https://github.com/laravel/framework/blob/646520ad682d98b5211c6e26092259cfbe130b5c/src/Illuminate/Collections/Arr.php#L562

GromNaN (changes requested):
Thank you @fuyuki0511 for contributing. I really appreciate.

Also, could you add a test in `QueryBuilderTest`. Something like this:
```php
    public function testPluckObjectId()
    {
        DB::table('users')->insert([
            ['id' => $id = new ObjectId(), 'name' => 'Jane Doe', 'age' => 20],
        ]);

        $names = DB::table('users')->pluck('name', 'id')->toArray();
        $this->assertCount(1, $names);
        $this->assertArrayHasKey((string) $id, $names);
        $this->assertEquals([(string) $id => 'Jane Doe'], $names);
    }
```

GromNaN:
I see you created a new PR https://github.com/mongodb/laravel-mongodb/pull/3169. You could have pushed to the same branch to update this PR and keep the discussion in the same place.

PR #3169, opened 2024-10-04
Fix `Query\Builder::pluck()` with `ObjectId` as key

fuyuki0511:
-Fix bug in pluck function.
Cannot use object of type stdClass as array

-Add Tests in Query\BuilderTest
-Remove the string convert of key , thanks @GromNaN


GromNaN (approved):
Thank you @fuyuki0511.

This bugfix is tracked by [PHPORM-249](https://jira.mongodb.org/browse/PHPORM-249)

PR #3182, opened 2024-10-17
PHPORM-238 Add support for `withCount` and other aggregations

GromNaN:
Fix [PHPORM-238](https://jira.mongodb.org/browse/PHPORM-238)

Implement this feature from Laravel: https://laravel.com/docs/11.x/eloquent-relationships#counting-related-models
Counting relations with SQL is done using a subquery. For MongoDB, there is 2 ways:
1. Making a query during eager loading, that count the objects for each result ID. This is an aggregation with a `$group` by foreign key.
2. Using `$lookup` to load references and count them. The process would be similar for embedded relations (without the `$lookup`). But this doesn't work with hybrid relations.

### Checklist

- [x] Support mongodb relationships
- [x] CANCELLED: Support hybrid relationships (PR reverted from Laravel: https://github.com/laravel/framework/pull/53679)
- [x] Support embedded relationships, contraint query not supported PHPORM-292


src/Query/Builder.php
```diff
@@ -556,6 +557,8 @@ public function generateCacheKey()
     /** @return ($function is null ? AggregationBuilder : mixed) */
     public function aggregate($function = null, $columns = ['*'])
     {
+        assert(is_array($columns), new TypeError(sprintf('Argument #2 ($columns) must be of type array, %s given', get_debug_type($columns))));
```
GromNaN: We cannot add the type to the argument as it's missing in the parent method. When the value is invalid here, the error comes very late in `Query\Builder::toMQL`.

src/Query/Builder.php
```diff
@@ -596,6 +599,11 @@ public function aggregate($function = null, $columns = ['*'])
         $this->columns            = $previousColumns;
         $this->bindings['select'] = $previousSelectBindings;
 
+        // When the aggregation is per group, we return the results as is.
+        if ($this->groups) {
+            return $results;
+        }
```
GromNaN: Trying to support this feature in Laravel, so that `withCount` on hybrid relations could work.
- https://github.com/laravel/framework/pull/53209

masterbater: Can you just not add withCount on hybrid relations, its a blocker.  Also a suggestion It would be better for hybrid to be a separate package instead of being integrated into laravel-mongodb if that could ease the maintenance and focus for other mongodb eloquent methods.

tests/Eloquent/EloquentWithAggregateTest.php
```diff
@@ -0,0 +1,354 @@
...
+
+        $results = EloquentWithAggregateModel1::withAvg(['twos' => $filter], 'value')->where('id', 2);
+        self::assertSameResults([
+            ['id' => 2, 'twos_avg' => 5.0],
+        ], $results->get());
+    }
+
+    public function testWithAggregateEmbedFiltered()
+    {
+        self::markTestSkipped('EmbedsMany does not support filtering. $filter requires an expression but the Query Builder generates query predicates.');
```
GromNaN: I tried to convert the query builder into a `$filter` expression, but did not succeed because this creates query predicate (`$match` syntax) while `$filter` expect an expression.

```php
    /**
     * Get the filter condition for the query.
     *
     * @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/filter/
     *
     * @param  string|array $expression Field name or expression that resolves to an array
     */
    public function getFilterField(string|array $expression): string|array
    {
        // Shortcut for simple field names
        if ($this->wheres === [] && ! $this->limit) {
            return $expression;
        }

        return [
            'input' => $expression,
            'as' => 'filtered',
            'cond' => $this->compileWheres('$$filtered.'),
            ...($this->limit ? ['limit' => $this->limit] : []),
        ];
    }
```

src/Query/Builder.php
GromNaN: Creating a new package would not help for the maintenance; we've decided not to support this feature with hybrid relations for now.

tests/HybridRelationsTest.php
```diff
@@ -206,6 +207,10 @@ public function testHybridWith()
             ->each(function ($user) {
                 $this->assertEquals($user->id, $user->books->count());
             });
+        //SqlUser::withCount('books')->get()
+        //    ->each(function ($user) {
+        //        $this->assertEquals($user->id, $user->books_count);
+        //    });
```
jmikola: Why is this commented out? If it's only intended to serve as an example then a short comment about it could be used to clarify that.

tests/QueryBuilderTest.php
```diff
@@ -575,6 +576,12 @@ public function testAggregate()
         $this->assertEquals(3, DB::table('items')->min('amount'));
         $this->assertEquals(34, DB::table('items')->max('amount'));
         $this->assertEquals(17.75, DB::table('items')->avg('amount'));
+        $this->assertTrue(DB::table('items')->exists());
+        $this->assertTrue(DB::table('items')->where('name', 'knife')->exists());
+        $this->assertFalse(DB::table('items')->where('name', 'ladle')->exists());
+        $this->assertFalse(DB::table('items')->doesntExist());
+        $this->assertFalse(DB::table('items')->where('name', 'knife')->doesntExist());
+        $this->assertTrue(DB::table('items')->where('name', 'ladle')->doesntExist());
```
jmikola: This looks like basic tests for `exists()` and `doesntExist()` methods. How is this related to the PR?

tests/HybridRelationsTest.php
```diff
@@ -157,6 +157,7 @@ public function testHybridWhereHas()
 
     public function testHybridWith()
     {
+        DB::connection('mongodb')->enableQueryLog();
```
jmikola: Is `enableQueryLog()` necessary for these test? This looks like it may have been left over from local testing.

tests/Eloquent/EloquentWithAggregateTest.php
```diff
@@ -0,0 +1,257 @@
...
+        self::assertSameResults([
+            ['id' => 2, 'twos_max' => 6],
+        ], $results->get());
+
+        $results = EloquentWithAggregateModel1::withMin(['twos' => $filter], 'value')->where('id', 2);
+        self::assertSameResults([
+            ['id' => 2, 'twos_min' => 4],
+        ], $results->get());
+
+        $results = EloquentWithAggregateModel1::withAvg(['twos' => $filter], 'value')->where('id', 2);
```
jmikola, line 152: In QueryBuilderTest, I noted there were two names for the same method: `avg()` and `average()`. Is there also a `withAverage()` method that should be tested?

tests/Eloquent/EloquentWithAggregateTest.php
```diff
@@ -0,0 +1,257 @@
...
+        ]);
+
+        self::assertSameResults([
+            ['id' => 1, 'twos_count' => 2],
+            ['id' => 2, 'twos_count' => 0],
+            ['id' => 3, 'twos_count' => 1],
+            ['id' => 4, 'twos_count' => 0],
+        ], $results->get());
+
+        self::assertSame(2, count($connection->getQueryLog()));
```
jmikola, line 206: It'd be helpful if this test explained why exactly two query log entries are expected. I assume the answer is the same for each aggregate method, so I'd only expect a comment here for the first case.

tests/Eloquent/EloquentWithAggregateTest.php
```diff
@@ -0,0 +1,257 @@
...
+        EloquentWithAggregateModel2::truncate();
+        EloquentWithAggregateModel3::truncate();
+        EloquentWithAggregateModel4::truncate();
+
+        parent::tearDown();
+    }
+
+    public function testWithAggregate()
+    {
+        EloquentWithAggregateModel1::create(['id' => 1]);
```
jmikola, line 33: Does this first fixture serve any purpose? All of the cases below use `where('id', 2)`.

This same question applies to `testWithAggregateEmbed()` and `testWithAggregateFiltered()` below.

tests/Eloquent/EloquentWithAggregateTest.php
```diff
@@ -0,0 +1,257 @@
...
+
+        EloquentWithAggregateModel1::withCount('hybrids')->where('id', 2)->get();
+    }
+
+    private static function assertSameResults(array $expected, Collection $collection)
+    {
+        $actual = $collection->toArray();
+
+        foreach ($actual as &$item) {
+            ksort($item);
```
jmikola, line 289: PHPLIB uses MultipleIterator for comparing an iterable with an expected result, but I can imagine mapping `ksort()` to both arguments and finishing with `assertSame()` is easier so long as both values are simple arrays and you don't need to worry about objects within.

src/Query/Builder.php
```diff
@@ -650,6 +650,14 @@ public function aggregateByGroup(string $function, array $columns = ['*'])
         return $this->aggregate($function, $columns);
     }
 
+    public function count($columns = '*')
+    {
+        // Can be removed when available in Laravel: https://github.com/laravel/framework/pull/53209
```
jmikola, line 655: It looks like this PR was merged, then reverted, and then superseded by https://github.com/laravel/framework/pull/53679. Should this be updated?

src/Eloquent/Builder.php
```diff
@@ -294,6 +306,91 @@ public function createOrFirst(array $attributes = [], array $values = [])
...
 
+    public function withAggregate($relations, $column, $function = null)
+    {
+        if (empty($relations)) {
+            return $this;
+        }
+
+        $relations = is_array($relations) ? $relations : [$relations];
+
+        foreach ($this->parseWithRelations($relations) as $name => $constraints) {
```
jmikola, line 340: I assume this calls [this method on the base class](https://github.com/illuminate/database/blob/eab4e652ebfabed02b33770a69391dd1aab759c6/Eloquent/Builder.php#L1645) but it's not clear what "parse a list of relations into individuals" means for the structure of the return value, which is only described as an array. In particular, it's not clear how `"<field> as <alias>"` would be returned.

Would a `$name` with exactly three space-delimited segments always have that structure?

src/Eloquent/Builder.php
```diff
@@ -294,6 +306,91 @@ public function createOrFirst(array $attributes = [], array $values = [])
...
+            $segments = explode(' ', $name);
+
+            $name = $segments[0];
+            $alias = (count($segments) === 3 && Str::lower($segments[1]) === 'as' ? $segments[2] : Str::snake($name) . '_' . $function);
```
jmikola: Are the only alternatives here `count($segments) === 3` or a single segment? If so, is it worth asserting on this instead of failing back to `"<name>_<function>"`?

src/Eloquent/Builder.php
```diff
@@ -294,6 +306,91 @@ public function createOrFirst(array $attributes = [], array $values = [])
...
+                $subQuery = $this->newQuery();
+                $constraints($subQuery);
+                if ($subQuery->getQuery()->wheres) {
+                    // @see https://jira.mongodb.org/browse/PHPORM-292
+                    throw new InvalidArgumentException('Constraints are not supported for embedded relations');
+                }
+
+                switch ($function) {
+                    case 'count':
+                        $this->getQuery()->project([$alias => ['$size' => ['$ifNull' => ['$' . $name, []]]]]);
```
jmikola, line 366: Does `$this->getQuery()` return an aggregation builder, to which you add a `$project` stage?

src/Eloquent/Builder.php
```diff
@@ -294,6 +306,91 @@ public function createOrFirst(array $attributes = [], array $values = [])
         }
     }
 
+    public function withAggregate($relations, $column, $function = null)
```
jmikola, line 330: I originally didn't find what this was based on, but traced it to [Illuminate\Database\Eloquent\Concerns\QueriesRelationships](https://github.com/laravel/framework/blob/f958f330417a46c72f35df9b49a629219d541749/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php#L690), which is then used by the base class.

The corresponding method in that trait has comments to help explain some of the logic, but there's none of that here. I won't advocate for copying all of those comments (although I think if any are relevant they'd do well to be added here), but we should at least reference the trait where this is adapted from so we know what it needs to be kept in sync with upstream.

src/Eloquent/Builder.php
```diff
@@ -294,6 +306,91 @@ public function createOrFirst(array $attributes = [], array $values = [])
...
+                        ->where($withAggregate['constraints'])
+                        ->whereIn($withAggregate['relation']->getForeignKeyName(), $modelIds)
+                        ->groupBy($withAggregate['relation']->getForeignKeyName())
+                        ->aggregate($withAggregate['function'], [$withAggregate['column']]);
+
+                    foreach ($models as $model) {
+                        $value = $withAggregate['function'] === 'count' ? 0 : null;
+                        foreach ($results as $result) {
+                            if ($model->getKey() === $result->{$withAggregate['relation']->getForeignKeyName()}) {
+                                $value = $result->aggregate;
```
jmikola, line 408: What assurance is there that `$result->aggregate` is set, or of the correct type?

src/Eloquent/Builder.php
```diff
@@ -294,6 +306,91 @@ public function createOrFirst(array $attributes = [], array $values = [])
...
+        return $this;
+    }
+
+    public function eagerLoadRelations(array $models)
+    {
+        if ($this->withAggregate) {
+            $modelIds = collect($models)->pluck($this->model->getKeyName())->all();
+
+            foreach ($this->withAggregate as $withAggregate) {
+                if ($withAggregate['relation'] instanceof HasMany) {
```
jmikola, line 397: Given that this only applies some processing for HasMany relations, under what circumstances would different relations get appended to `$this->withAggregate`? And are they OK being left as-is?

src/Eloquent/Builder.php
```diff
@@ -34,6 +43,9 @@ class Builder extends EloquentBuilder
     private const DUPLICATE_KEY_ERROR = 11000;
     use QueriesRelationships;
 
+    /** @var array{relation: Relation, function: string, constraints: array, column: string, alias: string}[] */
+    private array $withAggregate = [];
```
jmikola, line 56: It'd be helpful to leave a comment explaining that this property exists to share state between `withAggregate()` and `eagerLoadRelations()`.

It struck me as odd that in base Eloquent that trait and base method were seemingly unrelated, but they become closely related in this PR. I don't understand why that was necessary, but I'm more concerned about the lack of documentation here.

There's nothing that explains why this property was necessary, or exactly how it's used. If you'd rather not do that here, then perhaps within the conditional in `withAggregate()` where this is actually used.

src/Eloquent/Builder.php
```diff
@@ -294,6 +306,91 @@ public function createOrFirst(array $attributes = [], array $values = [])
...
+                    case 'min':
+                    case 'max':
+                    case 'avg':
+                        $this->getQuery()->project([$alias => ['$' . $function => '$' . $name . '.' . $column]]);
+                        break;
+                    default:
+                        throw new InvalidArgumentException(sprintf('Invalid aggregate function "%s"', $function));
+                }
+            } else {
+                $this->withAggregate[$alias] = [
```
jmikola, line 378: If this is the only place `withAggregate()` is appended to, it'd be helpful to explain why it's necessary to save this state in the property for future use by `eagerLoadRelations()`.

Also, per my other question about `eagerLoadRelations()` only handling HasMany, this may be the appropriate place to assert that you're only handling HasMany relations (if you decide everything else is rightfully ignored).

src/Query/Builder.php
GromNaN: Then reverted again: https://github.com/laravel/framework/pull/54196

tests/Eloquent/EloquentWithAggregateTest.php
GromNaN: This ensures that the initial query filter on `id` works (perhaps I'm being too creative about what might fail).

tests/HybridRelationsTest.php
GromNaN: Leftover. Removed.

tests/HybridRelationsTest.php
GromNaN: That was for debugging the query. Removed.

tests/QueryBuilderTest.php
GromNaN: I realized that was not tested when working on the aggregate method. That could be part of [PHPORM-285](https://jira.mongodb.org/browse/PHPORM-285) that I had to postpone.

src/Eloquent/Builder.php
GromNaN: The `aggregate($function, $column)` method always adds the `aggregate` field: https://github.com/mongodb/laravel-mongodb/blob/2b2c70a66279f9206085c857cb30f44ed52d8785/src/Query/Builder.php#L367-L378

src/Eloquent/Builder.php
GromNaN: Adding an exception in the "else" case. I thing the only other case if "HasOne", but I need to check "BelongsToMany".

src/Eloquent/Builder.php
GromNaN: This need to be there to be processed during the `eagerLoadRelations` step. https://github.com/laravel/framework/blob/f958f330417a46c72f35df9b49a629219d541749/src/Illuminate/Database/Eloquent/Builder.php#L766-L771

src/Eloquent/Builder.php
GromNaN: I added the doc to the method. The architecture is different as we use eager loading instead of an SQL subquery.

src/Eloquent/Builder.php
GromNaN: It's adding the constraint closure from the scope and the relationships.

src/Eloquent/Builder.php
GromNaN: Exception added.

src/Eloquent/Builder.php
GromNaN: It returns a `MongoDB\Laravel\Query\Builder`, with the method `project`: https://github.com/mongodb/laravel-mongodb/blob/2b2c70a66279f9206085c857cb30f44ed52d8785/src/Query/Builder.php#L207-L209

src/Eloquent/Builder.php
GromNaN: Comment added. I'll check the supported relationships.

tests/Eloquent/EloquentWithAggregateTest.php
GromNaN: Comment added.

GromNaN (commented):
Thanks, I addressed most comments but need to add tests on the new error cases.

src/Eloquent/Builder.php
```diff
@@ -306,19 +315,37 @@ public function createOrFirst(array $attributes = [], array $values = [])
...
+        assert(is_string($function), new TypeError('Argument 3 ($function) passed to withAggregate must be of the type string, ' . get_debug_type($function) . ' given'));
+
         $relations = is_array($relations) ? $relations : [$relations];
 
         foreach ($this->parseWithRelations($relations) as $name => $constraints) {
             $segments = explode(' ', $name);
 
+            $alias = match (true) {
+                count($segments) === 1 => Str::snake($segments[0]) . '_' . $function,
+                count($segments) === 3 && Str::lower($segments[1]) => $segments[2],
```
jmikola: Is this missing `Str::lower($segments[1]) === 'as'`?

GromNaN: Thanks for pointing, I restored the condition. And I added tests to the alias feature.

tests/QueryBuilderTest.php
jmikola: I'd agree that this is better off in a separate PR/commit. Or include it here but reference PHPORM-285 in the commit message (which can still be squashed at merge time, but at least we have a note pointing to the relevant issue).

src/Query/Builder.php
jmikola: In either case, the reference to https://github.com/laravel/framework/pull/53209 looks outdated. Please adjust accordingly and then feel free to resolve this thread.

tests/QueryBuilderTest.php
GromNaN: I removed this tests and created https://github.com/mongodb/laravel-mongodb/pull/3257
I realized, only doesntExist was not tested.

PR #3190, opened 2024-10-31
Supports disabling _id to id automatic conversion and disablling returning results as arrays in aggregate instead of bson object based on database configuration

khaleelrasakh:
This is fork supports disabling _id to id automatic conversion and disablling returning results as arrays in aggregate instead of bson object based on database configuration.
 'mongodb' => [
                  'driver' => 'mongodb',
                  'dsn' => env('DB_URI'),
                  'database' => env('DB_NAME'),
                  'options'  => [
                   
                    'ssl' => true,
                    'AggregateCollectionArray'=>true, //add this to disable aggregate array
                    'DisableAliasIdForResult'=>true, //add this to disable automatic _id conversion
                    'tlsAllowInvalidCertificates' => true, 
                ],
          ],
### Checklist
This is indeed required for me cause we have developed full fledged ERP Project with 20+ modules using v4 in completely Laravel. Changing all the 
code is not practical

### DIFF
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -181,7 +181,10 @@ class Builder extends EloquentBuilder
 
         // Convert MongoCursor results to a collection of models.
         if ($results instanceof CursorInterface) {
-            $results->setTypeMap(['root' => 'array', 'document' => 'array', 'array' => 'array']);
+            if(!config('database.connections.mongodb.options.AggregateCollectionArray')){
+                $results->setTypeMap(['root' => 'array', 'document' => 'array', 'array' => 'array']);
+            }
+            
             $results = $this->query->aliasIdForResult(iterator_to_array($results));
 
             return $this->model->hydrate($results);
--- a/README.md
+++ b/README.md
@@ -1,6 +1,18 @@
 Laravel MongoDB
 ===============
-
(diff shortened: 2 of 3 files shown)


GromNaN:
Sorry, but this option would make package maintenance very complex. We decided to make this alias automatic in order to support Laravel features that couldn't be supported before. We can't guarantee the same level of features by deactivating it. Stay with version 4 until you find a solution for renaming the fields in your database or use MongoDB library to query the database, without the Laravel Query abstraction.

src/Eloquent/Builder.php
```diff
@@ -181,7 +181,10 @@ public function raw($value = null)
 
         // Convert MongoCursor results to a collection of models.
         if ($results instanceof CursorInterface) {
-            $results->setTypeMap(['root' => 'array', 'document' => 'array', 'array' => 'array']);
+            if(!config('database.connections.mongodb.options.AggregateCollectionArray')){
+                $results->setTypeMap(['root' => 'array', 'document' => 'array', 'array' => 'array']);
```
GromNaN, line 185: If your callback does not return a `CursorInterface`, this behavior will be disabled. Use `Cursor::toArray()` for example.

```php
Model::raw(function (\MongoDB\Collection $collection): array {
    $cursor = $collection->find([], ['typeMap' => ['array' => 'bson']);

    return $cursor->toArray();
});
```

PR #3192, opened 2024-11-05
PHPORM-259 Register MongoDB Session Handler with `SESSION_DRIVER=mongodb`

GromNaN:
Fix [PHPORM-259](https://jira.mongodb.org/browse/PHPORM-259)

Adding the "mongodb" session driver that uses [Symfony MongoDbSessionHandler](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php) to store the sessions. This handler is optimized and maintained in the Symfony repository.

Registration:

1. Configure a [`mongodb` connection](https://www.mongodb.com/docs/drivers/php/laravel-mongodb/v5.x/quick-start/configure-mongodb/#set-the-connection-string-in-the-database-configuration) in `config/database.php`.

2. Select the session driver and connection with `.env` file
```ini
SESSION_DRIVER=mongodb
SESSION_CONNECTION=mongodb # optional, this is the default value
```

Or in `config/session.php`
```php
<?php return [
    'driver' => 'mongodb',     // required
    'connection' => 'mongodb', // the database connection name, default is "mongodb"
    'table' => 'sessions',     // the collection name, default is "sessions"
    'lifetime' => null,        // the ttl of session in minutes, default is 120 minutes
    'options' => []            // Additional driver options
];
```

List of additional options and some documentation in the class: https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php#L39-L46

### DIFF
--- a/src/MongoDBServiceProvider.php
+++ b/src/MongoDBServiceProvider.php
@@ -10,6 +10,7 @@ use Illuminate\Cache\Repository;
 use Illuminate\Filesystem\FilesystemAdapter;
 use Illuminate\Filesystem\FilesystemManager;
 use Illuminate\Foundation\Application;
+use Illuminate\Session\SessionManager;
 use Illuminate\Support\ServiceProvider;
 use InvalidArgumentException;
 use League\Flysystem\Filesystem;
@@ -20,6 +21,7 @@ use MongoDB\Laravel\Cache\MongoStore;
 use MongoDB\Laravel\Eloquent\Model;
 use MongoDB\Laravel\Queue\MongoConnector;
 use RuntimeException;
+use Symfony\Component\HttpFoundation\Session\Storage\Handler\MongoDbSessionHandler;
 
 use function assert;
 use function class_exists;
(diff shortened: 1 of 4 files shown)


src/MongoDBServiceProvider.php
```diff
@@ -20,6 +21,7 @@
 use MongoDB\Laravel\Eloquent\Model;
 use MongoDB\Laravel\Queue\MongoConnector;
 use RuntimeException;
+use Symfony\Component\HttpFoundation\Session\Storage\Handler\MongoDbSessionHandler;
```
alcaeus, line 24: Noted that Laravel itself has a dependency on `symfony/http-foundation`. Still, I wonder whether we should either add it to `require` or `require-dev`, and/or add a `conflict` entry to ensure we're getting a version of the session handler that we expect instead of relying on the transitive dependency.

alcaeus (commented):
Code LGTM, just wondering about the handling of the implicit dependency.

src/MongoDBServiceProvider.php
GromNaN: In fact, it's best to be explicit about this dependency. Laravel 10 requires symfony 6.4 while laravel 11 requires symfony 7.

alcaeus (commented):
Code changes LGTM. I'm taking a look at the static analysis check, which fails to upload the generated SARIF report.

alcaeus (approved):

PR #3205, opened 2024-11-13
PHPORM-28 Add Scout engine to index into MongoDB Search

GromNaN:
Fix PHPORM-28

- And "index" stands for a MongoDB collection with an Atlas Search index named "scout".
- The collection name MUST be prefixed to prevent error of indexing in the same MongoDB Collection as the source documents.

### Checklist

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -65,6 +65,9 @@ jobs:
                     until docker exec --tty mongodb mongosh 127.0.0.1:27017 --eval "db.runCommand({ ping: 1 })"; do
                       sleep 1
                     done
+                    until docker exec --tty mongodb mongosh 127.0.0.1:27017 --eval "db.createCollection('connection_test') && db.getCollection('connection_test').createSearchIndex({mappings:{dynamic: true}})"; do
+                      sleep 1
+                    done
 
             -   name: "Show MongoDB server status"
                 run: |
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,7 @@
     },
     "require-dev": {
         "mongodb/builder": "^0.2",
+        "laravel/scout": "^11",
         "league/flysystem-gridfs": "^3.28",
         "league/flysystem-read-only": "^3.0",
         "phpunit/phpunit": "^10.3",
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,8 @@ services:
         ports:
             - "27017:27017"
         healthcheck:
-            test: mongosh mongodb:27017 --quiet --eval 'db.runCommand("ping").ok'
+            test: mongosh --quiet --eval 'db.runCommand("ping").ok'
             interval: 10s
             timeout: 10s
             retries: 5
+s
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,4 +22,3 @@ services:
             interval: 10s
             timeout: 10s
             retries: 5
-s
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,3 @@
-version: '3.5'
-
 services:
     app:
         tty: true
@@ -16,11 +14,11 @@ services:
 
     mongodb:
         container_name: mongodb
-        image: mongo:latest
+        image: mongodb/mongodb-atlas-local:latest
         ports:
             - "27017:27017"
         healthcheck:
-            test: echo 'db.runCommand("ping").ok' | mongosh mongodb:27017 --quiet
+            test: mongosh mongodb:27017 --quiet --eval 'db.runCommand("ping").ok'
             interval: 10s
             timeout: 10s
             retries: 5
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -17,7 +17,7 @@
         </testsuite>
     </testsuites>
     <php>
-        <env name="MONGODB_URI" value="mongodb://mongodb/"/>
+        <env name="MONGODB_URI" value="mongodb://mongodb/?directConnection=true"/>
         <env name="MONGODB_DATABASE" value="unittest"/>
         <env name="SQLITE_DATABASE" value=":memory:"/>
         <env name="QUEUE_CONNECTION" value="database"/>
--- a/src/Scout/ScoutEngine.php
+++ b/src/Scout/ScoutEngine.php
@@ -27,7 +27,6 @@ use TypeError;
 
 use function array_column;
 use function array_flip;
-use function array_key_exists;
 use function array_map;
 use function array_merge;
 use function assert;
@@ -44,7 +43,6 @@ use function iterator_to_array;
 use function method_exists;
 use function sleep;
 use function sprintf;
-use function substr;
 use function time;
 
 /**
@@ -313,40 +311,67 @@ final class ScoutEngine extends Engine
      *
      * @see Engine::map()
      *
-     * @param array $results
-     * @param Model $model
+     * @param Builder      $builder
+     * @param array|Cursor $results
+     * @param Model        $model
+     *
+     * @return Collection
      */
     #[Override]
     public function map(Builder $builder, $results, $model): Collection
     {
-        assert(is_array($results), new TypeError(sprintf('Argument #2 ($results) must be of type array, %s given', get_debug_type($results))));
-        assert($model instanceof Model, new TypeError(sprintf('Argument #3 ($model) must be of type %s, %s given', Model::class, get_debug_type($model))));
+        return $this->performMap($builder, $results, $model, false);
+    }
 
+    /**
+     * Map the given results to instances of the given model via a lazy collection.
+     *
+     * @see Engine::lazyMap()
+     *
+     * @param Builder      $builder
+     * @param array|Cursor $results
+     * @param Model        $model
+     *
+     * @return LazyCollection
+     */
+    #[Override]
+    public function lazyMap(Builder $builder, $results, $model): LazyCollection
+    {
+        return $this->performMap($builder, $results, $model, true);
+    }
+
+    /** @return ($lazy is true ? LazyCollection : Collection)<mixed> */
+    private function performMap(Builder $builder, array $results, Model $model, bool $lazy): Collection|LazyCollection
+    {
         if (! $results) {
-            return $model->newCollection();
+            $collection = $model->newCollection();
+
+            return $lazy ? LazyCollection::make($collection) : $collection;
         }
 
         $objectIds = array_column($results, '_id');
         $objectIdPositions = array_flip($objectIds);
 
-        return $model->getScoutModelsByIds(
-            $builder,
-            $objectIds,
-        )->filter(function ($model) use ($objectIdPositions) {
-            return array_key_exists($model->getScoutKey(), $objectIdPositions);
-        })->map(function ($model) use ($results, $objectIdPositions) {
-            $result = $results[$objectIdPositions[$model->getScoutKey()]] ?? [];
+        return $model->queryScoutModelsByIds($builder, $objectIds)
+            ->{$lazy ? 'cursor' : 'get'}()
+            ->filter(function ($model) use ($objectIds) {
+                return in_array($model->getScoutKey(), $objectIds);
+            })
+            ->map(function ($model) use ($results, $objectIdPositions) {
+                $result = $results[$objectIdPositions[$model->getScoutKey()]] ?? [];
 
-            foreach ($result as $key => $value) {
-                if ($key[0] === '_') {
-                    $model->withScoutMetadata($key, $value);
+                foreach ($result as $key => $value) {
+                    if ($key[0] === '_' && $key !== '_id') {
+                        $model->withScoutMetadata($key, $value);
+                    }
                 }
-            }
 
-            return $model;
-        })->sortBy(function ($model) use ($objectIdPositions) {
-            return $objectIdPositions[$model->getScoutKey()];
-        })->values();
+                return $model;
+            })
+            ->sortBy(function ($model) use ($objectIdPositions) {
+                return $objectIdPositions[$model->getScoutKey()];
+            })
+            ->values();
     }
 
     /**
@@ -364,7 +389,7 @@ final class ScoutEngine extends Engine
             return 0;
         }
 
-        return $results[0]['__count'];
+        return $results[0]->__count;
     }
 
     /**
@@ -382,50 +407,6 @@ final class ScoutEngine extends Engine
         $collection = $this->getIndexableCollection($model);
 
         $collection->deleteMany([]);
-    }
-
-    /**
-     * Map the given results to instances of the given model via a lazy collection.
-     *
-     * @see Engine::lazyMap()
-     *
-     * @param Builder      $builder
-     * @param array|Cursor $results
-     * @param Model        $model
-     *
-     * @return LazyCollection
-     */
-    #[Override]
-    public function lazyMap(Builder $builder, $results, $model): LazyCollection
-    {
-        assert($results instanceof Cursor || is_array($results), new TypeError(sprintf('Argument #2 ($results) must be of type %s|array, %s given', Cursor::class, get_debug_type($results))));
-        assert($model instanceof Model, new TypeError(sprintf('Argument #3 ($model) must be of type %s, %s given', Model::class, get_debug_type($model))));
-
-        if (! $results) {
-            return LazyCollection::make($model->newCollection());
-        }
-
-        $objectIds = array_column($results, '_id');
-        $objectIdPositions = array_flip($objectIds);
-
-        return $model->queryScoutModelsByIds(
-            $builder,
-            $objectIds,
-        )->cursor()->filter(function ($model) use ($objectIds) {
-            return in_array($model->getScoutKey(), $objectIds);
-        })->map(function ($model) use ($results, $objectIdPositions) {
-            $result = $results[$objectIdPositions[$model->getScoutKey()]] ?? [];
-
-            foreach ($result as $key => $value) {
-                if (substr($key, 0, 1) === '_') {
-                    $model->withScoutMetadata($key, $value);
-                }
-            }
-
-            return $model;
-        })->sortBy(function ($model) use ($objectIdPositions) {
-            return $objectIdPositions[$model->getScoutKey()];
-        })->values();
     }
 
     /**
@@ -475,7 +456,7 @@ final class ScoutEngine extends Engine
     {
         assert(is_string($name), new TypeError(sprintf('Argument #1 ($name) must be of type string, %s given', get_debug_type($name))));
 
-        $this->database->selectCollection($name)->drop();
+        $this->database->dropCollection($name);
     }
 
     /** Get the MongoDB collection used to search for the provided model */
--- a/src/Scout/ScoutEngine.php
+++ b/src/Scout/ScoutEngine.php
@@ -21,6 +21,7 @@ use MongoDB\Driver\CursorInterface;
 use MongoDB\Exception\RuntimeException as MongoDBRuntimeException;
 use MongoDB\Laravel\Connection;
 use Override;
+use stdClass;
 use Traversable;
 use TypeError;
 
@@ -212,6 +213,13 @@ final class ScoutEngine extends Engine
             return $cursor->toArray();
         }
 
+        // Using compound to combine search operators
+        // https://www.mongodb.com/docs/atlas/atlas-search/compound/#options
+        // "should" specifies conditions that contribute to the relevance score
+        // at least one of them must match,
+        // - "text" search for the text including fuzzy matching
+        // - "wildcard" allows special characters like * and ?, similar to LIKE in SQL
+        // These are the only search operators to accept wildcard path.
         $compound = [
             'should' => [
                 [
@@ -236,7 +244,6 @@ final class ScoutEngine extends Engine
         // "filter" specifies conditions on exact values to match
         // "mustNot" specifies conditions on exact values that must not match
         // They don't contribute to the relevance score
-        // https://www.mongodb.com/docs/atlas/atlas-search/compound/#options
         foreach ($builder->wheres as $field => $value) {
             if ($field === '__soft_deleted') {
                 $value = (bool) $value;
@@ -378,8 +385,9 @@ final class ScoutEngine extends Engine
      * This is an estimate if the count is larger than 1000.
      *
      * @see Engine::getTotalCount()
+     * @see https://www.mongodb.com/docs/atlas/atlas-search/counting/
      *
-     * @param mixed $results
+     * @param stdClass[] $results
      */
     #[Override]
     public function getTotalCount($results): int
@@ -388,6 +396,8 @@ final class ScoutEngine extends Engine
             return 0;
         }
 
+        // __count field is added by the aggregation pipeline in performSearch()
+        // using the count.lowerBound in the $search stage
         return $results[0]->__count;
     }
 
--- a/src/Scout/ScoutEngine.php
+++ b/src/Scout/ScoutEngine.php
@@ -66,6 +66,8 @@ final class ScoutEngine extends Engine
         ],
     ];
 
+    private const TYPEMAP = ['root' => 'object', 'document' => 'bson', 'array' => 'bson'];
+
     public function __construct(
         private Database $database,
         private bool $softDelete,
@@ -96,28 +98,42 @@ final class ScoutEngine extends Engine
 
         $bulk = [];
         foreach ($models as $model) {
+            assert($model instanceof Model && method_exists($model, 'toSearchableArray'), new LogicException(sprintf('Model "%s" must use "%s" trait', $model::class, Searchable::class)));
+
             $searchableData = $model->toSearchableArray();
             $searchableData = self::serialize($searchableData);
 
-            if ($searchableData) {
-                unset($searchableData['_id']);
-
-                $searchableData = array_merge($searchableData, $model->scoutMetadata());
-                if (isset($searchableData['__soft_deleted'])) {
-                    $searchableData['__soft_deleted'] = (bool) $searchableData['__soft_deleted'];
-                }
-
+            // Skip/remove the model if it doesn't provide any searchable data
+            if (! $searchableData) {
                 $bulk[] = [
-                    'updateOne' => [
+                    'deleteOne' => [
                         ['_id' => $model->getScoutKey()],
-                        [
-                            '$setOnInsert' => ['_id' => $model->getScoutKey()],
-                            '$set' => $searchableData,
-                        ],
-                        ['upsert' => true],
                     ],
                 ];
+
+                continue;
             }
+
+            unset($searchableData['_id']);
+
+            $searchableData = array_merge($searchableData, $model->scoutMetadata());
+
+            /** Convert the __soft_deleted set by {@see Searchable::pushSoftDeleteMetadata()}
+             * into a boolean for efficient storage and indexing. */
+            if (isset($searchableData['__soft_deleted'])) {
+                $searchableData['__soft_deleted'] = (bool) $searchableData['__soft_deleted'];
+            }
+
+            $bulk[] = [
+                'updateOne' => [
+                    ['_id' => $model->getScoutKey()],
+                    [
+                        '$setOnInsert' => ['_id' => $model->getScoutKey()],
+                        '$set' => $searchableData,
+                    ],
+                    ['upsert' => true],
+                ],
+            ];
         }
 
         $this->getIndexableCollection($models)->bulkWrite($bulk);
@@ -133,7 +149,7 @@ final class ScoutEngine extends Engine
     #[Override]
     public function delete($models): void
     {
-        assert($models instanceof Collection, new TypeError(sprintf('Argument #1 ($models) must be of type %s, %s given', Collection::class, get_debug_type($models))));
+        assert($models instanceof EloquentCollection, new TypeError(sprintf('Argument #1 ($models) must be of type %s, %s given', Collection::class, get_debug_type($models))));
 
         if ($models->isEmpty()) {
             return;
@@ -149,7 +165,7 @@ final class ScoutEngine extends Engine
      *
      * @see Engine::search()
      *
-     * @return mixed
+     * @return array
      */
     #[Override]
     public function search(Builder $builder)
@@ -158,14 +174,14 @@ final class ScoutEngine extends Engine
     }
 
     /**
-     * Perform the given search on the engine.
+     * Perform the given search on the engine with pagination.
      *
      * @see Engine::paginate()
      *
      * @param int $perPage
      * @param int $page
      *
-     * @return mixed
+     * @return array
      */
     #[Override]
     public function paginate(Builder $builder, $perPage, $page)
@@ -182,20 +198,21 @@ final class ScoutEngine extends Engine
     /**
      * Perform the given search on the engine.
      */
-    protected function performSearch(Builder $builder, ?int $offset = null): array
+    private function performSearch(Builder $builder, ?int $offset = null): array
     {
         $collection = $this->getSearchableCollection($builder->model);
 
         if ($builder->callback) {
-            $result = call_user_func(
+            $cursor = call_user_func(
                 $builder->callback,
                 $collection,
                 $builder->query,
                 $offset,
             );
-            assert($result instanceof CursorInterface, new LogicException(sprintf('The search builder closure must return a MongoDB cursor, %s returned', get_debug_type($result))));
+            assert($cursor instanceof CursorInterface, new LogicException(sprintf('The search builder closure must return a MongoDB cursor, %s returned', get_debug_type($cursor))));
+            $cursor->setTypeMap(self::TYPEMAP);
 
-            return $result->toArray();
+            return $cursor->toArray();
         }
 
         $compound = [
@@ -270,11 +287,10 @@ final class ScoutEngine extends Engine
             $pipeline[] = ['$limit' => $builder->limit];
         }
 
-        $options = [
-            'typeMap' => ['root' => 'array', 'document' => 'array', 'array' => 'array'],
-        ];
+        $cursor = $collection->aggregate($pipeline);
+        $cursor->setTypeMap(self::TYPEMAP);
 
-        return $collection->aggregate($pipeline, $options)->toArray();
+        return $cursor->toArray();
     }
 
     /**
@@ -539,6 +555,6 @@ final class ScoutEngine extends Engine
             sleep(1);
         }
 
-        throw new MongoDBRuntimeException('Atlas search index operation time out');
+        throw new MongoDBRuntimeException(sprintf('Atlas search index operation time out after %s seconds', self::WAIT_TIMEOUT_SEC));
     }
 }
--- a/src/Scout/ScoutEngine.php
+++ b/src/Scout/ScoutEngine.php
@@ -126,7 +126,8 @@ final class ScoutEngine extends Engine
                 'updateOne' => [
                     ['_id' => $model->getScoutKey()],
                     [
-                        '$setOnInsert' => ['_id' => $model->getScoutKey()],
+                        // The _id field is added automatically when the document is inserted
+                        // Update all other fields
                         '$set' => $searchableData,
                     ],
                     ['upsert' => true],
--- a/src/Scout/ScoutEngine.php
+++ b/src/Scout/ScoutEngine.php
@@ -84,6 +84,8 @@ final class ScoutEngine extends Engine
     #[Override]
     public function update($models)
     {
+        assert($models instanceof EloquentCollection, new TypeError(sprintf('Argument #1 ($models) must be of type %s, %s given', EloquentCollection::class, get_debug_type($models))));
+
         if ($models->isEmpty()) {
             return;
         }
--- a/src/Scout/ScoutEngine.php
+++ b/src/Scout/ScoutEngine.php
@@ -1,0 +1,542 @@
+<?php
+
+namespace MongoDB\Laravel\Scout;
+
+use Closure;
+use DateTimeInterface;
+use Illuminate\Database\Eloquent\Collection as EloquentCollection;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\SoftDeletes;
+use Illuminate\Support\Collection;
+use Illuminate\Support\LazyCollection;
+use Laravel\Scout\Builder;
+use Laravel\Scout\Engines\Engine;
+use Laravel\Scout\Searchable;
+use LogicException;
+use MongoDB\BSON\Serializable;
+use MongoDB\BSON\UTCDateTime;
+use MongoDB\Collection as MongoDBCollection;
+use MongoDB\Database;
+use MongoDB\Driver\Cursor;
+use MongoDB\Driver\CursorInterface;
+use MongoDB\Exception\RuntimeException as MongoDBRuntimeException;
+use MongoDB\Laravel\Connection;
+use Override;
+use Traversable;
+use TypeError;
+
+use function array_column;
+use function array_flip;
+use function array_key_exists;
+use function array_map;
+use function array_merge;
+use function assert;
+use function call_user_func;
+use function class_uses_recursive;
+use function get_debug_type;
+use function hrtime;
+use function in_array;
+use function is_array;
+use function is_int;
+use function is_iterable;
+use function is_string;
+use function iterator_to_array;
+use function method_exists;
+use function sleep;
+use function sprintf;
+use function substr;
+use function time;
+
+/**
+ * In the context of this Laravel Scout engine, a "search index" refers to
+ * a MongoDB Collection with a Search Index.
+ */
+final class ScoutEngine extends Engine
+{
+    /** Name of the Atlas Search index. */
+    private const INDEX_NAME = 'scout';
+
+    // Atlas Search index management operations are asynchronous.
+    // They usually take less than 5 minutes to complete.
+    private const WAIT_TIMEOUT_SEC = 300;
+
+    private const DEFAULT_DEFINITION = [
+        'mappings' => [
+            'dynamic' => true,
+        ],
+    ];
+
+    public function __construct(
+        private Database $database,
+        private bool $softDelete,
+    ) {
+    }
+
+    /**
+     * Update the given model in the index.
+     *
+     * @see Engine::update()
+     *
+     * @param EloquentCollection $models
+     *
+     * @throws MongoDBRuntimeException
+     */
+    #[Override]
+    public function update($models)
+    {
+        if ($models->isEmpty()) {
+            return;
+        }
+
+        if ($this->softDelete && $this->usesSoftDelete($models)) {
+            $models->each->pushSoftDeleteMetadata();
+        }
+
+        $bulk = [];
+        foreach ($models as $model) {
+            $searchableData = $model->toSearchableArray();
+            $searchableData = self::serialize($searchableData);
+
+            if ($searchableData) {
+                unset($searchableData['_id']);
+
+                $searchableData = array_merge($searchableData, $model->scoutMetadata());
+                if (isset($searchableData['__soft_deleted'])) {
+                    $searchableData['__soft_deleted'] = (bool) $searchableData['__soft_deleted'];
+                }
+
+                $bulk[] = [
+                    'updateOne' => [
+                        ['_id' => $model->getScoutKey()],
+                        [
+                            '$setOnInsert' => ['_id' => $model->getScoutKey()],
+                            '$set' => $searchableData,
+                        ],
+                        ['upsert' => true],
+                    ],
+                ];
+            }
+        }
+
+        $this->getIndexableCollection($models)->bulkWrite($bulk);
+    }
+
+    /**
+     * Remove the given model from the index.
+     *
+     * @see Engine::delete()
+     *
+     * @param EloquentCollection $models
+     */
+    #[Override]
+    public function delete($models): void
+    {
+        assert($models instanceof Collection, new TypeError(sprintf('Argument #1 ($models) must be of type %s, %s given', Collection::class, get_debug_type($models))));
+
+        if ($models->isEmpty()) {
+            return;
+        }
+
+        $collection = $this->getIndexableCollection($models);
+        $ids = $models->map(fn (Model $model) => $model->getScoutKey())->all();
+        $collection->deleteMany(['_id' => ['$in' => $ids]]);
+    }
+
+    /**
+     * Perform the given search on the engine.
+     *
+     * @see Engine::search()
+     *
+     * @return mixed
+     */
+    #[Override]
+    public function search(Builder $builder)
+    {
+        return $this->performSearch($builder);
+    }
+
+    /**
+     * Perform the given search on the engine.
+     *
+     * @see Engine::paginate()
+     *
+     * @param int $perPage
+     * @param int $page
+     *
+     * @return mixed
+     */
+    #[Override]
+    public function paginate(Builder $builder, $perPage, $page)
+    {
+        assert(is_int($perPage), new TypeError(sprintf('Argument #2 ($perPage) must be of type int, %s given', get_debug_type($perPage))));
+        assert(is_int($page), new TypeError(sprintf('Argument #3 ($page) must be of type int, %s given', get_debug_type($page))));
+
+        $builder = clone $builder;
+        $builder->take($perPage);
+
+        return $this->performSearch($builder, $perPage * ($page - 1));
+    }
+
+    /**
+     * Perform the given search on the engine.
+     */
+    protected function performSearch(Builder $builder, ?int $offset = null): array
+    {
+        $collection = $this->getSearchableCollection($builder->model);
+
+        if ($builder->callback) {
+            $result = call_user_func(
+                $builder->callback,
+                $collection,
+                $builder->query,
+                $offset,
+            );
+            assert($result instanceof CursorInterface, new LogicException(sprintf('The search builder closure must return a MongoDB cursor, %s returned', get_debug_type($result))));
+
+            return $result->toArray();
+        }
+
+        $compound = [
+            'should' => [
+                [
+                    'text' => [
+                        'query' => $builder->query,
+                        'path' => ['wildcard' => '*'],
+                        'fuzzy' => ['maxEdits' => 2],
+                        'score' => ['boost' => ['value' => 5]],
+                    ],
+                ],
+                [
+                    'wildcard' => [
+                        'query' => $builder->query . '*',
+                        'path' => ['wildcard' => '*'],
+                        'allowAnalyzedField' => true,
+                    ],
+                ],
+            ],
+            'minimumShouldMatch' => 1,
+        ];
+
+        // "filter" specifies conditions on exact values to match
+        // "mustNot" specifies conditions on exact values that must not match
+        // They don't contribute to the relevance score
+        // https://www.mongodb.com/docs/atlas/atlas-search/compound/#options
+        foreach ($builder->wheres as $field => $value) {
+            if ($field === '__soft_deleted') {
+                $value = (bool) $value;
+            }
+
+            $compound['filter'][] = ['equals' => ['path' => $field, 'value' => $value]];
+        }
+
+        foreach ($builder->whereIns as $field => $value) {
+            $compound['filter'][] = ['in' => ['path' => $field, 'value' => $value]];
+        }
+
+        foreach ($builder->whereNotIns as $field => $value) {
+            $compound['mustNot'][] = ['in' => ['path' => $field, 'value' => $value]];
+        }
+
+        // Sort by field value only if specified
+        $sort = [];
+        foreach ($builder->orders as $order) {
+            $sort[$order['column']] = $order['direction'] === 'asc' ? 1 : -1;
+        }
+
+        $pipeline = [
+            [
+                '$search' => [
+                    'index' => self::INDEX_NAME,
+                    'compound' => $compound,
+                    'count' => ['type' => 'lowerBound'],
+                    ...($sort ? ['sort' => $sort] : []),
+                ],
+            ],
+            [
+                '$addFields' => [
+                    // Metadata field with the total count of documents
+                    '__count' => '$$SEARCH_META.count.lowerBound',
+                ],
+            ],
+        ];
+
+        if ($offset) {
+            $pipeline[] = ['$skip' => $offset];
+        }
+
+        if ($builder->limit) {
+            $pipeline[] = ['$limit' => $builder->limit];
+        }
+
+        $options = [
+            'typeMap' => ['root' => 'array', 'document' => 'array', 'array' => 'array'],
+        ];
+
+        return $collection->aggregate($pipeline, $options)->toArray();
+    }
+
+    /**
+     * Pluck and return the primary keys of the given results.
+     *
+     * @see Engine::mapIds()
+     *
+     * @param list<array|object> $results
+     */
+    #[Override]
+    public function mapIds($results): Collection
+    {
+        assert(is_array($results), new TypeError(sprintf('Argument #1 ($results) must be of type array, %s given', get_debug_type($results))));
+
+        return new Collection(array_column($results, '_id'));
+    }
+
+    /**
+     * Map the given results to instances of the given model.
+     *
+     * @see Engine::map()
+     *
+     * @param array $results
+     * @param Model $model
+     */
+    #[Override]
+    public function map(Builder $builder, $results, $model): Collection
+    {
+        assert(is_array($results), new TypeError(sprintf('Argument #2 ($results) must be of type array, %s given', get_debug_type($results))));
+        assert($model instanceof Model, new TypeError(sprintf('Argument #3 ($model) must be of type %s, %s given', Model::class, get_debug_type($model))));
+
+        if (! $results) {
+            return $model->newCollection();
+        }
+
+        $objectIds = array_column($results, '_id');
+        $objectIdPositions = array_flip($objectIds);
+
+        return $model->getScoutModelsByIds(
+            $builder,
+            $objectIds,
+        )->filter(function ($model) use ($objectIdPositions) {
+            return array_key_exists($model->getScoutKey(), $objectIdPositions);
+        })->map(function ($model) use ($results, $objectIdPositions) {
+            $result = $results[$objectIdPositions[$model->getScoutKey()]] ?? [];
+
+            foreach ($result as $key => $value) {
+                if ($key[0] === '_') {
+                    $model->withScoutMetadata($key, $value);
+                }
+            }
+
+            return $model;
+        })->sortBy(function ($model) use ($objectIdPositions) {
+            return $objectIdPositions[$model->getScoutKey()];
+        })->values();
+    }
+
+    /**
+     * Get the total count from a raw result returned by the engine.
+     * This is an estimate if the count is larger than 1000.
+     *
+     * @see Engine::getTotalCount()
+     *
+     * @param mixed $results
+     */
+    #[Override]
+    public function getTotalCount($results): int
+    {
+        if (! $results) {
+            return 0;
+        }
+
+        return $results[0]['__count'];
+    }
+
+    /**
+     * Flush all records from the engine.
+     *
+     * @see Engine::flush()
+     *
+     * @param Model $model
+     */
+    #[Override]
+    public function flush($model): void
+    {
+        assert($model instanceof Model, new TypeError(sprintf('Argument #1 ($model) must be of type %s, %s given', Model::class, get_debug_type($model))));
+
+        $collection = $this->getIndexableCollection($model);
+
+        $collection->deleteMany([]);
+    }
+
+    /**
+     * Map the given results to instances of the given model via a lazy collection.
+     *
+     * @see Engine::lazyMap()
+     *
+     * @param Builder      $builder
+     * @param array|Cursor $results
+     * @param Model        $model
+     *
+     * @return LazyCollection
+     */
+    #[Override]
+    public function lazyMap(Builder $builder, $results, $model): LazyCollection
+    {
+        assert($results instanceof Cursor || is_array($results), new TypeError(sprintf('Argument #2 ($results) must be of type %s|array, %s given', Cursor::class, get_debug_type($results))));
+        assert($model instanceof Model, new TypeError(sprintf('Argument #3 ($model) must be of type %s, %s given', Model::class, get_debug_type($model))));
+
+        if (! $results) {
+            return LazyCollection::make($model->newCollection());
+        }
+
+        $objectIds = array_column($results, '_id');
+        $objectIdPositions = array_flip($objectIds);
+
+        return $model->queryScoutModelsByIds(
+            $builder,
+            $objectIds,
+        )->cursor()->filter(function ($model) use ($objectIds) {
+            return in_array($model->getScoutKey(), $objectIds);
+        })->map(function ($model) use ($results, $objectIdPositions) {
+            $result = $results[$objectIdPositions[$model->getScoutKey()]] ?? [];
+
+            foreach ($result as $key => $value) {
+                if (substr($key, 0, 1) === '_') {
+                    $model->withScoutMetadata($key, $value);
+                }
+            }
+
+            return $model;
+        })->sortBy(function ($model) use ($objectIdPositions) {
+            return $objectIdPositions[$model->getScoutKey()];
+        })->values();
+    }
+
+    /**
+     * Create the MongoDB Atlas Search index.
+     *
+     * Accepted options:
+     *  - wait: bool, default true. Wait for the index to be created.
+     *
+     * @see Engine::createIndex()
+     *
+     * @param string            $name    Collection name
+     * @param array{wait?:bool} $options
+     */
+    #[Override]
+    public function createIndex($name, array $options = []): void
+    {
+        assert(is_string($name), new TypeError(sprintf('Argument #1 ($name) must be of type string, %s given', get_debug_type($name))));
+
+        // Ensure the collection exists before creating the search index
+        $this->database->createCollection($name);
+
+        $collection = $this->database->selectCollection($name);
+        $collection->createSearchIndex(
+            self::DEFAULT_DEFINITION,
+            ['name' => self::INDEX_NAME],
+        );
+
+        if ($options['wait'] ?? true) {
+            $this->wait(function () use ($collection) {
+                $indexes = $collection->listSearchIndexes([
+                    'name' => self::INDEX_NAME,
+                    'typeMap' => ['root' => 'bson'],
+                ]);
+
+                return $indexes->current() && $indexes->current()->status === 'READY';
+            });
+        }
+    }
+
+    /**
+     * Delete a "search index", i.e. a MongoDB collection.
+     *
+     * @see Engine::deleteIndex()
+     */
+    #[Override]
+    public function deleteIndex($name): void
+    {
+        assert(is_string($name), new TypeError(sprintf('Argument #1 ($name) must be of type string, %s given', get_debug_type($name))));
+
+        $this->database->selectCollection($name)->drop();
+    }
+
+    /** Get the MongoDB collection used to search for the provided model */
+    private function getSearchableCollection(Model|EloquentCollection $model): MongoDBCollection
+    {
+        if ($model instanceof EloquentCollection) {
+            $model = $model->first();
+        }
+
+        assert(method_exists($model, 'searchableAs'), sprintf('Model "%s" must use "%s" trait', $model::class, Searchable::class));
+
+        return $this->database->selectCollection($model->searchableAs());
+    }
+
+    /** Get the MongoDB collection used to index the provided model */
+    private function getIndexableCollection(Model|EloquentCollection $model): MongoDBCollection
+    {
+        if ($model instanceof EloquentCollection) {
+            $model = $model->first();
+        }
+
+        assert($model instanceof Model);
+        assert(method_exists($model, 'indexableAs'), sprintf('Model "%s" must use "%s" trait', $model::class, Searchable::class));
+
+        if (
+            $model->getConnection() instanceof Connection
+            && $model->getConnection()->getDatabaseName() === $this->database->getDatabaseName()
+            && $model->getTable() === $model->indexableAs()
+        ) {
+            throw new LogicException(sprintf('The MongoDB Scout collection "%s.%s" must use a different collection from the collection name of the model "%s". Set the "scout.prefix" configuration or use a distinct MongoDB database', $this->database->getDatabaseName(), $model->indexableAs(), $model::class));
+        }
+
+        return $this->database->selectCollection($model->indexableAs());
+    }
+
+    private static function serialize(mixed $value): mixed
+    {
+        if ($value instanceof DateTimeInterface) {
+            return new UTCDateTime($value);
+        }
+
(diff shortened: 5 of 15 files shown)


composer.json
```diff
@@ -70,7 +71,8 @@
         "laravel": {
             "providers": [
                 "MongoDB\\Laravel\\MongoDBServiceProvider",
-                "MongoDB\\Laravel\\MongoDBBusServiceProvider"
+                "MongoDB\\Laravel\\MongoDBBusServiceProvider",
+                "MongoDB\\Laravel\\Scout\\ScoutServiceProvider"
```
GromNaN: This class could be placed at the root of the package and given a name consistent with the other classes.
```suggestion
                "MongoDB\\Laravel\\MongoDBScoutServiceProvider"
```

tests/Scout/ScoutEngineTest.php
```diff
@@ -0,0 +1,394 @@
...
+                    '$search' => [
+                        'index' => 'scout',
+                        'text' => [
+                            'query' => '',
+                            'path' => [
+                                'wildcard' => '*',
+                            ],
+                            'fuzzy' => [
+                                'maxEdits' => 2,
+                            ],
```
GromNaN: `fuzzy` is required to return results for partial matches (search for "lar" must return "Laravel Framework").

tests/Scout/ScoutEngineTest.php
```diff
@@ -0,0 +1,394 @@
...
+            fn () => new Builder(new SearchableModel(), ''),
+            [
+                [
+                    '$search' => [
+                        'index' => 'scout',
+                        'text' => [
+                            'query' => '',
+                            'path' => [
+                                'wildcard' => '*',
+                            ],
```
GromNaN: By default we don't know on which fields to search.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,507 @@
...
+
+        if (! $results) {
+            return $model->newCollection();
+        }
+
+        $ids = array_column($results, '_id');
+
+        return $model->getScoutModelsByIds($builder, $ids)
+            ->filter(fn (Model $model) => in_array($model->getScoutKey(), $ids))
+            ->sortBy(fn (Model $model) => array_search($model->getScoutKey(), $ids))
```
GromNaN: This is suspicious.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,507 @@
...
+    }
+
+    /**
+     * Remove the given model from the index.
+     *
+     * @param EloquentCollection $models
+     */
+    public function delete($models): void
+    {
+        assert($models instanceof Collection, new TypeError(sprintf('Argument #1 ($models) must be of type %s, %s given', Collection::class, get_debug_type($models))));
```
GromNaN: `assert` calls compensate for the lack of native types in the parent class.

GromNaN:
Integration tests require an atlas or local-atlas cluster to run.
I need to find a way to test the result as the order is not the same for each reindexation.

composer.json
jmikola: Would "ScoutServiceProvider" likely conflict with other short class names? There looks to be redundancy in both suggestions here, but I'll defer to whatever you think is most consistent.

It's been a couple of weeks since I picked up the Telescope PR but IIRC I opted to use a namespace there as there would be multiple classes. Seemed preferable than keeping a service provider in a separate namespace from other classes associated with the feature.

src/Scout/ScoutEngine.php
jmikola: Can you elaborate?

tests/Models/SearchableModel.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Laravel\Tests\Models;
+
+use Laravel\Scout\Searchable;
+use MongoDB\Laravel\Eloquent\Model as Eloquent;
```
jmikola: Is there actually a convention for using "Eloquent" here instead of the short class name without an alias (or something like BaseModel)?

tests/Models/SearchableModel.php
```diff
@@ -0,0 +1,47 @@
...
+     * Overriding the `getScoutKey` method to ensure the custom key is used for indexing
+     * and searching the model.
+     */
+    public function getScoutKey(): string
+    {
+        return $this->getAttribute($this->getScoutKeyName()) ?: 'key_' . $this->getKey();
+    }
+
+    /**
+     * This method must be overridden when the `getScoutKey` method is also overridden,
```
jmikola: If `getScoutKey()` is overridden, isn't it possible that it no longer calls `getScoutKeyName()` at all? It's not clear to me why this was broken out to a separate method if it's only called from `getScoutKey()`, and a user could just as easily override _that_ on its own.

tests/Models/User.php
```diff
@@ -139,4 +141,30 @@ public function prunable(): Builder
...
+
+    public function toSearchableArray(): array
+    {
+        return [
+            'id' => (int) $this->id,
+            'name' => $this->name,
+        ];
+    }
+
+    public function getAtlasSearchDefinition(): array
```
jmikola: Is this method called anywhere?

tests/Scout/ScoutEngineTest.php
```diff
@@ -0,0 +1,552 @@
...
+
+        yield 'where conditions' => [
+            function () {
+                $builder = new Builder(new SearchableModel(), 'lar');
+                $builder->where('foo', 'bar');
+                $builder->where('key', 'value');
+
+                return $builder;
+            },
+            array_replace_recursive($defaultPipeline, [
```
jmikola, line 118: Correct me if I'm wrong, but in this case you're actually just replacing `$defaultPipeline[0]` since there's an equality matching between key `0` in `$defaultPipeline` and the lone key in the replacement array.

Basing this on my read of [`array_replace_recursive`](https://www.php.net/array_replace_recursive).

Does this actually need to use the recursive variant if you're really just replacing an aggregation pipeline stage in its entirety?

tests/Scout/ScoutEngineTest.php
```diff
@@ -0,0 +1,552 @@
...
+    }
+
+    #[DataProvider('provideResultsForMapIds')]
+    public function testMapIds(array $results): void
+    {
+        $engine = new ScoutEngine(m::mock(Database::class), softDelete: false, prefix: '');
+
+        $ids = $engine->mapIds($results);
+
+        $this->assertInstanceOf(\Illuminate\Support\Collection::class, $ids);
```
jmikola: Would this be better off having a `use` statement with its own alias? I'm not sure how often inline references to FQCNs are done in the library. I saw you used "EloquentCollection" elsewhere in the PR. Should this be "IlluminateCollection" to avoid conflict with the PHPLIB class?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,527 @@
...
+/**
+ * In the context of this Laravel Scout engine, a "search index" refers to
+ * a MongoDB Collection with a Search Index.
+ */
+final class ScoutEngine extends Engine
+{
+    /**
+     * Name of the Atlas Search index.
+     *
+     * @todo make this configurable
```
jmikola: Is this something you intend to resolve within the PR? If not, perhaps it warrants a ticket to be referenced here.

tests/TestCase.php
```diff
@@ -22,6 +26,8 @@ protected function getPackageProviders($app): array
         return [
             MongoDBServiceProvider::class,
             ValidationServiceProvider::class,
+            \Laravel\Scout\ScoutServiceProvider::class,
```
jmikola: Should this also have a `use` with alias? I suppose there may be no code standard checks to avoid inline FQCN references in this package. If that's intentional, feel free to disregard.

tests/Scout/ScoutIntegrationTest.php
```diff
@@ -0,0 +1,158 @@
...
+use function sleep;
+
+class ScoutIntegrationTest extends TestCase
+{
+    protected static string $userModel = SqlUser::class;
+
+    protected function defineEnvironment($app)
+    {
+        $app['config']->set('scout.driver', 'mongodb');
+        $app['config']->set('scout.prefix', 'scout_');
```
jmikola: How does this compare to the assignments done in `TestCase:getEnvironmentSetUp()`? Are these code paths entirely unrelated?

tests/Scout/ScoutIntegrationTest.php
```diff
@@ -0,0 +1,158 @@
...
+        });
+
+        UserFactory::new()
+            ->times(44)
+            ->state(new Sequence(...$collect->all()))
+            ->create();
+
+        self::assertSame(0, artisan($this, 'scout:delete-index', ['name' => SqlUser::class]));
+        self::assertSame(0, artisan($this, 'scout:index', ['name' => SqlUser::class]));
+        self::assertSame(0, artisan($this, 'scout:import', ['model' => SqlUser::class]));
```
jmikola: Are these asserting successful exit codes? It wasn't immediately clear.

tests/Scout/ScoutIntegrationTest.php
```diff
@@ -0,0 +1,158 @@
...
+
+        sleep(5);
```
jmikola: I'm not familiar with how `defineDatabaseMigrations()` is called within a test, but what is the significance of sleeping for only five seconds?

tests/Scout/ScoutIntegrationTest.php
```diff
@@ -0,0 +1,158 @@
...
+        })->get();
+
+        $this->assertSame([
+            42 => 'Dax Larkin',
+            43 => 'Dana Larson Sr.',
+            44 => 'Amos Larson Sr.',
+        ], $results->pluck('name', 'id')->all());
+    }
+
+    public function testItCanUseBasicSearchToFetchKeys()
```
jmikola, line 187: Is this test seems redundant in light of `testItCanUseBasicSearch()`?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,527 @@
...
+        return $mapping;
+    }
+
+    /**
+     * Wait for the callback to return true, use it for asynchronous
+     * Atlas Search index management operations.
+     */
+    private function wait(Closure $callback): void
+    {
+        $timeout = hrtime()[0] + self::WAIT_TIMEOUT_SEC;
```
jmikola: This is likely a very minor edge case, but `hrtime()` could be unsupported and return `false`:

- https://github.com/php/php-src/blob/php-8.4.1/ext/standard/basic_functions.stub.php#L2186
- https://github.com/php/php-src/blob/php-8.4.1/ext/standard/hrtime.c#L49
- https://github.com/php/php-src/blob/php-8.4.1/Zend/zend_hrtime.h#L57

But if static analysis doesn't care about this then perhaps you shouldn't either.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,527 @@
...
+        while (hrtime()[0] < $timeout) {
+            if ($callback()) {
+                return;
+            }
+
+            sleep(5);
```
jmikola: Why five seconds? Was this taken from PHPLIB tests?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,527 @@
...
+        if (method_exists($model, 'searchableMapping')) {
+            $mapping = $model->searchableMapping();
+        }
+
+        if ($this->usesSoftDelete($model)) {
+            // This field is a boolean represented with the integers 0 and 1
+            $mapping['fields']['__soft_deleted'] ??= [
+                'type' => 'number',
+                'representation' => 'int64',
+                'indexDoubles' => false,
```
jmikola: Noted this pertains to https://www.mongodb.com/docs/atlas/atlas-search/field-types/number-type/#configure-fts-field-type-field-properties

Consider including a comment that links to some docs on this, as it's not clear what this mapping means otherwise.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,527 @@
...
+    }
+
+    private function performSearchIndexOperation(Closure $closure): void
+    {
+        try {
+            $closure();
+        } catch (ServerException $exception) {
+            // @todo add error codes
+            if (in_array($exception->getCode(), [])) {
+                throw new \RuntimeException('Failed to perform search index operation. MongoDB Atlas cluster M10+ is required', 0, $exception);
```
jmikola: https://www.mongodb.com/docs/atlas/atlas-search/shared-tier-limitations/ suggests that Atlas Search is supported by lower tiers. Didn't this come up recently with regard to PHPLIB?

jmikola (commented):
Leaving an incomplete round of feedback. We'll need to schedule some time to actually go through this PR together if you need a meaningful review, as a lot of it is over my head.

tests/Models/SearchableModel.php
GromNaN: `getScoutKeyName` is necessary for searching original model objects from the results of the search engine.
Example: if in a SQL database, the "Scout Key" is `scout_key`, a search query will return a list of objects with the `scout_key` property and the following query is executed: `SELECT * FROM model WHERE scout_key IN (<ids>)`.

src/Scout/ScoutEngine.php
GromNaN: I think this name is specific enough to not conflict with anything. So I'll keep it non-configurable.

src/Scout/ScoutEngine.php
GromNaN: At first I didn't understood why the index was retrieved from an array, but now I understand that it's intended to sort in the initial order of the results. It's clever. That's copilot that refactored this code copied from MeilisearchEngine: https://github.com/laravel/scout/blob/13cb6460eb9802ff95a28dc11e178b1abebdcc02/src/Engines/MeilisearchEngine.php#L290

src/Scout/ScoutEngine.php
GromNaN: Exact, and we will remove all mentions to M10+ from the code and docs. https://jira.mongodb.org/browse/DRIVERS-3050

src/Scout/ScoutEngine.php
GromNaN: The indexation process is async, I'll try to run a query to detect when all documents are indexed un lucene.

tests/Models/SearchableModel.php
GromNaN: No specific reason, I removed the alias from this class and `HiddenAnimal`

tests/Scout/ScoutEngineTest.php
GromNaN: Yes, I'm using `array_replace_recursive` to replace the numeric key (instead of append in cas of `array_merge`), and this adds the `filter` key-value to  `$defaultPipeline['$search']['compound']`

tests/Scout/ScoutIntegrationTest.php
GromNaN: Indeed, it's redundant.

tests/Scout/ScoutIntegrationTest.php
GromNaN: Yes, 0 is the exit code of the command. Adding a comment.

tests/Scout/ScoutIntegrationTest.php
GromNaN: The indexation of the documents is async. I'll try to run a query in loop to wait for all document to be available.

tests/Models/SearchableModel.php
jmikola: That answer is over my head. My question was more basic, as I was just looking for what might actually cal `getScoutKeyName()`. I think I found my answer in the [Searchable trait](https://github.com/laravel/scout/blob/11.x/src/Searchable.php), though.

It'd have been helpful if any overridden methods mentioned that they were overriding something from the trait (maybe using `@see` with a reference to the base method).

tests/Models/User.php
```diff
@@ -139,4 +141,30 @@ public function prunable(): Builder
...
+
+    public function getScoutKey(): string
+    {
+        return (string) $this->getKey();
+    }
+
+    public function toSearchableArray(): array
+    {
+        return [
+            'id' => (int) $this->id,
```
jmikola: Why is the `$id` property documented as a string in the class docblock, but cast to an integer here? Is it expected to be an ObjectId within the database that's represented as a string in a hydrated model instance?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+        private string $prefix,
+    ) {
+    }
+
+    /**
+     * Update the given model in the index.
+     *
+     * @param EloquentCollection $models
+     *
+     * @throws Exception
```
jmikola: I think this is referring to the `MongoDB\Exception\Exception` interface, but it'd be helpful to alias that for clarity. Why would it be the general interface instead of a RuntimeException, though?

If this is implemented correctly, I expect we wouldn't get any non-runtime exceptions from PHPC.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+    public function update($models)
+    {
+        if ($models->isEmpty()) {
+            return;
+        }
+
+        $collection = $this->getIndexableCollection($models);
+
+        if ($this->softDelete && $this->usesSoftDelete($models)) {
+            $models->each->pushSoftDeleteMetadata();
```
jmikola, line 94: Noted that this calls the method in the [Searchable trait](https://github.com/laravel/scout/blob/11.x/src/Searchable.php#L367C21-L367C43).

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+        }
+
+        // Recursively serialize arrays
+        return array_map(self::serialize(...), $value);
+    }
+
+    private function usesSoftDelete(Model|EloquentCollection $model): bool
+    {
+        if ($model instanceof EloquentCollection) {
+            $model = $model->first();
```
jmikola, line 527: Is it safe to assume that SoftDeletes is used simply because the first model has it? I assume the trait is applied on the model class and an EloquentCollection can be assumed to always contain the same class. I dunno if Eloquent supports anything like Doctrine's multi-collection inheritance, but that might make it possible for there to be various classes here (considering an edge case where only some children utilize the trait).

But even if that were the case, would a single Scout collection cover _all_ of the classes in that hierarchy? Perhaps not.

tests/Models/SchemaVersion.php
```diff
@@ -5,9 +5,9 @@
 namespace MongoDB\Laravel\Tests\Models;
 
 use MongoDB\Laravel\Eloquent\HasSchemaVersion;
-use MongoDB\Laravel\Eloquent\Model as Eloquent;
+use MongoDB\Laravel\Eloquent\Model;
 
-class SchemaVersion extends Eloquent
+class SchemaVersion extends Model
```
jmikola, line 10: Is this class referenced at all?

tests/Scout/ScoutIntegrationTest.php
```diff
@@ -0,0 +1,256 @@
...
+use function env;
+use function in_array;
+use function Orchestra\Testbench\artisan;
+use function range;
+use function sprintf;
+use function usleep;
+
+class ScoutIntegrationTest extends TestCase
+{
+    /** Replace the User model that is used by the UserFactory. */
```
jmikola: I realize that the `Override` attribute isn't usable here, but it'd be helpful to reference the base method when overriding things (maybe with `@see`).

src/Scout/MongoDBScoutServiceProvider.php
```diff
@@ -0,0 +1,37 @@
...
+        $this->app->extend(EngineManager::class, function (EngineManager $engineManager) {
+            $engineManager->extend('mongodb', function (Container $app) {
+                $connectionName = $app->get('config')->get('scout.mongodb.connection', 'mongodb');
+                $connection = $app->get('db')->connection($connectionName);
+                $softDelete = (bool) $app->get('config')->get('scout.soft_delete', false);
+                $prefix = (string) $app->get('config')->get('scout.prefix');
+
+                assert($connection instanceof Connection, new InvalidArgumentException(sprintf('The connection "%s" is not a MongoDB connection.', $connectionName)));
+
+                return new ScoutEngine($connection->getMongoDB(), $softDelete, $prefix);
```
jmikola: Noted that the [Connection class](https://github.com/mongodb/laravel-mongodb/blob/5.x/src/Connection.php) has `getMongoDB()` and `getMongoClient()` methods that return a `MongoDB\Database` and `MongoDB\Client`, respectively. Not a fan of those names, as they smell like the legacy extension API.

I'm not sure why they couldn't just be `getDatabase()` and `getClient()`. That seems beyond the scope of this PR, though.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+    private const WAIT_TIMEOUT_SEC = 300;
+
+    private const DEFAULT_DEFINITION = [
+        'mappings' => [
+            'dynamic' => true,
+        ],
+    ];
+
+    public function __construct(
+        private Database $mongodb,
```
jmikola: Do you think that `$database` or just `$db` would be clearer? The name `$mongodb` sounds even more vague because it doesn't directly correspond to a PHPLIB class. This is similar to my objection to the naming of the `Connection::getMongoDB()` method.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+     *
+     * @throws Exception
+     */
+    public function update($models)
+    {
+        if ($models->isEmpty()) {
+            return;
+        }
+
+        $collection = $this->getIndexableCollection($models);
```
jmikola: This variable doesn't appear to be used until further down in the method, so would it make sense to initialize it there (before executing the bulk write)?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+
+            if ($searchableData) {
+                unset($searchableData['_id']);
+
+                $bulk[] = [
+                    'updateOne' => [
+                        ['_id' => $model->getScoutKey()],
+                        [
+                            '$setOnInsert' => ['_id' => $model->getScoutKey()],
+                            '$set' => array_merge($searchableData, $model->scoutMetadata()),
```
jmikola: If I understand correctly, `scoutMetadata` is just going to include the `__soft_deleted` integer field. Given that you had some later code that had to map this as an integer for indexing, would it make more sense to override `pushSoftDeleteMetadata()` in this package to use a real boolean value? Or is that not possible, since you'd expect applications to only use Scout's original Searchable trait (and it's not extensible)?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+                        ],
+                        ['upsert' => true],
+                    ],
+                ];
+            }
+        }
+
+        if ($bulk) {
```
jmikola: Earlier, you return early if `$models` is empty. If `$bulk` is also empty here, wouldn't that be a logic error of some sort?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+        $collection = $this->getSearchableCollection($builder->model);
+
+        if ($builder->callback) {
+            $result = call_user_func(
+                $builder->callback,
+                $collection,
+                $builder->query,
+                $offset,
+            );
+            assert($result instanceof CursorInterface || is_array($result), new LogicException(sprintf('The search builder closure must return an array or a MongoDB cursor, %s returned', get_debug_type($result))));
```
jmikola: Why would an array conditionally be returned instead of a Cursor?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+    protected function performSearch(Builder $builder, ?int $offset = null): array
+    {
+        $collection = $this->getSearchableCollection($builder->model);
+
+        if ($builder->callback) {
+            $result = call_user_func(
+                $builder->callback,
+                $collection,
+                $builder->query,
+                $offset,
```
jmikola, line 209: I looked at `ScoutEngineTest::testSearch()` and its data provider, but it's still not clear to me what the `$builder->callback` might be.

Am I correct in assume that Scout is also limited to doing offset/limit-based pagination? That sounds like it'd be quite inefficient for deep pagination, and may be something worth mention in the docs. We could mention that the `searchAfter` or `searchBefore` options for [`$search`](https://www.mongodb.com/docs/atlas/atlas-search/aggregation-stages/search/) won't be usable (assuming they're actually more efficient, as index-based pagination in a non-search MongoDB collection might be).

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+                    ],
+                ],
+            ],
+        ];
+
+        // "filter" specifies conditions on exact values to match
+        // "mustNot" specifies conditions on exact values that must not match
+        // They don't contribute to the relevance score
+        // https://www.mongodb.com/docs/atlas/atlas-search/compound/#options
+        foreach ($builder->wheres as $field => $value) {
```
jmikola, line 248: I noted that the [Scout Builder class](https://github.com/laravel/scout/blob/11.x/src/Builder.php) may add `wheres` for the `__soft_deleted` field. Is that going to be indexed for more efficient querying?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+                '$search' => [
+                    'index' => self::INDEX_NAME,
+                    'compound' => $compound,
+                    'count' => ['type' => 'lowerBound'],
+                    ...($sort ? ['sort' => $sort] : []),
+                ],
+            ],
+            [
+                '$addFields' => [
+                    'search_meta' => '$$SEARCH_META',
```
jmikola: Does this exist purely for `getTotalCount()`? If so, that may warrant a comment.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+        if ($offset) {
+            $pipeline[] = ['$skip' => $offset];
+        }
+
+        if ($builder->limit) {
+            $pipeline[] = ['$limit' => $builder->limit];
+        }
+
+        $options = [
+            'allowDiskUse' => true,
```
jmikola: Is this meaningful? The [aggregate](https://www.mongodb.com/docs/manual/reference/command/aggregate/) docs suggest that this would happen by default, unless Atlas overrides the [allowDiskUseByDefault](https://www.mongodb.com/docs/manual/reference/parameters/#mongodb-parameter-param.allowDiskUseByDefault) parameter. And if it does, presumably there's a good reason for that?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+            'typeMap' => ['root' => 'object', 'document' => 'array', 'array' => 'array'],
```
jmikola: Why is only the root-level document an object? Alternatively, why not arrays for everything?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+    /**
+     * Pluck and return the primary keys of the given results.
+     *
+     * @param list<array|object> $results
+     */
+    public function mapIds($results): Collection
+    {
+        assert(is_array($results), new TypeError(sprintf('Argument #1 ($results) must be of type array, %s given', get_debug_type($results))));
+
+        return new Collection(array_column($results, '_id'));
```
jmikola, line 313: Is this related to the typeMap above? I noted that [array_column](https://www.php.net/array_column) can pull public properties of objects (stdClass in that case).

src/Scout/ScoutEngine.php
jmikola: Wouldn't flipping the array (i.e. each `$id` becomes a key) be more efficient to derive a sort order? That would replace repetitive `array_search()` calls with HashTable lookups.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+     *
+     * @param mixed $results
+     */
+    public function getTotalCount($results): int
+    {
+        if (! $results) {
+            return 0;
+        }
+
+        return $results[0]->search_meta['count']['lowerBound'];
```
jmikola: Two questions:

* Should `search_meta` be a defined constant instead of repeated? I'll defer to you.
* Is the name likely to conflict? If so, would a double-underscore (as Laravel uses for `__soft_deleted`) be preferable?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+
+    /** Get the MongoDB collection used to index the provided model */
+    private function getIndexableCollection(Model|EloquentCollection $model): MongoDBCollection
+    {
+        if ($model instanceof EloquentCollection) {
+            $model = $model->first();
+        }
+
+        assert($model instanceof Model);
+        assert(in_array(Searchable::class, class_uses_recursive($model)), sprintf('Model "%s" must use "%s" trait', $model::class, Searchable::class));
```
jmikola: Since `class_uses_recursive()` is inefficient and `getIndexableCollection()` is called from various places, would it make sense to memoize this and cache results for `$model` lookups? This can maybe be a future optimization, but if so it'd be helpful to leave some comments behind around any `class_uses_recursive()` calls.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+        return array_map(self::serialize(...), $value);
+    }
+
+    private function usesSoftDelete(Model|EloquentCollection $model): bool
+    {
+        if ($model instanceof EloquentCollection) {
+            $model = $model->first();
+        }
+
+        return in_array(SoftDeletes::class, class_uses_recursive($model));
```
jmikola, line 530: This is another place where we can consider memorization for `class_uses_recursive()` (if it might be called multiple times per request).

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+    }
+
+    private function getMapping(Model $model): array
+    {
+        $mapping = self::DEFAULT_DEFINITION;
+
+        if (method_exists($model, 'searchableMapping')) {
+            $mapping = $model->searchableMapping();
```
jmikola: Is this a documented extension point for Scout? I didn't find this string in a [code search of the laravel/scout repo](https://github.com/search?q=repo%3Alaravel%2Fscout%20searchableMapping&type=code), and I don't think it was in this PR either.

If it is, do you have a reminder to ensure it gets documented?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+    }
+
+    /**
+     * Delete all "search indexes", i.e. all MongoDB collections.
+     */
+    public function deleteAllIndexes()
+    {
+        $collectionNames = $this->mongodb->listCollectionNames([
+            'filter' => [
+                'name' => new Regex('^' . preg_quote($this->prefix)),
```
jmikola: In ScoutEngineTest, you frequently construct this class with an empty `prefix` option. If `deleteAllIndexes()` were to be called, wouldn't that wipe the entire database?

Is there a safer way that you could query Eloquent for any Model classes, filter those down to anything implementing Searchable, and then derive their MongoDB namespaces from there?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+
+                return false;
+            });
+        }
+    }
+
+    /**
+     * Delete a "search index", i.e. a MongoDB collection.
+     */
+    public function deleteIndex($name): void
```
jmikola, line 465: I'm not sure where this would be called, but should there be some validation that it's actually an index name and not just an arbitrary MongoDB collection?

One sanity check would be to ensure it starts with `$this->prefix`, although that's still a bit risky per my comment below on `deleteAllIndexes()`.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+     */
+    public function createIndex($name, array $options = []): void
+    {
+        assert(is_string($name), new TypeError(sprintf('Argument #1 ($name) must be of type string, %s given', get_debug_type($name))));
+
+        // Ensure the collection exists before creating the search index
+        $this->mongodb->createCollection($name);
+        $collection = $this->mongodb->selectCollection($name);
+
+        $collection = $this->mongodb->selectCollection($name);
```
jmikola:
```suggestion
        $collection = $this->mongodb->selectCollection($name);

```

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+        /** @todo accept configuration for the mapping */
```
jmikola: Is this `@todo` referring to the undocumented `searchableMapping` method I asked about in `getMapping()`?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+        );
+
+        if ($options['wait'] ?? true) {
+            $this->wait(function () use ($collection) {
+                $indexes = $collection->listSearchIndexes([
+                    'name' => self::INDEX_NAME,
+                    'typeMap' => ['root' => 'bson'],
+                ]);
+
+                // Many indexes with the same name may exist on Atlas local
```
jmikola: Why would multiple indexes with the same name exist? [createSearchIndexes](https://www.mongodb.com/docs/manual/reference/command/createSearchIndexes/) says:

> You cannot create multiple indexes with the same name on a single collection.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,535 @@
...
+    {
+        assert($results instanceof Cursor || is_array($results), new TypeError(sprintf('Argument #2 ($results) must be of type %s|array, %s given', Cursor::class, get_debug_type($results))));
+        assert($model instanceof Model, new TypeError(sprintf('Argument #3 ($model) must be of type %s, %s given', Model::class, get_debug_type($model))));
+
+        if (! $results) {
+            return LazyCollection::make($model->newCollection());
+        }
+
+        $objectIds = array_column($results, '_id');
+        $objectIdPositions = array_flip($objectIds);
```
jmikola: I assume this was copied from another engine in the upstream library. Is this the more efficient approach to the `array_search()` algorithm I saw earlier (re: `sortBy()`)?

src/Scout/ScoutEngine.php
GromNaN: I had a bug with atlas local previously, but it seems to be resolved. When dropping a database without dropping the search indexes, the old indexes with were still listed for the new database.

src/Scout/ScoutEngine.php
GromNaN: Yes.

src/Scout/MongoDBScoutServiceProvider.php
GromNaN: I agree, but is it worth taking a BC break just for the sake of naming?
https://jira.mongodb.org/browse/PHPORM-278

src/Scout/ScoutEngine.php
GromNaN: Added an alias `use MongoDB\Exception\RuntimeException as MongoDBRuntimeException`

src/Scout/ScoutEngine.php
GromNaN: Ok, moved to the call to `bulkWrite`. My idea was to throw an exception early if the model was not compatible. But that should never be the case, then there is no optimisation to do.

src/Scout/ScoutEngine.php
GromNaN: Yes, and there is no interface for "simplicity" I guess. I don't even see how I could add an `assert(...)` with this sequence of magic function calls.

src/Scout/ScoutEngine.php
GromNaN: The Model class can use a regular Eloquent SQL connection. Introducing a custom trait would add complexity. I understand that storing an int for a bool is not optimal. All I can do is transform the value here.

src/Scout/ScoutEngine.php
GromNaN: `__soft_deleted` is explicitly added to the search index (otherwise the filter would not work). Converted to `bool` for performance.

src/Scout/ScoutEngine.php
GromNaN: The callback is provided by the developers, maybe they need to manipulate the result before returning. Let's wait if there is a feature request before making code complex. Especially since that is not covered by a test.

src/Scout/ScoutEngine.php
GromNaN: The callback is for "Customizing Engine Searches". https://laravel.com/docs/11.x/scout#customizing-engine-searches

Tracking request for pagination using searchAfter or searchBefore https://jira.mongodb.org/browse/PHPORM-279

src/Scout/ScoutEngine.php
GromNaN: Removing, that should not be necessary. And if necessary, this options should be set in the connection driver options.

src/Scout/ScoutEngine.php
GromNaN: That's right, the `$results` array contains the value returned by `search`  using the `typeMap` and objects are accepted by `array_column`. I used this typeMap for consistency with Eloquent, but I'm not sure this is necessary. `'root' => 'array'` would be fine.

src/Scout/ScoutEngine.php
GromNaN: Getting the list of Model classes is not natively supported by Laravel, there are [some solutions](https://stackoverflow.com/a/60310985) but that's a rabbit hole.
The only solution would be to use the list of indexes in the scout settings, once we setup this feature.

Only Meilisearch supports `deleteAllIndexes` we can remove it while it's not safe.

src/Scout/ScoutEngine.php
GromNaN: I removed this method. Additional feature tracked in PHPORM-268

tests/Models/SchemaVersion.php
GromNaN: Used in `SchemaVersionTest`, I updated for consistency.

composer.json
GromNaN: I choose `MongoDBScoutServiceProvider`. This is consistence with out other ServiceProvider classes.

We may have a single service provider class instead.

tests/Models/SearchableModel.php
GromNaN: I added `#[Override]` and `@see` annotations on all overridden methods.

tests/Scout/ScoutIntegrationTest.php
GromNaN: This is a distinct test. The `->keys()` method is used instead of `->get()`

src/Scout/ScoutEngine.php
GromNaN: Added fallback to `time()` if `hrtime()` is not supported

tests/Models/User.php
GromNaN: Removed.

src/Scout/ScoutEngine.php
GromNaN: Removed, something for a follow-up ticket: [PHPORM-268](https://jira.mongodb.org/browse/PHPORM-268)

src/Scout/ScoutEngine.php
GromNaN: It's called from the [`scout:delete-index` command](https://github.com/laravel/scout/blob/10.x/src/Console/DeleteIndexCommand.php). There is a sanity check for the name prefix.

src/Scout/ScoutEngine.php
GromNaN: I replaced `class_uses_recursive` calls by `method_exists`.

src/Scout/ScoutEngine.php
GromNaN: I added the projected field `__count` to the result.

src/Scout/ScoutEngine.php
GromNaN: Updated to `array`

src/Scout/ScoutEngine.php
GromNaN: Exact, I removed the `if`.

src/Scout/ScoutEngine.php
GromNaN: Renamed to `$database`.

src/Scout/ScoutEngine.php
GromNaN: An Eloquent collection contains only objects of the same class. This is not enforced by the code, but how they are used.

src/Scout/ScoutEngine.php
GromNaN: I removed this mapping, dynamic mapping if fine.

.github/workflows/build-ci.yml
```diff
@@ -11,6 +11,7 @@ jobs:
         name: "PHP ${{ matrix.php }} Laravel ${{ matrix.laravel }} MongoDB ${{ matrix.mongodb }} ${{ matrix.mode }}"
 
         strategy:
+            fail-fast: false
```
GromNaN: Tests with Atlas fail randomly (more than 1/4 of tries fail). fail-fast interrupt other jobs before they are successful, making it impossible to get all the tests successful simultaneously.

.github/workflows/build-ci.yml
```diff
@@ -63,6 +64,9 @@ jobs:
                     until docker exec --tty mongodb mongosh 127.0.0.1:27017 --eval "db.runCommand({ ping: 1 })"; do
                       sleep 1
                     done
+                    until docker exec --tty mongodb mongosh 127.0.0.1:27017 --eval "db.createCollection('connection_test') && db.getCollection('connection_test').createSearchIndex({mappings:{dynamic: true}})"; do
```
GromNaN, line 68: Ensure mongot is ready.

GromNaN (commented):
Update:
- Existing tests are not touched by the Scout tests. The service provider is registered only for the Scout tests and we use dedicated model classes.
- Sorting search results by `id` field as the order is not constant between index creation.
- CI and compose configuration fixed to make atlas search container more resilient (even if it still randomly crashes).
- The mapping is dynamic (soft-delete field not added). Index definition to be made configurable in a follow-up ticket: PHPORM-268

src/Scout/MongoDBScoutServiceProvider.php
jmikola: Resolving this, as [PHPORM-278](https://jira.mongodb.org/browse/PHPORM-278) was discussed during the last triage meeting.

src/Scout/ScoutEngine.php
jmikola: I did not consider that this might be used to index SQL models in MongoDB. Fair point.

composer.json
GromNaN: The additional service provider is not necessary, I merged the service definition into the main `MongoDBBusServiceProvider`.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+        private bool $softDelete,
+    ) {
+    }
+
+    /**
+     * Update the given model in the index.
+     *
+     * @see Engine::update()
+     *
+     * @param EloquentCollection $models
```
jmikola, line 80: Is this worth asserting? I saw you do that in other PRs where you couldn't enforce this via the method signature.

docker-compose.yml
```diff
@@ -16,11 +14,11 @@ services:
 
     mongodb:
         container_name: mongodb
-        image: mongo:latest
+        image: mongodb/mongodb-atlas-local:latest
```
jmikola, line 17: I just noted that https://www.mongodb.com/docs/atlas/cli/current/atlas-cli-deploy-docker/ advises the use of `directConnection=true`. Perhaps you'd like to add that to the `if [ "${{ matrix.mongodb }}" = "Atlas" ]` condition in `build-ci.yml` to be explicit.

Also, is `docker-compose.yml` only present for local testing? I assume this has no direct relation to `build-ci.yml`, since that already includes a "Create MongoDB Atlas Local" step that launches this same container.

docker-compose.yml
```diff
@@ -16,11 +14,11 @@ services:
...
         ports:
             - "27017:27017"
         healthcheck:
-            test: echo 'db.runCommand("ping").ok' | mongosh mongodb:27017 --quiet
+            test: mongosh mongodb:27017 --quiet --eval 'db.runCommand("ping").ok'
```
jmikola: Does the "mongodb" hostname here correspond to `container_name`, or something else?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+        $timeout = (hrtime()[0] ?? time()) + self::WAIT_TIMEOUT_SEC;
+        while ((hrtime()[0] ?? time()) < $timeout) {
+            if ($callback()) {
+                return;
+            }
+
+            sleep(1);
+        }
+
+        throw new MongoDBRuntimeException('Atlas search index operation time out');
```
jmikola: I suppose it's unlikely we'd reach this point, but the message could be improved by saying "timed out after %d seconds".

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+ * a MongoDB Collection with a Search Index.
+ */
+final class ScoutEngine extends Engine
+{
+    /** Name of the Atlas Search index. */
+    private const INDEX_NAME = 'scout';
+
+    // Atlas Search index management operations are asynchronous.
+    // They usually take less than 5 minutes to complete.
+    private const WAIT_TIMEOUT_SEC = 300;
```
jmikola, line 59: For a CLI process with not execution timeout, this is fine.

If this is ever called from a web SAPI, I doubt you'd ever hit this since [`max_execution_time`](https://www.php.net/manual/en/info.configuration.php#ini.max-execution-time) defaults to 30 seconds. I'll note that we have the same issue with the [Manager constructor's](https://www.php.net/manual/en/mongodb-driver-manager.construct.php) default value for `socketTimeoutMS`.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+            return;
+        }
+
+        if ($this->softDelete && $this->usesSoftDelete($models)) {
+            $models->each->pushSoftDeleteMetadata();
+        }
+
+        $bulk = [];
+        foreach ($models as $model) {
+            $searchableData = $model->toSearchableArray();
```
jmikola, line 101: I assume this corresponds to the method on the [Searchable trait](https://github.com/laravel/scout/blob/v10.11.9/src/Searchable.php#L327), but we can't easily assert that as there is no interface to rely on?

Did this require an exclusion in static analysis?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+            $searchableData = self::serialize($searchableData);
+
+            if ($searchableData) {
```
jmikola: Is this actually just testing for a non-empty array?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+                unset($searchableData['_id']);
+
+                $searchableData = array_merge($searchableData, $model->scoutMetadata());
+                if (isset($searchableData['__soft_deleted'])) {
+                    $searchableData['__soft_deleted'] = (bool) $searchableData['__soft_deleted'];
```
jmikola: Is this the hard cast to convert integers to booleans? I think that came up in a previous comment thread that was since resolved, unless I'm mixing this up with another PR.

Either way, a comment may be helpful here.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+                }
+
+                $bulk[] = [
+                    'updateOne' => [
+                        ['_id' => $model->getScoutKey()],
+                        [
+                            '$setOnInsert' => ['_id' => $model->getScoutKey()],
```
jmikola: Is `$setOnInsert` actually required? Quoting the "Set" tab in the [upsert docs](https://www.mongodb.com/docs/manual/reference/method/db.collection.update/#std-label-upsert-behavior):

> If no document matches the query criteria and the `<update>` parameter is a document with update operator expressions, then the operation creates a base document from the equality clauses in the `<query>` parameter and applies the expressions from the `<update>` parameter.

Therefore, `$set` should be applied atop `['_id' => $model->getScoutKey()]`. I don't think there's any reason to use [`$setOnInsert`](https://www.mongodb.com/docs/manual/reference/operator/update/setOnInsert/) and repeat the update filter.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+     * Remove the given model from the index.
+     *
+     * @see Engine::delete()
+     *
+     * @param EloquentCollection $models
+     */
+    #[Override]
+    public function delete($models): void
+    {
+        assert($models instanceof Collection, new TypeError(sprintf('Argument #1 ($models) must be of type %s, %s given', Collection::class, get_debug_type($models))));
```
jmikola: I assume this is checking for `Illuminate\Support\Collection`, but why does the doc block refer to `Illuminate\Database\Eloquent\Collection` (according to the alias)?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+     * @return mixed
+     */
+    #[Override]
+    public function search(Builder $builder)
+    {
+        return $this->performSearch($builder);
+    }
+
+    /**
+     * Perform the given search on the engine.
```
jmikola: This description looks copied from the previous doc block. Was that intentional?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+     */
+    protected function performSearch(Builder $builder, ?int $offset = null): array
+    {
+        $collection = $this->getSearchableCollection($builder->model);
+
+        if ($builder->callback) {
+            $result = call_user_func(
+                $builder->callback,
```
jmikola, line 206: Where is the `$builder->callback` signature enforced?

The only callback tests I see in ScoutIntegrationTest use a callback that takes a single `$query` parameter (with no declared type). But that doesn't seem to agree with the `MongoDB\Collection` returned by `getSearchableCollection()`.

The docs in [Customizing Engine Searches](https://laravel.com/docs/11.x/scout#customizing-engine-searches) didn't clear this up for me.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+                    ],
+                ],
+            ],
+            'minimumShouldMatch' => 1,
+        ];
+
+        // "filter" specifies conditions on exact values to match
+        // "mustNot" specifies conditions on exact values that must not match
+        // They don't contribute to the relevance score
+        // https://www.mongodb.com/docs/atlas/atlas-search/compound/#options
```
jmikola: I did read through this to try and understand how this logic is enforced. My understanding is that `should` is only a preference (mapping to `OR`), and by specifying `minimumShouldMatch: 1` you are requiring either [text](https://www.mongodb.com/docs/atlas/atlas-search/text/) or [wildcard](https://www.mongodb.com/docs/atlas/atlas-search/wildcard/) to match. Is that correct?

It's not clear to my why both need to be used, but I assume it's intended to mimic the functionality you'd get with other search backends. I think it'd be helpful to leave a comment above the `$compound` assignment explaining the thought process behind the `text` and `wildcard` sub-clauses.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+
+    /**
+     * Perform the given search on the engine.
+     *
+     * @see Engine::paginate()
+     *
+     * @param int $perPage
+     * @param int $page
+     *
+     * @return mixed
```
jmikola: Since this returns `$this->performSearch()`, which has an array return type, can we use that here? I understand if you can't use an actual return type due to the parent signature.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+        if ($offset) {
+            $pipeline[] = ['$skip' => $offset];
+        }
+
+        if ($builder->limit) {
+            $pipeline[] = ['$limit' => $builder->limit];
+        }
+
+        $options = [
+            'typeMap' => ['root' => 'array', 'document' => 'array', 'array' => 'array'],
```
jmikola: What's the reason for decoding everything as an array?

Since this is the branch for _not_ using a `$builder->callback`, I'm curious if there's any guidance for a callback doing the same here. Earlier, you assert that a MongoDB Cursor is returned by the callback, but you don't apply the same type map with [`Cursor::setTypeMap()`](https://www.php.net/manual/en/mongodb-driver-cursor.settypemap.php). Why is that?

In the [previously resolved comment](https://github.com/mongodb/laravel-mongodb/pull/3205/files#r1888839886) under `mapIds()`, you said:

> I used this typeMap for consistency with Eloquent, but I'm not sure this is necessary. 'root' => 'array' would be fine.

So I assume it may be required for the `$builder->callback` code path as well. We did previously talk about [`array_column`](https://www.php.net/array_column) also working on objects, but I understand that you'd want to be consistent with Eloquent and return a models as arrays.

If there's no requirement to convert sub-documents to arrays, I think they're better off as objects (to preserve typing); however, I'm not sure how else the return value of a search might get used. If you think arrays are less likely to cause problems, feel free to leave the type map as-is.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+        assert($model instanceof Model, new TypeError(sprintf('Argument #3 ($model) must be of type %s, %s given', Model::class, get_debug_type($model))));
+
+        if (! $results) {
+            return $model->newCollection();
+        }
+
+        $objectIds = array_column($results, '_id');
+        $objectIdPositions = array_flip($objectIds);
+
+        return $model->getScoutModelsByIds(
```
jmikola: Noted that this will translate to a `whereIn` query in [`Searchable::queryScoutModelsByIds()`](https://github.com/laravel/scout/blob/v10.11.9/src/Searchable.php#L256).

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+            $builder,
+            $objectIds,
+        )->filter(function ($model) use ($objectIdPositions) {
+            return array_key_exists($model->getScoutKey(), $objectIdPositions);
```
jmikola: It's not clear to me why this `filter()` step is required, but I see that all of the other engines in laravel/scout do the same.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+        })->map(function ($model) use ($results, $objectIdPositions) {
+            $result = $results[$objectIdPositions[$model->getScoutKey()]] ?? [];
+
+            foreach ($result as $key => $value) {
+                if ($key[0] === '_') {
```
jmikola: Wouldn't the `_id` field trigger this condition? If so, is that something you want set on the [scout metadata](https://github.com/laravel/scout/blob/v10.11.9/src/Searchable.php#L391)?

Noted that this was copied from other engines. AFAIK, the only metadata field used elsewhere is `__soft_deleted`. I'm curious as to why the engines don't check for a `__` prefix instead.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+     * @return LazyCollection
+     */
+    #[Override]
+    public function lazyMap(Builder $builder, $results, $model): LazyCollection
+    {
+        assert($results instanceof Cursor || is_array($results), new TypeError(sprintf('Argument #2 ($results) must be of type %s|array, %s given', Cursor::class, get_debug_type($results))));
+        assert($model instanceof Model, new TypeError(sprintf('Argument #3 ($model) must be of type %s, %s given', Model::class, get_debug_type($model))));
+
+        if (! $results) {
+            return LazyCollection::make($model->newCollection());
```
jmikola: Is this line the only difference between `lazyMap()` and `map()`?

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+     * Delete a "search index", i.e. a MongoDB collection.
+     *
+     * @see Engine::deleteIndex()
+     */
+    #[Override]
+    public function deleteIndex($name): void
+    {
+        assert(is_string($name), new TypeError(sprintf('Argument #1 ($name) must be of type string, %s given', get_debug_type($name))));
+
+        $this->database->selectCollection($name)->drop();
```
jmikola: Alternatively use [`Database::dropCollection()`](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBDatabase-dropCollection/):

```suggestion
        $this->database->dropCollection($name);
```

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,542 @@
...
+    }
+
+    /** Get the MongoDB collection used to search for the provided model */
+    private function getSearchableCollection(Model|EloquentCollection $model): MongoDBCollection
+    {
+        if ($model instanceof EloquentCollection) {
+            $model = $model->first();
+        }
+
+        assert(method_exists($model, 'searchableAs'), sprintf('Model "%s" must use "%s" trait', $model::class, Searchable::class));
```
jmikola, line 479: Is there any reason you didn't do similar assertions elsewhere before assuming a model used the Searchable trait?

tests/AtlasSearchTest.php
```diff
@@ -193,10 +194,13 @@ public function testDatabaseBuilderAutocomplete()
 
         self::assertInstanceOf(LaravelCollection::class, $results);
         self::assertCount(3, $results);
+        // Sort results, because order is not guaranteed
```
jmikola: This change looks unrelated to the Scout work. Should it instead of associated with [PHPORM-276](https://jira.mongodb.org/browse/PHPORM-276)?

tests/Scout/Models/SearchableModel.php
```diff
@@ -0,0 +1,50 @@
...
+     * @see Searchable::getScoutKey()
+     */
+    public function getScoutKey(): string
+    {
+        return $this->getAttribute($this->getScoutKeyName()) ?: 'key_' . $this->getKey();
+    }
+
+    /**
+     * This method must be overridden when the `getScoutKey` method is also overridden,
+     * to support model serialization for async indexation jobs.
```
jmikola:
```suggestion
     * to support model serialization for async indexing jobs.
```

TIL [indexation](https://www.merriam-webster.com/dictionary/indexation) is an economic term.

tests/Scout/ScoutEngineTest.php
```diff
@@ -0,0 +1,548 @@
...
+use PHPUnit\Framework\Attributes\DataProvider;
+
+use function array_replace_recursive;
+use function serialize;
+use function unserialize;
+
+/** Unit tests that do not require an Atlas Search cluster */
+class ScoutEngineTest extends TestCase
+{
+    private const EXPECTED_SEARCH_OPTIONS = ['typeMap' => ['root' => 'array', 'document' => 'array', 'array' => 'array']];
```
jmikola: Would `EXPECTED_AGGREGATE_OPTIONS` be more appropriate?

Alternatively, if this is an annoyance, there's no need to specify the type map when calling `Collection::aggregate()`. You can also do so on the Cursor object it returns _before_ then calling `toArray()`.

If it's not actually necessary to assert this in your mock expectations and only focus on the pipeline parameter, I'd be in favor of that.

tests/Scout/ScoutEngineTest.php
```diff
@@ -0,0 +1,548 @@
...
+        yield 'exclude soft-deleted' => [
+            function () {
+                return new Builder(new SearchableModel(), 'lar', softDelete: true);
+            },
+            array_replace_recursive($defaultPipeline, [
+                [
+                    '$search' => [
+                        'compound' => [
+                            'filter' => [
+                                ['equals' => ['path' => '__soft_deleted', 'value' => 0]],
```
jmikola: Doesn't the value here conflict with the bool cast you do in ScoutEngine? Likewise for the "only trashed" case below.

tests/Scout/ScoutEngineTest.php
```diff
@@ -0,0 +1,548 @@
...
+                        ['$setOnInsert' => ['_id' => 'key_1'], '$set' => ['id' => 1]],
+                        ['upsert' => true],
+                    ],
+                ],
+            ]);
+
+        $model = new SearchableModel(['id' => 1]);
+        $model->delete();
+
+        $engine = new ScoutEngine($database, softDelete: false);
```
jmikola: The test case is "testUpdateWithSoftDelete" but you're specifying `softDelete: false` here. Is that a mistake?

tests/Scout/ScoutIntegrationTest.php
```diff
@@ -0,0 +1,242 @@
...
+            ScoutUser::create($data)->save();
+        }
+
+        self::assertSame(44, ScoutUser::count());
+    }
+
+    /** This test create the search index for tests performing search */
+    public function testItCanCreateTheCollection()
+    {
+        $this->skipIfSearchIndexManagementIsNotSupported();
```
jmikola: Would it make more sense to check for this when the test itself is setup? I assume none of the integration tests can be run without support for search indexes.

tests/Scout/ScoutIntegrationTest.php
```diff
@@ -0,0 +1,242 @@
...
+        self::assertSame(0, artisan($this, 'scout:index', ['name' => ScoutUser::class]));
+        self::assertSame(0, artisan($this, 'scout:import', ['model' => ScoutUser::class]));
+
+        self::assertSame(44, $collection->countDocuments());
+
+        $searchIndexes = $collection->listSearchIndexes(['name' => 'scout']);
+        self::assertCount(1, $searchIndexes);
+
+        // Wait for all documents to be indexed asynchronously
+        $i = 100;
```
jmikola, line 108: Is this effectively waiting for 10 seconds? Since we have a 100,000μs sleep in each iteration?

tests/Scout/ScoutIntegrationTest.php
```diff
@@ -0,0 +1,242 @@
...
+        while (true) {
+            $indexedDocuments = $collection->aggregate([
+                ['$search' => ['index' => 'scout', 'exists' => ['path' => 'name']]],
+            ])->toArray();
+
+            if (count($indexedDocuments) > 0) {
```
jmikola: Is it an "all or nothing" approach, such that if `$search` returns at least one document we can safely assert that all 44 have been indexed?

If not, should this check for `>= 44`?

docker-compose.yml
GromNaN: `docker-compose.yml` is only for local development of the library. I updated the default `phpunit.xml.dist` to use `directConnection=true`. I'll try in GHA as part of [PHPORM-287](https://jira.mongodb.org/browse/PHPORM-287)

docker-compose.yml
GromNaN: The connection uri name is not necessary, I remove it to use the default (`localhost:27017`).

src/Scout/ScoutEngine.php
GromNaN: The indexation is typically done by CLI processes. The timeout is very long, I don't know how long it can take for large indexes to create.

src/Scout/ScoutEngine.php
GromNaN: Yes, all scout engines implement this pattern: skip the model if it results empty searchable data. https://github.com/laravel/scout/blob/2fe33e6e5eb02153fa3de0e1b521111cfc1219e3/src/Engines/MeilisearchEngine.php#L64-L66

src/Scout/ScoutEngine.php
GromNaN: I can add `/** @var Model&Searchable $model */`, this is understood by PHPStorm but phpcs wants to convert it to an invalid `assert($model instanceof Model && $model instanceof Searchable)`.

Adding an other assertion.
```php
assert($model instanceof Model && method_exists($model, 'toSearchableArray'), new LogicException(sprintf('Model "%s" must use "%s" trait', $model::class, Searchable::class)));
```

src/Scout/ScoutEngine.php
GromNaN: Adding comment 
> Convert the __soft_deleted set by `{@see Searchable::pushSoftDeleteMetadata()}` into a boolean for efficient storage and indexing.

src/Scout/ScoutEngine.php
GromNaN: Good catch.

src/Scout/ScoutEngine.php
GromNaN: Duplicated from the [parent class](https://github.com/laravel/scout/blob/f32c6188f2703d910b7c2cca92913f037a5cb244/src/Engines/Engine.php#L25-L41). You are better at review than Laravel team.

Removing this duplicate, I'll rely on parent comments. This classes are not directly.

src/Scout/ScoutEngine.php
GromNaN: I added a test for the collection name. I test the parameters passed to the callback, but I don't see the point of validating the callback signature. Using reflection to do that would be too much.

src/Scout/ScoutEngine.php
GromNaN: Updated to set the typemap on the cursor. So I can set it for the callback.

Also, update the typemap to `['root' => 'object', 'document' => 'bson', 'array' => 'bson']`, I only need the `root._id` extracted by the `map` function.

src/Scout/ScoutEngine.php
GromNaN: It should not be necessary. I think that's a double-check if the query returned results that are not in the `$objectIds` list.

src/Scout/ScoutEngine.php
GromNaN: I excluded `'_id'`, I don't want to make difference with other engine on this prefix.

src/Scout/ScoutEngine.php
GromNaN: I didn't notice! There are 2 difference: the default empty collection and the method `get`/`cursor`. I factorized into a new method `performMap` (to match the other method `performSearch`). The types are now checked by this new private method parameters.

src/Scout/ScoutEngine.php
GromNaN: Every method is covered by this check. I check `searchableAs` in `getSearchableCollection` and `indexableAs` in `getIndexableCollection`. Now I remember this is why I instanciated the `$collection` early in update (see [your comment](https://github.com/mongodb/laravel-mongodb/pull/3205#discussion_r1859820934)). But now I check `toSearchableArray` in update.

tests/AtlasSearchTest.php
GromNaN: This fixes the test. `Collection->sort()` keeps the keys, that fails the purpose or sorting.

tests/Scout/ScoutEngineTest.php
GromNaN: I switched to `EXPECTED_TYPEMAP`, set on the cursor directly.

tests/Scout/ScoutEngineTest.php
GromNaN: Good catch, I updated the test.

tests/Scout/ScoutIntegrationTest.php
GromNaN: It takes about 10 iterations on my machine. Waiting ~1s on my machine.

tests/Scout/ScoutIntegrationTest.php
GromNaN: Updated to `>= 44`, thanks.

src/Scout/ScoutEngine.php
jmikola: @GromNaN: Just want to confirm if this comment was missed or intentionally left as-is.

docker-compose.yml
```diff
@@ -16,11 +14,12 @@ services:
 
     mongodb:
         container_name: mongodb
-        image: mongo:latest
+        image: mongodb/mongodb-atlas-local:latest
         ports:
             - "27017:27017"
         healthcheck:
-            test: echo 'db.runCommand("ping").ok' | mongosh mongodb:27017 --quiet
+            test: mongosh --quiet --eval 'db.runCommand("ping").ok'
```
jmikola, line 21: I assume this relies on `mongosh` defaulting `localhost:27017"

.github/workflows/build-ci.yml
```diff
@@ -65,6 +65,9 @@ jobs:
...
```
jmikola, line 68: In `docker-compose.yml` you removed the host/port from the `mongosh` invocation. Should you do so here as well for consistency? I'm OK with either but wanted to ask.

phpunit.xml.dist
```diff
@@ -17,7 +17,7 @@
         </testsuite>
     </testsuites>
     <php>
-        <env name="MONGODB_URI" value="mongodb://mongodb/"/>
```
jmikola, line 20: Is the `mongodb` hostname a sensible hostname here? In https://github.com/mongodb/laravel-mongodb/pull/3205#discussion_r1914863583 you mentioned removing it in favor of the default of `localhost:27017`. I understand that you'd need to use an explicit value in this context, but I'd expect to see `localhost:27017` instead of `mongodb` in that case.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,540 @@
...
+     * @param mixed $results
+     */
+    #[Override]
+    public function getTotalCount($results): int
+    {
+        if (! $results) {
+            return 0;
+        }
+
+        return $results[0]->__count;
```
jmikola, line 402: This seems tightly coupled to the `$addFields` stage in `performSearch()`. Pointing that out in case you'd like to do something about that.

I assume `$results[0]` can be assumed to be a stdClass instance per the internal type map, but `@param mixed $results` in the doc block doesn't tell us much. You might be able to address that with an internal assertion even though the method signature cannot be changed.

src/Scout/ScoutEngine.php
```diff
@@ -0,0 +1,540 @@
...
+            })
+            ->sortBy(function ($model) use ($objectIdPositions) {
+                return $objectIdPositions[$model->getScoutKey()];
+            })
+            ->values();
+    }
+
+    /**
+     * Get the total count from a raw result returned by the engine.
+     * This is an estimate if the count is larger than 1000.
```
jmikola, line 386: A `@see` reference to https://www.mongodb.com/docs/atlas/atlas-search/counting/ may be helpful here. You can also mention that this pertains to the pipeline created in `performSearch()` since that's what ultimately determines this.

src/Scout/ScoutEngine.php
jmikola: @GromNaN: Pointing this out in case you lost track of this and agreed with leaving a comment above.

jmikola (approved):
Noted that most of my previous points were all addressed. Left a few more small things that you can address or ignore as you wish.

.github/workflows/build-ci.yml
GromNaN: I'll update it as part of [PHPORM-287](https://jira.mongodb.org/browse/PHPORM-287)

phpunit.xml.dist
GromNaN: It's the default localhost from the mongodb container, but from the other container, you need to specify the mongodb container name.

src/Scout/ScoutEngine.php
GromNaN: You are correct, I removed the explicit `$setOnInsert`

PR #3206, opened 2024-11-13
Increase connection timeouts to allow using Atlas shared clusters

GromNaN:
I wasted far too much time trying to run the tests using an Atlas M0 cluster. It turned out that the error was due to a delay too short to allow the connection.

### DIFF
--- a/tests/ConnectionTest.php
+++ b/tests/ConnectionTest.php
@@ -351,7 +351,7 @@ class ConnectionTest extends TestCase
             'database' => 'unittest',
             'options'  => [
                 'connectTimeoutMS'         => 1000,
-                'serverSelectionTimeoutMS' => 1000,
+                'serverSelectionTimeoutMS' => 6000,
             ],
         ];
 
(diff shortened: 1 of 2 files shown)


tests/ConnectionTest.php
```diff
@@ -350,8 +350,8 @@ public function testPingMethod()
             'dsn'      => env('MONGODB_URI', 'mongodb://127.0.0.1/'),
             'database' => 'unittest',
             'options'  => [
-                'connectTimeoutMS'         => 100,
-                'serverSelectionTimeoutMS' => 250,
+                'connectTimeoutMS'         => 1000,
+                'serverSelectionTimeoutMS' => 1000,
```
jmikola: I'd suggest using a slightly higher value here, as it's intended to outlast a replica set election. [Replica Set Elections](https://www.mongodb.com/docs/manual/core/replica-set-elections/) still notes that an election should not exceed 12 seconds, but I think you'd be OK to just throw another zero on this and use 10. If that's still too much, I'd at least consider bumping it to 5 or 6 seconds.

But using the same value as `connectTimeoutMS` may send the wrong message to any users that come across this. The connection timeout should logically be a smaller value.

GromNaN: Ok. I've set serverSelectionTimeoutMS to 6000ms.

jmikola (approved):

PR #3230, opened 2024-12-18
PHPORM-273 Add schema helpers to create search and vector indexes

GromNaN:
Fix PHPORM-273

The new helpers have advanced array-shape types for the index definition. The support for autocompletion of the array-shaped arguments in PHPStorm is not good for use-case 😞 (even if [it was improved in 2022](https://blog.jetbrains.com/phpstorm/2022/02/phpstorm-2022-1-eap-3/))

### Checklist

### DIFF
--- a/src/Schema/Blueprint.php
+++ b/src/Schema/Blueprint.php
@@ -17,12 +17,81 @@ use function is_string;
 use function key;
 
 /**
- * @phpstan-type SearchIndexField array{type: 'boolean'|'date'|'dateFacet'|'objectId'|'stringFacet'|'uuid'} | array{type: 'autocomplete', analyzer?: string, maxGrams?: int, minGrams?: int, tokenization?: 'edgeGram'|'rightEdgeGram'|'nGram', foldDiacritics?: bool} | array{type: 'document'|'embeddedDocuments', dynamic?:bool, fields: array<string, array>} | array{type: 'geo', indexShapes?: bool} | array{type: 'number'|'numberFacet', representation?: 'int64'|'double', indexIntegers?: bool, indexDoubles?: bool} | array{type: 'token', normalizer?: 'lowercase'|'none'} | array{type: 'string', analyzer?: string, searchAnalyzer?: string, indexOptions?: 'docs'|'freqs'|'positions'|'offsets', store?: bool, ignoreAbove?: int, multi?: array<string, array>, norms?: 'include'|'omit'}
- * @phpstan-type SearchIndexAnalyser array{name: string, charFilters?: list<array<string, mixed>>, tokenizer: array{type: string}, tokenFilters?: list<array<string, mixed>>}
- * @phpstan-type SearchIndexStoredSource bool | array{includes: array<string>} | array{excludes: array<string>}
- * @phpstan-type SearchIndexDefinition array{analyser?: string, analyzers?: SearchIndexAnalyser[], searchAnalyzer?: string, mappings: array{dynamic: true} | array{dynamic?: bool, fields: array<string, SearchIndexField>}, storedSource?: SearchIndexStoredSource}
- * @phpstan-type VectorSearchIndexField array{type: 'vector', path: string, numDimensions: int, similarity: 'euclidean'|'cosine'|'dotProduct', quantization?: 'none'|'scalar'|'binary'}
- * @phpstan-type VectorSearchIndexDefinition array{fields: array<string, VectorSearchIndexField>}
+ * @phpstan-type TypeSearchIndexField array{
+ *     type: 'boolean'|'date'|'dateFacet'|'objectId'|'stringFacet'|'uuid',
+ * } | array{
+ *     type: 'autocomplete',
+ *     analyzer?: string,
+ *     maxGrams?: int,
+ *     minGrams?: int,
+ *     tokenization?: 'edgeGram'|'rightEdgeGram'|'nGram',
+ *     foldDiacritics?: bool,
+ * } | array{
+ *     type: 'document'|'embeddedDocuments',
+ *     dynamic?:bool,
+ *     fields: array<string, array<mixed>>,
+ * } | array{
+ *     type: 'geo',
+ *     indexShapes?: bool,
+ * } | array{
+ *     type: 'number'|'numberFacet',
+ *     representation?: 'int64'|'double',
+ *     indexIntegers?: bool,
+ *     indexDoubles?: bool,
+ * } | array{
+ *     type: 'token',
+ *     normalizer?: 'lowercase'|'none',
+ * } | array{
+ *     type: 'string',
+ *     analyzer?: string,
+ *     searchAnalyzer?: string,
+ *     indexOptions?: 'docs'|'freqs'|'positions'|'offsets',
+ *     store?: bool,
+ *     ignoreAbove?: int,
+ *     multi?: array<string, array<mixed>>,
+ *     norms?: 'include'|'omit',
+ * }
+ * @phpstan-type TypeSearchIndexCharFilter array{
+ *      type: 'icuNormalize'|'persian',
+ *  } | array{
+ *     type: 'htmlStrip',
+ *     ignoredTags?: string[],
+ * } | array{
+ *     type: 'mapping',
+ *     mappings?: array<string, string>,
+ * }
+ * @phpstan-type TypeSearchIndexTokenFilter array{type: string, ...}
+ * @phpstan-type TypeSearchIndexAnalyzer array{
+ *     name: string,
+ *     charFilters?: TypeSearchIndexCharFilter,
+ *     tokenizer: array{type: string},
+ *     tokenFilters?: TypeSearchIndexTokenFilter,
+ * }
+ * @phpstan-type TypeSearchIndexStoredSource bool | array{
+ *     includes: array<string>,
+ * } | array{
+ *     excludes: array<string>,
+ * }
+ * @phpstan-type TypeSearchIndexDefinition array{
+ *     analyser?: string,
+ *     analyzers?: TypeSearchIndexAnalyzer[],
+ *     searchAnalyzer?: string,
+ *     mappings: array{dynamic: true} | array{dynamic?: bool, fields: array<string, TypeSearchIndexField>},
+ *     storedSource?: TypeSearchIndexStoredSource,
+ * }
+ * @phpstan-type TypeVectorSearchIndexField array{
+ *     type: 'vector',
+ *     path: string,
+ *     numDimensions: int,
+ *     similarity: 'euclidean'|'cosine'|'dotProduct',
+ *     quantization?: 'none'|'scalar'|'binary',
+ * } | array{
+ *     type: 'filter',
+ *      path: string,
+ * }
+ * @phpstan-type TypeVectorSearchIndexDefinition array{
+ *     fields: array<string, TypeVectorSearchIndexField>,
+ * }
  */
 class Blueprint extends SchemaBlueprint
 {
@@ -314,9 +383,9 @@ class Blueprint extends SchemaBlueprint
     /**
      * Create an Atlas Search Index.
      *
-     * @see https://www.mongodb.com/docs/atlas/atlas-search/
+     * @see https://www.mongodb.com/docs/manual/reference/command/createSearchIndexes/
      *
-     * @phpstan-param SearchIndexDefinition $definition
+     * @phpstan-param TypeSearchIndexDefinition $definition
      */
     public function searchIndex(array $definition, string $name = 'default'): static
     {
@@ -330,7 +399,7 @@ class Blueprint extends SchemaBlueprint
      *
      * @see https://www.mongodb.com/docs/atlas/atlas-vector-search/
      *
-     * @phpstan-param VectorSearchIndexDefinition $definition
+     * @phpstan-param TypeVectorSearchIndexDefinition $definition
      */
     public function vectorSearchIndex(array $definition, string $name = 'default'): static
     {
--- a/src/Schema/Blueprint.php
+++ b/src/Schema/Blueprint.php
@@ -16,6 +16,14 @@ use function is_int;
 use function is_string;
 use function key;
 
+/**
+ * @phpstan-type SearchIndexField array{type: 'boolean'|'date'|'dateFacet'|'objectId'|'stringFacet'|'uuid'} | array{type: 'autocomplete', analyzer?: string, maxGrams?: int, minGrams?: int, tokenization?: 'edgeGram'|'rightEdgeGram'|'nGram', foldDiacritics?: bool} | array{type: 'document'|'embeddedDocuments', dynamic?:bool, fields: array<string, array>} | array{type: 'geo', indexShapes?: bool} | array{type: 'number'|'numberFacet', representation?: 'int64'|'double', indexIntegers?: bool, indexDoubles?: bool} | array{type: 'token', normalizer?: 'lowercase'|'none'} | array{type: 'string', analyzer?: string, searchAnalyzer?: string, indexOptions?: 'docs'|'freqs'|'positions'|'offsets', store?: bool, ignoreAbove?: int, multi?: array<string, array>, norms?: 'include'|'omit'}
+ * @phpstan-type SearchIndexAnalyser array{name: string, charFilters?: list<array<string, mixed>>, tokenizer: array{type: string}, tokenFilters?: list<array<string, mixed>>}
+ * @phpstan-type SearchIndexStoredSource bool | array{includes: array<string>} | array{excludes: array<string>}
+ * @phpstan-type SearchIndexDefinition array{analyser?: string, analyzers?: SearchIndexAnalyser[], searchAnalyzer?: string, mappings: array{dynamic: true} | array{dynamic?: bool, fields: array<string, SearchIndexField>}, storedSource?: SearchIndexStoredSource}
+ * @phpstan-type VectorSearchIndexField array{type: 'vector', path: string, numDimensions: int, similarity: 'euclidean'|'cosine'|'dotProduct', quantization?: 'none'|'scalar'|'binary'}
+ * @phpstan-type VectorSearchIndexDefinition array{fields: array<string, VectorSearchIndexField>}
+ */
 class Blueprint extends SchemaBlueprint
 {
     /**
@@ -299,6 +307,34 @@ class Blueprint extends SchemaBlueprint
         $options['unique'] = true;
 
         $this->index($columns, null, null, $options);
+
+        return $this;
+    }
+
+    /**
+     * Create an Atlas Search Index.
+     *
+     * @see https://www.mongodb.com/docs/atlas/atlas-search/
+     *
+     * @phpstan-param SearchIndexDefinition $definition
+     */
+    public function searchIndex(array $definition, string $name = 'default'): static
+    {
+        $this->collection->createSearchIndex($definition, ['name' => $name, 'type' => 'search']);
+
+        return $this;
+    }
+
+    /**
+     * Create an Atlas Vector Search Index.
+     *
+     * @see https://www.mongodb.com/docs/atlas/atlas-vector-search/
+     *
+     * @phpstan-param VectorSearchIndexDefinition $definition
+     */
+    public function vectorSearchIndex(array $definition, string $name = 'default'): static
+    {
+        $this->collection->createSearchIndex($definition, ['name' => $name, 'type' => 'vectorSearch']);
 
         return $this;
     }
--- a/src/Schema/Blueprint.php
+++ b/src/Schema/Blueprint.php
@@ -16,98 +16,6 @@ use function is_int;
 use function is_string;
 use function key;
 
-/**
- * @link https://www.mongodb.com/docs/atlas/atlas-search/define-field-mappings/#std-label-fts-field-mappings
- * @phpstan-type TypeSearchIndexField array{
- *     type: 'boolean'|'date'|'dateFacet'|'objectId'|'stringFacet'|'uuid',
- * } | array{
- *     type: 'autocomplete',
- *     analyzer?: string,
- *     maxGrams?: int,
- *     minGrams?: int,
- *     tokenization?: 'edgeGram'|'rightEdgeGram'|'nGram',
- *     foldDiacritics?: bool,
- * } | array{
- *     type: 'document'|'embeddedDocuments',
- *     dynamic?:bool,
- *     fields: array<string, array<mixed>>,
- * } | array{
- *     type: 'geo',
- *     indexShapes?: bool,
- * } | array{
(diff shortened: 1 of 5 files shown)


tests/SchemaTest.php
```diff
@@ -514,4 +558,16 @@ protected function getIndex(string $collection, string $name)
 
         return false;
     }
+
+    protected function getSearchIndex(string $collection, string $name)
```
jmikola: Would it make more sense for this to return a nullable BSONDocument instance? IIRC, `Collection::listSearchIndexes()` inherits the default type map.

tests/SchemaTest.php
```diff
@@ -502,9 +504,51 @@ public function testGetIndexes()
...
+        Schema::create('newcollection', function (Blueprint $collection) {
+            $collection->vectorSearchIndex([
+                'fields' => [
+                    ['type' => 'vector', 'path' => 'foo', 'numDimensions' => 128, 'similarity' => 'euclidean', 'quantization' => 'none'],
+                ],
+            ], 'vector');
+        });
+
+        $index = $this->getSearchIndex('newcollection', 'vector');
+        self::assertNotFalse($index);
```
jmikola: Are the indexes immediately returned by `$listSearchIndexes`? I recall that the prose spec tests in PHPLIB use [`waitForIndexes()`](https://github.com/mongodb/mongo-php-library/blob/1.20.0/tests/SpecTests/SearchIndexSpecTest.php#L170).

tests/SchemaTest.php
```diff
@@ -502,9 +504,51 @@ public function testGetIndexes()
...
+    }
+
+    public function testVectorSearchIndex()
+    {
+        Schema::create('newcollection', function (Blueprint $collection) {
+            $collection->vectorSearchIndex([
+                'fields' => [
+                    ['type' => 'vector', 'path' => 'foo', 'numDimensions' => 128, 'similarity' => 'euclidean', 'quantization' => 'none'],
+                ],
+            ], 'vector');
```
jmikola, line 565: Why does only this test use a non-default name? Would it make sense for both to do so?

Also, is there anything in these tests that clean up to created indexes, or are they left around for the remainder of the test suite?

src/Schema/Blueprint.php
```diff
@@ -16,6 +16,14 @@
...
+ * @phpstan-type SearchIndexAnalyser array{name: string, charFilters?: list<array<string, mixed>>, tokenizer: array{type: string}, tokenFilters?: list<array<string, mixed>>}
+ * @phpstan-type SearchIndexStoredSource bool | array{includes: array<string>} | array{excludes: array<string>}
+ * @phpstan-type SearchIndexDefinition array{analyser?: string, analyzers?: SearchIndexAnalyser[], searchAnalyzer?: string, mappings: array{dynamic: true} | array{dynamic?: bool, fields: array<string, SearchIndexField>}, storedSource?: SearchIndexStoredSource}
+ * @phpstan-type VectorSearchIndexField array{type: 'vector', path: string, numDimensions: int, similarity: 'euclidean'|'cosine'|'dotProduct', quantization?: 'none'|'scalar'|'binary'}
```
jmikola: https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-type/ suggests that `type` may be either "vector" or "filter".

See also: https://www.mongodb.com/docs/manual/reference/command/createSearchIndexes/#vector-search-index-definition-syntax

src/Schema/Blueprint.php
```diff
@@ -303,6 +311,34 @@ public function sparse_and_unique($columns = null, $options = [])
         return $this;
     }
 
+    /**
+     * Create an Atlas Search Index.
+     *
+     * @see https://www.mongodb.com/docs/atlas/atlas-search/
```
jmikola: IMO, these gateway pages are an overwhelming starting point for someone looking for reference on how to defining mappings. This is why I referenced https://www.mongodb.com/docs/manual/reference/command/createSearchIndexes/ in the PHPLIB docs for the CreateSearchIndexes operation class.

Perhaps link to both, since the command documentation is much more clear in its presentation of the `$definition` structure.

The gateway pages are certainly something we'd want to references in the prose documentation or a tutorial, but I'm not convinced it's sufficient here. Likewise for the vector search method.

src/Schema/Blueprint.php
```diff
@@ -16,6 +16,14 @@
 use function is_string;
 use function key;
 
+/**
+ * @phpstan-type SearchIndexField array{type: 'boolean'|'date'|'dateFacet'|'objectId'|'stringFacet'|'uuid'} | array{type: 'autocomplete', analyzer?: string, maxGrams?: int, minGrams?: int, tokenization?: 'edgeGram'|'rightEdgeGram'|'nGram', foldDiacritics?: bool} | array{type: 'document'|'embeddedDocuments', dynamic?:bool, fields: array<string, array>} | array{type: 'geo', indexShapes?: bool} | array{type: 'number'|'numberFacet', representation?: 'int64'|'double', indexIntegers?: bool, indexDoubles?: bool} | array{type: 'token', normalizer?: 'lowercase'|'none'} | array{type: 'string', analyzer?: string, searchAnalyzer?: string, indexOptions?: 'docs'|'freqs'|'positions'|'offsets', store?: bool, ignoreAbove?: int, multi?: array<string, array>, norms?: 'include'|'omit'}
+ * @phpstan-type SearchIndexAnalyser array{name: string, charFilters?: list<array<string, mixed>>, tokenizer: array{type: string}, tokenFilters?: list<array<string, mixed>>}
```
jmikola: Suggest using "SearchIndexAnalyzer". American English spelling should be preferred for consistency with the MongoDB documentation.

I'll note you had `analyser?: string, analyzers?: SearchIndexAnalyser[]` below, which is even internally inconsistent.

----

`charFilters` seems correct based on https://www.mongodb.com/docs/atlas/atlas-search/analyzers/character-filters/#std-label-char-filters-ref, although you could break that out into a separate definition to mandate `type` if you wish. Doing so would also give you a space to directly link to that documentation page. But considering how varied the structures are, `mixed` is also entirely reasonable.

Same feedback on `tokenFilters` (https://www.mongodb.com/docs/atlas/atlas-search/analyzers/token-filters/#std-label-token-filters-ref). Actually, it looks like both of those could utilize the same common definition (e.g. "charOrTokenFilter"), although that might send the wrong signal to users. I'm OK to keep this as-is.

src/Schema/Blueprint.php
```diff
@@ -16,6 +16,14 @@
 use function is_string;
 use function key;
 
+/**
```
jmikola: Is it possible to break these definitions out into multiple lines for readability? Doing so would also make this more maintainable if we have to revisit in the future and compare with upstream changes.

src/Schema/Blueprint.php
```diff
@@ -16,6 +16,14 @@
...
+ * @phpstan-type SearchIndexAnalyser array{name: string, charFilters?: list<array<string, mixed>>, tokenizer: array{type: string}, tokenFilters?: list<array<string, mixed>>}
+ * @phpstan-type SearchIndexStoredSource bool | array{includes: array<string>} | array{excludes: array<string>}
+ * @phpstan-type SearchIndexDefinition array{analyser?: string, analyzers?: SearchIndexAnalyser[], searchAnalyzer?: string, mappings: array{dynamic: true} | array{dynamic?: bool, fields: array<string, SearchIndexField>}, storedSource?: SearchIndexStoredSource}
```
jmikola: Quoting the [docs](https://www.mongodb.com/docs/manual/reference/command/createSearchIndexes/#search-index-definition-syntax), `mappings.fields` is "required only if dynamic mapping is disabled." In that case, wouldn't it make sense to use `dynamic?: false` for that branch instead of a generic `bool`? I don't think there's any case where you'd specify `dynamic: true` _and_ define `fields`.

----

It looks like `synonyms` is missing here. See: https://www.mongodb.com/docs/atlas/atlas-search/synonyms/#std-label-synonyms-ref

src/Schema/Blueprint.php
```diff
@@ -16,6 +16,14 @@
 use function is_string;
 use function key;
 
+/**
+ * @phpstan-type SearchIndexField array{type: 'boolean'|'date'|'dateFacet'|'objectId'|'stringFacet'|'uuid'} | array{type: 'autocomplete', analyzer?: string, maxGrams?: int, minGrams?: int, tokenization?: 'edgeGram'|'rightEdgeGram'|'nGram', foldDiacritics?: bool} | array{type: 'document'|'embeddedDocuments', dynamic?:bool, fields: array<string, array>} | array{type: 'geo', indexShapes?: bool} | array{type: 'number'|'numberFacet', representation?: 'int64'|'double', indexIntegers?: bool, indexDoubles?: bool} | array{type: 'token', normalizer?: 'lowercase'|'none'} | array{type: 'string', analyzer?: string, searchAnalyzer?: string, indexOptions?: 'docs'|'freqs'|'positions'|'offsets', store?: bool, ignoreAbove?: int, multi?: array<string, array>, norms?: 'include'|'omit'}
+ * @phpstan-type SearchIndexAnalyser array{name: string, charFilters?: list<array<string, mixed>>, tokenizer: array{type: string}, tokenFilters?: list<array<string, mixed>>}
+ * @phpstan-type SearchIndexStoredSource bool | array{includes: array<string>} | array{excludes: array<string>}
```
jmikola: This seems correct based on https://www.mongodb.com/docs/atlas/atlas-search/stored-source-definition/#std-label-fts-stored-source-definition

If it's possible to break these definitions up and add links to the canonical docs for these shapes, I think that would be a boon for future maintainability (even if the URLs end up changing at some point).

src/Schema/Blueprint.php
```diff
@@ -16,6 +16,14 @@
 use function is_string;
 use function key;
 
+/**
+ * @phpstan-type SearchIndexField array{type: 'boolean'|'date'|'dateFacet'|'objectId'|'stringFacet'|'uuid'} | array{type: 'autocomplete', analyzer?: string, maxGrams?: int, minGrams?: int, tokenization?: 'edgeGram'|'rightEdgeGram'|'nGram', foldDiacritics?: bool} | array{type: 'document'|'embeddedDocuments', dynamic?:bool, fields: array<string, array>} | array{type: 'geo', indexShapes?: bool} | array{type: 'number'|'numberFacet', representation?: 'int64'|'double', indexIntegers?: bool, indexDoubles?: bool} | array{type: 'token', normalizer?: 'lowercase'|'none'} | array{type: 'string', analyzer?: string, searchAnalyzer?: string, indexOptions?: 'docs'|'freqs'|'positions'|'offsets', store?: bool, ignoreAbove?: int, multi?: array<string, array>, norms?: 'include'|'omit'}
```
jmikola: Corresponding docs for this type would be: https://www.mongodb.com/docs/atlas/atlas-search/define-field-mappings/#std-label-fts-field-mappings

I'm holding off reviewing this in case you're able to break it apart into multiple lines for readability.

tests/SchemaTest.php
```diff
@@ -502,9 +504,51 @@ public function testGetIndexes()
         $this->assertSame([], $indexes);
     }
 
+    /** @todo requires SearchIndex support */
```
jmikola: Note this TODO item. I presume both this and `testVectorSearchIndex()` below require some runtime checks to determine if the methods can be executed successfully.

tests/SchemaTest.php
GromNaN: I should have 1 test with default and one with custom name.

The collection `newcollection` is dropped in tearDown. Therefore, the associated search indexes too.

tests/SchemaTest.php
GromNaN: It's true that I simply copied the `getIndex` method. Should be named `hasIndex` and `hasSearchIndex`.

src/Schema/Blueprint.php
GromNaN: I'm not sure about creating so many types. But I can split the array shape in multi-lines. You can also appreciate how Psalm has better validation than PHPStan for this array shapes.
- https://phpstan.org/r/412cb41e-4002-44fa-86fc-386236eab654
- https://psalm.dev/r/058366489f

src/Schema/Blueprint.php
GromNaN: If it's allowed by the server, I think it's a bad idea to be more restrictive. In fact, `dynamic: true` and a list of fields is exactly what I do for [Laravel Scout](https://github.com/mongodb/laravel-mongodb/pull/3205/files#diff-e72108cfb23e6b3da7b9d7c5ce5342499ee7a89de22366aa17dbecf367b95f7cR494).

src/Schema/Blueprint.php
GromNaN: I had this other link: https://www.mongodb.com/docs/atlas/atlas-search/create-index/#create-an-fts-index

But the reference doc may be a more stable link.

src/Schema/Blueprint.php
GromNaN: phpcs don't let me add links between annotations.

src/Schema/Blueprint.php
jmikola: I'm not sure what file/line you were linking to above, as I have a lot of condensed files in that view. Was it the `DEFAULT_DEFINITION` using `dynamic: true`?

Even if you leave this as-is, I would inquire with the Atlas team about this as it's not clear from the documentation what `dynamic: true` with field mappings would actually imply. That it's accepted by the server might have been an oversight, or something they neglected to address in the docs.

src/Schema/Blueprint.php
jmikola: Grouping a bunch of links at the top might still be helpful, but I'll defer to you.

src/Schema/Blueprint.php
```diff
@@ -16,6 +16,83 @@
...
+ * }
+ * @phpstan-type TypeVectorSearchIndexField array{
+ *     type: 'vector',
+ *     path: string,
+ *     numDimensions: int,
+ *     similarity: 'euclidean'|'cosine'|'dotProduct',
+ *     quantization?: 'none'|'scalar'|'binary',
+ * } | array{
+ *     type: 'filter',
+ *      path: string,
```
jmikola:
```suggestion
 *     path: string,
```

Alignment looks off.

src/Schema/Blueprint.php
```diff
@@ -16,6 +16,83 @@
 use function is_string;
 use function key;
 
+/**
+ * @phpstan-type TypeSearchIndexField array{
+ *     type: 'boolean'|'date'|'dateFacet'|'objectId'|'stringFacet'|'uuid',
```
jmikola: Noted that these types have no field properties. For example: https://www.mongodb.com/docs/atlas/atlas-search/field-types/boolean-type/#configure-fts-field-type-field-properties

src/Schema/Blueprint.php
```diff
@@ -16,6 +16,83 @@
...
+ *     type: 'autocomplete',
+ *     analyzer?: string,
+ *     maxGrams?: int,
+ *     minGrams?: int,
+ *     tokenization?: 'edgeGram'|'rightEdgeGram'|'nGram',
+ *     foldDiacritics?: bool,
+ * } | array{
+ *     type: 'document'|'embeddedDocuments',
+ *     dynamic?:bool,
+ *     fields: array<string, array<mixed>>,
```
jmikola: Should this be `array<string, TypeSearchIndexField>`, or is a recursive definition unsupported?

src/Schema/Blueprint.php
```diff
@@ -16,6 +16,83 @@
...
+ * } | array{
+ *     type: 'geo',
+ *     indexShapes?: bool,
+ * } | array{
+ *     type: 'number'|'numberFacet',
+ *     representation?: 'int64'|'double',
```
jmikola: I'm not sure if there's a reason the docs used `string` instead of `enum` for this type: https://www.mongodb.com/docs/atlas/atlas-search/field-types/number-facet-type/#configure-fts-field-type-field-properties

Seems inconsistent with the `autocomplete` `tokenization` definition. Maybe something to alert the Atlas docs teams about. I'll defer to you if you'd rather relax this to `string` or restrict the values here.

src/Schema/Blueprint.php
```diff
@@ -16,6 +16,83 @@
...
+ *     indexIntegers?: bool,
+ *     indexDoubles?: bool,
+ * } | array{
+ *     type: 'token',
+ *     normalizer?: 'lowercase'|'none',
```
jmikola: Another case where the docs specify `string` instead of `enum`: https://www.mongodb.com/docs/atlas/atlas-search/field-types/token-type/#configure-fts-field-type-field-properties

src/Schema/Blueprint.php
GromNaN: I disabled the phpcs rule. It's better to be relevant than to just follow a rule.

src/Schema/Blueprint.php
```diff
@@ -16,6 +16,91 @@
...
+ * } | array{
+ *     type: 'string',
+ *     analyzer?: string,
+ *     searchAnalyzer?: string,
+ *     indexOptions?: 'docs'|'freqs'|'positions'|'offsets',
```
jmikola: Also not an `enum` in https://www.mongodb.com/docs/atlas/atlas-search/field-types/string-type/#configure-fts-field-type-field-properties

src/Schema/Blueprint.php
```diff
@@ -16,6 +16,91 @@
...
+ *     store?: bool,
+ *     ignoreAbove?: int,
+ *     multi?: array<string, array<mixed>>,
```
jmikola: The [multi example](https://www.mongodb.com/docs/atlas/atlas-search/field-types/string-type/#try-an-example-for-the-fts-field-type-type) suggests this would be a map of strings to string field definitions. Since those don't exist, I suppose you could use `TypeSearchIndexField`, but maybe not possible since it's recursive or undesirable as it would suggest that users could use _other_ field definition types.

src/Schema/Blueprint.php
```diff
@@ -16,6 +16,91 @@
...
+ *     norms?: 'include'|'omit',
```
jmikola: Another field that could probably be an `enum` in the docs.

src/Schema/Blueprint.php
```diff
@@ -16,6 +16,91 @@
...
+ *      type: 'icuNormalize'|'persian',
+ *  } | array{
+ *     type: 'htmlStrip',
+ *     ignoredTags?: string[],
+ * } | array{
+ *     type: 'mapping',
+ *     mappings?: array<string, string>,
+ * }
+ * @link https://www.mongodb.com/docs/atlas/atlas-search/analyzers/token-filters/
+ * @phpstan-type TypeSearchIndexTokenFilter array{type: string, ...}
```
jmikola: I assume `...` is the easiest way to avoid mapping out the complicated structure presented in https://www.mongodb.com/docs/atlas/atlas-search/analyzers/token-filters/

src/Schema/Blueprint.php
```diff
@@ -16,6 +16,91 @@
...
+ *     mappings: array{dynamic: true} | array{dynamic?: bool, fields: array<string, TypeSearchIndexField>},
+ *     storedSource?: TypeSearchIndexStoredSource,
+ * }
+ * @link https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-type/#atlas-vector-search-index-fields
+ * @phpstan-type TypeVectorSearchIndexField array{
+ *     type: 'vector',
+ *     path: string,
+ *     numDimensions: int,
+ *     similarity: 'euclidean'|'cosine'|'dotProduct',
+ *     quantization?: 'none'|'scalar'|'binary',
```
jmikola: Noted that these also aren't `enum` in the docs. Given how prevalent this is, I wonder if the use of `enum` in the [autocomplete](https://www.mongodb.com/docs/atlas/atlas-search/field-types/autocomplete-type/) docs are actually what's inconsistent.

tests/SchemaTest.php
jmikola: If this is supposed to be `hasSearchIndex()`, shouldn't it then just return a `bool`?

If there is actually a need to return the index definition, I'd suggest always returning a nullable BSONDocument and working with the return value accordingly in the calling context. But it looks like something should still be changed with this method.

tests/TestCase.php
```diff
@@ -64,4 +65,23 @@ protected function getEnvironmentSetUp($app): void
...
         $app['config']->set('queue.failed.driver', 'mongodb');
     }
+
+    public function skipIfSearchIndexManagementIsNotSupported(): void
+    {
+        try {
+            $this->getConnection('mongodb')->getCollection('test')->listSearchIndexes(['name' => 'just_for_testing']);
+        } catch (ServerException $e) {
+            switch ($e->getCode()) {
+                // MongoDB 6: Unrecognized pipeline stage name: '$listSearchIndexes'
```
jmikola: Does CI only use MongoDB 6.0+? Just curious if this needs to support anything further back.

tests/SchemaTest.php
GromNaN: Indexes are created immediately, but they are "ready" after about a minute.

src/Schema/Blueprint.php
GromNaN: Recursive definition is not supported. That's why I stopped here.

src/Schema/Blueprint.php
GromNaN: The example looks like a hashmap of hashmap. I can't use the type `TypeSearchIndexField` recursively.
```json
        "multi": {
          "english": {
            "type": "string",
            "analyzer": "lucene.english"
          },
```

src/Schema/Blueprint.php
GromNaN: Yes. I was too lazy to map all the types of this field.

src/Schema/Blueprint.php
GromNaN: Maybe the other types are open to new values, not `autocomplete.tokenization`. Just an hypothesis.

tests/TestCase.php
GromNaN: We test from 4.4 to 7.0
https://github.com/mongodb/laravel-mongodb/blob/a257a9fc4aff5138b276268fa8d734186379a4ed/.github/workflows/build-ci.yml#L17-L21

src/Schema/Blueprint.php
jmikola: I'll leave this thread open for the time being but I'm going to close my other comments about `string` and `enum` consistency.

src/Schema/Blueprint.php
jmikola: Closing this per: https://github.com/mongodb/laravel-mongodb/pull/3230#discussion_r1899780846

jmikola (approved):
Looks like all of my previous comments have been addressed.

PR #3232, opened 2024-12-23
PHPORM-275 PHPORM-276 Add `Query\Builder::search()` and `autocomplete()`

GromNaN:
Fix PHPORM-275 and PHPORM-276

### Checklist

- [x] `Model::search()` returns a collection of model instances
- [x] Rebase on https://github.com/mongodb/laravel-mongodb/pull/3233

### DIFF
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -5,7 +5,10 @@ declare(strict_types=1);
 namespace MongoDB\Laravel\Eloquent;
 
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
+use Illuminate\Database\Eloquent\Collection;
+use Illuminate\Database\Eloquent\Model;
 use MongoDB\BSON\Document;
+use MongoDB\Builder\Type\SearchOperatorInterface;
 use MongoDB\Driver\CursorInterface;
 use MongoDB\Driver\Exception\WriteException;
 use MongoDB\Laravel\Connection;
@@ -21,7 +24,10 @@ use function is_object;
 use function iterator_to_array;
 use function property_exists;
 
-/** @method \MongoDB\Laravel\Query\Builder toBase() */
+/**
+ * @method \MongoDB\Laravel\Query\Builder toBase()
+ * @template TModel of Model
+ */
 class Builder extends EloquentBuilder
 {
     private const DUPLICATE_KEY_ERROR = 11000;
@@ -70,9 +76,27 @@ class Builder extends EloquentBuilder
         return $result ?: $this;
     }
 
-    public function search(...$args)
-    {
-        $results = $this->toBase()->search(...$args);
+    /**
+     * Performs a full-text search of the field or fields in an Atlas collection.
+     *
+     * @see https://www.mongodb.com/docs/atlas/atlas-search/aggregation-stages/search/
+     *
+     * @return Collection<int, TModel>
+     */
+    public function search(
+        SearchOperatorInterface|array $operator,
+        ?string $index = null,
+        ?array $highlight = null,
+        ?bool $concurrent = null,
+        ?string $count = null,
+        ?string $searchAfter = null,
+        ?string $searchBefore = null,
+        ?bool $scoreDetails = null,
+        ?array $sort = null,
+        ?bool $returnStoredSource = null,
+        ?array $tracking = null,
+    ): Collection {
+        $results = $this->toBase()->search($operator, $index, $highlight, $concurrent, $count, $searchAfter, $searchBefore, $scoreDetails, $sort, $returnStoredSource, $tracking);
 
         return $this->model->hydrate($results->all());
     }
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace MongoDB\Laravel\Eloquent;
 
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
+use Illuminate\Support\Collection;
 use MongoDB\BSON\Document;
 use MongoDB\Driver\CursorInterface;
 use MongoDB\Driver\Exception\WriteException;
@@ -16,6 +17,7 @@ use MongoDB\Model\BSONDocument;
 use function array_key_exists;
 use function array_merge;
 use function collect;
+use function compact;
 use function is_array;
 use function is_object;
 use function iterator_to_array;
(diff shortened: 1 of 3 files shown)


src/Eloquent/Builder.php
```diff
@@ -49,6 +50,7 @@ class Builder extends EloquentBuilder
         'insertusing',
         'max',
         'min',
+        'autocomplete',
```
jmikola, line 58: I noted how [Builder::__call()](https://github.com/illuminate/database/blob/482726b5a5f444321341bae5a1be01212223c1cc/Eloquent/Builder.php#L2042) handles this.

Why does only `autocomplete` get `$passthru`? Is that because it doesn't perform any hydration, so the implementation is identical? That would be in contrast to `search()`, since the method in this class needs to hydrate.

src/Eloquent/Builder.php
```diff
@@ -69,6 +71,42 @@ public function aggregate($function = null, $columns = ['*'])
...
+        ?bool $concurrent = null,
+        ?string $count = null,
+        ?string $searchAfter = null,
+        ?string $searchBefore = null,
+        ?bool $scoreDetails = null,
+        ?array $sort = null,
+        ?bool $returnStoredSource = null,
+        ?array $tracking = null,
+    ) {
+        $results = $this->toBase()->search($operator, $index, $highlight, $concurrent, $count, $searchAfter, $searchBefore, $scoreDetails, $sort, $returnStoredSource, $tracking);
```
jmikola, line 99: I noted the `@method \MongoDB\Laravel\Query\Builder toBase()` annotation on this class but I'm failing to see how `toBase()` actually returns this library's class instead of `Illuminate\Database\Query\Builder`. I imagine there's some magic or trait involved that isn't clear.

Noted this probably has nothing to do with the PR, so feel free to ignore this complaint.

src/Eloquent/Builder.php
```diff
@@ -69,6 +71,42 @@ public function aggregate($function = null, $columns = ['*'])
...
+     * @param ?string                       $searchAfter        Reference point for retrieving results. searchAfter returns documents starting immediately following the specified reference point.
+     * @param ?string                       $searchBefore       Reference point for retrieving results. searchBefore returns documents starting immediately before the specified reference point.
+     * @param ?bool                         $scoreDetails       Flag that specifies whether to retrieve a detailed breakdown of the score for the documents in the results. If omitted, defaults to false.
+     * @param ?array                        $sort               Document that specifies the fields to sort the Atlas Search results by in ascending or descending order.
+     * @param ?bool                         $returnStoredSource Flag that specifies whether to perform a full document lookup on the backend database or return only stored source fields directly from Atlas Search.
+     * @param ?array                        $tracking           Document that specifies the tracking option to retrieve analytics information on the search terms.
```
jmikola: The `@param` lines here seem like they're only useful for the one-line descriptions. Is this ultimately something that belongs in documentation? I bring this up because most methods have very little in the way of doc blocks.

Alternatively, you could link to https://www.mongodb.com/docs/atlas/atlas-search/aggregation-stages/search/ or some other docs for reference.

src/Eloquent/Builder.php
```diff
@@ -69,6 +71,42 @@ public function aggregate($function = null, $columns = ['*'])
...
+        ?array $highlight = null,
+        ?bool $concurrent = null,
+        ?string $count = null,
+        ?string $searchAfter = null,
+        ?string $searchBefore = null,
+        ?bool $scoreDetails = null,
+        ?array $sort = null,
+        ?bool $returnStoredSource = null,
+        ?array $tracking = null,
+    ) {
```
jmikola: Does Laravel not allow you to define a return type hint here?

src/Query/Builder.php
```diff
@@ -1490,6 +1493,55 @@ public function options(array $options)
...
+        ?bool $concurrent = null,
+        ?string $count = null,
+        ?string $searchAfter = null,
+        ?string $searchBefore = null,
+        ?bool $scoreDetails = null,
+        ?array $sort = null,
+        ?bool $returnStoredSource = null,
+        ?array $tracking = null,
+    ): Collection {
+        return $this->aggregate()->search(...array_filter(func_get_args(), fn ($arg) => $arg !== null))->get();
```
jmikola: Is `array_filter` only necessary because [FluentFactoryTrait::aggregate()](https://github.com/mongodb/mongo-php-library/blob/8b2e9fae3c9583eaeaa4b72c6a2724f67955deb2/src/Builder/Stage/FluentFactoryTrait.php#L624) would otherwise require `Optional::Undefined` to differentiate unset parameters?

src/Query/Builder.php
```diff
@@ -1490,6 +1493,55 @@ public function options(array $options)
...
+    }
+
+    /** @return Collection<string> */
+    public function autocomplete(string $path, string $query, bool|array $fuzzy = false, string $tokenOrder = 'any'): Collection
```
jmikola, line 1544: If you're going to document the `search()` params, it'd probably make sense to do so for these as well. Alternatively, link to https://www.mongodb.com/docs/atlas/atlas-search/autocomplete/ or some other relevant documentation.

src/Eloquent/Builder.php
GromNaN: - by default, the `Eloquent\Builder` ignores the result of the call to the `Query\Builder` method, and return the `Eloquent\Builder` instance for "fluent API".
- `passthru` is to get the unmodified result of the `Query\Builder` method call.
- by overloading `Query\Builder::search` as `Eloquent\Builder::search`, I can get and hydrate the results.

src/Eloquent/Builder.php
GromNaN: I refrain from adding return types when overriding a Laravel method because I don't know what will be used later. But in this case it's an oversight. I've added it.

src/Eloquent/Builder.php
GromNaN: Yes, it's easier to maintain. Fixed.

src/Query/Builder.php
GromNaN: You got it. [`Optional::Undefined` is internal](https://github.com/mongodb/mongo-php-library/blob/8b2e9fae3c9583eaeaa4b72c6a2724f67955deb2/src/Builder/Type/Optional.php#L11)  and I don't need it here as no parameter accepts `null`.

src/Query/Builder.php
jmikola: I think this would benefit from a docs link, like you did for `search()` in Eloquent\Builder.

src/Query/Builder.php
```diff
@@ -1490,6 +1493,55 @@ public function options(array $options)
...
+     * @param ?string                       $searchAfter        Reference point for retrieving results. searchAfter returns documents starting immediately following the specified reference point.
+     * @param ?string                       $searchBefore       Reference point for retrieving results. searchBefore returns documents starting immediately before the specified reference point.
+     * @param ?bool                         $scoreDetails       Flag that specifies whether to retrieve a detailed breakdown of the score for the documents in the results. If omitted, defaults to false.
+     * @param ?array                        $sort               Document that specifies the fields to sort the Atlas Search results by in ascending or descending order.
+     * @param ?bool                         $returnStoredSource Flag that specifies whether to perform a full document lookup on the backend database or return only stored source fields directly from Atlas Search.
+     * @param ?array                        $tracking           Document that specifies the tracking option to retrieve analytics information on the search terms.
```
jmikola: Did you intend to leave these `@param` lines in place or replace them with the docs link you had in Eloquent\Builder?

GromNaN: Updated, thanks for the reminder.

jmikola (approved):

PR #3233, opened 2024-12-31
PHPORM-274 List search indexes in `Schema::getIndexes()` introspection method

GromNaN:
Fix PHPORM-274

Search indexes are a distinct type of index that are exposed using a specific method. Listing the search indexes in `Schema::getIndexes()` so they are listed by the command `php artisan db:table`.

### DIFF
--- a/src/Schema/Builder.php
+++ b/src/Schema/Builder.php
@@ -5,9 +5,12 @@ declare(strict_types=1);
 namespace MongoDB\Laravel\Schema;
 
 use Closure;
+use MongoDB\Collection;
+use MongoDB\Driver\Exception\ServerException;
 use MongoDB\Model\CollectionInfo;
 use MongoDB\Model\IndexInfo;
 
+use function array_column;
 use function array_fill_keys;
 use function array_filter;
 use function array_keys;
@@ -225,9 +228,11 @@ class Builder extends \Illuminate\Database\Schema\Builder
 
     public function getIndexes($table)
     {
-        $indexes = $this->connection->getMongoDB()->selectCollection($table)->listIndexes();
-
+        $collection = $this->connection->getMongoDB()->selectCollection($table);
+        assert($collection instanceof Collection);
         $indexList = [];
+
+        $indexes = $collection->listIndexes();
         foreach ($indexes as $index) {
             assert($index instanceof IndexInfo);
             $indexList[] = [
@@ -242,6 +247,26 @@ class Builder extends \Illuminate\Database\Schema\Builder
                 },
                 'unique' => $index->isUnique(),
             ];
+        }
+
+        try {
+            $indexes = $collection->listSearchIndexes(['typeMap' => ['root' => 'array', 'array' => 'array', 'document' => 'array']]);
+            foreach ($indexes as $index) {
+                $indexList[] = [
+                    'name' => $index['name'],
+                    'columns' => match ($index['type']) {
+                        'search' => array_keys($index['latestDefinition']['mappings']['fields'] ?? []),
+                        'vectorSearch' => array_column($index['latestDefinition']['fields'], 'path'),
+                    },
+                    'type' => $index['type'],
+                    'primary' => false,
+                    'unique' => false,
+                ];
+            }
+        } catch (ServerException $exception) {
+            if (! self::isAtlasSearchNotSupportedException($exception)) {
+                throw $exception;
+            }
         }
 
         return $indexList;
@@ -289,5 +314,11 @@ class Builder extends \Illuminate\Database\Schema\Builder
         }
 
         return $collections;
+    }
+
+    /** @internal */
+    public static function isAtlasSearchNotSupportedException(ServerException $e): bool
+    {
+        return in_array($e->getCode(), [59, 40324, 115, 6047401, 31082], true);
     }
 }
--- a/src/Schema/Builder.php
+++ b/src/Schema/Builder.php
@@ -15,6 +15,7 @@ use function array_fill_keys;
 use function array_filter;
 use function array_keys;
 use function array_map;
+use function array_merge;
 use function assert;
 use function count;
 use function current;
@@ -243,7 +244,7 @@ class Builder extends \Illuminate\Database\Schema\Builder
                     $index->isText() => 'text',
                     $index->is2dSphere() => '2dsphere',
                     $index->isTtl() => 'ttl',
-                    default => 'default',
+                    default => null,
                 },
                 'unique' => $index->isUnique(),
             ];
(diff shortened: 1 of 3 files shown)


tests/AtlasSearchTest.php
```diff
@@ -0,0 +1,122 @@
...
+use Illuminate\Support\Facades\Schema;
+use MongoDB\Collection as MongoDBCollection;
+use MongoDB\Driver\Exception\ServerException;
+use MongoDB\Laravel\Schema\Builder;
+use MongoDB\Laravel\Tests\Models\Book;
+
+use function assert;
+use function usleep;
+
+class AtlasSearchTest extends TestCase
```
GromNaN, line 15: I plan to test all the search index features in this class. It is also in https://github.com/mongodb/laravel-mongodb/pull/3232

src/Schema/Builder.php
```diff
@@ -244,6 +249,26 @@ public function getIndexes($table)
...
         }
 
+        try {
+            $indexes = $collection->listSearchIndexes(['typeMap' => ['root' => 'array', 'array' => 'array', 'document' => 'array']]);
+            foreach ($indexes as $index) {
+                $indexList[] = [
+                    'name' => $index['name'],
+                    'columns' => match ($index['type']) {
+                        'search' => array_keys($index['latestDefinition']['mappings']['fields'] ?? []),
+                        'vectorSearch' => array_column($index['latestDefinition']['fields'], 'path'),
```
jmikola, line 263: Per [Atlas Vector Search Index Fields](https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-type/#atlas-vector-search-index-fields), I expect this might return field paths instead of field names. Is that going to cause problems with whatever consumes this?

I assume it may be consistent with how the basic `listIndexes()` query above runs (using `array_keys($index->getKey())` but I'm not sure.

src/Schema/Builder.php
```diff
@@ -244,6 +249,26 @@ public function getIndexes($table)
             ];
         }
 
+        try {
+            $indexes = $collection->listSearchIndexes(['typeMap' => ['root' => 'array', 'array' => 'array', 'document' => 'array']]);
+            foreach ($indexes as $index) {
+                $indexList[] = [
+                    'name' => $index['name'],
+                    'columns' => match ($index['type']) {
+                        'search' => array_keys($index['latestDefinition']['mappings']['fields'] ?? []),
```
jmikola: Does this need better reporting for dynamic indexes, or is there no good way to express that given the expected format of `$indexList` elements?

tests/AtlasSearchTest.php
```diff
@@ -0,0 +1,122 @@
...
+            ['title' => 'Algorithms'],
+            ['title' => 'Understanding Machine Learning: From Theory to Algorithms'],
+            ['title' => 'Deep Learning'],
+            ['title' => 'Pattern Recognition and Machine Learning'],
+        ]);
+
+        $collection = $this->getConnection('mongodb')->getCollection('books');
+        assert($collection instanceof MongoDBCollection);
+        try {
+            $collection->createSearchIndex([
```
jmikola, line 47: An index with dynamic mappings should probably be included so you can assert its output.

src/Schema/Builder.php
```diff
@@ -225,9 +228,11 @@ public function getColumns($table)
 
     public function getIndexes($table)
     {
-        $indexes = $this->connection->getMongoDB()->selectCollection($table)->listIndexes();
-
+        $collection = $this->connection->getMongoDB()->selectCollection($table);
+        assert($collection instanceof Collection);
         $indexList = [];
+
+        $indexes = $collection->listIndexes();
```
jmikola, line 236: This is outside the diff but I think it's worth mentioning since I see you use an `===` to check for `['_id' => 1]`. If there is ever an issue with that, you may need to relax the comparison on the numeric value. IIRC, mongos used to sometimes return command results as `1.0` instead of `1` and I _think_ there may be similar inconsistencies for index definitions.

That said, it's probably fine to just wait and see if this is ever reported before going out of your way to address it now. It's quite possible that it's a non-issue on recent server versions.

jmikola (approved):
Some questions but LGTM. Feel free to add a test for dynamic field mappings if you wish.

src/Schema/Builder.php
GromNaN: The triple `===` is enforced by phpcs. We will fix it if a new server version produces an incompatible value, but I doubt it.

tests/AtlasSearchTest.php
GromNaN: Dynamic index added.

src/Schema/Builder.php
GromNaN: All I can do is returning "dynamic" as a column name. This is purely informative.

src/Schema/Builder.php
GromNaN: As soon as it is a string, it can be displayed.

GromNaN:
Example of output:

```
❯ php artisan db:table comments

  comments ................................................................................  
  Columns ............................................................................... 7  
  Size ........................................................................... 52.00 KB  

  Column ............................................................................. Type  
  id objectId, objectId .......................................................... objectId  
  author_id string, nullable ....................................................... string  
  content string, nullable ......................................................... string  
  created_at date, nullable .......................................................... date  
  post_id string, nullable ......................................................... string  
  published_at date, nullable ........................................................ date  
  updated_at date, nullable .......................................................... date  

  Index ...................................................................................  
  _id_ _id ........................................................................ primary  
  vector_index my_vector ..................................................... vectorSearch  
  default dynamic .................................................................. search  
  search dynamic, body ............................................................. search  


```

src/Schema/Builder.php
```diff
@@ -238,12 +244,35 @@ public function getIndexes($table)
...
         }
 
+        try {
+            $indexes = $collection->listSearchIndexes(['typeMap' => ['root' => 'array', 'array' => 'array', 'document' => 'array']]);
+            foreach ($indexes as $index) {
+                $indexList[] = [
+                    'name' => $index['name'],
+                    'columns' => match ($index['type']) {
+                        'search' => array_merge(
+                            $index['latestDefinition']['mappings']['dynamic'] ? ['dynamic'] : [],
```
GromNaN, line 260: - `latestDefinition.mappings.dynamic` is always defined.
- `dynamic` word is added to the list for columns names it there is also explicit mapping.

src/Schema/Builder.php
jmikola: Returning `dynamic` as a column name seems undesirable, as then it might be indistinguishable from a real index with a `dynamic` field mapped.

GromNaN: The alternative is replace the type with `search, dynamic`

PR #3235, opened 2025-01-02
PHPORM-283 Add `Schema::dropSearchIndex()`

GromNaN:
Fix PHPORM-283

Same helper for both "search" and "vector search" indexes.

### Checklist

### DIFF
--- a/src/Schema/Blueprint.php
+++ b/src/Schema/Blueprint.php
@@ -342,7 +342,7 @@ class Blueprint extends SchemaBlueprint
     /**
      * Drop an Atlas Search or Vector Search index
      */
-    public function dropSearchIndex(string $name = 'default'): static
+    public function dropSearchIndex(string $name): static
     {
         $this->collection->dropSearchIndex($name);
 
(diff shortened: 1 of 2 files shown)


src/Schema/Blueprint.php
```diff
@@ -339,6 +339,16 @@ public function vectorSearchIndex(array $definition, string $name = 'default'):
         return $this;
     }
 
+    /**
+     * Drop an Atlas Search or Vector Search index
+     */
+    public function dropSearchIndex(string $name = 'default'): static
```
jmikola: I think it's sensible to require an explicit value for `$name` since this is a destructive operation. It'd be consistent with PHPLIB's existing API.

GromNaN: I assumed that if you don't need to specify the name to create or use the default index, it must be the same for deletion. But as a precaution, it can be required. It's now modified.

jmikola (approved):

src/Schema/Blueprint.php
jmikola: Requiring explicit values for potentially destructive operations is the same reason the CRUD spec mandates a `filter` parameter for deletes. More generally, this logic can apply to any write operation (we also wouldn't want to default to an empty filter for an `updateMany()`).

PR #3236, opened 2025-01-02
Fix tests on Schema index helpers

GromNaN:
Find index by its actual name instead of the key fields.
Fix wrong assertions checking indexes doesn't exist with the wrong name.

### DIFF
--- a/tests/SchemaTest.php
+++ b/tests/SchemaTest.php
@@ -16,6 +16,7 @@ use MongoDB\Model\IndexInfo;
 use function assert;
 use function collect;
 use function count;
+use function sprintf;
 
 class SchemaTest extends TestCase
 {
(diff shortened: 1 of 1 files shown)


tests/SchemaTest.php
```diff
@@ -149,35 +150,37 @@ public function testDropIndex(): void
...
+        $this->assertNull($index);
 
+        $indexName = 'field_a_-1_field_b_1';
         Schema::table('newcollection', function ($collection) {
             $collection->index(['field_a' => -1, 'field_b' => 1]);
         });
 
-        $index = $this->getIndex('newcollection', 'field_a_-1_field_b_1');
+        $index = $this->getIndex('newcollection', $indexName);
         $this->assertNotNull($index);
```
GromNaN: Here the assertion was always true, `$index` was `false` for missing index.

tests/SchemaTest.php
```diff
@@ -573,23 +585,23 @@ public function testVectorSearchIndex()
         self::assertSame('vector', $index['latestDefinition']['fields'][0]['type']);
     }
 
-    protected function getIndex(string $collection, string $name)
+    protected function getIndex(string $collection, string $name): ?IndexInfo
```
GromNaN, line 580: Should I cast `IndexInfo` to an array for future compatibility with mongodb/mongodb v2.0?

jmikola: I'm not sure what you're suggesting. https://github.com/mongodb/mongo-php-library/commit/3d46eaf593df9337a23288368eda147c14cfc59c only removed the iterator classes. IndexInfo still remains.

tests/SchemaTest.php
jmikola: Having `assertIndexExists()` and `assertIndexNotExists()` would make the tests more readable if you wanted to go that route. I realize you can't get rid of `getIndex()` entirely, since some later tests expect to make assertions on the IndexInfo structure.

GromNaN: Thanks for the suggestion, that's more explicit with the new assert methods.

jmikola (approved):

PR #3242, opened 2025-01-09
PHPORM-277 Add `Builder::vectorSearch()`

GromNaN:
Fix PHPORM-277

Adding the field `vectorSearchScore` by default, as this is a helpful and [presented in the documentation](https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-stage/#atlas-vector-search-score).

For tests, the vectors are generated randomly, using a fixed seed.

### Checklist

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -25,6 +25,7 @@ use MongoDB\BSON\Regex;
 use MongoDB\BSON\UTCDateTime;
 use MongoDB\Builder\Search;
 use MongoDB\Builder\Stage\FluentFactoryTrait;
+use MongoDB\Builder\Type\QueryInterface;
 use MongoDB\Builder\Type\SearchOperatorInterface;
 use MongoDB\Driver\Cursor;
 use Override;
@@ -1530,6 +1531,40 @@ class Builder extends BaseBuilder
         ], fn ($arg) => $arg !== null);
 
         return $this->aggregate()->search(...$args)->get();
+    }
+
+    /**
+     * Performs a semantic search on data in your Atlas Vector Search index.
+     * NOTE: $vectorSearch is only available for MongoDB Atlas clusters, and is not available for self-managed deployments.
+     *
+     * @see https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-stage/
+     *
+     * @return Collection<object|array>
+     */
+    public function vectorSearch(
+        string $index,
+        array|string $path,
+        array $queryVector,
+        int $limit,
+        bool $exact = false,
+        QueryInterface|array|null $filter = null,
+        int|null $numCandidates = null,
+    ): Collection {
+        // Forward named arguments to the vectorSearch stage, skip null values
+        $args = array_filter([
+            'index' => $index,
+            'limit' => $limit,
+            'path' => $path,
+            'queryVector' => $queryVector,
+            'exact' => $exact,
+            'filter' => $filter,
+            'numCandidates' => $numCandidates,
+        ], fn ($arg) => $arg !== null);
+
+        return $this->aggregate()
+            ->vectorSearch(...$args)
+            ->addFields(vectorSearchScore: ['$meta' => 'vectorSearchScore'])
+            ->get();
     }
 
     /**
--- a/tests/AtlasSearchTest.php
+++ b/tests/AtlasSearchTest.php
@@ -251,7 +251,7 @@ class AtlasSearchTest extends TestCase
         );
     }
 
-    /** Generate random vectors */
+    /** Generate random vectors using fixed seed to make tests deterministic */
     private function addVector(array $items): array
     {
         srand(1);
--- a/tests/AtlasSearchTest.php
+++ b/tests/AtlasSearchTest.php
@@ -5,23 +5,31 @@ namespace MongoDB\Laravel\Tests;
 use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use Illuminate\Support\Collection as LaravelCollection;
 use Illuminate\Support\Facades\Schema;
+use MongoDB\Builder\Query;
 use MongoDB\Builder\Search;
 use MongoDB\Collection as MongoDBCollection;
 use MongoDB\Driver\Exception\ServerException;
 use MongoDB\Laravel\Schema\Builder;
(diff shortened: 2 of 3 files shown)


src/Query/Builder.php
```diff
@@ -1532,6 +1533,40 @@ public function search(
...
+        // Forward named arguments to the vectorSearch stage, skip null values
+        $args = array_filter([
+            'index' => $index,
+            'limit' => $limit,
+            'path' => $path,
+            'queryVector' => $queryVector,
+            'exact' => $exact,
+            'filter' => $filter,
+            'numCandidates' => $numCandidates,
+        ], fn ($arg) => $arg !== null);
```
jmikola, line 1562: This looks similar to the `array_filter` call you had in [#3232](https://github.com/mongodb/laravel-mongodb/pull/3232/files#diff-bdf045bbd1faa398417a788236b3b7da49c553bc4c6e7d6403568ecd775d2034R1518). Is there an opportunity to use a common helper here?

It doesn't seem like PHP has built-in function to conveniently access all function args with their names. The [note within `func_get_args()`](https://www.php.net/manual/en/function.func-get-args.php#refsect1-function.func-get-args-notes) specifically talks about how the function is transparent with respect to named arguments.

https://stackoverflow.com/q/74436227 suggests using [`get_defined_vars()`](https://www.php.net/get_defined_vars), if you think that'd be an improvement. You're already using `...$args` below, so `get_defined_args()` might let you replace these array literals.

Thinking about this some more, I suppose eliminating the explicit array here would make this logic _more_ brittle. Right now, the parameter names on these laravel-mongodb builder methods don't actually need to be in sync with those on the `vectorSearch()` method below. Switching over to `get_defined_args()`, or something similar, would then tightly couple the signatures. Given that, I think we're probably better off keeping this as-is, despite the code being a bit more verbose.

tests/AtlasSearchTest.php
```diff
@@ -131,7 +140,7 @@ public function testGetIndexes()
             ],
             [
                 'name' => 'vector',
-                'columns' => ['vector16', 'vector32'],
+                'columns' => ['vector4', 'vector32', 'title'],
```
jmikola, line 143: Does `vector32` serve any purpose? I only see it referenced here in `testGetIndexes()` and the original `setUp()` function above.

tests/AtlasSearchTest.php
```diff
@@ -194,9 +203,62 @@ public function testDatabaseBuilderAutocomplete()
...
+        ], $results->sort()->values()->all());
+    }
+
+    public function testDatabaseBuilderVectorSearch()
+    {
+        $results = $this->getConnection('mongodb')->table('books')
+            ->vectorSearch(
+                index: 'vector',
+                path: 'vector4',
+                queryVector: $this->vectors[7], // This is an exact match of the vector
```
jmikola, line 218: Prefacing this with the fact that I haven't a clue how Atlas Vector Search is implemented, it's not clear to me how this API works in practice. The `queryVector` parameter here originates from inserting the Book document. In an actual application where the records already exist in the database prior to the PHP process being executed, how would the developer obtain a `queryVector` value to use here?

tests/AtlasSearchTest.php
```diff
@@ -194,9 +203,62 @@ public function testDatabaseBuilderAutocomplete()
...
+        self::assertInstanceOf(Book::class, $results->first());
+        self::assertNotSame('The Art of Computer Programming', $results->first()->title);
+        self::assertSame('The Mythical Man-Month: Essays on Software Engineering', $results->first()->title);
+        self::assertThat(
+            $results->first()->vectorSearchScore,
+            self::logicalAnd(self::isType('float'), self::greaterThan(0.9), self::lessThan(1.0)),
+        );
+    }
+
+    /** Generate random vectors */
```
jmikola: I think this comment would benefit from noting that you're using a fixed seed to make the tests deterministic.

Also, am I correct in assuming the `vector4` generation here pertains to [How to Ingest Pre-Quantized Vectors](https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-quantization/#how-to-ingest-pre-quantized-vectors) in the Atlas docs? That may be worth referencing as well.

tests/AtlasSearchTest.php
```diff
@@ -194,9 +203,62 @@ public function testDatabaseBuilderAutocomplete()
...
+    private function addVector(array $items): array
+    {
+        srand(1);
+        foreach ($items as &$item) {
+            $this->vectors[] = $item['vector4'] = array_map(fn () => rand() / mt_getrandmax(), range(0, 3));
```
jmikola, line 259: Consider leaving a TODO to revisit this once [PHPC-2474](https://jira.mongodb.org/browse/PHPC-2474) is implemented. It looks like that ticket entails more than just adding a new subtype constant, as there are utility functions to convert to/from vectors (as described in [the corresponding spec](https://github.com/mongodb/specifications/blob/master/source/bson-binary-vector/bson-binary-vector.md#api-guidance)).

src/Query/Builder.php
GromNaN: Yes, I've tried solutions that were a bit automatic with `compact` for example (which is forbidden by our phpcs configuration), but in the end it's just fine to be explicit and exhaustive.

tests/AtlasSearchTest.php
GromNaN: It's only for `getIndexes` tests.

tests/AtlasSearchTest.php
GromNaN: The vectors are generated using an ["embedding model"](https://www.mongodb.com/docs/atlas/atlas-vector-search/create-embeddings/). In real applications, the database vector field would be filled by generating the "embedding" of every document (usually for each sentences for texts, or from images).
And the same AI model is used to generate the "embedding" vector for the query. The database is "simply" comparing vectors to find the nearest using the specified similarity function.

tests/AtlasSearchTest.php
GromNaN: Not sure for the todo, I create a ticket to add support in the aggregation builder first. https://jira.mongodb.org/browse/PHPLIB-1603

jmikola (approved):

PR #3243, opened 2025-01-10
PHPORM-286 Add `Query::countByGroup()` and other `aggregateByGroup()` functions

GromNaN:
Fix PHPORM-286

Required by https://github.com/mongodb/laravel-mongodb/pull/3182
New fonctions from Laravel v11.38+ by https://github.com/laravel/framework/pull/53679

### Checklist

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -11,6 +11,8 @@ jobs:
         name: "PHP ${{ matrix.php }} Laravel ${{ matrix.laravel }} MongoDB ${{ matrix.mongodb }} ${{ matrix.mode }}"
 
         strategy:
+            # Tests with Atlas fail randomly
+            fail-fast: false
             matrix:
                 os:
                     - "ubuntu-latest"
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -30,6 +30,7 @@ use MongoDB\Driver\Cursor;
 use Override;
 use RuntimeException;
 use stdClass;
+use TypeError;
 
 use function array_fill_keys;
 use function array_filter;
@@ -324,8 +325,10 @@ class Builder extends BaseBuilder
                     // this mimics SQL's behaviour a bit.
                     $group[$column] = ['$last' => '$' . $column];
                 }
+            }
 
-                // Do the same for other columns that are selected.
+            // Add the last value of each column when there is no aggregate function.
+            if ($this->groups && ! $this->aggregate) {
                 foreach ($columns as $column) {
                     $key = str_replace('.', '_', $column);
 
@@ -349,7 +352,7 @@ class Builder extends BaseBuilder
 
                     $aggregations = blank($this->aggregate['columns']) ? [] : $this->aggregate['columns'];
 
-                    if (in_array('*', $aggregations) && $function === 'count') {
+                    if (in_array('*', $aggregations) && $function === 'count' && empty($group['_id'])) {
                         $options = $this->inheritConnectionOptions($this->options);
 
                         return ['countDocuments' => [$wheres, $options]];
@@ -559,6 +562,8 @@ class Builder extends BaseBuilder
     /** @return ($function is null ? AggregationBuilder : mixed) */
     public function aggregate($function = null, $columns = ['*'])
     {
+        assert(is_array($columns), new TypeError(sprintf('Argument #2 ($columns) must be of type array, %s given', get_debug_type($columns))));
+
         if ($function === null) {
             if (! trait_exists(FluentFactoryTrait::class)) {
                 // This error will be unreachable when the mongodb/builder package will be merged into mongodb/mongodb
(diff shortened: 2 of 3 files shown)


tests/QueryBuilderTest.php
```diff
@@ -617,6 +619,43 @@ public function testSubdocumentArrayAggregate()
...
+            ['name' => 'Jane Doe', 'role' => 'admin', 'score' => 2],
+            ['name' => 'Robert Roe', 'role' => 'user', 'score' => 4],
+        ]);
+
+        $results = DB::table('users')->groupBy('role')->orderBy('role')->aggregateByGroup('count');
+        $this->assertInstanceOf(LaravelCollection::class, $results);
+        $this->assertEquals([(object) ['role' => 'admin', 'aggregate' => 2], (object) ['role' => 'user', 'aggregate' => 1]], $results->toArray());
+
+        if (! method_exists(Builder::class, 'countByGroup')) {
+            $this->markTestSkipped('countBy* function require Laravel v11.38+');
```
jmikola: Based on the description of [PHPORM-286](https://jira.mongodb.org/browse/PHPORM-286), I assume it's the "*ByGroup" functions that require Laravel 11.38+.

Also fixes subject/verb agreement.

```suggestion
            $this->markTestSkipped('*byGroup functions require Laravel v11.38+');
```

tests/QueryBuilderTest.php
```diff
@@ -617,6 +619,43 @@ public function testSubdocumentArrayAggregate()
...
 
+    public function testAggregateGroupBy()
+    {
+        DB::table('users')->insert([
+            ['name' => 'John Doe', 'role' => 'admin', 'score' => 1],
+            ['name' => 'Jane Doe', 'role' => 'admin', 'score' => 2],
+            ['name' => 'Robert Roe', 'role' => 'user', 'score' => 4],
+        ]);
+
+        $results = DB::table('users')->groupBy('role')->orderBy('role')->aggregateByGroup('count');
```
jmikola, line 630: This only tests the first argument of `aggregateByGroup()`. Is that sufficient?

src/Query/Builder.php
```diff
@@ -559,6 +562,8 @@ public function generateCacheKey()
     /** @return ($function is null ? AggregationBuilder : mixed) */
     public function aggregate($function = null, $columns = ['*'])
     {
+        assert(is_array($columns), new TypeError(sprintf('Argument #2 ($columns) must be of type array, %s given', get_debug_type($columns))));
+
         if ($function === null) {
             if (! trait_exists(FluentFactoryTrait::class)) {
                 // This error will be unreachable when the mongodb/builder package will be merged into mongodb/mongodb
```
jmikola, line 581: Is this now dead code since the builder was merged into PHPLIB? Assuming this isn't the only reference, this may be something to clean up en masse via another PR.

src/Query/Builder.php
```diff
@@ -349,7 +352,7 @@ public function toMql(): array
 
                     $aggregations = blank($this->aggregate['columns']) ? [] : $this->aggregate['columns'];
 
-                    if (in_array('*', $aggregations) && $function === 'count') {
+                    if (in_array('*', $aggregations) && $function === 'count' && empty($group['_id'])) {
```
jmikola: Noted that `$group['_id']` is only set within the conditional for `$this->groups` above. I assume this is the code path for `aggregateByGroup('count')`, but it's not clear to me how `aggregate()` ends up calling `toMql()`. Does that happen via the call chain of `get()` and `getFresh()`?

----

Looking a few lines up (beyond the diff, as I couldn't comment there directly), I don't see `$column` used after it is assigned with `implode('.', $splitColumns);`. Is that dead code?

src/Query/Builder.php
```diff
@@ -599,13 +604,32 @@ public function aggregate($function = null, $columns = ['*'])
         $this->columns            = $previousColumns;
         $this->bindings['select'] = $previousSelectBindings;
 
+        // When the aggregation is per group, we return the results as is.
+        if ($this->groups) {
+            return $results->map(function (object $result) {
+                unset($result->id);
```
jmikola, line 622: What is `$result->id` in this context? If this pertains to a `$group` stage, then wouldn't you mean to filter out `_id`? Or has some hydration already taken place by this point?

jmikola (commented):
Tests look fine, but I have some questions about the Builder.

src/Query/Builder.php
GromNaN: This is not released yet.

src/Query/Builder.php
GromNaN: Yes, `aggregateByGroup` calls `aggregate`, which calls `get`, which calls `getFresh`, which calls `toMql`.

---

This code is necessary, feature tested by [`testSubdocumentArrayAggregate`](https://github.com/mongodb/laravel-mongodb/blob/d6d8004b675369d2bc0f9c1b3091cbbd9c6057d8/tests/QueryBuilderTest.php#L597);

src/Query/Builder.php
GromNaN: In the results, `_id` field is renamed to `id` by [`aliasIdForResult`](https://github.com/mongodb/laravel-mongodb/blob/d6d8004b675369d2bc0f9c1b3091cbbd9c6057d8/src/Query/Builder.php#L1733)

tests/QueryBuilderTest.php
GromNaN: Added support for the 2nd parameter, the count will count distinct values.

.github/workflows/build-ci.yml
```diff
@@ -11,6 +11,8 @@ jobs:
         name: "PHP ${{ matrix.php }} Laravel ${{ matrix.laravel }} MongoDB ${{ matrix.mongodb }} ${{ matrix.mode }}"
 
         strategy:
+            # Tests with Atlas fail randomly
```
jmikola, line 14: Consider making an open ticket with the "Build Failure" type so you don't lose track of this. Assuming it's something that can be researched further down the line.

jmikola (approved):

.github/workflows/build-ci.yml
GromNaN: [PHPORM-287](https://jira.mongodb.org/browse/PHPORM-287)

PR #3244, opened 2025-01-13
PHPORM-209 Add query builder helper to set read preference

GromNaN:
Fix PHPORM-209

- Add a new method `Query\Builder::readPreference()` to set the read preference option for find and aggregate. 
- Add tests to the timeout option, to accept decimal number of seconds

### Checklist

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -103,7 +103,7 @@ class Builder extends BaseBuilder
     /**
      * The maximum amount of seconds to allow the query to run.
      *
-     * @var int
+     * @var int|float
      */
     public $timeout;
 
@@ -214,7 +214,7 @@ class Builder extends BaseBuilder
     /**
      * The maximum amount of seconds to allow the query to run.
      *
-     * @param  int $seconds
+     * @param  int|float $seconds
      *
      * @return $this
      */
@@ -457,7 +457,7 @@ class Builder extends BaseBuilder
 
         // Apply order, offset, limit and projection
         if ($this->timeout) {
-            $options['maxTimeMS'] = $this->timeout * 1000;
+            $options['maxTimeMS'] = (int) ($this->timeout * 1000);
         }
 
         if ($this->orders) {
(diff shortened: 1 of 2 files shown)


src/Query/Builder.php
```diff
@@ -113,6 +114,8 @@ class Builder extends BaseBuilder
      */
     public $hint;
 
+    private ReadPreference $readPreference;
```
GromNaN, line 117: I don't get why other properties are public. Using private by default to not add the new property to the API for no reason.

src/Query/Builder.php
```diff
@@ -1534,6 +1537,24 @@ public function options(array $options)
...
+     *
+     * @see https://www.php.net/manual/en/class.mongodb-driver-readpreference.php
+     *
+     * @param  string $mode
+     * @param  array  $tagSets
+     * @param  array  $options
+     *
+     * @return $this
+     */
+    public function readPreference(string $mode, ?array $tagSets = null, ?array $options = null): static
```
jmikola, line 1551: Just want to confirm that you don't need the ability to _clear_ a ReadPreference previously assigned to the builder.

tests/Query/BuilderTest.php
```diff
@@ -1416,6 +1417,31 @@ function (Builder $elemMatchQuery): void {
...
+        ];
+
+        yield 'hint' => [
+            ['find' => [[], ['hint' => ['foo' => 1]]]],
+            fn (Builder $builder) => $builder->hint(['foo' => 1]),
+        ];
+
+        yield 'timeout' => [
+            ['find' => [[], ['maxTimeMS' => 2345]]],
+            fn (Builder $builder) => $builder->timeout(2.3456),
```
jmikola, line 1443: Noted that this tests integer truncation.

jmikola (approved):

src/Query/Builder.php
GromNaN: Most of the operations using the query builder cannot be unset. 
For this config, it's still possible to overwrite with an other read preference.

PR #3262, opened 2025-02-05
Add template types to relation classes.

mad-briller:
This allows users of the package to properly define template types for their relations without receiving false positive warnings from phpstan about the relations not being generic, as well as allowing larastan to infer the model types returned from relation property accesses.

```php
/**
 * @return EmbedsMany<Items, ItemCollection>
 */
public function items(): EmbedsMany
{
  return $this->embedsMany(Item::class);
}
```

As a result this also cleans up the packages phpstan baseline a bit too :tada: 

### Checklist

### DIFF
--- a/src/Relations/BelongsTo.php
+++ b/src/Relations/BelongsTo.php
@@ -8,8 +8,8 @@ use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 
 /**
- * @template TRelatedModel of \MongoDB\Laravel\Eloquent\Model
- * @template TDeclaringModel of \MongoDB\Laravel\Eloquent\Model
+ * @template TRelatedModel of Model
+ * @template TDeclaringModel of Model
  * @extends \Illuminate\Database\Eloquent\Relations\BelongsTo<TRelatedModel, TDeclaringModel>
  */
 class BelongsTo extends \Illuminate\Database\Eloquent\Relations\BelongsTo
--- a/src/Relations/BelongsTo.php
+++ b/src/Relations/BelongsTo.php
@@ -7,6 +7,11 @@ namespace MongoDB\Laravel\Relations;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 
+/**
+ * @template TRelatedModel of \MongoDB\Laravel\Eloquent\Model
+ * @template TDeclaringModel of \MongoDB\Laravel\Eloquent\Model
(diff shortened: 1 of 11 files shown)


src/Relations/BelongsTo.php
```diff
@@ -7,6 +7,11 @@
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 
+/**
+ * @template TRelatedModel of \MongoDB\Laravel\Eloquent\Model
```
GromNaN: A Model class that extends the `Illuminate\Database\Eloquent\Model` base class and uses the `MongoDB\Laravel\Eloquent\DocumentModel` also works. Unfortunately we cannot typehint against the trait.

Also we have hybrid relationships (SQL <-> MongoDB).

Would it work to use `Illuminate\Database\Eloquent\Model` here instead?

mad-briller: yeah that would be fine, i can quickly change that

mad-briller: okay so just checked it, it works to allow us to add template types to the relations but it surfaces lots of new phpstan errors in the package, for example:

```
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   Relations/MorphToMany.php
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
         ✏️  Relations/MorphToMany.php
  236    Method Illuminate\Database\Eloquent\Model::push() invoked with 3 parameters, 0 required.
         ✏️  Relations/MorphToMany.php
  245    Method Illuminate\Database\Eloquent\Model::push() invoked with 3 parameters, 0 required.
         ✏️  Relations/MorphToMany.php
  267    Method Illuminate\Database\Eloquent\Model::push() invoked with 3 parameters, 0 required.
         ✏️  Relations/MorphToMany.php
  290    Method Illuminate\Database\Eloquent\Model::push() invoked with 3 parameters, 0 required.
         ✏️  Relations/MorphToMany.php
  333    Call to an undefined method TDeclaringModel of Illuminate\Database\Eloquent\Model::pull().
         ✏️  Relations/MorphToMany.php
  350    Call to an undefined method TDeclaringModel of Illuminate\Database\Eloquent\Model::pull().
         ✏️  Relations/MorphToMany.php
```
         
i can baseline them but thought it was worth bringing up.

I tried specifying a union of the mongo Model and eloquent Model and that didn't resolve them either :'(

GromNaN (approved):

GromNaN:
Thank you @mad-briller

PR #3255, opened 2025-01-28
DOCSP-46269: atlas search & atlas vector search pages

rustagir:
documents api for atlas search & vector search
https://jira.mongodb.org/browse/DOCSP-46269

Staging -

- https://deploy-preview-156--docs-laravel.netlify.app/fundamentals/atlas-search/
- https://deploy-preview-156--docs-laravel.netlify.app/fundamentals/vector-search/

### Checklist

- [x] Add tests and ensure they pass

PR #3279, opened 2025-02-17
DOCSP-46269 Fix doc examples on atlas search

GromNaN:
Fix doc examples introduced in https://github.com/mongodb/laravel-mongodb/pull/3255#pullrequestreview-2621081819

### DIFF
--- a/docs/includes/fundamentals/as-avs/AtlasSearchTest.php
+++ b/docs/includes/fundamentals/as-avs/AtlasSearchTest.php
@@ -10,7 +10,9 @@ use MongoDB\Builder\Query;
 use MongoDB\Builder\Search;
 use MongoDB\Driver\Exception\ServerException;
 use MongoDB\Laravel\Schema\Builder;
+use MongoDB\Laravel\Tests\Models\Book;
 use MongoDB\Laravel\Tests\TestCase;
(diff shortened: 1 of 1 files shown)


docs/includes/fundamentals/as-avs/AtlasSearchTest.php
```diff
@@ -127,7 +130,7 @@ public function autocompleteSearchTest(): void
      * @runInSeparateProcess
      * @preserveGlobalState disabled
      */
-    public function vectorSearchTest(): void
+    public function testVectorSearch(): void
```
GromNaN, line 132: This test is still returning an empty result collection, I don't get why.

rustagir: Maybe this is because the titles of the sample data aren't related to the plot field values? I can investigate.

rustagir (approved):
code changes lgtm, thanks a lot for fixing. We can wait to merge while I look into the vector Search test more

docs/includes/fundamentals/as-avs/AtlasSearchTest.php
GromNaN: The issue was in the model class that was used.

PR #3280, opened 2025-02-17
PHPORM-296 Enable support for Scout v10

GromNaN:
The Scout engine introduced in https://github.com/mongodb/laravel-mongodb/pull/3205 works with Scout v10, which is the latest released version. V11 is not released yet.

Fix [PHPORM-296](https://jira.mongodb.org/browse/PHPORM-296)


composer.json
```diff
@@ -35,12 +35,12 @@
     },
     "require-dev": {
         "mongodb/builder": "^0.2",
-        "laravel/scout": "^11",
+        "laravel/scout": "^10.3",
```
GromNaN, line 38: Version 10.3 required for `whereNotIn` feature that we implement. https://github.com/laravel/scout/commit/9b45a429a54d9ad1c2117e4fb4fcc75a2a0cacf9

composer.json
```diff
@@ -35,12 +35,12 @@
...
         "league/flysystem-gridfs": "^3.28",
         "league/flysystem-read-only": "^3.0",
         "phpunit/phpunit": "^10.3",
         "orchestra/testbench": "^8.0|^9.0",
-        "mockery/mockery": "^1.4.4",
+        "mockery/mockery": "^1.4.4@stable",
```
GromNaN, line 43: This prevents installing unstable release that fails with PHP 8.4

jmikola: I assume this is needed because of `minimum-stability: dev` below. Why does this library have that to begin with?

jmikola (approved):

composer.json
GromNaN: I think that was to test upcoming versions of Laravel. If found issues before the release thanks to that.

PR #3281, opened 2025-02-17
PHPORM-268 Add configuration for scout search indexes

GromNaN:
Fix PHPORM-268

The configuration looks like this:

```php
<?php # scout.php

return [
    'driver' => env('SCOUT_DRIVER', 'mongodb'),
    'prefix' => env('SCOUT_PREFIX', 'scout_'),

    'mongodb' => [
        'connection' => env('SCOUT_MONGODB_CONNECTION', 'mongodb'),
        'index-definitions' => [
            'scout_books' => [
                // The definition here
                'mappings' => ['dynamic' => true],
            ],
        ]
    ],
],
```

### DIFF
--- a/src/Scout/ScoutEngine.php
+++ b/src/Scout/ScoutEngine.php
@@ -9,6 +9,7 @@ use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\SoftDeletes;
 use Illuminate\Support\Collection;
 use Illuminate\Support\LazyCollection;
+use InvalidArgumentException;
 use Laravel\Scout\Builder;
 use Laravel\Scout\Engines\Engine;
 use Laravel\Scout\Searchable;
@@ -439,7 +440,7 @@ final class ScoutEngine extends Engine
 
         $definition = $this->indexDefinitions[$name] ?? self::DEFAULT_DEFINITION;
         if (! isset($definition['mappings'])) {
-            throw new LogicException(sprintf('The search index definition for collection "scout.mongodb.index-definition.%s" must contain a "mappings" key. Find documentation at https://www.mongodb.com/docs/atlas/atlas-search/create-index/', $name));
+            throw new InvalidArgumentException(sprintf('Invalid search index definition for collection "%s", the "mappings" key is required. Find documentation at https://www.mongodb.com/docs/manual/reference/command/createSearchIndexes/#search-index-definition-syntax', $name));
         }
 
         // Ensure the collection exists before creating the search index
--- a/src/Scout/ScoutEngine.php
+++ b/src/Scout/ScoutEngine.php
@@ -66,9 +66,11 @@ final class ScoutEngine extends Engine
 
     private const TYPEMAP = ['root' => 'object', 'document' => 'bson', 'array' => 'bson'];
 
+    /** @param array<string, array> $indexDefinitions */
     public function __construct(
         private Database $database,
         private bool $softDelete,
+        private array $indexDefinitions = [],
     ) {
     }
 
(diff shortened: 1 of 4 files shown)


src/Scout/ScoutEngine.php
```diff
@@ -435,14 +437,16 @@ public function createIndex($name, array $options = []): void
     {
         assert(is_string($name), new TypeError(sprintf('Argument #1 ($name) must be of type string, %s given', get_debug_type($name))));
 
+        $definition = $this->indexDefinitions[$name] ?? self::DEFAULT_DEFINITION;
+        if (! isset($definition['mappings'])) {
+            throw new LogicException(sprintf('The search index definition for collection "scout.mongodb.index-definition.%s" must contain a "mappings" key. Find documentation at https://www.mongodb.com/docs/atlas/atlas-search/create-index/', $name));
```
GromNaN: I'm not sure about adding a link here. Can be useful, maybe update it when the laravel docs is written.

jmikola: Would InvalidArgumentException be preferable, or is there prior art for using LogicException directly?

https://www.mongodb.com/docs/atlas/atlas-search/create-index/ seems fine if you're in the habit of adding doc links to exception messages.

https://www.mongodb.com/docs/manual/reference/command/createSearchIndexes/#search-index-definition-syntax may be a more direct reference, though.

tests/Scout/ScoutEngineTest.php
```diff
@@ -31,6 +34,82 @@ class ScoutEngineTest extends TestCase
 {
     private const EXPECTED_TYPEMAP = ['root' => 'object', 'document' => 'bson', 'array' => 'bson'];
 
+    public function testCreateIndexInvalidDefinition(): void
+    {
+        $database = m::mock(Database::class);
+        $engine = new ScoutEngine($database, false, ['collection_invalid' => ['foo' => 'bar']]);
+
+        $this->expectException(LogicException::class);
+        $this->expectExceptionMessage('The search index definition for collection "scout.mongodb.index-definition.collection_invalid" must contain a "mappings" key');
```
jmikola: Does this attempt to match the entire message, or is it only a prefix?

tests/Scout/ScoutIntegrationTest.php
```diff
@@ -38,6 +39,9 @@ protected function getEnvironmentSetUp($app): void
 
         $app['config']->set('scout.driver', 'mongodb');
         $app['config']->set('scout.prefix', 'prefix_');
+        $app['config']->set('scout.mongodb.index-definitions', [
+            'prefix_scout_users' => ['mappings' => ['dynamic' => true, 'fields' => ['bool_field' => ['type' => 'boolean']]]],
```
jmikola, line 43: Is it sensible to combine `dynamic: true` with explicit field mappings? I don't recall seeing anything like this in the server manual or Atlas docs.

GromNaN: It's allowed. It forces some field types but enable automatic indexation for others. That's useful for me here as I want to ensure the dynamic mapping works with my search query, but also want to test a custom definition.

jmikola (approved):
Will defer to you on the exception message.

tests/Scout/ScoutEngineTest.php
GromNaN: `str_contains` is used by PHPUnit for this.

src/Scout/ScoutEngine.php
GromNaN: I updated the exception class and the message with your link.

PR #3283, opened 2025-02-19
PHPORM-289 Support Laravel 12

### LINKED ISSUE #3288: Add Laravel 12 Support
Unable to upgrade to laravel 12 because of dependency conflicts

Add laravel 12 alongside with 10 & 11

N/A

### Additional context
N/A

GromNaN:
Fix PHPORM-289

Replace https://github.com/mongodb/laravel-mongodb/pull/3278
Closes https://github.com/mongodb/laravel-mongodb/issues/3288

Needed:
- [x] https://github.com/mongodb/laravel-mongodb/pull/3280
- [x] Release of https://github.com/spatie/laravel-query-builder/pull/994 (not branch-alias)
- [x] #3289

**Breaking changes**
- ~Add types to properties in `MongoDB\Laravel\Schema\BluePrint`: `$connection` and `$collection`. The property was added by https://github.com/laravel/framework/pull/51821~
- Add parameters to the methods in `MongoDB\Laravel\Schema`: `getTables` and `getTableListing`. This were added to support multi-databases in https://github.com/laravel/framework/pull/54274, but we decided not implementing this feature for MongoDB as this would introduce an other breaking change if collection name uses a `.`.


.github/workflows/build-ci.yml
```diff
@@ -28,19 +28,18 @@ jobs:
...
                 include:
                     - php: "8.1"
                       laravel: "10.*"
                       mongodb: "5.0"
                       mode: "low-deps"
                       os: "ubuntu-latest"
-                    - php: "8.4"
-                      laravel: "11.*"
-                      mongodb: "7.0"
-                      os: "ubuntu-latest"
```
GromNaN, line 40: This was duplicating an existing case.

phpstan-baseline.neon
```diff
@@ -1,5 +1,15 @@
...
+		-
+			message: "#^Class MongoDB\\\\Laravel\\\\Query\\\\Grammar does not have a constructor and must be instantiated without any parameters\\.$#"
+			count: 1
+			path: src/Connection.php
+
+		-
+			message: "#^Class MongoDB\\\\Laravel\\\\Schema\\\\Grammar does not have a constructor and must be instantiated without any parameters\\.$#"
+			count: 1
+			path: src/Connection.php
+
```
GromNaN, line 12: It tried using the `@phpstan-ignore new.noConstructor` in the code, but that triggers an other error `ignore.unmatchedIdentifier` on Laravel 12, where the constructor has this parameter.

src/Connection.php
```diff
@@ -114,13 +114,31 @@ public function getSchemaBuilder()
     /**
      * Get the MongoDB database object.
      *
+     * @deprecated Use getDatabase() instead
```
GromNaN: This is the beginning of [PHPORM-278](https://jira.mongodb.org/browse/PHPORM-278), but I should create a distinct PR.

src/Connection.php
```diff
@@ -327,13 +345,15 @@ protected function getDefaultPostProcessor()
...
+        // Argument added in Laravel 12
+        return new Query\Grammar($this);
     }
 
     /** @inheritdoc */
     protected function getDefaultSchemaGrammar()
     {
-        return new Schema\Grammar();
+        // Argument added in Laravel 12
+        return new Schema\Grammar($this);
```
GromNaN, line 366: The constructor was introduced by https://github.com/laravel/framework/commit/c78fa3d0684206f721b9a3b76e8c596aa8a08cd0#diff-13a4975d935d982e846853d4328df60c7ef014b5a4010441277b2ef4de31ed2aR27

tests/RelationsTest.php
```diff
@@ -35,6 +35,7 @@ public function tearDown(): void
         Photo::truncate();
         Label::truncate();
         Skill::truncate();
+        Soft::truncate();
```
GromNaN, line 38: This was a latent bug when the test suite was run in a particular order, `MassPrunableTest::testPruneSoftDelete` failed.

src/Connection.php
GromNaN: https://github.com/mongodb/laravel-mongodb/pull/3289

src/Schema/Blueprint.php
```diff
@@ -18,19 +20,19 @@
...
+    private static bool $hasConnectionInConstructor;
+
     /**
-     * The MongoConnection object for this blueprint.
-     *
-     * @var Connection
+     * The MongoDB connection object for this blueprint.
      */
-    protected $connection;
+    protected Connection $connection;
```
GromNaN: Type added in Laravel 12. This is technically a breaking change, but I hope nobody extends this class.

.github/workflows/build-ci.yml
```diff
@@ -28,19 +28,18 @@ jobs:
...
                 exclude:
                     - php: "8.1"
                       laravel: "11.*"
+                    - php: "8.1"
+                      laravel: "12.*"
```
jmikola, line 42: Is this excluded because Laravel 11.x and 12.x require PHP 8.2+? If so, considering adding a comment above for future reference.

src/Connection.php
```diff
@@ -114,13 +114,31 @@ public function getSchemaBuilder()
...
+     *
+     * @return Database
+     */
+    public function getDatabase(?string $name = null): Database
+    {
+        if ($name === null) {
+            return $this->db;
+        }
+
+        return $this->connection->{$name};
```
jmikola: Would it be preferable to explicitly invoke `Client::getDatabase()` here instead of going through `__get()`, which will do so on its own?

src/Schema/Blueprint.php
jmikola: May be worth adding a comment in the doc block above to explain this. Just a note that Laravel 12 added the type hint and that necessitated doing so here.

src/Schema/Blueprint.php
```diff
@@ -18,19 +20,19 @@
...
 
     /**
-     * The Collection object for this blueprint.
+     * The MongoDB collection object for this blueprint.
      *
      * @var Collection
      */
-    protected $collection;
+    protected Collection $collection;
```
jmikola: I assume this wasn't required but you were just taking advantage of the opportunity since the Connection type hint was added above.

src/Connection.php
```diff
@@ -327,13 +345,15 @@ protected function getDefaultPostProcessor()
     /** @inheritdoc */
     protected function getDefaultQueryGrammar()
     {
-        return new Query\Grammar();
+        // Argument added in Laravel 12
+        return new Query\Grammar($this);
```
jmikola, line 359: I assume this relies on the relative namespace of the file. Is that preferable to explicitly using these classes? Perhaps, since doing so would require aliases.

src/Schema/Blueprint.php
```diff
@@ -42,9 +44,19 @@ class Blueprint extends SchemaBlueprint
     /**
      * Create a new schema blueprint.
      */
-    public function __construct(Connection $connection, string $collection)
+    public function __construct(Connection $connection, string $collection, ?Closure $callback = null)
     {
-        parent::__construct($collection);
+        // Parent constructor signature was changed in Laravel 12
+        // https://github.com/laravel/framework/commit/f29df4740d724f1c36385c9989569e3feb9422df#diff-68f714a9f1b751481b993414d3f1300ad55bcef12084ec0eb8f47f350033c24bR107
+        self::$hasConnectionInConstructor ??= (new ReflectionMethod(parent::class, '__construct'))->getParameters()[0]->getName() === 'connection';
```
jmikola: Is there a reason you didn't try this approach for the construction of Query\Grammar and Schema\Grammar objects in Connection.php?

Would it not have worked to avoid the PHPStan issue?

src/Schema/Builder.php
```diff
@@ -137,9 +142,10 @@ public function dropAllTables()
         }
     }
 
-    public function getTables()
+    /** @param string|null $schema Database name */
+    public function getTables($schema = null)
```
jmikola, line 146: Are the signature changes here also a potential BC break for Laravel 11 users, or would you not expect anyone to extend this class?

src/Schema/Builder.php
```diff
@@ -165,9 +171,23 @@ public function getTables()
         return $collections;
     }
 
-    public function getTableListing()
+    public function getTableListing($schema = null, $schemaQualified = true)
```
jmikola: Would this function benefit from a doc block to document `$schema` as `string|array|null`, and `$schemaQualified` as a `bool`?

src/Schema/Builder.php
```diff
@@ -165,9 +171,23 @@ public function getTables()
...
+            foreach ($schema as $db) {
+                $collections[$db] = iterator_to_array($this->connection->getDatabase($db)->listCollectionNames());
+            }
+        }
+
+        if ($schema && $schemaQualified) {
+            $collections = array_map(fn ($db, $collections) => array_map(static fn ($collection) => $db . '.' . $collection, $collections), array_keys($collections), $collections);
+        }
+
+        $collections = array_merge(...array_values($collections));
```
jmikola, line 197: If `$schemaQualified` is false and the database name is never prepended to yield full namespaces, would this statement potentially clobber collection names?

.github/workflows/build-ci.yml
GromNaN: PHP 8.1 is not supported since laravel 11.
https://github.com/laravel/framework/blob/11.x/composer.json#L18

src/Connection.php
GromNaN: This was to ensure compatibility before the upgrade to mongodb/mongodb v1.21

src/Schema/Blueprint.php
GromNaN: Yes.

src/Connection.php
GromNaN: In find it more explicit than having to alias one of `Query\Grammar` and `Schema\Grammar`.

src/Schema/Blueprint.php
GromNaN: For the construction of the Grammar classes, PHP allows passing extra arguments that are not used by the contructor/method. I used this approach to not rely too much on reflection at runtime.

src/Schema/Builder.php
GromNaN: Yes, this is a breaking change if someone extends our class... I hope nobody does. We should make more classes final.

src/Schema/Builder.php
GromNaN: Yes, but that's how the feature is designed.

jmikola (approved):
LGTM, as I'm not sure what else would need to be done here; however, you may want to consider whether these BC breaks warrant a major version bump.

src/Schema/BlueprintLaravelCompatibility.php
```diff
@@ -0,0 +1,48 @@
...
+if (! property_exists(BaseBlueprint::class, 'connection')) {
+    /** @internal For compatibility with Laravel 10 and 11 */
+    trait BlueprintLaravelCompatibility
+    {
+        /**
+         * The MongoDB connection object for this blueprint.
+         *
+         * @var Connection
+         */
+        protected $connection;
```
jmikola, line 29: Does this also need to be declared in the Laravel 12+ trait?

The earlier version of this PR also added a type hint on the `$collection` property (https://github.com/mongodb/laravel-mongodb/pull/3283#discussion_r1971818758). I'm not sure if you intended to include that here.

GromNaN: That's not necessary as the parent class declares it. And we don't use it in the trait.

marlonbasten:
@jmikola So what's the holdup here? 😅

BlackOrder:
this build need the update of mongodb. but it's a breaking change because it needs: pecl install mongodb

PR #3286, opened 2025-02-24
PHPORM-299 Enable PHPUnit 11

GromNaN:
Fix PHPORM-299

Ensures the `parent::tearDown` method is always called to remove the error/exception handlers registered by Laravel.

### DIFF
--- a/tests/AuthTest.php
+++ b/tests/AuthTest.php
@@ -17,10 +17,10 @@ class AuthTest extends TestCase
 {
     public function tearDown(): void
     {
-        parent::setUp();
-
         User::truncate();
         DB::table('password_reset_tokens')->truncate();
+
+        parent::tearDown();
     }
 
     public function testAuthAttempt()
--- a/tests/SchemaTest.php
+++ b/tests/SchemaTest.php
@@ -26,6 +26,8 @@ class SchemaTest extends TestCase
         assert($database instanceof Database);
         $database->dropCollection('newcollection');
         $database->dropCollection('newcollection_two');
+
+        parent::tearDown();
     }
 
     public function testCreate(): void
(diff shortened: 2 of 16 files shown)


tests/AuthTest.php
```diff
@@ -17,10 +17,10 @@ class AuthTest extends TestCase
 {
     public function tearDown(): void
     {
-        parent::setUp();
```
jmikola, line 20: Was this a longstanding bug? I don't understand why a `setUp()` method should ever have been called during `tearDown()`.

tests/SchemaTest.php
```diff
@@ -66,14 +66,12 @@ public function testDrop(): void
...
-        Schema::table('newcollection', function ($collection) use ($instance) {
-            $instance->assertInstanceOf(Blueprint::class, $collection);
+        Schema::table('newcollection', static function ($collection) {
+            self::assertInstanceOf(Blueprint::class, $collection);
         });
 
-        Schema::table('newcollection', function ($collection) use ($instance) {
-            $instance->assertInstanceOf(Blueprint::class, $collection);
+        Schema::table('newcollection', static function ($collection) {
+            self::assertInstanceOf(Blueprint::class, $collection);
```
jmikola, line 74: Alternatively, you can consider importing the PHPUnit functions as we do in various PHPLIB tests. That would allow removing `self::` here.

jmikola (approved):
Defer to you if you'd like to use imported functions to replace the `self::` prefixes for assertions.

tests/SchemaTest.php
GromNaN: I'm fine with the `self::` prefix, that's more common than global functions.

tests/AuthTest.php
GromNaN: That was added by https://github.com/mongodb/laravel-mongodb/commit/bdb15c948cb1f3d703df0996753d7e5336950bee#diff-ebd98368c673d521f95a7f40fff9ea0f20763fdd85dc6071fb7f710343804bcf, I can't explain.

PR #3287, opened 2025-02-24
PHPORM-303 Require mongodb library v1.21 with aggregation builder

GromNaN:
Fix PHPORM-303

The new search features https://github.com/mongodb/laravel-mongodb/pull/3242 and https://github.com/mongodb/laravel-mongodb/pull/3232 use the aggregation builder introduced by https://github.com/mongodb/mongo-php-library/pull/1537

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,11 +30,10 @@
         "illuminate/database": "^10.30|^11",
         "illuminate/events": "^10.0|^11",
         "illuminate/support": "^10.0|^11",
-        "mongodb/mongodb": "^1.18",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -54,6 +53,7 @@
         "mongodb/builder": "Provides a fluent aggregation builder for MongoDB pipelines"
     },
     "minimum-stability": "dev",
+    "prefer-stable": true,
```
GromNaN, line 56: So that we don't use unstable versions unless necessary.

jmikola (approved):

PR #3289, opened 2025-02-24
PHPORM-278 Introduce `Connection::getDatabase()` and `getClient()`

GromNaN:
And deprecate getMongoDB and get MongoClient

Fix PHPORM-278

### Checklist

### DIFF
--- a/src/Concerns/ManagesTransactions.php
+++ b/src/Concerns/ManagesTransactions.php
@@ -12,15 +12,18 @@ use Throwable;
 
 use function MongoDB\with_transaction;
 
-/** @see https://docs.mongodb.com/manual/core/transactions/ */
+/**
+ * @internal
+ *
+ * @see https://docs.mongodb.com/manual/core/transactions/
+ */
 trait ManagesTransactions
 {
     protected ?Session $session = null;
 
     protected $transactions = 0;
 
-    /** @return Client */
-    abstract public function getMongoClient();
+    abstract public function getClient(): ?Client;
 
     public function getSession(): ?Session
     {
@@ -30,7 +33,7 @@ trait ManagesTransactions
     private function getSessionOrCreate(): Session
     {
         if ($this->session === null) {
-            $this->session = $this->getMongoClient()->startSession();
+            $this->session = $this->getClient()->startSession();
         }
 
         return $this->session;
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -71,7 +71,7 @@ class Connection extends BaseConnection
         $this->database = $this->getDefaultDatabaseName($dsn, $config);
 
         // Select database
-        $this->db = $this->connection->selectDatabase($this->database);
+        $this->db = $this->connection->getDatabase($this->database);
 
         $this->tablePrefix = $config['prefix'] ?? '';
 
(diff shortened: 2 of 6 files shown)


src/Connection.php
```diff
@@ -65,9 +68,10 @@ public function __construct(array $config)
 
         // Create the connection
         $this->connection = $this->createConnection($dsn, $config, $options);
+        $this->database = $this->getDefaultDatabaseName($dsn, $config);
 
         // Select database
-        $this->db = $this->connection->selectDatabase($this->getDefaultDatabaseName($dsn, $config));
+        $this->db = $this->connection->selectDatabase($this->database);
```
alcaeus:
```suggestion
        $this->db = $this->connection->getDatabase($this->database);
```

src/Concerns/ManagesTransactions.php
```diff
@@ -12,15 +12,18 @@
 
 use function MongoDB\with_transaction;
 
-/** @see https://docs.mongodb.com/manual/core/transactions/ */
+/**
+ * @internal
```
jmikola, line 16: Was this class always internal? Just curious if there's a BC concern here.

tests/ConnectionTest.php
```diff
@@ -196,14 +196,51 @@ public static function dataConnectionConfig(): Generator
...
 
+    public function testLegacyGetMongoClient(): void
+    {
+        $connection = DB::connection('mongodb');
+        $expected = $connection->getClient();
+
+        $this->assertSame($expected, $connection->getMongoClient());
+    }
+
+    public function testLegacyGetMongoDB(): void
```
jmikola, line 215: Do these legacy tests need to expect a deprecation message?

src/Connection.php
```diff
@@ -65,9 +68,10 @@ public function __construct(array $config)
 
         // Create the connection
         $this->connection = $this->createConnection($dsn, $config, $options);
+        $this->database = $this->getDefaultDatabaseName($dsn, $config);
```
jmikola, line 71: Is this dynamically assigning a class property, or is this declared on BaseConnection?

Either way, this appears to be the first time it's being used.

src/Concerns/ManagesTransactions.php
GromNaN: We try to do better than [Laravel's `ManagesTransactions` trait](https://github.com/laravel/framework/blob/12.x/src/Illuminate/Database/Concerns/ManagesTransactions.php), that does not declare its dependencies. This trait is not meant to be used outside of the `Connection` class (it should be merged in it but we kept Laravel's design).

Technically, that's a breaking change that is easily overcome by adding a `getClient` method to the class that uses it.

tests/ConnectionTest.php
GromNaN: Something catches deprecation messages and hides them. I couldn't add an assertion.

src/Connection.php
GromNaN: The property is defined [in the parent class](https://github.com/laravel/framework/blob/45f0347e4baaab341fbab27031026b60ff3e9ffb/src/Illuminate/Database/Connection.php#L55), it was not set previously.

jmikola (approved):

PR #3299, opened 2025-03-04
Fix releasing from development branch

alcaeus:
This PR fixes a syntax error in the release workflow that prevents releasing minor versions from their development branch.


GromNaN (approved):
Thanks

PR #3301, opened 2025-03-04
Merges the read operation reorganization into 5.2

rustagir:
Merges the read operation reorganization into 5.2 (seems like there was an issue where the 5.2 branch was created but not included in the front-ports)

### Checklist

### DIFF
--- a/docs/fundamentals/read-operations/search-text.txt
+++ b/docs/fundamentals/read-operations/search-text.txt
@@ -1,0 +1,157 @@
+.. _laravel-fundamentals-search-text:
+.. _laravel-retrieve-text-search:
+
+===========
+Search Text
+===========
+
+.. facet::
+   :name: genre
+   :values: reference
+
+.. meta::
+   :keywords: filter, string, CRUD, code example
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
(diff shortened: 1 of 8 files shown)


norareidy (approved):

docs/fundamentals/read-operations/search-text.txt
```diff
@@ -0,0 +1,157 @@
+.. _laravel-fundamentals-search-text:
+.. _laravel-retrieve-text-search:
```
jmikola, line 2: These anchors seem inconsistent with the names in `retrieve.txt`:

```
.. _laravel-fundamentals-retrieve-documents:
.. _laravel-fundamentals-read-retrieve:
```

Is there even a need for multiple anchors at the same location? Presumably, the references to these could just be updated to use the same name.

jmikola (approved):
One question about inconsistent and/or redundant anchors but LGTM.

docs/fundamentals/read-operations/search-text.txt
rustagir: We have some guidelines around moving pages/content - https://wiki.corp.mongodb.com/spaces/DE/pages/201997355/How+To+Add+or+Rename+an+rST+Page+or+Section

That's the reason that this PR contains multiple anchors - preserving the original section headings and adding more targeted ones that match the current file tree structure. We could probably develop better guidelines around the specific words we use in these anchors though!

Issue #3015, opened 2024-06-28
[Bug fix] BelongToMany relation with ObjectId


florianJacques:
- Laravel-mongodb Version: 4.5
- PHP Version: 8.3
- Database Driver & Version: MongoDB extension version => 1.18.0

### Description:

When we use a BelongsToMany relationship with objectIf in foreignKey, this one are incorrectly cast.

Create Planet model:

```php
<?php

declare(strict_types=1);

namespace App\Models\Test;

use MongoDB\Laravel\Eloquent\Model;
use MongoDB\Laravel\Relations\BelongsToMany;

class Planet extends Model
{
    protected $connection = 'mwe_mongodb';
    protected $collection = 'TestPlanet';

    public function getIdAttribute($value = null)
    {
        return $value;
    }

    public function visitors(): BelongsToMany
    {
        return $this->belongsToMany(SpaceExplorer::class, null, 'planetId', 'visitorId');
    }
}

```

Create SpaceExplorerModel

```php
<?php

declare(strict_types=1);

namespace App\Models\Test;

use MongoDB\Laravel\Eloquent\Model;
use MongoDB\Laravel\Relations\BelongsToMany;

class SpaceExplorer extends Model
{
    protected $connection = 'mwe_mongodb';
    protected $collection = 'TestSpaceExplorer';

    public function getIdAttribute($value = null)
    {
        return $value;
    }

    public function planetsVisited(): BelongsToMany
    {
        return $this->belongsToMany(Planet::class, null, 'visitorIds', 'planetIds');
    }
}

```

And attach planet to spaceExplorer:

```php
<?php

use App\Models\Test\Planet;
use App\Models\Test\SpaceExplorer;

$planetEarth = new Planet();
$planetEarth->name = 'Earth';
$planetEarth->save();
$planetJupiter = new Planet();
$planetJupiter->name = 'Jupiter';
$planetJupiter->save();
$explorerTanya = new SpaceExplorer();
$explorerTanya->name = 'Tanya Kirbuk';
$explorerTanya->save();
$explorerTanya->planetsVisited()->attach($planetEarth);
$explorerTanya->planetsVisited()->attach($planetJupiter);
```

### Expected behaviour
The expected result

```json
{
    "_id" : ObjectId("667e95f3759cbea7c80d1a24"),
    "name" : "Tanya Kirbuk",
    "updated_at" : ISODate("2024-06-28T10:52:35.735+0000"),
[...]

PR #3310, opened 2025-03-06
Fix BelongsToMany Relation to Support Array Foreign Keys in Laravel MongoDB

peymanaslani70:
### **🚀 What’s in This PR?**  
This PR fixes the `BelongsToMany` relationship to support **array-based foreign keys**, allowing multiple related ObjectIds to be stored in a single field.  

### **🐞 The Issue**  
- Previously, `BelongsToMany` expected foreign keys to be a **single value**, causing issues when an array was used.  
- This resulted in errors like `Illegal offset type` when working with MongoDB.  

🔍 Issue Reference
Fixes [#3015]

### **🔧 Changes Made**  
✅ Updated `attach()` in `BelongsToMany.php` to store ObjectIds as an **array** instead of a single value.  
✅ Modified `buildDictionary()` to correctly **handle arrays** when resolving relationships.  
✅ Added a new unit test: `testBelongsToManyRelationSupportsArrayForeignKeys` (in `tests/RelationsTest.php`).  
   - Ensures that array-based foreign keys work as expected.  
   - Confirms that relationships retrieve the correct data.  

### **🧪 How to Test It?**  
1. Run the tests with:  
   ```sh
   vendor/bin/phpunit tests/RelationsTest.php
   ```  
2. All tests should pass ✅  

---

### **📜 Documentation Update Checklist**  
✔️ Updated **Many-to-Many Relationship** section in the docs.  
✔️ Explained how to use **array-based foreign keys** with `BelongsToMany()`.  
✔️ Added an example showing how to define relationships with an array field.  

### Checklist

- [x] Updated **Many-to-Many Relationship** section in the docs.

### DIFF
--- a/docs/eloquent-models/relationships.txt
+++ b/docs/eloquent-models/relationships.txt
@@ -255,6 +255,26 @@ in the Laravel documentation.
 The following section shows an example of how to create a many to many
 relationship between model classes.
 
+.. note::
+
+   In previous versions, the `foreignPivotKey` in a `BelongsToMany` relationship
+   was expected to be a single `ObjectId` or `string`. However, in some MongoDB
+   schemas, it is common to store multiple related IDs as an array. As of this update,
+   `BelongsToMany` now supports array values for the `foreignPivotKey`.
+
+   Example:
+
+   .. code-block:: php
+
+       class User extends Model {
+           public function roles() {
+               return $this->belongsToMany(Role::class, null, 'role_ids', '_id');
+           }
+       }
+
+   Here, `role_ids` can be an array of `ObjectId`s, and the relationship will work correctly.
+
+
 Many to Many Example
 ~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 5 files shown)


GromNaN (changes requested):
Sorry for the delay in the review, it's because I don't understand the feature implemented.
The new test passes, even without the `BelongsToMany` modification. And the data stored in the database is identical with or without your modification. Which makes me think that the test doesn't cover the intended functionality.

Tested with Laravel 12.11.0, the user document contains an array of ids represented in with `string` type.
image

Jira tracking: PHPORM-116

peymanaslani70:
> Sorry for the delay in the review, it's because I don't understand the feature implemented. The new test passes, even without the `BelongsToMany` modification. And the data stored in the database is identical with or without your modification. Which makes me think that the test doesn't cover the intended functionality.
> 
> Tested with Laravel 12.11.0, the user document contains an array of ids represented in with `string` type. image
> 
> Jira tracking: [PHPORM-116](https://jira.mongodb.org/browse/PHPORM-116)

Thanks a lot for the detailed feedback! 🙏
I understand the points you've mentioned and will work on aligning the implementation with the package's structure. I'll also add a dedicated test case as suggested.
Will update the PR soon.

GromNaN:
Closing this PR due to lack of follow-up. Please feel free to open a new PR if you manage to add a non-regression test for the proposed patch.

PR #3319, opened 2025-03-11
PHPORM-306 Test with MongoDB Driver v2

GromNaN:
Fix PHPORM-306

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -4,11 +4,15 @@ on:
     push:
     pull_request:
 
+env:
+  MONGODB_EXT_V1: mongodb-1.21.0
+  MONGODB_EXT_V2: mongodb-mongodb/mongo-php-driver@v2.x
+
 jobs:
     build:
         runs-on: "${{ matrix.os }}"
 
-        name: "PHP ${{ matrix.php }} ext ${{ matrix.driver }} Laravel ${{ matrix.laravel }} MongoDB ${{ matrix.mongodb }} ${{ matrix.mode }}"
+        name: "PHP/${{ matrix.php }} Laravel/${{ matrix.laravel }} Driver/${{ matrix.driver }} Server/${{ matrix.mongodb }} ${{ matrix.mode }}"
 
         strategy:
             matrix:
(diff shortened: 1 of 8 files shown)


tests/Scout/ScoutEngineTest.php
```diff
@@ -115,26 +115,28 @@ public function testCreateIndexCustomDefinition(): void
...
-        $database = m::mock(Database::class);
-        $collection = m::mock(Collection::class);
-        $database->shouldReceive('selectCollection')
+        $database = $this->createMock(Database::class);
+        $collection = $this->createMock(Collection::class);
+        $database->expects($this->once())
+            ->method('selectCollection')
             ->with('collection_searchable')
-            ->andReturn($collection);
-        $cursor = m::mock(CursorInterface::class);
```
GromNaN, line 123: Mocking the `CursorInterface` with Mockery doesn't work: https://github.com/mockery/mockery/issues/1459

src/Eloquent/Builder.php
```diff
@@ -285,7 +285,7 @@ public function createOrFirst(array $attributes = [], array $values = [])
 
         try {
             return $this->create(array_merge($attributes, $values));
-        } catch (WriteException $e) {
```
GromNaN, line 288: Exception removed in https://github.com/mongodb/mongo-php-driver/pull/1685

tests/Scout/ScoutEngineTest.php
jmikola: Did that necessitate changing the entire file to use PHPUnit for mocking? It looks like only two test methods mocked a CursorInterface. Would it have been possible to just use PHPUnit for the cursor assertions?

I don't feel strongly about this either way. Just curious.

GromNaN: Actually, there was no need to change everywhere. I did it for the sake of consistency throughout this test class.
Personally, I prefer PHPUnit's Mock Objects.

.github/workflows/build-ci.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     build:
         runs-on: "${{ matrix.os }}"
 
-        name: "PHP ${{ matrix.php }} Laravel ${{ matrix.laravel }} MongoDB ${{ matrix.mongodb }} ${{ matrix.mode }}"
+        name: "PHP ${{ matrix.php }} ext ${{ matrix.driver }} Laravel ${{ matrix.laravel }} MongoDB ${{ matrix.mongodb }} ${{ matrix.mode }}"
```
GromNaN: @alcaeus the required check will have to be updated once we agree on the new pattern.

jmikola: I assume that's related to the two outstanding pending checks in the CI status?

Are MongoDB 4.4 and 5.0 (with PHP 8.2 and Laravel 11.*) the _only_ required checks? If so, perhaps the server versions should be updated to something more recent. I say this irrespective of the extension change in this PR.

alcaeus: We'll want to use something for the label that does not require changing every time we update the driver version, so the full version string is probably a bad idea. We could try to only list the major version number to make sure this doesn't need changing again until we test with version 3 of the driver.

alcaeus (approved):
LGTM. I'll make the changes to the branch rules.

Issue #3320, opened 2025-03-12
exclude System Collections in listCollectionNames()


masterbater:
MongoDB stores system information in collections that use the <database>.system.* [namespace](https://www.mongodb.com/docs/manual/reference/glossary/#std-term-namespace), which MongoDB reserves for internal use. Do not create collections that begin with system.

https://www.mongodb.com/docs/manual/reference/system-collections/

<database>.system.profile
<database>.system.js
<database>.system.views

### Is your feature request related to a problem?
system.views is a systems collection but in laravel includes it in the result which causes issue with testing traits that drops all tables

Other system collection that could in the testing database 
<database>.system.profile
<database>.system.js

### Describe the solution you'd like
Exclude it when dropping all tables, like using migrate:fresh, testing traits like RefreshDatabase, DatabaseMigration

### Additional context

<img width="733" alt="Image" src="https://github.com/user-attachments/assets/e551f26a-8e43-4166-9021-913c4455b544" />

Temporary solution add a custom command, but I strongly think  that system collections should be excluded in listCollectionNames
```php
$database = DB::connection('mongodb')->getMongoDB();

    foreach ($database->listCollectionNames() as $collectionName) {

        if (in_array($collectionName, ['system.views', 'system.profile'])) {
            continue;
        }
        DB::table($collectionName)->delete();
    }
```

masterbater:
Its seems in mongodb driver for nodejs like getCollectionNames like listCollectionNames in php driver will return system collections, so I think its up to laravel-mongodb package to handle this issue and not in the driver 

<img width="786" alt="Image" src="https://github.com/user-attachments/assets/c4818a01-858e-4bea-83eb-b4fd9167da3d" />

PR #3321, opened 2025-03-12
feat: exclude system collection in getAllCollection

masterbater:
Exclude system collection  in getAllCollections which only use by dropAllTables
solves this 
https://github.com/mongodb/laravel-mongodb/issues/3320

### Checklist

### DIFF
--- a/src/Schema/Builder.php
+++ b/src/Schema/Builder.php
@@ -5,27 +5,41 @@ declare(strict_types=1);
 namespace MongoDB\Laravel\Schema;
 
 use Closure;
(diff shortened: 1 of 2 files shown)


GromNaN:
That's a good idea. In think it should also be done in `getTableListing` and `getTables`.
Also, could you add some tests please?

masterbater:
> That's a good idea. In think it should also be done in `getTableListing` and `getTables`. Also, could you add some tests please?


Its moved here
https://github.com/mongodb/laravel-mongodb/pull/3327

PR #3327, opened 2025-03-24
PHPORM-325 Add `getViews` and categorize table types

### LINKED ISSUE #3320: exclude System Collections in listCollectionNames()

masterbater:
Fix #3320
This fixes `db:show` and `migrate:fresh` for database that has virtual view type collections

- add `Connection::getViews()`
- `php artisan db:show  --views`
- fix collection fields introspection with qualified name

### Checklist


masterbater:
Laravel has 
```php
php artisan db:show  --views
$views = Schema::getViews();
```

To follow laravel convention Ill adjust and add getViews

src/Schema/Builder.php
```diff
@@ -142,13 +143,28 @@ public function dropAllTables()
...
+            // Skip system collections
+            if (str_starts_with($collectionName, 'system.')) {
+                continue;
+            }
+
+            // Skip views it doesnt suport aggregate
+            $isView = ($collectionInfo['type'] ?? '') === 'view';
+            $stats = null;
+
+            if ($isView) {
```
GromNaN: The `type` property is always set, you can use [`getType`](https://github.com/mongodb/mongo-php-library/blob/9803a12f6569c86255fab1c213da21bc3408ba4d/src/Model/CollectionInfo.php#L126C21-L126C28). 

And the variable `$stats` is always set line 168, isn't it?
```suggestion
            if ($collectionInfo->getType() === 'view') {
```

src/Schema/Builder.php
```diff
@@ -165,9 +181,45 @@ public function getTables($schema = null)
...
+            // Skip system collections
+            if (str_starts_with($collectionName, 'system.')) {
+                continue;
+            }
+
+            // Skip views it doesnt suport aggregate
+            $isView = ($collectionInfo['type'] ?? '') === 'view';
+            $stats = null;
+
+            if (! $isView) {
```
GromNaN:
```suggestion
            if ($collectionInfo->getType() !== 'view') {
```

src/Schema/Builder.php
```diff
@@ -346,7 +401,14 @@ protected function getAllCollections()
     {
         $collections = [];
         foreach ($this->connection->getDatabase()->listCollections() as $collection) {
-            $collections[] = $collection->getName();
+            $name = $collection->getName();
+
+            // Skip system collections
+            if (str_starts_with($name, 'system.')) {
```
GromNaN: This will preserve system collection from being dropped 👍🏻 

Even for the gridfs bucket metadata, I think that's good to not remove this collection when `dropAllTables` is called.
https://www.mongodb.com/docs/manual/reference/system-collections/#database-specific-collections

tests/SchemaTest.php
```diff
@@ -397,6 +399,13 @@ public function testGetTables()
         DB::connection('mongodb')->table('newcollection_two')->insert(['test' => 'value']);
         $dbName = DB::connection('mongodb')->getDatabaseName();
 
+        // Create a view (this creates system.views)
+        DB::connection('mongodb')->getDatabase()->command([
+            'create' => 'test_view',
+            'viewOn' => 'newcollection',
+            'pipeline' => [],
+        ]);
```
GromNaN: That's a good idea to create a view to initialize a system collection. You could also create a GridFS bucket.

You can use the `createCollection` method instead of the raw command. The pipeline is optional.
```suggestion
        DB::connection('mongodb')->getDatabase()->createCollection('test_view', ['viewOn' => 'newcollection']);
```

GromNaN (commented):
Sounds perfect to me. A few minor comments.

tests/SchemaTest.php
```diff
@@ -413,24 +418,67 @@ public function testGetTables()
                 $this->assertEquals($dbName . '.newcollection', $table['schema_qualified_name']);
                 $found = true;
             }
+
+            // Ensure system collections are excluded
+            $this->assertFalse(str_starts_with($table['name'], 'system.'));
```
jmikola: [`assertStringStartsNotWith()`](https://docs.phpunit.de/en/10.5/assertions.html#assertstringstartswith) can be used here. In the event of a failure the error message would have more context.

tests/SchemaTest.php
```diff
@@ -454,6 +502,28 @@ public function testGetTableListingBySchema()
...
+        // Get all collections directly from MongoDB
+        $allCollections = DB::connection('mongodb')->getDatabase()->listCollectionNames();
+
+        // Ensure the system.views collection exists in MongoDB
+        $this->assertContains('system.views', $allCollections);
+
+        // Ensure Schema::getTables does NOT include system collections
+        $tables = Schema::getTables();
+        foreach ($tables as $table) {
+            $this->assertFalse(str_starts_with($table['name'], 'system.'));
```
jmikola: [`assertStringStartsNotWith()`](https://docs.phpunit.de/en/10.5/assertions.html#assertstringstartswith) can be used here as well.

src/Schema/Builder.php
jmikola: [`system.buckets`](https://www.mongodb.com/docs/manual/reference/system-collections/#mongodb-data--database-.system.buckets) stores data for time series collections. It's unrelated to GridFS, which stores its data in `<bucketName>.fs` (metadata) and `<bucketName.files>` (chunk data) collections.

I'm not sure how `dropAllTables()` is used, but I don't see why you'd need system collections to persist if everything else is being removed. Moreover, if `Blueprint::drop()` (called by `dropAllTables()`) is going to invoke `Collection::drop()` for each collection, you could conceivably just drop the entire database to be more efficient.

phpunit.xml.dist
```diff
@@ -17,7 +17,7 @@
         </testsuite>
     </testsuites>
     <php>
-        <env name="MONGODB_URI" value="mongodb://mongodb/?directConnection=true"/>
+        <env name="MONGODB_URI" value="mongodb://127.0.0.1/?directConnection=true"/>
```
jmikola: This change seems unrelated to the PR.

The introduction of the `mongodb` hostname here dates back to https://github.com/mongodb/laravel-mongodb/commit/74d85f8194109dde4a121adcb0467b0f22223a6a and seems relevant for the Docker test environment.

src/Schema/Builder.php
```diff
@@ -142,13 +143,25 @@ public function dropAllTables()
...
-        foreach ($db->listCollectionNames() as $collectionName) {
+        foreach ($db->listCollections() as $collectionInfo) {
+            $collectionName = $collectionInfo->getName();
+
+            // Skip system collections
+            if (str_starts_with($collectionName, 'system.')) {
+                continue;
+            }
+
+            // Skip views it doesnt suport aggregate
```
jmikola: Fixing the grammar in this comment:

> Skip views, which don't support aggregate

I haven't used [time series collections](https://www.mongodb.com/docs/manual/core/timeseries-collections/#behavior) myself, but they are documented as a special type of writable view; however, it appears they [do support aggregations](https://www.mongodb.com/docs/manual/core/timeseries/timeseries-procedures/#run-aggregations-on-a-time-series-collection).

src/Schema/Builder.php
```diff
@@ -142,13 +143,25 @@ public function dropAllTables()
...
     public function getTables($schema = null)
     {
         $db = $this->connection->getDatabase($schema);
         $collections = [];
 
-        foreach ($db->listCollectionNames() as $collectionName) {
+        foreach ($db->listCollections() as $collectionInfo) {
+            $collectionName = $collectionInfo->getName();
+
+            // Skip system collections
```
jmikola: The system collections are actual collections. In the case of `system.profile`, that might actually be something an application would want to query.

I'm not certain it makes sense to exclude these from _all_ output through the integration.

GromNaN (approved):
Thanks for the corrections. 
@jmikola could you also review please.

src/Schema/Builder.php
```diff
@@ -165,9 +178,42 @@ public function getTables($schema = null)
...
+    /** @param  string|null $schema Database name */
+    public function getViews($schema = null)
+    {
+        $db = $this->connection->getDatabase($schema);
+        $collections = [];
+
+        foreach ($db->listCollections() as $collectionInfo) {
+            $collectionName = $collectionInfo->getName();
+
+            // Skip system collections
```
jmikola: Is this necessary? I wouldn't expect any system collection to have a "view" type, so the conditional below should already filter these out.

src/Schema/Builder.php
```diff
@@ -165,9 +178,42 @@ public function getTables($schema = null)
...
+                continue;
+            }
+
+            $collections[] = [
+                'name' => $collectionName,
+                'schema' => $db->getDatabaseName(),
+                'schema_qualified_name' => $db->getDatabaseName() . '.' . $collectionName,
+                'size' => null,
+                'comment' => null,
+                'collation' => null,
```
jmikola, line 211: If this is actually relevant, the information should be available via the `options` field for each collection info result. See: [`listCollections`](https://www.mongodb.com/docs/v7.0/reference/command/listCollections/#mongodb-data-listCollections.cursor)

src/Schema/Builder.php
```diff
@@ -165,9 +178,42 @@ public function getTables($schema = null)
...
+            // Skip normal type collection
+            if ($collectionInfo->getType() !== 'view') {
+                continue;
+            }
+
+            $collections[] = [
+                'name' => $collectionName,
+                'schema' => $db->getDatabaseName(),
+                'schema_qualified_name' => $db->getDatabaseName() . '.' . $collectionName,
+                'size' => null,
```
jmikola, line 209: If this is relevant, the information can be provided via [`$collStats`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/collStats/) (MongoDB 6.2+). I'm not sure that'd be worth the overhead to obtain this for all collections/views, though.

masterbater:
This is now a simple PR, separating tables(collections in mongodb) and views(standard views). No more segregating SYSTEM collections. 

For dropAllTables, I followed the recommendation to drop the entire database instead of deleting collections individually. However, this approach may lead to errors, such as "database is in the process of being dropped." While I’m unsure if tthere any production workloads that could affect this, it does cause issues in parallel testing when using the same database—an incorrect setup that should instead use separate databases for each test.

jmikola:
> For dropAllTables, I followed the recommendation to drop the entire database instead of deleting collections individually. However, this approach may lead to errors, such as "database is in the process of being dropped." While I’m unsure if tthere any production workloads that could affect this, it does cause issues in parallel testing when using the same database—an incorrect setup that should instead use separate databases for each test.

If the library needs to work around this, it might be sufficient to catch a [ServerException](https://www.php.net/manual/en/class.mongodb-driver-exception-serverexception.php) and no-op for [DatabaseDropPending code 215](https://www.mongodb.com/docs/manual/reference/error-codes/#mongodb-error-215).

I reckon that the original approach to delete collections individually would still be subject to a race condition if a collection was created by another connection between `dropAllTables()` listing the collections and dropping them individually.

src/Schema/Builder.php
```diff
@@ -134,21 +134,34 @@ public function drop($table)
...
+ *
+ * Drops the entire database instead of deleting each collection individually.
+ *
+ * In MongoDB, dropping the whole database is much faster than dropping collections
+ * one by one. The database will be automatically recreated when a new connection
+ * writes to it.
+ */
     public function dropAllTables()
     {
-        foreach ($this->getAllCollections() as $collection) {
```
jmikola, line 140: @GromNaN: This is now outside the scope of this PR, but I noted that `getAllCollections()` could more easily use [`Database::listCollectionNames()`](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBDatabase-listCollectionNames/).

That said, I'm not even sure that method is still necessary unless you're overriding a base method. It was protected, but this seemed like the only reference to it (in this file at least).

src/Schema/Builder.php
```diff
@@ -134,21 +134,34 @@ public function drop($table)
...
-    /** @inheritdoc */
+/**
+ * @inheritdoc
+ *
+ * Drops the entire database instead of deleting each collection individually.
+ *
+ * In MongoDB, dropping the whole database is much faster than dropping collections
+ * one by one. The database will be automatically recreated when a new connection
+ * writes to it.
+ */
```
jmikola: This doc block and its method have inconsistent indentation.

jmikola (commented):
@GromNaN: Let me know if there's anything else for me to review, but I think you can take it from here.

src/Schema/Builder.php
GromNaN: I'm adding a deprecation notice.

src/Schema/Builder.php
GromNaN: If someone finds it useful, we can do it, but we'll have to check the impact on performance if there are a lot of collections.

Tracked in [PHPORM-326](https://jira.mongodb.org/browse/PHPORM-326)

src/Schema/Builder.php
GromNaN: Tracked in [PHPORM-326](https://jira.mongodb.org/browse/PHPORM-326)

GromNaN:
Thank you @masterbater

src/Schema/Builder.php
paulinevos: FYI these line comments are on `getViews` while the task refers to collections (which already have the size displayed). I'm gonna go ahead and assume the task is about collections as the code comments mention views don't support aggregation.

PR #3329, opened 2025-03-30
Remove manual dirty _id check when updating a model

### LINKED ISSUE #3326: saveQuietly() fails inside created event listener
- Laravel-mongodb Version: ~5.1.0
- PHP Version: 8.1.32
- Database Driver & Version: 
ext-mongodb: [1.20.1](https://pecl.php.net/package/mongodb/1.20.1)
mongodb/mongodb: ^1.8
- Laravel version: 10.48.29

### Description:
I'm having an issue using this package with models that have a "created" event listener.
I've created a minimal repository with a clean Laravel 10 installation where the error occurs:
🔗 [GitHub Repo](https://github.com/andresantos-tech/laravel10-mongodb-PoC)

I have the following in my model:

```php
static::created(function (MyModelWithCreatedEventNotWorking $model) {
    $model->bla = 'bla';
    $model->saveQuietly();
});
```
However, when saving the model, I get the following error:

```
InvalidArgumentException  Cannot update "id" field
```
This exception is thrown in [Builder.php#L742](https://github.com/mongodb/laravel-mongodb/blob/5.1.1/src/Query/Builder.php#L742).

#### Example
This code triggers the error:

```php
$model = new \App\Models\MyModelWithCreatedEventNotWorking();
$model->value = 'value';
$model->save();
// InvalidArgumentException  Cannot update "id" field
```

However, if I call saveQuietly() **outside** the static::created event, it works as expected:

```php
$model = new \App\Models\MyModelWithoutCreatedEvent();
$model->value = 'value';
$model->bla = 'bla';
$model->saveQuietly();
// Works fine
```

#### Workaround
To make it work inside the static::created event, I need to call refresh() before saving:

```php
static::created(function (MyModelWithCreatedEventWorking $model) {
    // Fix
    $model->refresh();
    $model->bla = 'bla';
    $model->saveQuietly();
});
```

Now, the model saves without issues:

```php
$model = new \App\Models\MyModelWithCreatedEventWorking();
$model->value = 'value';
$model->save();
// Works fine
```

This behavior seems to be a bug because it worked correctly in Laravel 9.* with "jenssegers/mongodb": "^3.9".

### Steps to reproduce
1. Clone the repo 🔗 [GitHub Repo](https://github.com/andresantos-tech/laravel10-mongodb-PoC)
2. Install dependencies
3. Run the examples above in `artisan tinker`

### Expected behaviour
`->saveQuietly()` inside the `static::created` event should work without needing to call `$model->refresh()`.

### Actual behaviour
`->saveQuietly()` inside the `static::created` event only works after calling `$model->refresh()`. Without refreshing, it throws the exception:
`InvalidArgumentException Cannot update "id" field.`

Thanks! 😊

itodor:
Fixes an issue where calling `save()` or `saveQuietly()` inside a model's `created` event throws:

`InvalidArgumentException: Cannot update "id" field
`

### Related Issue

Closes [#3326](https://github.com/mongodb/laravel-mongodb/issues/3326)

### Background

After inserting the model, Laravel's internal state still sees `_id` as dirty — so `save()` tries to update it again, which MongoDB doesn't allow.

### Fix

The explicit check for dirty _id has been removed, and the behavior now relies on MongoDB itself to throw an error if _id is mutated.

### Checklist

- [x] Tests added

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -783,13 +783,6 @@ class Builder extends BaseBuilder
             unset($values[$key]);
         }
 
-        // Since "id" is an alias for "_id", we prevent updating it
-        foreach ($values as $fields) {
-            if (array_key_exists('id', $fields)) {
-                throw new InvalidArgumentException('Cannot update "id" field.');
-            }
-        }
-
         return $this->performUpdate($values, $options);
     }
 
--- a/tests/Ticket/GH3326Test.php
+++ b/tests/Ticket/GH3326Test.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Laravel\Tests\Ticket;
+
+use MongoDB\Laravel\Eloquent\Model;
+use MongoDB\Laravel\Tests\TestCase;
+
+/**
+ * @see https://github.com/mongodb/laravel-mongodb/issues/3326
+ * @see https://jira.mongodb.org/browse/PHPORM-309
+ */
+class GH3326Test extends TestCase
+{
+    public function testCreatedEventCanSafelyCallSave(): void
+    {
+        $model = new GH3326Model();
+        $model->foo = 'bar';
+        $model->save();
+
+        $fresh = $model->fresh();
+
+        $this->assertEquals('bar', $fresh->foo);
+        $this->assertEquals('written-in-created', $fresh->extra);
+    }
+}
+
+class GH3326Model extends Model
+{
+    protected $connection = 'mongodb';
+    protected $collection = 'test_gh3326';
+    protected $guarded = [];
+
+    protected static function booted(): void
+    {
+        static::created(function ($model) {
+            $model->extra = 'written-in-created';
+            $model->saveQuietly();
+        });
+    }
+}


src/Eloquent/Model.php
```diff
@@ -57,4 +58,62 @@ final public static function isDocumentModel(string|object $class): bool
...
+     *
+     * Ensures syncOriginal() is called immediately after insert, so that
+     * subsequent save() calls (e.g., inside the "created" model event)
+     * don't treat the primary key (_id) as dirty.
+     *
+     * @param Builder $query
+     *
+     * @return bool
+     */
+    protected function performInsert(Builder $query)
```
GromNaN: Overriding and duplicating a big method like this makes the code more difficult to maintain with Laravel updates. I would prefer to avoid.

tests/Eloquent/ModelTest.php
```diff
@@ -21,6 +23,18 @@ public function testIsDocumentModel(bool $expected, string|object $classOrObject
         $this->assertSame($expected, Model::isDocumentModel($classOrObject));
     }
 
+    public function testCreatedEventCanSafelyCallSave(): void
```
GromNaN: You should move this test is a dedicated class, in `tests/Ticket`.

GromNaN (changes requested):
As a simpler fix, we can remove the exception in the `Query\Builder::update` method, and let the MongoDB server throw an exception itself when someone tries to update the `_id` field. https://github.com/mongodb/laravel-mongodb/blob/730a2aceb715463eec2a84d9322d52d935f414a5/src/Query/Builder.php#L786-L791

GromNaN (approved):
SGTM. Thanks for the very descriptive issue report, the test case and the fix.

PR #3332, opened 2025-04-01
PHPORM-255 Enable disabling the `id` to `_id` field rename in embedded documents

### LINKED ISSUE #3184: [Feature Request] Options to ignore auto conversion id to _id
We have a `merchant.id` field in our mongodb document, and after upgrading to mongodb driver 5.1 we encountered issue that mongodb driver modify the where query from `merchant.id` to `merchant._id` which results to unexpected query

It would be good if there's mechanism to whitelisting the fields from auto conversion `.id` to `._id`

-

### Additional context
This part of code modify all `.id` to `._id`
image

GromNaN:
Fix PHPORM-255 
Fix #3184

To disable the automatic conversion of `id` to `_id` and store `id` fields in the database for embedded documents, set the connection setting using the new method:
```php
DB::connection('mongodb')->setRenameEmbeddedIdField(false);
```

### Checklist

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -126,8 +126,6 @@ class Builder extends BaseBuilder
      */
     public $options = [];
 
-    private ?bool $renameEmbeddedIdField;
-
     /**
      * All of the available clause operators.
      *
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -29,6 +29,7 @@ use MongoDB\Builder\Type\QueryInterface;
 use MongoDB\Builder\Type\SearchOperatorInterface;
 use MongoDB\Driver\Cursor;
 use MongoDB\Driver\ReadPreference;
+use MongoDB\Laravel\Connection;
 use Override;
 use RuntimeException;
 use stdClass;
@@ -83,6 +84,7 @@ use function substr;
 use function trait_exists;
 use function var_export;
 
+/** @method Connection getConnection() */
 class Builder extends BaseBuilder
 {
     private const REGEX_DELIMITERS = ['/', '#', '~'];
@@ -123,6 +125,8 @@ class Builder extends BaseBuilder
      * @var array
      */
     public $options = [];
+
+    private ?bool $renameEmbeddedIdField;
 
     /**
      * All of the available clause operators.
@@ -1764,9 +1768,9 @@ class Builder extends BaseBuilder
         throw new BadMethodCallException('This method is not supported by MongoDB');
     }
 
-    private function aliasIdForQuery(array $values): array
+    private function aliasIdForQuery(array $values, bool $root = true): array
     {
-        if (array_key_exists('id', $values)) {
+        if (array_key_exists('id', $values) && ($root || $this->getConnection()->getRenameEmbeddedIdField())) {
             if (array_key_exists('_id', $values) && $values['id'] !== $values['_id']) {
                 throw new InvalidArgumentException('Cannot have both "id" and "_id" fields.');
             }
@@ -1793,7 +1797,7 @@ class Builder extends BaseBuilder
             }
 
             // ".id" subfield are alias for "._id"
-            if (str_ends_with($key, '.id')) {
+            if (str_ends_with($key, '.id') && ($root || $this->getConnection()->getRenameEmbeddedIdField())) {
                 $newkey = substr($key, 0, -3) . '._id';
                 if (array_key_exists($newkey, $values) && $value !== $values[$newkey]) {
                     throw new InvalidArgumentException(sprintf('Cannot have both "%s" and "%s" fields.', $key, $newkey));
@@ -1806,7 +1810,7 @@ class Builder extends BaseBuilder
 
         foreach ($values as &$value) {
             if (is_array($value)) {
-                $value = $this->aliasIdForQuery($value);
+                $value = $this->aliasIdForQuery($value, false);
             } elseif ($value instanceof DateTimeInterface) {
                 $value = new UTCDateTime($value);
             }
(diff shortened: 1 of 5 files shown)


src/Query/Builder.php
```diff
@@ -1763,9 +1767,9 @@ public function orWhereIntegerNotInRaw($column, $values, $boolean = 'and')
         throw new BadMethodCallException('This method is not supported by MongoDB');
     }
 
-    private function aliasIdForQuery(array $values): array
+    private function aliasIdForQuery(array $values, bool $root = true): array
     {
-        if (array_key_exists('id', $values)) {
+        if (array_key_exists('id', $values) && ($root || $this->getConnection()->getRenameEmbeddedIdField())) {
```
stakx: Perhaps I am overly cautious, but I was wondering whether repeatedly calling `$this->getConnection()` could posssibly introduce a noticeable performance penalty worth worrying about? (This method appears to get called from inside loops, after all.) Or is this just a simple property lookup?

stakx (commented):
Thanks for doing this PR, really happy to see any progress regarding this issue!

src/Query/Builder.php
GromNaN: You're right, it's better to use the property since it's available.

src/Query/Builder.php
```diff
@@ -123,6 +125,8 @@ class Builder extends BaseBuilder
      */
     public $options = [];
 
+    private ?bool $renameEmbeddedIdField;
```
katjackson: If this property is only set on the connection, should it be defined here on the builder?

GromNaN: There are builder instances created in many places, you don't always have access to them to change the setting on each specific instance. 
If we used a static variable, it would apply to every connection. 
So I think that the connection is the good level for this setting.

katjackson: I agree, which is why I'm confused that this property is also defined here on the builder. I guess I meant, "this shouldn't be defined here on the builder, because it's never set on the builder".

GromNaN: Oh, right. I forgot to remove it after trying to define it in the builder. Which did not work well.

src/Query/Builder.php
```diff
@@ -1763,9 +1767,9 @@ public function orWhereIntegerNotInRaw($column, $values, $boolean = 'and')
         throw new BadMethodCallException('This method is not supported by MongoDB');
     }
 
-    private function aliasIdForQuery(array $values): array
+    private function aliasIdForQuery(array $values, bool $root = true): array
     {
-        if (array_key_exists('id', $values)) {
+        if (array_key_exists('id', $values) && ($root || $this->connection->getRenameEmbeddedIdField())) {
```
mshamaseen, line 1771: would it be easier to check:
`if (!$root && !$this->connection->getRenameEmbeddedIdField()) return $values;` at the beginning instead of checking them multiple times in the function?

mshamaseen (commented):
It would be great if we could add an option in the config file to disable this behaviour.

src/Query/Builder.php
GromNaN: `array_key_exists('id', $values)` is checked before, and can be `true` only once. I thins the current implementation is more optimized and clear.

src/Query/Builder.php
```diff
@@ -1805,7 +1807,7 @@ private function aliasIdForQuery(array $values): array
 
         foreach ($values as &$value) {
             if (is_array($value)) {
-                $value = $this->aliasIdForQuery($value);
+                $value = $this->aliasIdForQuery($value, false);
```
jmikola, line 1811: Note that this appears to be the only recursive call to `aliasIdForQuery()`. All other calls (e.g. preparing projections, wheres) are invoked at the root level.

src/Query/Builder.php
```diff
@@ -1823,10 +1825,13 @@ private function aliasIdForQuery(array $values): array
      *
      * @template T of array|object
      */
-    public function aliasIdForResult(array|object $values): array|object
+    public function aliasIdForResult(array|object $values, bool $root = true): array|object
```
jmikola, line 1829: There appear to be two calls to `aliasIdForResult()` from `MongoDB\Laravel\Eloquent\Builder::raw()` (see: [search results](https://github.com/search?q=repo%3Amongodb%2Flaravel-mongodb%20aliasIdForResult&type=code)).

The context suggests that you're only processing a single document, but in one case you call `iterator_to_array()` on a cursor (which suggests it'd be handling _multiple_ results instead of just one).

Neither of the `aliasIdForResult()` calls there seem affected by this change, since `$root: false` is only used when recursing, but it'd be worth confirming.

tests/Query/BuilderTest.php
```diff
@@ -1600,12 +1599,38 @@ public static function getEloquentMethodsNotSupported()
         yield 'orWhereIntegerNotInRaw' => [fn (Builder $builder) => $builder->orWhereIntegerNotInRaw('id', ['1a', 2])];
     }
 
-    private static function getBuilder(): Builder
+    public function testRenameEmbeddedIdFieldCanBeDisabled()
```
jmikola, line 1602: Does this test exercise both `aliasIdForQuery` _and_ `aliasIdForResult()`?

jmikola (approved):
I don't think my questions warrant any changes, but feel free to ping me for a second look if needed.

tests/Query/BuilderTest.php
GromNaN: Thanks for this question, I added a full test to cover both.

src/Query/Builder.php
GromNaN: Good catch, that's fixed by using `array_map`.

jmikola: Just to confirm, this fix was necessary to get the test case for multiple results passing?

jmikola (approved):
One question so I can understand what the `array_map()` addition actually fixed, but this LGTM.

src/Query/Builder.php
GromNaN: Previously, the function was called with `$root = true` on the list of document, and then each document root had `$root = false` because the function was called recursively. Using `array_map`, I'm calling `aliasIdForResult` on the documents directly, instead of the collection of documents.

muhammadshazada:
Hi, I'm still getting the "id" instead of "_id" in both root level and nested level, tried the 'rename_embedded_id_field' => false but not getting the " _id" even in nested level.

PR #3336, opened 2025-04-08
PHPORM-311 Fix Update of numeric field name

GromNaN:
Fix PHPORM-311
Fix https://github.com/mongodb/laravel-mongodb/discussions/3335 @vipinpapnai

`array_merge` changes the index of numeric keys, whereas `array_replace` keeps the keys.

### Checklist

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -40,6 +40,7 @@ use function array_is_list;
 use function array_key_exists;
 use function array_map;
 use function array_merge;
+use function array_replace;
 use function array_values;
 use function assert;
 use function blank;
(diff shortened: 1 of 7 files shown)


vipinpapnai:
@GromNaN 
I am using this code
$employee = Employee::findById($user_id, true);
$employee->{'38'} = 'Php';
$employee->save();
array_merge or array_replace is not there just updating numeric key in database but got no luck

src/Query/Builder.php
```diff
@@ -426,7 +427,7 @@ public function toMql(): array
 
             // Add custom query options
             if (count($this->options)) {
-                $options = array_merge($options, $this->options);
+                $options = array_replace($options, $this->options);
```
jmikola, line 430: There's likely no concern for numeric keys here, but as we discussed it's more expressive to use `array_replace()` when merging options as well.

jmikola (approved):

Issue #3341, opened 2025-04-09
Exception when using CarbonImmutable


saineshmamgain:
- Laravel-mongodb Version: 5.2.2
- PHP Version: 8.2.28
- Database Driver & Version:

### Steps to reproduce
1. Add `Date::use(CarbonImmutable::class);` in your `AppServiceProvider`'s boot method.

### Expected behaviour
There should not be any exception

### Actual behaviour
An exception occurs with the message:
`MongoDB\\Laravel\\Eloquent\\Model::asDateTime(): Return value must be of type Carbon\\Carbon, Carbon\\CarbonImmutable returned`

This can be solved if we change the return type of `asDateTime` to `CarbonInterface` or `DateTimeInterface`.

<details><summary><b>Logs</b>:</summary>

```json
{
    "message": "MongoDB\\Laravel\\Eloquent\\Model::asDateTime(): Return value must be of type Carbon\\Carbon, Carbon\\CarbonImmutable returned",
    "exception": "TypeError",
    "file": "/var/www/html/sg-ui-backend/vendor/mongodb/laravel-mongodb/src/Eloquent/DocumentModel.php",
    "line": 136
}
```

</details>

GromNaN:
I'm tracking this issue in Jira as PHPORM-313.

Since you identified the fix, would you like to open a pull request adding a test?

saineshmamgain:
@GromNaN Sure! I will raise a PR.

saineshmamgain:
@GromNaN I've created the PR and added a test. Please review it and let me know if anything else is required.

PR #3342, opened 2025-04-09
Fixes: getting immutable_datetime property fails if `Date::use(CarbonImmutable::class)` is set

saineshmamgain:
This PR fixes the issue #3341 .

Basically when `Date::use(CarbonImmutable::class);` is added. The method `asDateTime` in `DocumentModel` trait throws an exception. I've changed the return type of `asDateTime` method from `Carbon` to `DateTimeInterface`.

### Checklist

### DIFF
--- a/tests/DateTimeImmutableTest.php
+++ b/tests/DateTimeImmutableTest.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Laravel\Tests\Eloquent;
+
+use Carbon\CarbonImmutable;
+use Illuminate\Support\Facades\Date;
+use MongoDB\Laravel\Eloquent\DocumentModel;
+use MongoDB\Laravel\Tests\Models\Anniversary;
+use MongoDB\Laravel\Tests\TestCase;
+
+use function assert;
+
+final class DateTimeImmutableTest extends TestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        Date::use(CarbonImmutable::class);
+        Anniversary::truncate();
+    }
+
+    public function testCanReturnCarbonImmutableObject(): void
+    {
+        Date::use(CarbonImmutable::class);
+
+        Anniversary::create([
+            'name' => 'John',
+            'anniversary' => new CarbonImmutable('2020-01-01 00:00:00'),
+        ]);
+
+        $anniversary = Anniversary::sole();
+        assert($anniversary instanceof Anniversary);
+        self::assertInstanceOf(CarbonImmutable::class, $anniversary->anniversary);
+
+        Date::useDefault();
+    }
+}
--- a/tests/DateTimeImmutableTest.php
+++ b/tests/DateTimeImmutableTest.php
@@ -23,6 +23,8 @@ final class DateTimeImmutableTest extends TestCase
     protected function tearDown(): void
     {
         Date::useDefault();
+
+        parent::tearDown();
     }
 
     public function testCanReturnCarbonImmutableObject(): void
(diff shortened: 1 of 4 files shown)


tests/DateTimeImmutableTest.php
```diff
@@ -0,0 +1,38 @@
...
+        Anniversary::create([
+            'name' => 'John',
+            'anniversary' => new CarbonImmutable('2020-01-01 00:00:00'),
+        ]);
+
+        $anniversary = Anniversary::sole();
+        assert($anniversary instanceof Anniversary);
+        self::assertInstanceOf(CarbonImmutable::class, $anniversary->anniversary);
+
+        Date::useDefault();
```
GromNaN: The reset must be in the tearDown method, to be done even if there is an error in the test method.

tests/Models/Anniversary.php
```diff
@@ -0,0 +1,30 @@
...
+use MongoDB\Laravel\Eloquent\Model as Eloquent;
+use MongoDB\Laravel\Query\Builder;
+
+/**
+ * @property string $name
+ * @property string $anniversary
+ * @mixin Eloquent
+ * @method static Builder create(...$values)
+ * @method static Builder truncate()
+ * @method static Eloquent sole(...$parameters)
```
GromNaN, line 18: These annotations are interesting, I think we should document this practice (something I will ask the doc team).

GromNaN (commented):
I added a minor comment, but the overall seems good to me. Thanks!

tests/Models/Anniversary.php
saineshmamgain: credit goes to @Treggats. I followed his `HiddenAnimal` model.

tests/DateTimeImmutableTest.php
saineshmamgain: Updated the PR

tests/Models/Anniversary.php
Treggats: Happy to help :)

Treggats: @GromNaN the thing with Laravel Models is that by default they have the type `Model`. Which makes sense, as the model extends that class, but then the IDE doesn't know about any of the methods, relations, properties, etc.

These docblocks provide the IDE with the context it should have. Note; the `@method` docblock can not be used if the method already exists in the class.

Personally I also use `@property` for the columns of the database table and the relation once fetched.
The `@property-read` docblock is used for the `Builder` return type of a relation.

A little gotcha is that sometimes you need a additional docblock on the variable. Like this.
```php
use Tests\Models\Anniversary;

/** @var Anniversary $anniversary */
$anniversary = Anniversary::query()->where('name', 'John')->first();
```
It then knows about the properties, relations, etc.

Hope this helps.

GromNaN: Thanks, I'm tracking this to work on it later PHPORM-316

GromNaN:
@saineshmamgain can you look at the test failure:
```
1) MongoDB\Laravel\Tests\Eloquent\DateTimeImmutableTest::testCanReturnCarbonImmutableObject
* Test code or tested code did not remove its own error handlers

* Test code or tested code did not remove its own exception handlers

/home/runner/work/laravel-mongodb/laravel-mongodb/tests/DateTimeImmutableTest.php:28
```

saineshmamgain:
@GromNaN actually this error comes when I do the reset `Date::useDefault();` in `tearDown` Method.

saineshmamgain:
@GromNaN My bad missed call to `parent::tearDown()` in my method. Updated the PR.

GromNaN (approved):

GromNaN:
Thank you @saineshmamgain

PR #3347, opened 2025-04-11
PHPORM-319 Make branch 4.x compatible with driver v2

GromNaN:
Fix PHPORM-319
Asked in https://github.com/mongodb/laravel-mongodb/pull/3343#issuecomment-2795692645

PR #3353, opened 2025-04-14
PHPORM-320 Fix aliasing .id field path to ._id

### LINKED ISSUE #3351: Nested ids remapped even if renameEmbeddedIdField is false
- Laravel-mongodb Version: 5.3
- PHP Version: 8.3
- Database Driver & Version: -

### Description:
I'm trying to query based on the `id` field of a nested array document. When attempting to query the `id` field is remapped to `_id` regardless of the `renameEmbeddedIdField` setting. 

If I modify my query to include another `where()` clause, the `id` field is not remapped to `_id`
`Model::query()->where('id', 1)->where('nested.id', 1)->get()` -> does not remap `nested.id`
`Model::query()->where('nested.id', 1)->get()` -> remaps `nested.id`

### Steps to reproduce
1. `'rename_embedded_id_field' => false` or `DB::connection('mongodb')->setRenameEmbeddedIdField(false);`
2. `Model::query()->where('nested.id', 1)->get();`

### Expected behaviour
`nested.id` should not be remapped to `nested._id` in the query because `renameEmbeddedIdField` is set to `false`

### Actual behaviour
In the query that is executed `nested.id` is remapped to `nested._id` in the query

GromNaN:
Fix PHPORM-320
Fix #3351

### Checklist

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -34,20 +34,24 @@ jobs:
                     - "11.*"
                     - "12.*"
                 driver:
-                    - 1
+                    - 2
                 include:
                     - php: "8.1"
                       laravel: "10.*"
                       mongodb: "5.0"
                       mode: "low-deps"
                       os: "ubuntu-latest"
-                      driver: 1.x
-                      driver_version: "1.21.0"
+                      driver: 1
+                    - php: "8.3"
+                      laravel: "11.*"
+                      mongodb: "8.0"
+                      os: "ubuntu-latest"
+                      driver: 1
                     - php: "8.4"
                       laravel: "12.*"
                       mongodb: "8.0"
                       os: "ubuntu-latest"
-                      driver: 2
+                      driver: 1
                 exclude:
                     - php: "8.1"
                       laravel: "11.*"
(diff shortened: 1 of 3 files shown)


.github/workflows/build-ci.yml
```diff
@@ -34,20 +34,19 @@ jobs:
                     - "11.*"
                     - "12.*"
                 driver:
-                    - 1
+                    - 2
```
GromNaN, line 37: I have to change the default driver version that is tested to match the ruleset from https://github.com/mongodb/laravel-mongodb/pull/3347

tests/Query/BuilderTest.php
```diff
@@ -1599,29 +1599,86 @@ public static function getEloquentMethodsNotSupported()
...
+        ];
+
+        yield 'nested $and in $or' => [
+            [
+                'find' => [
+                    [
+                        '$or' => [
+                            [
+                                '$and' => [
+                                    ['_id' => 10],
```
GromNaN, line 1655: There are flaws in the conversion if you use advanced queries. This transformation from `id` to `_id` is done in the `compileWheres` from the “column name”: https://github.com/mongodb/laravel-mongodb/blob/bdae18d817483d14d200735f00b3d9d43d75e3f4/src/Query/Builder.php#L1209-L1212

alcaeus (approved):

PR #3358, opened 2025-04-15
ManageTransactions compatible with Illuminate\DB

xmaydin:
### Checklist

### DIFF
--- a/src/Concerns/ManagesTransactions.php
+++ b/src/Concerns/ManagesTransactions.php
@@ -78,41 +78,57 @@ trait ManagesTransactions
     }
 
     /**
-     * Static transaction function realize the with_transaction functionality provided by MongoDB.
-     *
-     * @param  int $attempts
+     * @param Closure $callback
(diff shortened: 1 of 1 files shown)


alcaeus:
Thank you for your attempt at fixing this. However, we can't accept the pull request in this form for multiple reasons.

First of all, while we do need to change our method to be compatible with the overridden one, we want to do so with as little impact to users as possible. The PR here completely removes transaction options, which means that we're again breaking user code.
More importantly though, transactions in MongoDB are significantly different from those in other databases. When error occurs, the type of error determines the correct behaviour: either retry the commit (e.g. if a network error prevented us from receiving the result of the commit operation), retry the entire transaction (in case of a transient error that is expected to be resolved on the next attempt, e.g. a write conflict in a locked document), or abort the transaction entirely (e.g. because of a duplicate key error).

I'm closing this PR, but we're already working on a fix for the issue that ensures that downstream code continues to work.

PR #3361, opened 2025-04-16
Fix namespace of test classes

GromNaN:
Spotted with `composer dump-autoload --strict-psr --optimize`


alcaeus (approved):
Nice! Can we include this in our CI checks somehow?

PR #3366, opened 2025-04-20
[5.x] Replaced a too strict assertion in DatabaseSessionHandler.php that prevented to use the sessions at all with a more simple sanity check. Test already failed on that, now they pass.

tonivega:
Replaced a too strict assertion in DatabaseSessionHandler.php that prevented to use the sessions at all with a more simple sanity check. Test already failed on that, now they pass.

Fixed the next error : 
image

Now tests are passing : 
image

### Checklist

- [x] Add tests and ensure they pass -> There was a test covering this issue that failed, this pull request fixes it.


GromNaN:
It’s my bad. I added the setting so the assertions are always evaluated.

GromNaN (approved):

PR #3373, opened 2025-04-26
`QueryBuilder` multiply and divide support

faissaloux:
Hello 👋🏻 

In this PR I have added `multiply` and `divide` to `QueryBuilder` .

### Usage
```php
DB::table('users')->where('name', 'John Doe')->multiply('salary');    // "salary" * 1
DB::table('users')->where('name', 'John Doe')->multiply('salary', 2); // "salary" * 2
DB::table('users')->where('name', 'John Doe')->divide('salary');      // "salary" / 1
DB::table('users')->where('name', 'John Doe')->divide('salary', 2);   // "salary" / 2
```

### Checklist

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -890,7 +890,13 @@ class Builder extends BaseBuilder
         return $this->incrementEach($decrement, $extra, $options);
     }
 
-    /** @inheritdoc */
+    /**
+     * Multiply a column's value by a given amount.
+     *
+     * @param  string     $column
+     * @param  float|int  $amount
+     * @return int
+     */
     public function multiply($column, $amount = 1, array $extra = [], array $options = [])
     {
         $query = ['$mul' => [(string) $column => $amount]];
@@ -911,7 +917,13 @@ class Builder extends BaseBuilder
         return $this->performUpdate($query, $options);
     }
 
-    /** @inheritdoc */
+    /**
+     * Divide a column's value by a given amount.
+     *
+     * @param  string     $column
+     * @param  float|int  $amount
+     * @return int
+     */
     public function divide($column, $amount = 1, array $extra = [], array $options = [])
     {
         return $this->multiply($column, 1 / $amount, $extra, $options);
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -897,7 +897,7 @@ class Builder extends BaseBuilder
      * @param  float|int  $amount
      * @return int
      */
-    public function multiply($column, $amount = 1, array $extra = [], array $options = [])
+    public function multiply($column, $amount, array $extra = [], array $options = [])
     {
         $query = ['$mul' => [(string) $column => $amount]];
 
@@ -924,7 +924,7 @@ class Builder extends BaseBuilder
      * @param  float|int  $amount
      * @return int
      */
-    public function divide($column, $amount = 1, array $extra = [], array $options = [])
+    public function divide($column, $amount, array $extra = [], array $options = [])
     {
         return $this->multiply($column, 1 / $amount, $extra, $options);
     }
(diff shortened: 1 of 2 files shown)


src/Query/Builder.php
```diff
@@ -890,6 +890,33 @@ public function decrementEach(array $columns, array $extra = [], array $options
         return $this->incrementEach($decrement, $extra, $options);
     }
 
+    /** @inheritdoc */
+    public function multiply($column, $amount = 1, array $extra = [], array $options = [])
```
GromNaN: I don't find this method in the parent class. I only find [`Illuminate\Collections\Collection::multiply`](https://github.com/laravel/framework/blob/bcc92201ab6d786477fed578cc2e3ea4c4e1b093/src/Illuminate/Collections/Collection.php#L881). 

Could you remove the `@inheritdoc` annotation and add a description for all `@param` and `@return`.

GromNaN (changes requested):
Thanks for this nice feature that leverage MongoDB [update operator](https://www.mongodb.com/docs/manual/reference/operator/update/mul/).

Could you tell me a bit more about your use-case?

src/Query/Builder.php
faissaloux: Done 👌🏻

faissaloux:
> Could you tell me a bit more about your use-case?

#### Case of storing price in cents
We need to get the price converted from cents to USD.

```php
DB::table('products')->where('title', 'socks')->divide('price', 100);
```

#### Case of adding tax

```php
DB::table('products')->where('title', 'socks')->multiply('price', 1.2);
```

src/Query/Builder.php
```diff
@@ -890,6 +890,45 @@ public function decrementEach(array $columns, array $extra = [], array $options
...
     }
 
+    /**
+     * Multiply a column's value by a given amount.
+     *
+     * @param  string     $column
+     * @param  float|int  $amount
+     * @return int
+     */
+    public function multiply($column, $amount = 1, array $extra = [], array $options = [])
```
GromNaN: `$amount` should not have a default value. Multiplying by 1 is no-op, unless you have extra values.

tests/QueryBuilderTest.php
```diff
@@ -1052,6 +1052,55 @@ public function testIncrement()
...
+
+        DB::table('users')->multiply('salary');
+        $user = DB::table('users')->where('name', 'John Doe')->first();
+        $this->assertEquals(44000, $user->salary);
+        $user = DB::table('users')->where('name', 'Jane Doe')->first();
+        $this->assertEquals(640000, $user->salary);
+        $user = DB::table('users')->where('name', 'Robert Roe')->first();
+        $this->assertNull($user->salary);
+        $user = DB::table('users')->where('name', 'Mark Moe')->first();
+        $this->assertEquals(0, $user->salary);
```
GromNaN: When the field doesn't exist, this operation should not create it as it will always be `0`.

```suggestion
        $this->assertNull($user->salary);
```

GromNaN (changes requested):

src/Query/Builder.php
faissaloux: Done on https://github.com/mongodb/laravel-mongodb/pull/3373/commits/87b0d4818e2225f38c5d6a40cdcc210eafb977e9

tests/QueryBuilderTest.php
faissaloux: Done on https://github.com/mongodb/laravel-mongodb/pull/3373/commits/60c6487f440bac3890d185224f201112ef9a314d

GromNaN:
Thank you @faissaloux

GromNaN (approved):

PR #3374, opened 2025-04-29
PHPORM-330 Fix: Convert query duration time to milliseconds

AmirRezaM75:
Related Issue: https://github.com/laravel/telescope/issues/1587

This merge request addresses an issue in the `logQuery` method of Laravel, where the execution time was being passed in microseconds instead of the expected milliseconds.

https://github.com/laravel/framework/blob/797c2bbd6701d13650a24697ab62529dc3df9fb9/src/Illuminate/Database/Connection.php#L847-L860

https://github.com/laravel/framework/blob/797c2bbd6701d13650a24697ab62529dc3df9fb9/src/Illuminate/Database/Events/QueryExecuted.php#L21-L26

I verified that using `sleep` function in my query.

```php
public function create()
{
    $start = microtime(true);

    Post::whereRaw(['$where' => 'sleep(600) ||  true'])->first();

    $end = microtime(true);

    $diff = number_format(($end - $start) * 1000, 3);

    info("Execution time: $diff ms");

    return response()->noContent();
}
```

Before:

before

After:

after

> [2025-04-29 07:03:42] local.INFO: Execution time: 615.043 ms

### DIFF
--- a/src/CommandSubscriber.php
+++ b/src/CommandSubscriber.php
@@ -48,6 +48,6 @@ final class CommandSubscriber implements CommandSubscriberInterface
             }
         }
 
-        $this->connection->logQuery(Document::fromPHP($command)->toCanonicalExtendedJSON(), [], $event->getDurationMicros());
+        $this->connection->logQuery(Document::fromPHP($command)->toCanonicalExtendedJSON(), [], $event->getDurationMicros() / 1000);
     }
 }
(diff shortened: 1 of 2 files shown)


GromNaN (approved):
Good catch. Thanks for the investigation and the fix @AmirRezaM75.
I added a non-regression test.
Tracked as PHPORM-330

AmirRezaM75:
@GromNaN 
My pleasure

One of the CI failed, is it because of my change?

GromNaN:
Test failure is not related to this PR. It will be fixed by https://github.com/mongodb/laravel-mongodb/pull/3375

AmirRezaM75:
Hey @GromNaN 
Hope you are doing well.
when do you plan to tag it?

this change still not available in 5.4.3

https://github.com/mongodb/laravel-mongodb/blob/5.4.3/src/CommandSubscriber.php

PR #3382, opened 2025-05-15
Add schema validation before saving.

florianJacques:
I suggest introducing an automatic schema validation mechanism before saving data to the database, inspired by the schema validation system in [Mongoose (Node.js)](https://mongoosejs.com/).
This validation would be handled at the model or persistence layer using Laravel’s native validation system (Illuminate\Validation\Validator).

Goals
	•	Prevent invalid or incomplete data from being saved.
	•	Centralize validation rules close to the model or schema definition.
	•	Improve maintainability and ensure data integrity.

Example Implementation
```php
<?php

namespace App\Models;

use MongoDB\Laravel\Eloquent\SchemaValidation;

class Planet extends Model
{
    use SchemaValidation;

    public function schemaRules(): array
    {
        return [
            'name' => 'required|string',
            'distance' => 'integer|between:10,65535',
        ];
    }
}

```
```php
<?php

$planet = \App\Models\Planet::query()->first();
$planet->name = 'Earth';
$planet->withValidation()->save();
```

### Checklist

### DIFF
--- a/src/Eloquent/SchemaValidation.php
+++ b/src/Eloquent/SchemaValidation.php
@@ -1,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Laravel\Eloquent;
+
+use Illuminate\Support\Facades\Validator;
+use MongoDB\Laravel\Exception\SchemaValidationException;
+
+trait SchemaValidation
+{
+    protected bool $validateBeforeSave = false;
+
+    /**
+     * @param bool $validateBeforeSave
+     * @return $this
+     */
+    public function markForValidation(bool $validateBeforeSave = true): static
+    {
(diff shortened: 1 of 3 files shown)


GromNaN:
Hello @florianJacques, thanks for bringing this idea.

Does this feature require to be part of the `laravel-mongodb` package, or is it something that could be provided by a third party package? On a quick search, I found [watson/validating](https://github.com/dwightwatson/validating) that is quite popular and seems to be very similar to what you describe.

florianJacques:
Indeed, I didn't know this package. Thank you for taking the time to look at my proposal.

GromNaN:
Let me know if you have any compatibility issue with MongoDB.

PR #3394, opened 2025-05-28
PHPORM-326 Add collation to `db:table` command

paulinevos:
Fix PHPORM-326

Adds `collation` to the `db:table` command. Actually also adds it everywhere else in the command layer (i.e. in `ShowCommand`, both for views and collections), but those are not displayed when you run `db:show`.

### Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
     },
     "authors": [
         { "name": "Andreas Braun", "email": "andreas.braun@mongodb.com", "role": "Leader" },
+        { "name": "Pauline Vos", "email": "pauline.vos@mongodb.com", "role": "Maintainer" },
         { "name": "Jérôme Tamarelle", "email": "jerome.tamarelle@mongodb.com", "role": "Maintainer" },
         { "name": "Jeremy Mikola", "email": "jmikola@gmail.com", "role": "Maintainer" },
         { "name": "Jens Segers", "homepage": "https://jenssegers.com", "role": "Creator" }
--- a/src/Schema/Builder.php
+++ b/src/Schema/Builder.php
@@ -15,6 +15,7 @@ use MongoDB\Model\IndexInfo;
 use function array_column;
 use function array_fill_keys;
 use function array_filter;
+use function array_key_exists;
 use function array_keys;
 use function array_map;
 use function array_merge;
@@ -26,6 +27,7 @@ use function explode;
 use function implode;
 use function in_array;
 use function is_array;
+use function is_bool;
 use function is_string;
 use function iterator_to_array;
 use function sort;
@@ -380,6 +382,9 @@ class Builder extends \Illuminate\Database\Schema\Builder
                 continue;
             }
 
+            $options = $collectionInfo->getOptions();
+            $collation = $options['collation'] ?? [];
+
             // Aggregation is not supported on views
             $stats = $collectionType !== 'view' ? $db->selectCollection($collectionName)->aggregate([
                 ['$collStats' => ['storageStats' => ['scale' => 1]]],
@@ -392,7 +397,7 @@ class Builder extends \Illuminate\Database\Schema\Builder
                 'schema_qualified_name' => $db->getDatabaseName() . '.' . $collectionName,
                 'size' => $stats[0]?->storageStats?->totalSize ?? null,
                 'comment' => null,
-                'collation' => null,
+                'collation' => $this->collationToString($collation),
                 'engine' => null,
             ];
         }
@@ -400,5 +405,31 @@ class Builder extends \Illuminate\Database\Schema\Builder
         usort($collections, fn ($a, $b) => $a['name'] <=> $b['name']);
 
         return $collections;
+    }
+
+    private function collationToString(array $collation): string
+    {
+        $map = [
+            'locale' => 'l',
+            'strength' => 's',
+            'caseLevel' => 'cl',
+            'caseFirst' => 'cf',
+            'numericOrdering' => 'no',
+            'alternate' => 'a',
+            'maxVariable' => 'mv',
+            'normalization' => 'n',
+            'backwards' => 'b',
+        ];
+
+        $parts = [];
+        foreach ($collation as $key => $value) {
+            if (array_key_exists($key, $map)) {
+                $shortKey = $map[$key];
+                $shortValue = is_bool($value) ? ($value ? '1' : '0') : $value;
+                $parts[] = $shortKey . '=' . $shortValue;
+            }
+        }
+
+        return implode(';', $parts);
     }
 }
--- a/src/Schema/Builder.php
+++ b/src/Schema/Builder.php
@@ -6,6 +6,7 @@ namespace MongoDB\Laravel\Schema;
 
 use Closure;
 use MongoDB\Collection;
+use MongoDB\Database;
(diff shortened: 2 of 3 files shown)


composer.json
```diff
@@ -18,7 +18,8 @@
         { "name": "Andreas Braun", "email": "andreas.braun@mongodb.com", "role": "Leader" },
         { "name": "Jérôme Tamarelle", "email": "jerome.tamarelle@mongodb.com", "role": "Maintainer" },
         { "name": "Jeremy Mikola", "email": "jmikola@gmail.com", "role": "Maintainer" },
-        { "name": "Jens Segers", "homepage": "https://jenssegers.com", "role": "Creator" }
+        { "name": "Jens Segers", "homepage": "https://jenssegers.com", "role": "Creator" },
+        { "name": "Pauline Vos", "email": "pauline.vos@mongodb.com", "role": "Maintainer" }
```
paulinevos: Wasn't sure if this was alphabetical order or by role 🤔

tests/SchemaTest.php
```diff
@@ -20,12 +20,15 @@
...
 {
+    private const COLL_WITH_COLLATION = 'collection_with_collation';
+
     public function tearDown(): void
     {
         $database = $this->getConnection('mongodb')->getDatabase();
         assert($database instanceof Database);
         $database->dropCollection('newcollection');
         $database->dropCollection('newcollection_two');
+        $database->dropCollection(self::COLL_WITH_COLLATION);
```
GromNaN, line 33: I'm not against the idea of using a constant, but it's inconsistent with other collection names. Can you harmonize?

composer.json
GromNaN: I'd say in order of importance, put your name after Andreas 😉

src/Schema/Builder.php
```diff
@@ -154,31 +157,36 @@ public function dropAllTables()
     }
 
     /** @param string|null $schema Database name */
-    public function getTables($schema = null)
-    {
+    private function getCollectionRows(
```
GromNaN: Generally, private methods are placed after public ones (this is enforced in the library, but not here).

src/Schema/Builder.php
```diff
@@ -188,34 +196,49 @@ public function getTables($schema = null)
...
-        usort($collections, fn ($a, $b) => $a['name'] <=> $b['name']);
+        return implode(';', $parts);
+    }
 
-        return $collections;
+    /** @param string|null $schema Database name */
+    public function getTables($schema = null)
+    {
+        $aggregateStats = function (Database $db, string $collectionName): array {
+            return $db->selectCollection($collectionName)->aggregate([
```
GromNaN: We renamed this method to `getCollection`: https://github.com/mongodb/mongo-php-library/pull/1494
```suggestion
            return $db->getCollection($collectionName)->aggregate([
```

src/Schema/Builder.php
```diff
@@ -154,31 +157,36 @@ public function dropAllTables()
     }
 
     /** @param string|null $schema Database name */
-    public function getTables($schema = null)
-    {
+    private function getCollectionRows(
+        $schema = null,
+        ?string $collectionType = null,
```
GromNaN: `$collectionType` is always set, you don't need to make it nullable.

```suggestion
        string $collectionType,
```

src/Schema/Builder.php
```diff
@@ -154,31 +157,36 @@ public function dropAllTables()
...
-                ['$collStats' => ['storageStats' => ['scale' => 1]]],
-                ['$project' => ['storageStats.totalSize' => 1]],
-            ])->toArray();
+            $options = $collectionInfo->getOptions();
+            $collation = $options['collation'] ?? [];
+
+            // Aggregation is not supported on views
+            $shouldAggregateStats = $aggregateStats !== null && $collectionInfo->getType() !== 'view';
+
+            $stats = $shouldAggregateStats ? $aggregateStats($db, $collectionName) : null;
```
GromNaN: If `$aggregateStats` is set only when `$collectionType == 'collection'`, this condition is the same as `$aggregateStats !== null`:

```suggestion
            $stats = $aggregateStats ? $aggregateStats($db, $collectionName) : null;
```

src/Schema/Builder.php
```diff
@@ -154,31 +157,36 @@ public function dropAllTables()
...
+        ?callable $aggregateStats = null,
+    ) {
         $db = $this->connection->getDatabase($schema);
         $collections = [];
 
         foreach ($db->listCollections() as $collectionInfo) {
             $collectionName = $collectionInfo->getName();
 
-            // Skip views, which don't support aggregate
-            if ($collectionInfo->getType() === 'view') {
```
GromNaN, line 166: From memory, we have other types than `collection` and `view`. Timeseries maybe https://github.com/mongodb/laravel-mongodb/pull/3327#discussion_r2010890697

GromNaN (changes requested):
Great start!

src/Schema/Builder.php
paulinevos: True, although I was thinking it might be nice if we ever want to get everything. Bit premature probably.

src/Schema/Builder.php
paulinevos: Yes, but then it expects this method to know about how it's called. If we do change it I think I'd prefer not to pass the callable at all and just conditionally aggregate based on the passed collection type (essentially like what was done before).

I think that probably makes sense. The reason I passed the optional callable is to avoid having conditionals in a generic-ish method but I guess we need the conditional one way or another 😅

GromNaN: Removing the callable would make the code more straightforward.

src/Schema/Builder.php
paulinevos: Yes, what I could find was those three. I had a hard time finding comprehensive documentation detailing which collection types there are and which of them support aggregation, but from the discussion it looks like only `view` doesn't support it? Might be good to suggest doc improvement re: this

docs/quick-start/view-data.txt
rustagir, line 1: Not sure why these changes are showing up in this PR when they should already be present in 5.x per #3379

rustagir (approved):

docs/quick-start/view-data.txt
paulinevos: Branch needs to be rebased probably :)

composer.json
```diff
@@ -16,9 +16,10 @@
     },
     "authors": [
         { "name": "Andreas Braun", "email": "andreas.braun@mongodb.com", "role": "Leader" },
+        { "name": "Pauline Vos", "email": "pauline.vos@mongodb.com", "role": "Maintainer" },
         { "name": "Jérôme Tamarelle", "email": "jerome.tamarelle@mongodb.com", "role": "Maintainer" },
         { "name": "Jeremy Mikola", "email": "jmikola@gmail.com", "role": "Maintainer" },
-        { "name": "Jens Segers", "homepage": "https://jenssegers.com", "role": "Creator" }
+        { "name": "Jens Segers", "homepage": "https://jenssegers.com", "role": "Creator" },
```
GromNaN: I've received an email from packagist for this typo.
For the next PR, you should work from your own fork.

GromNaN (approved):
LGTM

PR #3395, opened 2025-06-02
Add conditional return type to `Builder::raw()`

paulinevos:
for both the Eloquent and the Query builder, so that developers can understand that the return type will either be a `Collection` or an `Expression` based on the argument that's passed to `raw()`

### Checklist

- [ ] ~~Add tests and ensure they pass~~ (not applicable)

### DIFF
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -4,10 +4,12 @@ declare(strict_types=1);
 
 namespace MongoDB\Laravel\Eloquent;
 
+use Closure;
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
 use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Database\Eloquent\Model;
 use MongoDB\BSON\Document;
+use MongoDB\Builder\Expression;
 use MongoDB\Builder\Type\QueryInterface;
 use MongoDB\Builder\Type\SearchOperatorInterface;
 use MongoDB\Driver\CursorInterface;
@@ -229,7 +231,13 @@ class Builder extends EloquentBuilder
         return parent::decrement($column, $amount, $extra);
     }
 
-    /** @inheritdoc */
+    /**
+     * @param (Closure():T)|Expression|null $value
+     *
+     * @return ($value is Closure ? T : ($value is null ? Collection : Expression))
+     *
+     * @template T
+     */
     public function raw($value = null)
     {
         // Get raw results from the query builder.
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -964,7 +964,13 @@ class Builder extends BaseBuilder
         return $this->pluck($column, $key);
     }
 
-    /** @inheritdoc */
+    /**
+     * @param (Closure():T)|Expression|null $value
+     *
+     * @return ($value is Closure ? T : ($value is null ? Collection : Expression))
+     *
+     * @template T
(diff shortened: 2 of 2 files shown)


src/Eloquent/Builder.php
```diff
@@ -229,7 +230,11 @@ public function decrement($column, $amount = 1, array $extra = [])
         return parent::decrement($column, $amount, $extra);
     }
 
-    /** @inheritdoc */
+    /**
+     * @return ($value is callable ? Collection : ($value is null ? Collection : Expression))
+     *
+     * @inheritdoc
```
paulinevos: Was unsure if I should leave the `@inheritdoc` because it might override the conditional return type?

alcaeus: According to the [`@inheritdoc` documentation](https://docs.phpdoc.org/guide/guides/inheritance.html#methods):
> [...] each of the above will only be inherited if the redefined method's DocBlock does not have the element that is to be inherited. So, for example, if the DocBlock of the redefined method has a summary then it will not receive the overridden method's summary.

With that in mind, I'd say it's fine to use `@inheritdoc` to inherit the description and parameter type definitions, and use `@return` for the more specific return type.

src/Eloquent/Builder.php
```diff
@@ -229,7 +230,11 @@ public function decrement($column, $amount = 1, array $extra = [])
         return parent::decrement($column, $amount, $extra);
     }
 
-    /** @inheritdoc */
+    /**
+     * @return ($value is callable ? Collection : ($value is null ? Collection : Expression))
```
alcaeus: If I understand correctly, this should be `Closure`, not `callable` as the check in `Query::raw` only checks for closures.

src/Query/Builder.php
```diff
@@ -964,7 +964,11 @@ public function lists($column, $key = null)
         return $this->pluck($column, $key);
     }
 
-    /** @inheritdoc */
+    /**
+     * @return ($value is callable ? Collection : ($value is null ? Collection : Expression))
```
alcaeus: Same as above with `Closure` vs. `callable`. As a second question, since the `$value` type is only defined as `mixed`, can we rely on every Closure `$value` to return a `Collection` instance? If not, we could improve the types significantly by using `@template` to indicate that we're returning the return type of the closure:
```php
/**
 * @template T
 * @param (Closure():T)|Expression|null $value
 * @return ($value is callable ? T : ($value is null ? Collection : Expression))
 */
```

Note that I haven't tested this at all.

src/Eloquent/Builder.php
```diff
@@ -229,7 +230,11 @@ public function decrement($column, $amount = 1, array $extra = [])
...
+     *
+     * @inheritdoc
+     * */
```
alcaeus:
```suggestion
     */
```

src/Query/Builder.php
```diff
@@ -964,7 +964,11 @@ public function lists($column, $key = null)
...
+     *
+     * @inheritdoc
+     * */
```
alcaeus:
```suggestion
     */
```

alcaeus (changes requested):

src/Query/Builder.php
paulinevos: Agreed, I guess they could return god knows what with their own closure.

alcaeus (approved):

PR #3397, opened 2025-06-03
Support adding schema validation

paulinevos:
To support the `'$jsonSchema'` operation on collections.
Closes PHPORM-72.

### Checklist

### DIFF
--- a/src/Schema/Blueprint.php
+++ b/src/Schema/Blueprint.php
@@ -9,6 +9,7 @@ use MongoDB\Collection;
 use MongoDB\Laravel\Connection;
 
 use function array_flip;
+use function array_merge;
 use function implode;
 use function in_array;
 use function is_array;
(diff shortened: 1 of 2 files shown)


alcaeus (approved):
LGTM. I'll note that this uses the default settings for [`validationAction`](https://www.mongodb.com/docs/manual/core/schema-validation/handle-invalid-documents/#std-label-schema-validation-handle-invalid-docs) and [`validationLevel`](https://www.mongodb.com/docs/manual/core/schema-validation/specify-validation-level/), which causes an error when attempting to insert or update an invalid document.

We could allow users to supply their own `$validationLevel` and `$validationAction` to customise this behaviour, but I'm fine with adding this in a separate step.

paulinevos:
@alcaeus Oh no that seems like a pretty crucial step and arbitrary to add so will do

alcaeus (approved):

PR #3399, opened 2025-06-04
Bump PHP to 8.2 in Docker test container

paulinevos:
### Checklist

- [ ] ~~Add tests and ensure they pass~~
- [x] Ran test suite with new container image and ensured it passed


GromNaN:
This PR doesn't have any change. Should it be closed?

paulinevos:
@GromNaN No, I reverted the change to check if it was my change that screwed up the build (it's not). Seems to be an issue with the runner somehow but I can't quite figure it out. I'll have another look some time today.

docker-compose.yml
```diff
@@ -14,7 +14,7 @@ services:
 
     mongodb:
         container_name: mongodb
-        image: mongodb/mongodb-atlas-local:latest
+        image: mongodb/mongodb-atlas-local:8.0.8
```
GromNaN: We can safely use the latest of version 8 instead of locking to this fix version.

```suggestion
        image: mongodb/mongodb-atlas-local:8
```

GromNaN (approved):

PR #3400, opened 2025-06-05
DOCSP-50472: schema validation

rustagir:
https://jira.mongodb.org/browse/DOCSP-50472

Adds documentation for the `jsonSchema()` method.

Staging: https://deploy-preview-193--docs-laravel.netlify.app/eloquent-models/schema-builder/#implement-schema-validation

### Checklist


docs/eloquent-models/schema-builder.txt
```diff
@@ -117,6 +112,63 @@ To learn more about Laravel migrations, see
...
+
+Implement Schema Validation
+---------------------------
+
+You can use the ``jsonSchema()`` method to implement :manual:`schema
+validation </core/schema-validation/>` when using the following schema
+builder methods:
+
+- ``Schema::create()``: When creating a new collection
+- ``Schema::table()``: When updating collection properties
```
mongoKart, line 125: Q: can you only implement schema validation when using these methods for these specific purposes? or is the info after the colons just a reminder of the method's purpose?

docs/eloquent-models/schema-builder.txt
```diff
@@ -117,6 +112,63 @@ To learn more about Laravel migrations, see
...
+
+After you implement schema validation, the server allows you to run only
+those write operations which follow the validation rules. Use schema
```
mongoKart: [that/which](https://www.mongodb.com/docs/meta/style-guide/writing/use-that-which-correctly/)

```suggestion
those write operations that follow the validation rules. Use schema
```

docs/eloquent-models/schema-builder.txt
```diff
@@ -117,6 +112,63 @@ To learn more about Laravel migrations, see
...
+validation to restrict data types and value ranges of document fields in
+a specified collection.
```
mongoKart: S: i wonder if the connection between these two sentences would be clearer if you switched them. something like:

```suggestion
You can use schema validation to restrict data types and value ranges of document fields in
a specified collection. After you implement schema validation, the server allows you to run only
those write operations which follow the validation rules. 
```

docs/eloquent-models/schema-builder.txt
```diff
@@ -117,6 +112,63 @@ To learn more about Laravel migrations, see
...
+
+Before creating a collection with schema validation rules, you must
+define a JSON schema.
+
+The schema is a JSON object that contains key-value pairs
+specifying the validation rules for your collection. At the top level,
+this object must include a ``$jsonSchema`` object. The ``$jsonSchema``
+object can include the following fields:
```
mongoKart: S: break up paragraphs differently. could also then remove first 'JSON'

```suggestion
Before creating a collection with schema validation rules, you must
define the schema.
A schema is a JSON object that contains key-value pairs
specifying the validation rules for your collection.
At the top level, this object must include a ``$jsonSchema`` object. The ``$jsonSchema``
object can include the following fields:
```

docs/eloquent-models/schema-builder.txt
```diff
@@ -117,6 +112,63 @@ To learn more about Laravel migrations, see
...
+
+- ``title``: Sets an optional title for the schema.
+- ``required``: Specifies a list of required fields for each document in your collection.
+- ``properties``: Sets property requirements for individual fields.
```
mongoKart: I: no periods

```suggestion
- ``title``: Sets an optional title for the schema
- ``required``: Specifies a list of required fields for each document in your collection
- ``properties``: Sets property requirements for individual fields
```

docs/eloquent-models/schema-builder.txt
```diff
@@ -117,6 +112,63 @@ To learn more about Laravel migrations, see
...
+- ``validationLevel``: Sets the level of validation enforcement.
+  Accepted values are ``"strict"`` (default) and ``"moderate"``.
+- ``validationAction``: Specifies the action to take when invalid
+  operations are attempted. Accepted values are ``"error"`` (default) and
+  ``"warn"``.
+
+This example demonstrates how to pass a JSON schema to the
+``jsonSchema()`` method when creating a collection. The schema
+validation specifies that documents in the ``pilots`` collection must
+contain the ``license_number`` field with an integer value between
```
mongoKart: S:

```suggestion
contain the ``license_number`` field, and the field must have an integer value between
```

mongoKart (approved):
some small things then LGTM

docs/eloquent-models/schema-builder.txt
paulinevos: These are the only two methods in Laravel's `Schema` that accept a callback to alter columns, so you can only implement schema validation on either creating a collection or updating it.

docs/eloquent-models/schema-builder.txt
```diff
@@ -117,6 +112,63 @@ To learn more about Laravel migrations, see
...
+those write operations which follow the validation rules. Use schema
+validation to restrict data types and value ranges of document fields in
+a specified collection.
+
+Before creating a collection with schema validation rules, you must
+define a JSON schema.
+
+The schema is a JSON object that contains key-value pairs
+specifying the validation rules for your collection. At the top level,
+this object must include a ``$jsonSchema`` object. The ``$jsonSchema``
```
paulinevos: `$jsonSchema` doesn't need to be included in the schema argument (see the example). Since the method is `Blueprint::jsonSchema`, the `'$jsonSchema'` operation is already implied. 

So this documentation may be valid for MongoDB server but not for Laravel usage

docs/eloquent-models/schema-builder.txt
```diff
@@ -117,6 +112,63 @@ To learn more about Laravel migrations, see
...
+object can include the following fields:
+
+- ``title``: Sets an optional title for the schema.
```
paulinevos: Should we actually include this documentation? Seems to me the `$jsonSchema` operation is covered by the Mongo server docs. The Laravel ORM passes the schema through as is. I think specifying allowed fields here puts us in a situation of having to update this documentation when Mongo server specs change, when the user could just use the Mongo server docs as the single source of truth for jsonSchema usage.

paulinevos (changes requested):

docs/eloquent-models/schema-builder.txt
paulinevos: Oh also! Since this is PHP, the schema argument is not an object, but an associative array.

docs/eloquent-models/schema-builder.txt
rustagir: makes sense

docs/eloquent-models/schema-builder.txt
rustagir: changed to a list

docs/eloquent-models/schema-builder.txt
rustagir: That makes sense - I'll cut out some of this text and apply the suggestions from your prev comment.

docs/eloquent-models/schema-builder.txt
rustagir: Removed this paragraph per PV suggestion

docs/eloquent-models/schema-builder.txt
rustagir: removed this text

paulinevos (approved):
Great! 🙏

norareidy (approved):

PR #3404, opened 2025-06-12
Improve error handling on unsupported hybrid queries

paulinevos:
Hybrid belongs-to-many relationships are not supported for query constraints. However, the support check was done downstream of a bunch of Eloquent stuff, resulting in the user getting an exception that didn't tell them anything about the usage being unsupported.

This moves that check further up the chain so that the user is alerted to the lack of support before we do anything else.

### Checklist

### DIFF
--- a/src/Helpers/QueriesRelationships.php
+++ b/src/Helpers/QueriesRelationships.php
@@ -12,6 +12,7 @@ use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 use Illuminate\Database\Eloquent\Relations\HasOneOrMany;
 use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Support\Collection;
+use LogicException;
 use MongoDB\Laravel\Eloquent\Model;
 use MongoDB\Laravel\Relations\MorphToMany;
 
@@ -104,6 +105,8 @@ trait QueriesRelationships
      */
     public function addHybridHas(Relation $relation, $operator = '>=', $count = 1, $boolean = 'and', ?Closure $callback = null)
     {
+        $this->assertHybridRelationSupported($relation);
+
         $hasQuery = $relation->getQuery();
         if ($callback) {
             $hasQuery->callScope($callback);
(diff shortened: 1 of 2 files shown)


src/Helpers/QueriesRelationships.php
```diff
@@ -128,6 +130,26 @@ public function addHybridHas(Relation $relation, $operator = '>=', $count = 1, $
...
+    {
+        if (
+            $relation instanceof HasOneOrMany
+            || $relation instanceof BelongsTo
+            || ($relation instanceof BelongsToMany && ! $this->isAcrossConnections($relation))
+        ) {
+            return;
+        }
+
+        throw new Exception(class_basename($relation) . ' is not supported for hybrid query constraints.');
```
GromNaN: You can be more specific, this is a logic exception (developer fault).
```suggestion
        throw new \LogicException(class_basename($relation) . ' is not supported for hybrid query constraints.');
```

src/Helpers/QueriesRelationships.php
```diff
@@ -128,6 +130,26 @@ public function addHybridHas(Relation $relation, $operator = '>=', $count = 1, $
...
     }
 
+    /**
+     * @param Relation $relation
+     *
+     * @return void
+     *
+     * @throws Exception
+     */
+    public function assertHybridRelationSupported(Relation $relation): void
```
GromNaN: If this method don't need to be public, make it private so that we don't have to maintain backward compatibility on it.

```suggestion
    private function assertHybridRelationSupported(Relation $relation): void
```

In PHPLIB we enforce private methods to be at after the public ones, I think that would be a good idea for this method to move it at the end.

src/Helpers/QueriesRelationships.php
```diff
@@ -221,7 +245,7 @@ protected function getRelatedConstraintKey(Relation $relation)
             return $relation->getForeignKeyName();
         }
 
-        if ($relation instanceof BelongsToMany && ! $this->isAcrossConnections($relation)) {
```
GromNaN, line 224: Is this a behavior change? The value is now returned even if this is a cross-connection relationship.

GromNaN (changes requested):

src/Helpers/QueriesRelationships.php
paulinevos: Agree, I made it public cause I thought the other usage was outside of the class (it's not)

src/Helpers/QueriesRelationships.php
paulinevos: It's not, because the assertion added before already checks for this. If that's not explicit enough I can add it back, just didn't want to duplicate the logic but both approaches are valid.

GromNaN: Perfect. No need to duplicate the logic more than necessary.

GromNaN (approved):

PR #3408, opened 2025-06-16
PHPORM-264: Deprecate Mongo soft deletes trait

paulinevos:
In favor of the Laravel one. Ours is now obsolete; the SoftDeletes trait is only necessary to remove the call to qualifyColumn in the parent trait. But the DocumentModel::qualifyColumn is already disabled.

Also replaces some internal/test usages with the recommended trait.

Closes PHPORM-264 

### Checklist

- [x] Ensure tests still pass

### DIFF
--- a/src/Eloquent/SoftDeletes.php
+++ b/src/Eloquent/SoftDeletes.php
@@ -4,7 +4,14 @@ declare(strict_types=1);
 
 namespace MongoDB\Laravel\Eloquent;
 
-/** @deprecated 6.0.0 in favor of \Illuminate\Database\Eloquent\SoftDeletes */
+use function sprintf;
+use function trigger_error;
+
+use const E_USER_DEPRECATED;
+
+trigger_error(sprintf('Since mongodb/laravel-mongodb:5.5, trait "%s" is deprecated, use "%s" instead.', SoftDeletes::class, \Illuminate\Database\Eloquent\SoftDeletes::class), E_USER_DEPRECATED);
+
+/** @deprecated since mongodb/laravel-mongodb:5.5, use \Illuminate\Database\Eloquent\SoftDeletes instead */
 trait SoftDeletes
 {
     use \Illuminate\Database\Eloquent\SoftDeletes;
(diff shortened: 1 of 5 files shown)


rustagir (approved):
changes to docs tests lgtm

src/Eloquent/SoftDeletes.php
```diff
@@ -4,6 +4,7 @@
 
 namespace MongoDB\Laravel\Eloquent;
 
+/** @deprecated 6.0.0 in favor of \Illuminate\Database\Eloquent\SoftDeletes */
```
GromNaN: We can trigger a deprecation when the PHP file is loaded.

In the `@deprecated` annotation, we usually give the version of the deprecation. 

```suggestion
trigger_error(sprintf('Since mongodb/laravel-mongodb:5.5, trait "%s" is deprecated, use "%s" instead.', SoftDeletes::class, \Illuminate\Database\Eloquent\SoftDeletes::class), E_USER_DEPRECATED);

/** @deprecated since mongodb/laravel-mongodb:5.5, use \Illuminate\Database\Eloquent\SoftDeletes instead */
```

Example of an other deprecation I put in the library: https://github.com/mongodb/laravel-mongodb/blob/b02b40c120be6dee9ebf63ff0b7bc42464c8fce4/src/Connection.php#L126-L132

An other example of deprecating a class in Symfony: https://github.com/symfony/symfony/blob/138109a63e2202674d81f31481da7c6defe4c0ce/src/Symfony/Component/Serializer/CacheWarmer/CompiledClassMetadataCacheWarmer.php#L19

GromNaN (changes requested):

paulinevos:
Sorry this took me so long @GromNaN. Question, why do we trigger an `error` instead of a deprecation here?

GromNaN:
> Question, why do we trigger an `error` instead of a deprecation here?

[`trigger_deprecation`](https://github.com/symfony/symfony/blob/c50f24712a0a8ec2b1e9e8217185fa0742810167/src/Symfony/Contracts/Deprecation/function.php#L23) is a Symfony function to call `trigger_error` with the `E_USER_DEPRECATED` level. "deprecated" is just one level among others such as "error" and "warning". I don't know of any other way to do it.

GromNaN (approved):

paulinevos:
Gotcha, thank you!

GromNaN:
The CI is stuck because of the github-actions[bot] commit. You can rebase/squash an push.

PR #3410, opened 2025-06-17
[Fix] embed relations serialization

florianJacques:
Fixed the serialization of embed relations.
These relationships were not necessarily loaded, and could not use their attribute mutators.

### Checklist

### DIFF
--- a/src/Eloquent/EmbedsRelations.php
+++ b/src/Eloquent/EmbedsRelations.php
@@ -8,6 +8,7 @@ use Illuminate\Support\Str;
 use MongoDB\Laravel\Relations\EmbedsMany;
 use MongoDB\Laravel\Relations\EmbedsOne;
 
+use MongoDB\Laravel\Relations\EmbedsOneOrMany;
 use function class_basename;
 use function debug_backtrace;
 
@@ -84,5 +85,35 @@ trait EmbedsRelations
         $instance = new $related();
 
         return new EmbedsOne($query, $this, $instance, $localKey, $foreignKey, $relation);
+    }
+
+    /**
+     * Determine if the given key is an embed relationship method on the model.
+     *
+     * @param  string  $key
+     * @return bool
+     */
+    public function isEmbedRelation($key)
+    {
+        return $this->isRelation($key)
+            && is_a($this->{$key}(), EmbedsOneOrMany::class, true);
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function toArray()
+    {
+        $embeds = [];
+
+        foreach (array_keys($this->getAttributes()) as $key) {
+            if ($this->isEmbedRelation($key)) {
+                $embeds[] = $key;
+            }
+        }
+
+        $this->loadMissing($embeds);
+
+        return parent::toArray();
     }
 }
--- a/src/Eloquent/EmbedsRelations.php
+++ b/src/Eloquent/EmbedsRelations.php
@@ -4,13 +4,11 @@ declare(strict_types=1);
 
 namespace MongoDB\Laravel\Eloquent;
 
-use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Str;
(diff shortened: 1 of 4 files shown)


florianJacques:
Hello,
Why [Coding Standards / phpcs (pull_request)](https://github.com/mongodb/laravel-mongodb/actions/runs/15702940104/job/44241933447?pr=3410)
Fail ?

GromNaN:
In order to review the PR properly, could you add some tests that are fixed by the change?

> Why Coding Standards / phpcs (pull_request) Fail?

GitHub Action fails to commit the fixes that are found by phpcs.

```
----------------------------------------------------------------------
FILE                                                  FIXED  REMAINING
----------------------------------------------------------------------
src/Eloquent/EmbedsRelations.php                      2      0
src/Eloquent/DocumentModel.php                        1      0
----------------------------------------------------------------------
A TOTAL OF 3 ERRORS WERE FIXED IN 2 FILES
----------------------------------------------------------------------
```

florianJacques:
Yes of course, I will write the tests in the next few days
For CI errors, can I do something?

GromNaN:
> For CI errors, can I do something?

You can run `vendor/bin/phpcbf` locally to fix the CS issues.

florianJacques:
Thanks, Done! I'll write the tests when I have some time...

florianJacques:
Adding test ok

tests/EmbeddedRelationsTest.php
```diff
@@ -972,4 +974,18 @@ public function testUnsetPropertyOnEmbed()
...
         $this->assertSame('Kyoto', $user->addresses->get(1)->city);
     }
+
+    public function testEmbedManyToArrayCast()
+    {
+        $user = UserWithEmbeds::create(['name' => 'John Doe']);
+        $user->addresses()->saveMany([new Address(['city' => 'London'])]);
+
+        //Reload document
+        $user = UserWithEmbeds::where('name', 'John Doe')->first();
```
GromNaN, line 984: In order to get the `address` property correctly casted, you must call `with('address')`:
```suggestion
        $user = UserWithEmbeds::where('name', 'John Doe')->with('addresses')->first();
```

tests/Models/UserWithEmbeds.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Laravel\Tests\Models;
+
+class UserWithEmbeds extends User
+{
+    protected $withEmbeds = ['addresses'];
```
GromNaN, line 9: Having this new configuration to "fix" the bugged behavior of the embed relationship seems wrong.

The default `$with` property can be leveraged:

```suggestion
    protected $with = ['addresses'];
```

src/Eloquent/EmbedsRelations.php
```diff
@@ -85,4 +101,17 @@ protected function embedsOne($related, $localKey = null, $foreignKey = null, $re
...
     }
+
+    /**
+     * Determine if the given key is an embed relationship method on the model.
+     *
+     * @param string $key
+     *
+     * @return bool
+     */
+    public function isEmbeddedRelation($key)
```
GromNaN, line 112: This new method is not used and not tested. It doesn't seems to be part of the fix.

GromNaN (changes requested):
It seems to me that this is more of a documentation issue that something we should fix in the code.

As for any relationship, if you want to load the data from an embedded relationship you have to use the `Model::$with` property or call `Builder::with()`.

GromNaN:
Thank you for the contribution and for adding tests.

I verified that using the standard `$with` property achieves exactly the same result: replacing `$withEmbeds` with `$with` in the test model makes the same test pass unchanged. Embedded relations are standard Eloquent relations and load correctly via the built-in eager loading mechanism, without needing a dedicated property.

Closing in favour of documenting the correct use of `$with` with embedded relations.

PR #3412, opened 2025-06-18
PHPORM-146: Add override attribute everywhere

paulinevos:
Closes PHPORM-146

Adds the override attribute in every class that extends a Laravel base class, so that PHP will show errors if ever the base methods are renamed or removed.

Also adds in a few corrections to phpdoc declarations where necessary.

### Checklist

- [ ] ~~Add tests and ensure they pass~~ (not applicable)

### DIFF
--- a/src/Bus/MongoBatchRepository.php
+++ b/src/Bus/MongoBatchRepository.php
@@ -216,6 +216,7 @@ class MongoBatchRepository extends DatabaseBatchRepository implements PrunableBa
     }
 
     /** Prune all the unfinished entries older than the given date. */
+    #[Override]
     public function pruneUnfinished(DateTimeInterface $before): int
     {
         $result = $this->collection->deleteMany(
@@ -229,6 +230,7 @@ class MongoBatchRepository extends DatabaseBatchRepository implements PrunableBa
     }
 
     /** Prune all the cancelled entries older than the given date. */
+    #[Override]
     public function pruneCancelled(DateTimeInterface $before): int
     {
         $result = $this->collection->deleteMany(
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -16,6 +16,7 @@ use MongoDB\Driver\Exception\RuntimeException;
 use MongoDB\Driver\ReadPreference;
 use MongoDB\Laravel\Concerns\ManagesTransactions;
 use OutOfBoundsException;
+use Override;
 use Throwable;
 
 use function filter_var;
@@ -95,6 +96,7 @@ class Connection extends BaseConnection
      *
      * @return Query\Builder
      */
+    #[Override]
     public function table($table, $as = null)
     {
         $query = new Query\Builder($this, $this->getQueryGrammar(), $this->getPostProcessor());
@@ -115,6 +117,7 @@ class Connection extends BaseConnection
     }
 
     /** @inheritdoc */
+    #[Override]
     public function getSchemaBuilder()
     {
         return new Schema\Builder($this);
@@ -172,6 +175,8 @@ class Connection extends BaseConnection
         return $this->connection;
     }
 
+    /** @inheritdoc  */
+    #[Override]
     public function enableQueryLog()
     {
         parent::enableQueryLog();
@@ -182,6 +187,7 @@ class Connection extends BaseConnection
         }
     }
 
+    #[Override]
     public function disableQueryLog()
     {
         parent::disableQueryLog();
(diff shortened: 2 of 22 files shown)


paulinevos:
Apparently `git commit -m` does not like backticks... will rectify commit messages I guess.

src/Bus/MongoBatchRepository.php
GromNaN, line 1: The `use Override;` is missing. You an run `vendor/bin/phpcbf` to fix it.

src/Connection.php
```diff
@@ -182,6 +187,8 @@ public function enableQueryLog()
         }
     }
 
+    /** @inheritdoc */
```
GromNaN: `@inheritdoc` alone isn't useful, the php doc readers (IDE, SA) will automatically read on the parent method when a method is not documented.
This was useful to indicate to the human reader that this method was overriding a parent one, but the new `#[Override]` attribute fulfills this task.

GromNaN (commented):
This is huge, and will be very helpful to detect when a parent method is removed.
I spotted some tiny changes.

GromNaN:
> Apparently `git commit -m` does not like backticks... will rectify commit messages I guess.

We squash-merge. So you can refine the commit message in the merge box in GitHub.

.github/workflows/build-ci-atlas.yml
GromNaN, line 1: Do you know why this changes from unrelated commits are included in the PR?

src/Relations/EmbedsOneOrMany.php
```diff
@@ -105,13 +109,8 @@ public function match(array $models, Collection $results, $relation)
...
     }
 
-    /**
-     * Shorthand to get the results of the relationship.
-     *
-     * @param  array $columns
-     *
-     * @return Collection
-     */
+    /** Shorthand to get the results of the relationship. */
```
GromNaN: Without `@inheritdoc` you lose the phpdoc for the parameter and the return type that are described in the parent method; isn't it?

.github/workflows/build-ci-atlas.yml
paulinevos: Probably just needs a rebase

src/Relations/EmbedsOneOrMany.php
paulinevos: Hm maybe I misunderstood your previous comment, but didn't you say there's no point in just having `inheritdoc` when we already have `#[Override]`?

paulinevos: Ah wait, you said "when a method is not documented". So you're saying it won't parse the parent's phpdoc in this case because there is a comment.

I think the comment is identical to the parent's, so I should probably just remove it then

GromNaN: Yes, it that's the same comment as the parent method, no need to repeat it.

src/Relations/MorphMany.php
```diff
@@ -14,13 +15,8 @@
...
 class MorphMany extends EloquentMorphMany
 {
-    /**
-     * Get the name of the "where in" method for eager loading.
-     *
-     * @param string $key
-     *
-     * @return string
-     */
+    /** Get the name of the "where in" method for eager loading. */
```
GromNaN: Same here, the comment can be removed entirely. https://github.com/laravel/framework/blob/1831e267177acf34fabd121fc0b7eaf32d61c06d/src/Illuminate/Database/Eloquent/Relations/Relation.php#L423

alcaeus:
Would it make sense to enable the [`checkMissingOverrideMethodAttribute` config setting](https://phpstan.org/config-reference#checkmissingoverridemethodattribute) in PHPStan? That would help us keep the list up-to-date 🤔

GromNaN:
> Would it make sense to enable the [`checkMissingOverrideMethodAttribute` config setting](https://phpstan.org/config-reference#checkmissingoverridemethodattribute) in PHPStan? That would help us keep the list up-to-date 🤔 

Some methods are added in newer versions of Laravel, and are not available in Laravel 9, 10 or 11.0.
I don't think there is any method that has been removed. We could add this restriction only when phpstan run with the latest Laravel version.

paulinevos:
>We could add this restriction only when phpstan run with the latest Laravel version.

Is there a way we can actually do that? I can only find a parameter for the PHP version in the phpstan docs, but nothing about a specific version of a dependency/Laravel in either the phpstan or larastan docs

GromNaN (approved):

PR #3416, opened 2025-06-24
Consolidate repetitive model instantiations across tests

zigzagdev:
### Description  
This pull request refactors the test suite to eliminate redundant model instantiations. Several test cases repeatedly constructed identical `User`, `Item`, and `Book` model objects inline, leading to code duplication and reduced maintainability.

To address this:

- A set of **factory-like helper methods** has been introduced to create consistent test data.
- All repeated inline instantiations were replaced by these shared methods.
- This improves readability, enforces consistency, and simplifies future updates.

This is a non-functional change (refactor only) and does **not** alter the behaviour or scope of any existing tests.

### Checklist

スクリーンショット 2025-06-24 10 38 00

### DIFF
--- a/tests/ModelTest.php
+++ b/tests/ModelTest.php
@@ -52,6 +52,7 @@ class ModelTest extends TestCase
     protected function setUp(): void
     {
         parent::setUp();
+
         Carbon::setTestNow();
     }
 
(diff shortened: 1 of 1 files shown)


tests/ModelTest.php
```diff
@@ -81,12 +87,24 @@ public function testQualifyColumn(): void
         $this->assertEquals('users.name', $sqlUser->qualifyColumn('name'));
     }
 
+    private function makeUser(array $overrides = []): User
```
GromNaN: The `$overrides` is never used. I don't think we need it.

GromNaN: You can rename this method `createUser()` and include the `save()` call that is always called from what I see.

GromNaN (changes requested):

tests/ModelTest.php
zigzagdev: @GromNaN 

Thanks for replying !!

i fixed [it](https://github.com/mongodb/laravel-mongodb/pull/3416/commits/6ecd5279422805597c2caa1713672ca4bf35f133) ! 🙌

zigzagdev: @GromNaN 

Thanks for replying !!

i fixed [it](https://github.com/mongodb/laravel-mongodb/pull/3416/commits/6ecd5279422805597c2caa1713672ca4bf35f133) ! 🙌

GromNaN (approved):

GromNaN (approved):

GromNaN:
Thank you @zigzagdev

PR #3417, opened 2025-06-25
Add Mongo builder mixin to `DocumentModel`

paulinevos:
It helps with autocompletion as IDEs will recognize Mongo builder methods that are not present on the base builder.

### Checklist

- [ ] ~Add tests and ensure they pass~ N/A

### DIFF
--- a/src/Eloquent/DocumentModel.php
+++ b/src/Eloquent/DocumentModel.php
@@ -50,6 +50,7 @@ use function strcmp;
 use function strlen;
 use function var_export;
 
+/** @mixin Builder */
 trait DocumentModel
 {
     use HybridRelations;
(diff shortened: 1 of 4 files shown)


src/Eloquent/DocumentModel.php
```diff
@@ -50,6 +50,7 @@
 use function strlen;
 use function var_export;
 
+/** @mixin Builder */
```
GromNaN, line 53: I had to remove the `@mixin` annotation from the Model class (#2981) because the return types of the methods is not the same in the Builder and the Model. I must have made a mistake.

GromNaN (approved):
Let's make this change. It improves type understanding in PHPStorm. If a issue arises with Larastan, we will investigate it.

Issue #3306, opened 2025-03-05
Model `array` cast saved as string


alanondra:
- Laravel-mongodb Version: 5.2.0
- PHP Version: 8.2.27
- Database Driver & Version: 6.0.20

### Steps to reproduce
1. Create Model with property in `casts` method defined as `array`
2. Save Model 
3. Query Model from Collection in Studio 3T

### Expected behaviour
I expected it to save as a MongoDB array.

### Actual behaviour
It saved as a PHP string representation of an array.

<details><summary><b>Logs</b>:</summary>
Insert log.txt here (if necessary)
</details>

GromNaN:
You can omit the `array` cast and don't apply any cast to the field. MongoDB natively stores PHP arrays as BSON Array or Document.

We could do something to modify the behavior of this cast and make it no-op.

Tracking this feature request PHPORM-305.

masterbater:
Do you think its related?

https://github.com/mongodb/laravel-mongodb/discussions/3298

PR #3418, opened 2025-06-29
Fix array casts as string

hans-thomas:
Hello everyone :)
As described here #3306, when casting an array attribute as `array`, it will save as a JSON-encoded string.
In this PR, I prevent casting an array to a string.

### DIFF
--- a/src/Eloquent/DocumentModel.php
+++ b/src/Eloquent/DocumentModel.php
@@ -11,7 +11,6 @@ use DateTimeZone;
 use Illuminate\Contracts\Queue\QueueableCollection;
 use Illuminate\Contracts\Queue\QueueableEntity;
 use Illuminate\Contracts\Support\Arrayable;
-use Illuminate\Database\Eloquent\Casts\Json;
 use Illuminate\Database\Eloquent\Concerns\HasAttributes;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\Relation;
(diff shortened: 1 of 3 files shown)


hans-thomas:
Hi @paulinevos, could you please review my PR?

alcaeus:
@hans-thomas in general, the changes look good. I am a little bit worried about the BC impact of this change, as any fields that previously used an array cast will have been saved as JSON string and won't be read successfully in future. To mitigate this, we need to do two things:
- Create a new `json` cast that people should use instead of `array` (ignore if something like that already exists in Laravel)
- Emit a deprecation notice in `isJsonCastable` if someone is using an `array` cast. If they are, we should tell them to either use the `json` cast if they want to store the array as a JSON-encoded string, or to remove the array cast if they want to store it as an array. We can then remove that deprecation notice in the next major version.

With the `_id` vs. `id` issue fresh in my mind, I don't want to create another data migration problem for users, so we need to give people the option to retain the old behaviour, no matter how useless it may be in MongoDB.

GromNaN:

I rebased on 5.x, added a `@var` annotation in the test to fix PHPStan errors, and added a backward-compatibility test showing that data previously stored as a JSON-encoded string is still readable after the fix.

However, as @alcaeus noted, this is a silent breaking change for users who intentionally rely on the `array` cast storing as a JSON string (to share data with other systems that expect a string, for example). We will create an alternative PR that adds a deprecation notice pointing users toward the `json` cast, to give them a clear migration path before the behavior changes. Closing this PR in favor of that one https://github.com/mongodb/laravel-mongodb/pull/3495.

Issue #3428, opened 2025-07-16
Setting rename_embedded_id_field to false prevents application setting the main id/_id field


ctadlock:
- Laravel-mongodb Version: 5.4.1
- PHP Version: 8.4
- Laravel: 12.20.0

### Steps to reproduce
1. Set rename_embedded_id_field  to false in config
2. Set a documents id field to an application value "12345"
3. insert the model

### Expected behaviour
The `_id` field should be "12345"

### Actual behaviour
A generated ObjectID is used

<details><summary><b>Logs</b>:</summary>
Insert log.txt here (if necessary)
</details>

Im setting this value because I have a nested object with id `google.id` that I dont want to be `_id`

<img width="1355" height="794" alt="Image" src="https://github.com/user-attachments/assets/ccc2cc3c-2ade-44b7-be28-f141e59ca29f" />

<img width="610" height="199" alt="Image" src="https://github.com/user-attachments/assets/a9edc95b-8750-4418-8043-a3d9a3983c8b" />

ctadlock:
I just noticed that both `_id` and `id` are being set in my Account document.

alcaeus:
As the name implies, the `rename_embedded_id_field` option allows disabling ID field renaming in embedded documents. Since you are using `id` within the `google` embedded document, this is indeed the right option to enable so that the field will continue to be written and read as `id`. This setting however does not affect root documents.

In my quick testing, I was able to confirm that setting an `id` field on a model correctly maps this to `_id`, and no `id` field is being created:
```php
$user        = new User();
$user->id    = 12345;
$user->name  = 'John Doe';
$user->title = 'admin';
$user->age   = 35;

$user->save();
```

<img width="318" height="111" alt="Image" src="https://github.com/user-attachments/assets/2ef85fd6-aae9-4ecb-9ecb-a5f836f4473f" />

While this behaviour is different from previous major versions of this library, it is the intended behaviour as the primary key in MongoDB is always `_id`. Since there are some Laravel libraries and modules that use a hardcoded `id` primary key, we had no other choice than to map `id` to `_id` - anything else would create documents that use a potentially unindexed (and thus not guaranteed to be unique) field as primary keys.

ctadlock:
Well in my code its pretty clear that setting `rename_embedded_id_field` totally changes the behavior of the main primary key resulting in both _id and id.

I just worked around the problem.

alcaeus:
[...]

PR #3429, opened 2025-07-18
PHPORM-369: Fix ID handling when using insert method instead of save

alcaeus:
PHPORM-369

### Checklist

This addresses an issue highlighted in a [discussion about `rename_embedded_id_field`](https://github.com/mongodb/laravel-mongodb/issues/3428#issuecomment-3087725396): when using `DB::table()->insert`, we pass `$values` directly to `aliasIdForQuery`. Since `$values` contains a list of documents at this point, every document is treated as an embedded document and misses out on the ID handling for root documents. To work around this, we call `aliasIdForQuery` on each document through `array_map`.
I also added a test with a model that has `$incrementing` set to `false`, as the wrong behaviour is bypassed when `$incrementing` is its default `true` value.

### DIFF
--- a/tests/Models/NonIncrementing.php
+++ b/tests/Models/NonIncrementing.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Laravel\Tests\Models;
+
+use Illuminate\Database\Eloquent\Model;
+use MongoDB\Laravel\Eloquent\DocumentModel;
+
+/**
+ * @property string $id
+ * @property string $name
+ */
+class NonIncrementing extends Model
+{
+    use DocumentModel;
+
+    protected $keyType = 'string';
+    protected $connection = 'mongodb';
+
+    protected $fillable = [
+        'name',
+    ];
+    protected static $unguarded = true;
+    public $incrementing = false;
+}
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -745,7 +745,10 @@ class Builder extends BaseBuilder
             $values = [$values];
         }
 
-        $values = $this->aliasIdForQuery($values);
+        $values = array_map(
+            $this->aliasIdForQuery(...),
+            $values,
+        );
 
         $options = $this->inheritConnectionOptions();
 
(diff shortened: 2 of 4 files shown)


tests/Models/NonIncrementing.php
```diff
@@ -0,0 +1,26 @@
...
+    use DocumentModel;
+
+    protected $keyType = 'string';
+    protected $connection = 'mongodb';
+
+    protected $fillable = [
+        'name',
+    ];
+    protected static $unguarded = true;
+    public $incrementing = false;
```
jmikola, line 25: I noted in https://github.com/mongodb/laravel-mongodb/issues/3428#issuecomment-3089276047 you mentioned that `$incrementing = true` (Eloquent's default per [Primary Keys](https://laravel.com/docs/12.x/eloquent#primary-keys)) was responsible for a different code path being utilized.

Assuming this library is going to generate ObjectIds regardless, I would think it'd be ideal if `$incrementing` had absolutely no effect on our behavior. Is this something we have no control over (within reason)?

jmikola: Noted that this goes back to [`Illuminate\Database\Eloquent\Model::performInsert()`](https://github.com/laravel/framework/blob/v12.20.0/src/Illuminate/Database/Eloquent/Model.php#L1395-L1408).

When `$incrementing` is `true`, [`Model::insertAndSetId()`](https://github.com/laravel/framework/blob/v12.20.0/src/Illuminate/Database/Eloquent/Model.php#L1429) relies on [`Builder::insertGetId()`](https://github.com/mongodb/laravel-mongodb/blob/5.4.1/src/Query/Builder.php#L758) (where `$values` appears to be a single document). Otherwise, it delegates to [`Builder::insert()`](https://github.com/mongodb/laravel-mongodb/blob/5.4.1/src/Query/Builder.php#L724), which has the logic to wrap the singular document in an array before applying `aliasIdForQuery()`.

So the bug fix looks correct. I think there's still an edge case in the "document or documents?" detection within `Builder::insert()`, but that' outside the scope of this PR. [This assumption](https://github.com/mongodb/laravel-mongodb/blob/5.4.1/src/Query/Builder.php#L736-L737) could be problematic if all fields in a document were arrays (e.g. `_id` is an embedded document):

> As soon as we find a value that is not an array we assume the user is inserting a single document.

jmikola (approved):

Issue #3328, opened 2025-03-27
Incorrect behavior of the transaction method


babaduk47:
- Laravel-mongodb Version: 5.2.0
- PHP Version: 8.3.17
- Database Driver & Version: 1.21.1

### Description:
The transaction method of the `mongodb` connection does not call the `beginTransaction` method. If dispatch job with `afterCommit`=true, the queue will be sent before `commit`. This behavior differs from `mysql` connection

### Steps to reproduce
```php
try {
    DB::connection('mysql')->transaction(static function (): void {
        dispatch(fn() => dump('dump from mysql dispatch'))
            ->onConnection('sync')
            ->afterCommit();

        throw new \Exception('Command not executed.');
    });
} catch (\Exception) {
}

try {
    DB::connection('mongodb')->transaction(static function (): void {
        dispatch(fn() => dump('dump from mongodb dispatch'))
            ->onConnection('sync')
            ->afterCommit();

        throw new \Exception('Command not executed.');
    });
} catch (\Exception) {
}
```

### Expected behaviour
Jobs will not be completed

### Actual behaviour
We get dump = `dump from mongodb dispatch`

### Additional debug
```php
DB::connection('mysql')->transaction(static function (): void {
    $connection = DB::connection('mysql');

    $transactionsManager = (fn () => $this->transactionsManager)->bindTo($connection, $connection)();
    $pendingTransactions = (fn () => $this->pendingTransactions->all())->bindTo($transactionsManager, $transactionsManager)();

    dump('mysql transactionLevel: ' . DB::connection('mysql')->transactionLevel());
    dump('mysql pendingTransactions: ', $pendingTransactions);
});

DB::connection('mongodb')->transaction(static function (): void {
    $connection = DB::connection('mongodb');

    $transactionsManager = (fn () => $this->transactionsManager)->bindTo($connection, $connection)();
    $pendingTransactions = (fn () => $this->pendingTransactions->all())->bindTo($transactionsManager, $transactionsManager)();

    dump('mongodb transactionLevel: ' . DB::connection('mongodb')->transactionLevel());
    dump('mongodb pendingTransactions: ', $pendingTransactions);
});
```

<img width="741" alt="Image" src="https://github.com/user-attachments/assets/e0ce298c-2092-4318-ac06-e25ab4508ff2" />

GromNaN:
Issue tracked in Jira [PHPORM-373](https://jira.mongodb.org/browse/PHPORM-373)

[...]

PR #3443, opened 2025-09-04
PHPORM-373 Use `DatabaseTransactionsManager` in `ManagesTransactions` trait

paulinevos:
`DatabaseTransactionsManager` was introduced in Laravel to keep track of things like pending transactions and fire transaction events, so that e.g. dispatches could hook into transaction state.

Our trait was not using it yet, resulting in missing `afterCommit` behavior (see PHPORM-373 and #3328)

### Checklist

### DIFF
--- a/src/Concerns/ManagesTransactions.php
+++ b/src/Concerns/ManagesTransactions.php
@@ -8,9 +8,12 @@ use Closure;
 use MongoDB\Client;
 use MongoDB\Driver\Exception\RuntimeException;
 use MongoDB\Driver\Session;
+use MongoDB\Laravel\Connection;
 use Throwable;
 
+use function max;
 use function MongoDB\with_transaction;
+use function property_exists;
 
 /**
  * @internal
@@ -55,8 +58,23 @@ trait ManagesTransactions
      */
     public function beginTransaction(array $options = []): void
     {
+        $this->runCallbacksBeforeTransaction();
+
         $this->getSessionOrCreate()->startTransaction($options);
+
+        $this->handleInitialTransactionState();
+    }
+
+    private function handleInitialTransactionState(): void
+    {
         $this->transactions = 1;
+
+        $this->transactionsManager?->begin(
+            $this->getName(),
+            $this->transactions,
+        );
+
+        $this->fireConnectionEvent('beganTransaction');
     }
 
     /**
@@ -64,8 +82,26 @@ trait ManagesTransactions
      */
     public function commit(): void
     {
+        $this->fireConnectionEvent('committing');
         $this->getSessionOrThrow()->commitTransaction();
-        $this->transactions = 0;
+
+        $this->handleCommitState();
+    }
+
+    private function handleCommitState(): void
+    {
+        [$levelBeingCommitted, $this->transactions] = [
+            $this->transactions,
+            max(0, $this->transactions - 1),
+        ];
+
+        $this->transactionsManager?->commit(
+            $this->getName(),
+            $levelBeingCommitted,
+            $this->transactions,
+        );
+
+        $this->fireConnectionEvent('committed');
     }
 
     /**
@@ -73,14 +109,42 @@ trait ManagesTransactions
      */
     public function rollBack($toLevel = null): void
     {
-        $this->getSessionOrThrow()->abortTransaction();
+        $session = $this->getSessionOrThrow();
+        if ($session->isInTransaction()) {
+            $session->abortTransaction();
+        }
+
+        $this->handleRollbackState();
+    }
+
+    private function handleRollbackState(): void
+    {
         $this->transactions = 0;
+
+        $this->transactionsManager?->rollback(
+            $this->getName(),
+            $this->transactions,
+        );
+
+        $this->fireConnectionEvent('rollingBack');
+    }
+
+    private function runCallbacksBeforeTransaction(): void
+    {
+        // ToDo: remove conditional once we stop supporting Laravel 10.x
+        if (property_exists(Connection::class, 'beforeStartingTransaction')) {
+            foreach ($this->beforeStartingTransaction as $beforeTransactionCallback) {
+                $beforeTransactionCallback($this);
+            }
+        }
     }
 
     /**
      * Static transaction function realize the with_transaction functionality provided by MongoDB.
      *
-     * @param  int $attempts
+     * @param int $attempts
+     *
+     * @throws Throwable
      */
     public function transaction(Closure $callback, $attempts = 1, array $options = []): mixed
     {
@@ -93,15 +157,20 @@ trait ManagesTransactions
 
             if ($attemptsLeft < 0) {
                 $session->abortTransaction();
+                $this->handleRollbackState();
 
                 return;
             }
+
+            $this->runCallbacksBeforeTransaction();
+            $this->handleInitialTransactionState();
 
             // Catch, store, and re-throw any exception thrown during execution
             // of the callable. The last exception is re-thrown if the transaction
             // was aborted because the number of callback attempts has been exceeded.
             try {
                 $callbackResult = $callback($this);
+                $this->fireConnectionEvent('committing');
             } catch (Throwable $throwable) {
                 throw $throwable;
             }
@@ -110,8 +179,11 @@ trait ManagesTransactions
         with_transaction($this->getSessionOrCreate(), $callbackFunction, $options);
 
         if ($attemptsLeft < 0 && $throwable) {
+            $this->handleRollbackState();
             throw $throwable;
         }
+
+        $this->handleCommitState();
 
         return $callbackResult;
     }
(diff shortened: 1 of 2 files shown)


src/Concerns/ManagesTransactions.php
```diff
@@ -88,6 +92,8 @@ public function transaction(Closure $callback, $attempts = 1, array $options = [
         $callbackResult = null;
         $throwable      = null;
 
+        $this->transactionsManager?->begin($this->getName(), 0);
```
paulinevos: Unfortunately have to call this stuff here manually again, unlike what the Laravel trait does (which is to call its own `beginTransaction` method here), as we are using the `with_transaction` helper function and passing the callback. The helper function, in turn, calls `WithTransaction::execute` which handles starting and committing the actual transaction in the session.

Calling `beginTransaction` here instead of calling `DatabaseTransactionsManager::begin` directly would therefore result in attempting to start the transaction twice.

Same goes for the `commit` call

src/Concerns/ManagesTransactions.php
```diff
@@ -113,6 +119,8 @@ public function transaction(Closure $callback, $attempts = 1, array $options = [
             throw $throwable;
         }
 
+        $this->transactionsManager?->commit($this->getName(), 0, 0);
```
paulinevos: Wasn't sure if if it was worth not hard-coding this for future proofing reasons in case we ever start supporting nested transactions? 😬

src/Concerns/ManagesTransactions.php
```diff
@@ -55,6 +55,8 @@ private function getSessionOrThrow(): Session
      */
     public function beginTransaction(array $options = []): void
     {
+        $this->transactionsManager?->begin($this->getName(), 0);
```
GromNaN: Looking at Laravel's ManagesTransactions, the call to `transactionsManager?->begin` is done after the transaction is started.

https://github.com/laravel/framework/blob/d09f1345e25d8d04c67ffa4cdd3656f5ebbce57a/src/Illuminate/Database/Concerns/ManagesTransactions.php#L124-L136

src/Concerns/ManagesTransactions.php
```diff
@@ -55,6 +55,8 @@ private function getSessionOrThrow(): Session
...
+
         $this->getSessionOrCreate()->startTransaction($options);
         $this->transactions = 1;
     }
```
GromNaN, line 78: Should the `$this->fireConnectionEvent('beganTransaction');` also be called at like in Laravel: https://github.com/laravel/framework/blob/d09f1345e25d8d04c67ffa4cdd3656f5ebbce57a/src/Illuminate/Database/Concerns/ManagesTransactions.php#L138

That was added in Laravel 5.4, I don't know which feature it is.

src/Concerns/ManagesTransactions.php
jmikola: Noted that the second parameter here is `int $level` per [`DatabaseTransactionsManager::begin()`](https://github.com/laravel/framework/blob/v12.28.1/src/Illuminate/Database/DatabaseTransactionsManager.php#L46).

In the base [`ManagesTransactions::begin()`](https://github.com/laravel/framework/blob/v12.28.1/src/Illuminate/Database/Concerns/ManagesTransactions.php#L124) implementation, the level is incremented _before_ invoking `DatabaseTransactionsManager::begin()`. I'm not certain, but I expect that `1` would be the smallest value passed in that case. Here, you're passing `0` but setting `$this->transactions` to `1` a few lines below.

Also, it may be a subtle change, but it looks like the base implementation invokes `PDO::commit()` _before_ `DatabaseTransactionsManager::begin()`. Therefore, would it make more sense to invoke `Session::startTransaction()` first here?

Consider:

```php
$this->getSessionOrCreate()->startTransaction($options);
$this->transactions = 1;
$this->transactionsManager?->begin($this->getName(), $this->transactions);
```

src/Concerns/ManagesTransactions.php
```diff
@@ -65,6 +67,8 @@ public function beginTransaction(array $options = []): void
     public function commit(): void
     {
         $this->getSessionOrThrow()->commitTransaction();
+
+        $this->transactionsManager?->commit($this->getName(), 0, 0);
         $this->transactions = 0;
```
jmikola: The flow here does match [`ManagesTransactions::commit()`](https://github.com/laravel/framework/blob/v12.28.1/src/Illuminate/Database/Concerns/ManagesTransactions.php#L203] in that you invoke the driver's commit method first (analogous to `PDO::commit()`).

[`DatabaseTransactionsManager::commit()`](https://github.com/laravel/framework/blob/v12.28.1/src/Illuminate/Database/DatabaseTransactionsManager.php#L67) might expect two different values for `int $levelBeingCommitted` and `int $newTransactionLevel`, though. Consider the following in the base method:

```php
[$levelBeingCommitted, $this->transactions] = [
    $this->transactions,
    max(0, $this->transactions - 1),
];
```

Here, you pass `0` as _both_ the committed and new level, which is a subtle behavioral difference. While this could be replaced with `1, 0` literals, I think the code would be more self-documenting if copied the above assignment or instead did something like:

```
$levelBeingCommitted = $this->transactions;
$this->transactions = 0;
```

Either way, a comment to remind readers that MongoDB only supports a single transaction level might be helpful (if not here then perhaps in `begin()` where you manually assign `1` instead of incrementing).

src/Concerns/ManagesTransactions.php
jmikola: Per my earlier comment in `beginTransaction()`, should you actually pass `1` as the transaction level? Also, if you can't call `beginTransaction()` yourself perhaps you'd also want to manually assign `$this->transactions = 1` immediately before as well:

```php
$this->transactions = 1;
$this->transactionsManager?->begin($this->getName(), $this->transactions);
```

Since you're deviating from the base implementation, I think a block comment explaining why this inline duplication is necessary would help.

src/Concerns/ManagesTransactions.php
jmikola: Echoing my comments above about keeping this in sync with any changes to the `commit()` method and ensuring that `$this->transactions` also gets set back to `0`. I'd suggest using the same temporary variable for clarity:

```php
$levelBeingCommitted = $this->transactions;
$this->transactions = 0;
```

src/Concerns/ManagesTransactions.php
paulinevos: That can work for this method, but not for the `transaction` method, as `WithTransaction::execute()` both starts the transaction and calls the callback, which shouldn't happen (and is what this call handles)

So I'm struggling to think if we should keep the order consistent on our library side or have one method consistent with Laravel and the other not...

src/Concerns/ManagesTransactions.php
paulinevos: Ah, yes, forgot that one!

src/Concerns/ManagesTransactions.php
paulinevos: Yeah, I figured it would be appropriate to pas 0 here twice as we have no levels, and the levels wouldn't dip below 0. But it did feel a bit icky to me so I might do what you've suggested. Also just in case we somehow do start supporting nested transactions many years down the line, this won't become one of those things causing weird behavior because we forgot about these hardcoded values.

jmikola: I'm not sure if the current transaction level gets passed to the event listeners, but that was actually what prompted my comment (not so much the possibility of nested transactions in MongoDB 12.0—don't quote me on that). Something downstream of these events might be surprised to see a transaction get committed at level zero.

paulinevos: good point

src/Concerns/ManagesTransactions.php
jmikola: Note that there are a few other `fireConnectionEvent` calls for `committing`, `committed`, and `rollingBack`. Since we create our own callback in `transaction()`, which wraps the user's callback, I think you'll be able to work this into the code path that uses `with_transaction` as well.

tests/Ticket/GH3328Test.php
```diff
@@ -0,0 +1,267 @@
...
+        );
+
+        if (! interface_exists('\Illuminate\Contracts\Database\ConcurrencyErrorDetector')) {
+            // Earlier versions of Laravel use a trait instead of DI to detect concurrency errors
+            // That would increase the scope of this comparison dramatically and is probably not worth it.
+            return;
+        }
+
+        // phpcs:ignore
+        $this->app->bind(\Illuminate\Contracts\Database\ConcurrencyErrorDetector::class, FakeConcurrencyErrorDetector::class);
```
GromNaN: This class doesn't exist in the Laravel version you run phpcs?
You can make `phpcs:ignore` ore specific by setting the name of the rule to ignore.

Regarless if the class exists or not, the `use` statement should be used.

Or use the string version: `'Illuminate\Contracts\Database\ConcurrencyErrorDetector'`

GromNaN (approved):

PR #3459, opened 2025-12-03
PHPORM-418 Test with PHP 8.5 and reduce test matrix

GromNaN:
Fix PHPORM-418

- Add PHP 8.5 to the tests
- Reduce test matrix for less resource consumption while keeping a coverage for every MongoDB  server, driver, PHP and Laravel versions.
- Don't run tests in duplicate for `feature/*` branches, they must be used to PR version branches.

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - "[0-9]+.[0-9x]+"
-      - "feature/*"
   pull_request:
     branches:
       - "[0-9]+.[0-9x]+"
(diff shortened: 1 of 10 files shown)


GromNaN:
Tests failing on Atlas. I have to digg if something change in the atlas local docker image.

GromNaN:
Waiting for https://github.com/mongodb/laravel-mongodb/pull/3460

.github/workflows/build-ci.yml
```diff
@@ -16,53 +15,63 @@ env:
...
+          - php: "8.4"
+            laravel: "10.*"
+            mongodb: "8.0"
+          # Test Laravel 11
+          - php: "8.4"
             laravel: "11.*"
-          - php: "8.1"
+            mongodb: "8.0"
+          # Test all server versions
+          - php: "8.4"
```
alcaeus: Do we want to use PHP 8.5 (i.e. always the latest PHP version) for these tests?

alcaeus (approved):
LGTM with a suggestion to always use the latest PHP version for extra matrix builds.

.github/workflows/build-ci.yml
GromNaN: It's not easy to decide which versions to test in the matrix. Should we use the lowest PHP version to ensure there is no invalid syntax? Or should we use the highest version to ensure there is no depreciation?

We don't have any code path specific to a particular PHP version. We should probably use PHP 8.2.

PR #3464, opened 2025-12-08
Fix test that randomly crashes

GromNaN:
This test is inconsistently [crashing](https://github.com/mongodb/laravel-mongodb/actions/runs/20030747326/job/57439273219?pr=3463):

```
1) MongoDB\Laravel\Tests\SchemaTest::testGetTables Failed asserting that 40960 matches expected 8192.
```

### DIFF
--- a/tests/SchemaTest.php
+++ b/tests/SchemaTest.php
@@ -458,7 +458,7 @@ class SchemaTest extends TestCase
             $this->assertNotEquals('test_view', $table['name'], 'Standard views should not be included in the result of getTables.');
 
             if ($table['name'] === self::COLL_1) {
-                $this->assertEquals(8192, $table['size']);
+                $this->assertGreaterThanOrEqual(8192, $table['size']);
                 $this->assertEquals($dbName, $table['schema']);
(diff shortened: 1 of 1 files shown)


tests/SchemaTest.php
```diff
@@ -458,7 +458,7 @@ public function testGetTables()
             $this->assertNotEquals('test_view', $table['name'], 'Standard views should not be included in the result of getTables.');
 
             if ($table['name'] === self::COLL_1) {
-                $this->assertEquals(8192, $table['size']);
+                $this->assertGreaterThanOrEqual(8192, $table['size']);
```
alcaeus, line 461: Nit: if we want to go purely defensive, the check should be that the size is an int larger than `0` -- everything else is already assuming details of the underlying system.

alcaeus (approved):
LGTM with a nit about being even more defensive, but this works just fine.

tests/SchemaTest.php
GromNaN: We'll do that if we get another error.

PR #3467, opened 2025-12-15
Fix MongoStore increment behavior for expired cache entries

zigzagdev:
### Description

This PR fixes the behavior of `MongoStore::increment()` so that expired cache entries are no longer incremented.

### what has changed

- Before changed
    - Increment the value first
    - Then check whether the cache entry was expired
    - Potentially delete the entry afterward

This caused expired cache values to be modified briefly, which is inconsistent with Laravel’s cache semantics.

- The method now:
    - Filters out expired entries at the query level using expires_at > now
    - Returns false when the cache entry is missing or expired
    - Avoids modifying expired values entirely

### Test Results
スクリーンショット 2025-12-15 10 39 13

- A regression test was added to proved that
    - `increment()` returns `false` for expired cache entries
    - The stored value remains unchanged when the cache is already expired

### Checklist

### DIFF
--- a/src/Cache/MongoStore.php
+++ b/src/Cache/MongoStore.php
@@ -179,25 +179,23 @@ final class MongoStore implements LockProvider, Store
     #[Override]
     public function increment($key, $value = 1): int|float|false
     {
+        $now = $this->getUTCDateTime();
+
         $result = $this->collection->findOneAndUpdate(
             [
                 '_id' => $this->prefix . $key,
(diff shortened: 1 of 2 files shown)


GromNaN (approved):

GromNaN:

I rebased the branch on 5.7, removed the unused `expires_at` field from the projection (it is no longer read since the post-update expiry check was removed), and added a regression test covering `decrement` on expired entries.

Merging now.

PR #3469, opened 2026-01-14
Fix MongoDB presence verification for values containing regex delimiters

zigzagdev:
### Description

This PR fixes an issue in the MongoDB presence verifier where validation queries could break (or behave unexpectedly) when the input contained regex delimiter characters such as /.

### Motivation

The verifier builds a MongoDB regex from user input to perform a case-insensitive exact match. However, the previous implementation used preg_quote without specifying a delimiter and also mixed regex options in a way that could lead to incorrect patterns. This becomes visible when validating values like Foo/Bar.

###What changed
- Escaped input using preg_quote((string) $value, '/') so / (and other delimiter-related characters) are safely handled.
- Constructed MongoDB\BSON\Regex using the correct options argument ('i') rather than embedding flags in the pattern string.
- Queried using the Regex object directly (where($column, $regex)) instead of passing 'regex' as an operator.
- Applied the same escaping approach for getMultiCount() when building a regex alternation.
- Added a regression test covering Foo/Bar as well as case-insensitive matching and “partial should not match”.

### Why passing the Regex object directly

The MongoDB query builder expects a `MongoDB\BSON\Regex` instance to be provided as the value for a regex comparison.
Using `where($column, 'regex', $regex)` relies on an operator string that may not be interpreted consistently across the MongoDB driver / query builder implementation, whereas passing the Regex object directly (`where($column, $regex)`) is the explicit and supported way to perform regex matching.

### Test plan

./vendor/bin/phpunit

Specifically verified the new cases added in tests/ValidationTest.php (values containing /, case-insensitive exact match, and partial mismatch).

スクリーンショット 2026-01-14 20 34 21

### Checklist

### DIFF
--- a/src/Validation/DatabasePresenceVerifier.php
+++ b/src/Validation/DatabasePresenceVerifier.php
@@ -17,7 +17,10 @@ class DatabasePresenceVerifier extends \Illuminate\Validation\DatabasePresenceVe
     #[Override]
     public function getCount($collection, $column, $value, $excludeId = null, $idColumn = null, array $extra = [])
     {
-        $query = $this->table($collection)->where($column, new Regex('^' . preg_quote($value) . '$', '/i'));
+        $quoted = preg_quote((string) $value, '/');
+        $regex = new Regex('^' . $quoted . '$', 'i');
+
+        $query = $this->table($collection)->where($column, $regex);
 
         if ($excludeId !== null && $excludeId !== 'NULL') {
             $query->where($idColumn ?: 'id', '<>', $excludeId);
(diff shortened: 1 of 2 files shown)


GromNaN:

After investigation, this PR does not fix an observable behavioral bug. The key point is that `preg_quote` without a delimiter already escapes all regex metacharacters (`.`, `*`, `+`, ...) and the only additional character escaped when specifying `'/'` as delimiter is `/` itself, which has no special meaning in a MongoDB BSON regex pattern.

The following tests pass both **before and after** the fix, confirming no behavioral change:

```php
User::create(['name' => 'Foo/Bar']);

$validator = Validator::make(['name' => 'Foo/Bar'], ['name' => 'required|exists:users']);
$this->assertFalse($validator->fails()); // passes before AND after fix

$validator = Validator::make(['name' => 'foo/bar'], ['name' => 'required|exists:users']);
$this->assertFalse($validator->fails()); // passes before AND after fix

$validator = Validator::make(['name' => 'Foo'], ['name' => 'required|exists:users']);
$this->assertTrue($validator->fails()); // passes before AND after fix
```

No test can be written that fails before the fix and passes after. Closing for that reason.

Issue #3392, opened 2025-05-26
Bug Report: Inability to Disable Top-Level _id Aliasing in Laravel MongoDB Package


Danilim04:
## Bug Report: Inability to Disable Top-Level `_id` Aliasing in Laravel MongoDB Package

**Description:**
The `mongodb/laravel-mongodb` package currently provides a configuration option `rename_embedded_id_field` to disable automatic renaming of `_id` fields **only** for embedded documents. However, there is **no** corresponding option to disable top-level `_id` aliasing. As a result:

* Models with a custom `protected $primaryKey = 'id';` still have their primary key queries (`find()`, `findOrFail()`, `whereKey()`, etc.) hard-coded to use the MongoDB `_id` field.
* Disabling `rename_embedded_id_field` has no effect on the top-level `_id` behavior.
* Route Model Binding and direct lookups by `$model->id` do not work as expected when using a string or UUID primary key named `id`.

**Steps to Reproduce:**

1. Set up a Laravel model extending `Jenssegers\Mongodb\Eloquent\Model` with:

   ```php
   class Invoice extends Model {
       protected $primaryKey = 'id';
   }
   ```
2. In `config/database.php`, configure:

   ```php
   'connections' => [
       'mongodb' => [
           'driver' => 'mongodb',
           // ...
           'rename_embedded_id_field' => false,
       ],
   ],
   ```
3. Attempt to retrieve a document via `Invoice::find($idString)` or via route model binding on the `id` field.

**Expected Behavior:**

* The package should respect the model’s `$primaryKey` and use the `id` field for top-level queries when `rename_embedded_id_field` is set to `false`.

**Actual Behavior:**

* The package continues to query against the `_id` field, ignoring the custom `primaryKey`, making it impossible to use a string or UUID `id` as the primary key without overriding core builder methods.

**Impact:**
This limitation prevents the use of non-ObjectId primary keys at the top level and breaks Laravel features like route model binding for string or UUID keys.

**Workarounds:**
Currently, the only way to work around this issue is to override multiple internal builder methods (`find`, `whereKey`, `findOrFail`, etc.) or to run native MongoDB queries manually, which is not sustainable.

**Request:**
Please introduce a new configuration option (e.g., `rename_id_field`) or enhance existing logic so that top-level `_id` aliasing can be disabled, allowing the package to use the model’s `$primaryKey` for all primary key operations.

GromNaN:
Thanks for the report!

A few clarifications:
[...]

PR #3472, opened 2026-01-28
Add 'alias_id' config option to disable top-level _id/id aliasing

AymenReIn:
## Summary

Adds a new `alias_id` connection config option (default: `true`) that allows disabling the automatic top-level `_id` ↔ `id` aliasing introduced in v5.0.

- When `alias_id` is `false`, root-level `id` is treated as a regular field and not converted to/from `_id`
- When `alias_id` is `true` (default), behavior is identical to current — zero breaking changes
- When disabling `alias_id`, also set `rename_embedded_id_field` to `false` to prevent compound query operators (`$and`/`$or`) from incorrectly aliasing root-level `id` fields in nested where clauses

## Problem

Applications upgrading from v4 to v5 that have documents with a **business `id` field** (e.g., external API IDs, legacy system identifiers) alongside MongoDB's `_id` experience data corruption because the aliasing overwrites the custom `id` field with `_id`.

The existing `rename_embedded_id_field` config (v5.3+) only controls embedded document aliasing. There is currently no way to disable top-level aliasing, which has been reported in:
- #3392 — Requesting reconsideration with this minimal, backward-compatible implementation
- Community forum: https://www.mongodb.com/community/forums/t/issue-with-id-to-id-conversion-when-upgrading-to-laravel-mongodb-v5-laravel-12-support/321882
- Community forum: https://www.mongodb.com/community/forums/t/id-vs-id-with-laravel-mongodb-v5/316408

## Changes

| File | Change |
|---|---|
| `src/Connection.php` | New `aliasId` property, getter/setter, config init (`'alias_id' => true`) |
| `src/Query/Builder.php` | `compileWheres()`: gate `id` → `_id` column rename on `getAliasId()` |
| `src/Query/Builder.php` | `aliasIdForQuery()`: separate root aliasing (`alias_id`) from embedded aliasing (`rename_embedded_id_field`) |
| `src/Query/Builder.php` | `aliasIdForResult()`: same separation for result processing |
| `src/Eloquent/DocumentModel.php` | `getIdAttribute()`: return actual `id` attribute without `_id` fallback when disabled |
| `src/Eloquent/DocumentModel.php` | `getKeyName()`: return `_id` as primary key when disabled |
| `src/Eloquent/DocumentModel.php` | `getKey()`: return raw `_id` attribute directly, bypassing accessor |

## Config usage

```php
// config/database.php
'mongodb' => [
    'driver' => 'mongodb',
    'dsn' => env('MONGODB_URI'),
    'database' => env('MONGODB_DATABASE'),
    'alias_id' => false,
    'rename_embedded_id_field' => false,
],
```

Or at runtime:
```php
DB::connection('mongodb')->setAliasId(false);
DB::connection('mongodb')->setRenameEmbeddedIdField(false);
```

### Why both config keys?

The `aliasIdForQuery()` method recurses into nested arrays with `root=false`. Compound query operators like `$and`/`$or` (automatically added by SoftDeletes or multiple `where()` clauses) create nested array structures that are incorrectly treated as embedded documents during recursion. Setting `rename_embedded_id_field` to `false` prevents this from converting root-level `id` to `_id` inside these query structures.

## Test plan
- [x] All existing tests pass unchanged (default `alias_id => true`)
- [x] Verified with integration tests: model lifecycle (create, refresh, query, update), query builder insert, raw MongoDB document verification, soft deletes with compound where clauses

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -57,6 +57,9 @@ class Connection extends BaseConnection
     /** @var bool Whether to rename the rename "id" into "_id" for embedded documents. */
     private bool $renameEmbeddedIdField;
 
+    /** @var bool Whether to alias "id" to "_id" at the top level for queries and results. */
+    private bool $aliasId;
+
     /**
      * Create a new database connection instance.
      */
@@ -86,6 +89,7 @@ class Connection extends BaseConnection
         $this->useDefaultQueryGrammar();
 
         $this->renameEmbeddedIdField = $config['rename_embedded_id_field'] ?? true;
+        $this->aliasId = $config['alias_id'] ?? true;
     }
 
     /**
@@ -421,6 +425,18 @@ class Connection extends BaseConnection
     public function getRenameEmbeddedIdField(): bool
     {
         return $this->renameEmbeddedIdField;
+    }
+
+    /** Set whether to alias "id" to "_id" at the top level for queries and results. */
+    public function setAliasId(bool $alias): void
+    {
+        $this->aliasId = $alias;
+    }
+
+    /** Get whether to alias "id" to "_id" at the top level for queries and results. */
+    public function getAliasId(): bool
+    {
+        return $this->aliasId;
     }
 
     /**
--- a/src/Eloquent/DocumentModel.php
+++ b/src/Eloquent/DocumentModel.php
@@ -101,26 +101,38 @@ trait DocumentModel
 
     /**
      * Get the primary key for the model.
-     * When alias_id is disabled, always use '_id' as the primary key
-     * to prevent MongoDB's _id from overwriting a custom 'id' field.
+     *
+     * When alias_id is disabled and the model uses Eloquent's default 'id' key,
+     * returns '_id' to prevent MongoDB's _id from overwriting a custom 'id' field.
+     * Models with a custom primary key are always respected regardless of alias_id.
+     *
+     * @return string
      */
     public function getKeyName()
     {
-        return $this->getConnection()->getAliasId()
-            ? parent::getKeyName()
-            : '_id';
+        if (! $this->getConnection()->getAliasId() && parent::getKeyName() === 'id') {
+            return '_id';
+        }
+
+        return parent::getKeyName();
     }
 
     /**
      * Get the value of the model's primary key.
-     * When alias_id is disabled, return the raw '_id' attribute directly
-     * to bypass the getIdAttribute accessor.
+     *
+     * When alias_id is disabled and the model uses Eloquent's default 'id' key,
+     * returns the raw '_id' attribute directly to bypass the getIdAttribute accessor.
+     * Models with a custom primary key are always respected regardless of alias_id.
+     *
+     * @return mixed
      */
     public function getKey()
     {
-        return $this->getConnection()->getAliasId()
-            ? parent::getKey()
-            : $this->attributes['_id'] ?? null;
+        if (! $this->getConnection()->getAliasId() && parent::getKeyName() === 'id') {
+            return $this->attributes['_id'] ?? null;
+        }
+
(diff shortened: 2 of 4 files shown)


AymenReIn:
Hi maintainers,

I'm aware that #3392 was closed as top-level aliasing was considered by-design. I'd like to respectfully request reconsideration with this implementation. Here's why:

### Real-world use case

We maintain a large Laravel application that has been running on mongodb/laravel-mongodb v4 for years. Our MongoDB collections contain documents with a business `id` field (external system identifiers from APIs) that is separate from MongoDB's `_id`. This is a common pattern for systems that sync data from external sources.

When upgrading to v5, the automatic `id` <-> `_id` aliasing corrupts these documents:
- On **write**: our custom `id` value gets moved into `_id`, replacing the auto-generated ObjectId
- On **read**: MongoDB's `_id` overwrites our business `id` field

This makes v5 upgrade impossible without a risky data migration across millions of production documents.

### Why this approach is minimal and safe

1. **New config key** (`alias_id`), not reusing `rename_embedded_id_field` — no semantic change to existing config
2. **Default `true`** — zero behavioral change for any existing user
3. **Follows the existing pattern** — `rename_embedded_id_field` already proved that configurable aliasing is the right approach for embedded docs. This extends it consistently to root level
4. **Small diff** — 48 lines of source changes, 109 lines of tests

### Community demand

Multiple users have reported this exact pain point:
- https://www.mongodb.com/community/forums/t/issue-with-id-to-id-conversion-when-upgrading-to-laravel-mongodb-v5-laravel-12-support/321882
- https://www.mongodb.com/community/forums/t/id-vs-id-with-laravel-mongodb-v5/316408
- https://www.mongodb.com/community/forums/t/mongodb-laravel-mongodb-package-converting-id-to-id/326835

Happy to iterate on the implementation based on feedback. Thanks for considering!

adamwiteszczak:
Is there an ETA on this?

GromNaN:
Thank you for the detailed write-up and the careful implementation.

After consideration, we won't be adding an `alias_id` config option. Having both an `id` business field and a MongoDB `_id` field in the same document is a footgun: the library's aliasing exists precisely to provide a seamless `id` experience on top of `_id`, and supporting both simultaneously adds significant complexity and confusion.

For users who need fully custom aliasing logic, the recently merged #3476 exposes `prepareFieldsForQuery` and `prepareFieldsForResult` as public overridable methods on the `Grammar` class, which is the right extension point for this kind of customization.

Closing in favor of that approach.

ahmdadl:
@GromNaN Thanks for the feedback and for pointing out the new `Grammar` hooks in #3476. 

Unfortunately, overriding `prepareFieldsForQuery` doesn't solve this specific issue because the `Builder` intercepts and mutates the column name before the `Grammar` class ever receives the payload. 

Specifically, in `Query\Builder::toMql()`, `compileWheres()` is executed *before* `prepareFieldsForQuery()`:

```php
// src/Query/Builder.php

public function toMql(): array
{
    // ...

    $wheres = $this->compileWheres(); // 1. Hardcoded mutation happens here
    $wheres = $this->grammar->prepareFieldsForQuery($wheres); // 2. Grammar hook is called too late
    
    // ...
}
```

Because compileWheres() contains a hardcoded fallback that forcefully swaps id to _id, the Grammar class receives a payload where the original id key has already been lost:


```php
protected function compileWheres(): array
    {
        // ...

        if ($where['column'] === 'id') {
            $where['column'] = '_id'; // This needs to be replaced with a new exposed method in Grammar class
        }

        // ...
    }

```

GromNaN:
Thanks for the feedback, that's a valid point. We'll investigate the proper fix in [PHPORM-499](https://jira.mongodb.org/browse/PHPORM-499).

Issue #3453, opened 2025-10-01
[Feature Request] Ability to disable root-level _id and id mapping


mcnika:
Currently, **mongodb/laravel-mongodb** automatically maps MongoDB’s _id field to Laravel’s id attribute on Eloquent models.
This makes it difficult to use a custom id field alongside MongoDB’s native _id.

**Example use case:**
I want to store documents like this:
`{
  "_id": ObjectId("6518e9c4c6a3d1f9b3a2f1a7"),
  "id": "lyqxei40yhyw",
  "title": "Some custom HTML block"
}`

But right now, if I set id, it overrides the default _id mapping or causes conflicts.

**Proposed solution:**
Add a configuration option or model-level property to disable the automatic mapping of _id to id, so developers can:
- Keep _id as the real MongoDB ObjectId
- Define their own id (or any other field like customId) without collisions

**Migration issue:**
I am migrating from the package jenssegers/mongodb, where having both _id and id fields was possible.
Since I already have large amounts of existing data with both fields present, the current forced mapping causes serious migration issues and makes it impossible to work with my existing dataset.

SMFloris:
Already possible, add `'rename_embedded_id_field' => false` in your database.php  config, like so:

```php
'mongodb' => [
..
'host' => "HOST",
'rename_embedded_id_field' => false
]
```

mcnika:
> Already possible, add `'rename_embedded_id_field' => false` in your database.php config, like so:
> 
> 'mongodb' => [
> ..
> 'host' => "HOST",
> 'rename_embedded_id_field' => false
> ]

That option only affects embedded documents. I need to disable the root id field mapping, not the embedded one.

SMFloris:
Weirdly, it does work for me for root id field mappings.

mcnika:
No, it doesn’t actually work.

**rename_embedded_id_field** only affects embedded documents, not the root document.
You can also see that clearly from its naming — it doesn’t apply to the root id field.

When you try to insert the document like this:

```
$model->id = 'lyqxei40yhyw';
$model->title = 'Some custom HTML block';
$model->save();
```

you’ll get the error:

**Cannot have both "id" and "_id" fields.**

Also, if you try to query using id, it won’t return any results — because the query will look for _id, not id.

SMFloris:
I don't particularly use models, but this works:

```
        $dbCon = DB::connection('mongodb');
        $dbCon->table('collection_name_here')
[...]

PR #3476, opened 2026-02-10
Move field aliasing methods to Grammar class as public overridable API

dercoder:
"Changed the visibility of aliasIdForQuery from private to public. 
This allows developers to override the method or call it directly to normalize query data (e.g. id => _id) beforehand, resolving limitations like https://github.com/mongodb/laravel-mongodb/issues/3453

### Checklist

- [X] Add tests and ensure they pass

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -90,9 +90,9 @@ class Builder extends BaseBuilder
     /**
      * The database collection.
      *
-     * @var Collection
+     * @var \MongoDB\Collection
      */
-    protected Collection $collection;
+    protected $collection;
 
     /**
      * The column projections.
@@ -1391,7 +1391,7 @@ class Builder extends BaseBuilder
             $operator = $operator === 'like' ? '=' : 'not';
             // phpcs:ignore Squiz.ControlStructures.ControlSignature.SpaceAfterCloseBrace
         }
-        
+
         // Manipulate regex operations.
         elseif (in_array($operator, ['regex', 'not regex'])) {
             // Automatically convert regular expression strings to Regex objects.
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -8,14 +8,11 @@ use ArgumentCountError;
 use BadMethodCallException;
 use Carbon\CarbonPeriod;
 use Closure;
-use DateTimeInterface;
-use DateTimeZone;
 use Illuminate\Database\Query\Builder as BaseBuilder;
 use Illuminate\Database\Query\Expression;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Carbon;
 use Illuminate\Support\Collection;
-use Illuminate\Support\Facades\Date;
 use Illuminate\Support\LazyCollection;
 use InvalidArgumentException;
 use LogicException;
@@ -32,7 +29,6 @@ use MongoDB\Driver\ReadPreference;
 use MongoDB\Laravel\Connection;
 use Override;
 use RuntimeException;
-use stdClass;
 use TypeError;
 
 use function array_fill_keys;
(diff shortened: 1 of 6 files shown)


src/Query/Builder.php
```diff
@@ -1861,7 +1861,16 @@ public function orWhereIntegerNotInRaw($column, $values, $boolean = 'and')
         throw new BadMethodCallException('This method is not supported by MongoDB');
     }
 
-    private function aliasIdForQuery(array $values, bool $root = true): array
+    /**
+     * @internal
```
GromNaN: An `@internal` method is like a private method; you cannot call or override it. This defeats the purpose of this PR.

GromNaN (changes requested):
I don't think this is the appropriate extension point for this.

But if we decide to add the methods `aliasIdForQuery` and `aliasIdForResult` to the public API, we have to find better names for them.

An idea might be to move this logic to the Grammar class?

src/Query/Builder.php
dercoder: I agree, semantically it is contradictory. 
I only added it to match the existing convention found elsewhere in the repo (e.g. https://github.com/mongodb/laravel-mongodb/blob/5.x/src/Query/Builder.php#L1897). 

I’ll remove the tag to make the intention clear.

src/Query/Builder.php
```diff
@@ -84,24 +76,27 @@
...
     private const REGEX_DELIMITERS = ['/', '#', '~'];
 
     /**
      * The database collection.
      *
-     * @var \MongoDB\Collection
+     * @var Collection
      */
-    protected $collection;
+    protected Collection $collection;
```
GromNaN: Adding the property type is a breaking change in case anyone extends this class and override the property. Revert this change please.

src/Query/Builder.php
```diff
@@ -1386,9 +1381,7 @@ protected function compileWhereBasic(array $where): array
             // For inverse like operations, we can just use the $not operator with the Regex
             $operator = $operator === 'like' ? '=' : 'not';
             // phpcs:ignore Squiz.ControlStructures.ControlSignature.SpaceAfterCloseBrace
-        }
-
-        // Manipulate regex operations.
+        } // Manipulate regex operations.
```
GromNaN:
```suggestion
        }
        
        // Manipulate regex operations.
```

src/Query/Builder.php
dercoder: Yes, my bad but should we not set type here? This time the right one?

GromNaN: That's something to do for a major release. But I would wait for Laravel to do that in their codebase before.

dercoder:
In Laravel 11, the connection is optional in the Grammar class. Any suggestions here?

dercoder:
> In Laravel 11, the connection is optional in the Grammar class. Any suggestions here?

Laravel 12 passes the connection to Grammar via the constructor, but in Laravel 11 the constructor doesn't accept it 
-> setConnection() must be called explicitly. 

Added method_exists guards to support both versions.

GromNaN:
Thank you for implementing this refactoring!

The field aliasing logic now lives in `Query\Grammar` as public overridable methods, which is the right place for it. The implementation is a faithful extraction from `Builder`, and the 22 new tests in `GrammarTest` provide solid coverage.

A few notes on the review:
- `aliasIdForResult` was marked `@internal` on `Builder`, so removing it is not a breaking change.
- The `method_exists($grammar, 'setConnection')` guard in `Connection` correctly handles the difference between Laravel 11 (where `setConnection()` must be called explicitly) and Laravel 12 (where the connection is injected via the constructor).

Merging as-is.

GromNaN (approved):

PR #3480, opened 2026-02-11
Fix chunked order query bug

### LINKED ISSUE #3474: Incorrect sorting in MongoDB when using chunkById
- Laravel-mongodb Version: 12.47.0
- PHP Version: 8.4
- Database Driver & Version: 

### Description:
When using the `chunkById()` method with a model using the MongoDB driver, an incorrect sort query is generated with a duplicate `_id` key and a numeric key "0".

### Steps to reproduce
```php
Item::query()
    ->orderBy('id')
    ->chunkById(100, fn() => throw new \Exception('test'));
```

### Expected behaviour
The correct sort should look like this:
```json
{
  "find": "items",
  "filter": {
    "$and": [
      { "_id": { "$ne": null } },
      { "deleted_at": null }
    ]
  },
  "limit": { "$numberInt": "100" },
  "sort": {
    "_id": { "$numberInt": "1" }
  }
}
```

### Actual behaviour
```json
{
  "find": "items",
  "filter": {
    "$and": [
      { "_id": { "$ne": null } },
      { "deleted_at": null }
    ]
  },
  "limit": { "$numberInt": "100" },
  "sort": {
    "0": { "$numberInt": "1" },
    "_id": { "$numberInt": "1" }
  }
}
```

Issue: The sort section contains an incorrect key `"0"` along with _id.

### Root Cause
The problem arises from differences in the `orders` storage structure between Illuminate and MongoDB:

Illuminate: `orders` is an array of objects with `column` and `direction` keys
MongoDB: `orders` is an object (associative array)

In the `removeExistingOrdersFor()` method of the `Builder` class:

```php
protected function removeExistingOrdersFor($column)
{
    return (new Collection($this->orders))
        ->reject(fn ($order) => isset($order['column']) && $order['column'] === $column)
        ->values()
        ->all();
}
```

Problems:

* The `reject()` method doesn't filter orders correctly for MongoDB (since the data structure is different)
* Calling `values()` reindexes the array with numeric keys, causing the loss of original object keys in orders
* As a result, we get the key "0" instead of removing the existing sort

paulinevos:
Fix https://github.com/mongodb/laravel-mongodb/issues/3474

Illuminate's query builder ordering was conflicting with our own, resulting in incorrect ordering for chunked queries.

### Checklist

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -39,6 +39,7 @@ use function array_fill_keys;
 use function array_filter;
 use function array_is_list;
 use function array_key_exists;
+use function array_keys;
 use function array_map;
 use function array_merge;
 use function array_replace;
@@ -702,6 +703,33 @@ class Builder extends BaseBuilder
         }
 
         return $this;
+    }
+
+    /**
+     * Override Illuminate's removeExistingOrdersFor to support associative order storage used by MongoDB.
+     *
+     * @inheritdoc
+     */
+    #[Override]
+    protected function removeExistingOrdersFor($column): array
+    {
+        $orders = $this->orders ?? [];
+
+        $toUnset = array_filter(
+            array_keys($orders),
+            function ($orderColumn) use ($column) {
+                return $orderColumn === $column
+                    || ($orderColumn === 'id' && $column === '_id')
+                    || ($orderColumn === '_id' && $column === 'id'
+                );
+            },
+        );
+
+        foreach ($toUnset as $column) {
+            unset($orders[$column]);
+        }
+
+        return $orders;
     }
 
     /** @inheritdoc */
--- a/tests/Query/BuilderTest.php
+++ b/tests/Query/BuilderTest.php
@@ -531,6 +531,36 @@ class BuilderTest extends TestCase
                 ->orderBy('age', 1),
         ];
 
+        yield 'chunked ordering' => [
+            [
+                'find' => [
+                    ['_id' => ['$gt' => 0]],
+                    ['sort' => ['_id' => 1, 'name' => 1], 'limit' => 2],
+                ],
+            ],
+            function (Builder $builder) {
+                $builder->orderBy('_id')->orderBy('name');
+                $builder->forPageAfterId(2);
+
+                return $builder;
+            },
+        ];
+
+        yield 'chunked ordering with id alias' => [
+            [
+                'find' => [
+                    ['_id' => ['$gt' => 0]],
+                    ['sort' => ['_id' => 1, 'name' => 1], 'limit' => 2],
+                ],
+            ],
+            function (Builder $builder) {
+                $builder->orderBy('id')->orderBy('name');
+                $builder->forPageAfterId(2);
+
+                return $builder;
+            },
+        ];
+
         yield 'orderByDesc' => [
             ['find' => [[], ['sort' => ['email' => -1]]]],
             fn (Builder $builder) => $builder->orderByDesc('email'),


tests/Query/BuilderTest.php
```diff
@@ -531,6 +532,16 @@ function (Builder $builder) {
                 ->orderBy('age', 1),
         ];
 
+        yield 'chunked ordering' => [
+            ['find' => [[], ['sort' => ['_id' => 1]]]],
+            function (Builder $builder) {
+                $builder->orderBy('id');
+                $builder->chunkById(2, fn($results, $page) => var_dump($page));
```
GromNaN: Instead of testing `chunkById` which runs the query, you can call `$builder->forPageAfterId()` that configures the query builder without running it. That's closer to your fix.


```suggestion
                $builder->forPageAfterId();
```

src/Query/Builder.php
```diff
@@ -685,6 +704,41 @@ public function orderBy($column, $direction = 'asc')
...
+                ->all();
+        }
+
+        // Our standard shape: associative ['column' => direction]
+        $column = (string) $column;
+
+        // Remove direct key
+        if (array_key_exists($column, $orders)) {
+            unset($orders[$column]);
+        }
```
GromNaN: The condition is not necessary, you can remove an offset that is not set.
```suggestion
        unset($orders[$column]);
```

src/Query/Builder.php
```diff
@@ -685,6 +704,41 @@ public function orderBy($column, $direction = 'asc')
...
+
+        // Also remove alias variant between 'id' and '_id' if present
+        if ($column === 'id' && array_key_exists('_id', $orders)) {
+            unset($orders['_id']);
+        } elseif ($column === '_id' && array_key_exists('id', $orders)) {
+            unset($orders['id']);
+        }
```
GromNaN:
```suggestion
        if ($column === 'id') {
            unset($orders['_id']);
        } elseif ($column === '_id') {
            unset($orders['id']);
        }
```

tests/Query/BuilderTest.php
paulinevos: That worked perfect, thanks

tests/Query/BuilderTest.php
```diff
@@ -1689,6 +1700,6 @@ private function getBuilder(bool $renameEmbeddedIdField = true): Builder
         $connection->method('getSession')->willReturn(null);
         $connection->method('getQueryGrammar')->willReturn(new Grammar($connection));
 
-        return new Builder($connection, null, $processor);
+        return (new Builder($connection, null, $processor))->from(Birthday::class);
```
GromNaN: This change is not necessary.
```suggestion
        return new Builder($connection, null, $processor);
```

src/Query/Builder.php
```diff
@@ -685,6 +704,39 @@ public function orderBy($column, $direction = 'asc')
...
+    protected function removeExistingOrdersFor($column): array
+    {
+        $orders = $this->orders ?? [];
+
+        // If orders are in the base shape (list of arrays with 'column'), mimic base behavior sans values()
+        if ($orders && array_is_list($orders) && isset($orders[0]) && is_array($orders[0]) && array_key_exists('column', $orders[0])) {
+            return (new Collection($orders))
+                ->reject(fn ($order) => isset($order['column']) && $order['column'] === $column)
+                ->all();
+        }
```
GromNaN: The `_id`/`id` alias is not supported in that case here.

tests/Query/BuilderTest.php
paulinevos: Ah yeah still left over from before, whoops

GromNaN (approved):

PR #3491, opened 2026-04-07
Fix relation name passed to morphedByMany()

bskl:
Fix parameter name passed to morphToMany function.


GromNaN (approved):
Thank you @bskl
I retargeted to PR to 5.7, since it's a bugfix; and added a test.

PR #3496, opened 2026-04-14
Fix: Sync cancellation timestamps

zigzagdev:
### Summary
Refactored the cancel method to ensure that cancelled_at and finished_at share an identical timestamp. Previously, these fields were populated by separate method calls, potentially causing millisecond-level discrepancies.

### Changes
MongoBatchRepository.php
Captured a single UTC datetime instance into a $now variable.

Updated the $set operation to use $now for both fields, ensuring atomic consistency within the database update.

MongoBatchRepositoryTest.php
Added testCancelSetsSameTimestampForCheckedAtAndFinishedAt to verify the fix.

The test ensures that both fields are not only present but are also strictly identical at the BSON timestamp level.

### Technical Note
By unifying the timestamp, we eliminate non-deterministic behavior in the repository layer. This is particularly important for audit logs and state-machine transitions where "cancelled" and "finished" are considered to have occurred as part of the same logical event.

### Checklist

### Test Results
スクリーンショット 2026-04-14 22 45 30

### DIFF
--- a/src/Bus/MongoBatchRepository.php
+++ b/src/Bus/MongoBatchRepository.php
@@ -170,12 +170,14 @@ class MongoBatchRepository extends DatabaseBatchRepository implements PrunableBa
     public function cancel(string $batchId): void
     {
         $batchId = new ObjectId($batchId);
+
+        $now = $this->getUTCDateTime();
         $this->collection->updateOne(
             ['_id' => $batchId],
             [
                 '$set' => [
-                    'cancelled_at' => $this->getUTCDateTime(),
-                    'finished_at' => $this->getUTCDateTime(),
+                    'cancelled_at' => $now,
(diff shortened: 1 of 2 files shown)


GromNaN (approved):

GromNaN:
Thank you for the contribution! The fix is correct - using a single `$now` variable ensures both `cancelled_at` and `finished_at` are set to the identical timestamp, avoiding any potential millisecond discrepancy.

I applied `phpcbf` to fix minor code style issues (spacing around the `(string)` cast and trailing comma in the test).

zigzagdev:
@GromNaN 

Thank you for the review and for fixing the code style issues. I'll make sure to run phpcbf before my next push to align with the project's standards. 
Glad to hear the timestamp consistency fix was helpful!

PR #3499, opened 2026-04-15
Use DBX PR app to create merge-up pull requests

alcaeus:
As with https://github.com/mongodb/mongo-php-driver/pull/1972, this change modifies the merge-up workflow to no longer rely on a token, but rather uses the DBX PR app to create pull requests.

### DIFF
--- a/.github/workflows/merge-up.yml
+++ b/.github/workflows/merge-up.yml
@@ -5,22 +5,24 @@ on:
     branches:
       - '[0-9]+.[0-9x]+'
 
-env:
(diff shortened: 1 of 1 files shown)


.github/workflows/merge-up.yml
GromNaN, line 29: Would you move this action to mongodb-labs/drivers-github-tools too?

GromNaN (approved):

.github/workflows/merge-up.yml
alcaeus: I haven't yet decided where to put it, but gh-tools would be a good place.

PR #3501, opened 2026-04-15
Fix scalar type change not detected as dirty in MongoDB models

GromNaN:
Fixes a bug in `originalIsEquivalent()` where changing a field from `int(1)` to `string('1')` was not reported as dirty, even though MongoDB stores types as-is and the document would change on the next save.

## Changes

Replace the SQL-oriented logic (numeric coercion via `is_numeric` + `strcmp`, date/cast special cases) with BSON-aware comparison:
- Scalars: identical values are caught by the `===` early return; any non-identical scalar pair returns `false` (dirty).
- Complex types (UTCDateTime, arrays, embedded documents): compared via `Document::fromPHP()` serialization, which mirrors what MongoDB actually stores.

This is a breaking change: code that was relying on `int(1)` and `string('1')` being equivalent will now see them as dirty.

## Supersede #2515

This is a rework of #2515. The approach is the same but adapted to the current codebase (`DocumentModel` trait, `MongoDB\Laravel` namespace, `Document::fromPHP()` instead of the deprecated `fromPHP()` global function).


src/Eloquent/DocumentModel.php
```diff
@@ -363,22 +363,11 @@ public function originalIsEquivalent($key)
...
-        if ($this->hasCast($key, static::$primitiveCastTypes)) {
-            return $this->castAttribute($key, $attribute) ===
-                $this->castAttribute($key, $original);
+        if (is_scalar($attribute) || is_scalar($original)) {
+            return false;
         }
 
-        return is_numeric($attribute) && is_numeric($original)
-            && strcmp((string) $attribute, (string) $original) === 0;
+        return (string) Document::fromPHP(['v' => $attribute]) === (string) Document::fromPHP(['v' => $original]);
```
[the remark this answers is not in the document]
GromNaN: The same driver is used to encode both parts. No version issue.

src/Eloquent/DocumentModel.php
```diff
@@ -393,26 +392,11 @@ public function originalIsEquivalent($key)
...
-
-        if ($this->isClassCastable($key)) {
-            return ! is_object($attribute) ? $attribute === $original : $attribute == $original;
+        if (is_scalar($attribute) || is_scalar($original)) {
+            return false;
         }
 
-        return is_numeric($attribute) && is_numeric($original)
-            && strcmp((string) $attribute, (string) $original) === 0;
+        return (string) Document::fromPHP(['v' => $attribute]) === (string) Document::fromPHP(['v' => $original]);
```
paulinevos: I understand including these changes while we're at it, but it would be good to have this in a separate commit with an explanation. If anyone looks in the commit history trying to understand this change all they're gonna see is the thing about scalar types.

src/Eloquent/DocumentModel.php
```diff
@@ -392,6 +392,15 @@ public function originalIsEquivalent($key)
             return false;
         }
 
+        if ($this->isDateAttribute($key)) {
```
paulinevos: Ah okay, so his is added back in a subsequent commit. I would instead amend the prior commit that removes this.

src/Eloquent/DocumentModel.php
GromNaN: You're right, that's the thing I want to clarify. I'm not sure this is covered by a test

src/Eloquent/DocumentModel.php
```diff
@@ -389,30 +388,23 @@ public function originalIsEquivalent($key)
...
-        if ($this->isClassCastable($key)) {
-            return ! is_object($attribute) ? $attribute === $original : $attribute == $original;
+        if ($original instanceof DateTimeInterface) {
+            $original = new UTCDateTime($original);
         }
 
-        return is_numeric($attribute) && is_numeric($original)
-            && strcmp((string) $attribute, (string) $original) === 0;
+        // phpcs:disable SlevomatCodingStandard.Operators.DisallowEqualOperators.DisallowedEqualOperator
+        return Document::fromPHP(['v' => $attribute]) == Document::fromPHP(['v' => $original]);
```
alcaeus, line 440: Good idea encoding to BSON to check equality!

src/Eloquent/DocumentModel.php
```diff
@@ -389,30 +388,23 @@ public function originalIsEquivalent($key)
...
-            $original  = $original instanceof UTCDateTime ? $this->asDateTime($original) : $original;
-
-            // Comparison on DateTimeInterface values
-            // phpcs:disable SlevomatCodingStandard.Operators.DisallowEqualOperators.DisallowedEqualOperator
-            return $attribute == $original;
-        }
-
-        if ($this->hasCast($key, static::$primitiveCastTypes)) {
-            return $this->castAttribute($key, $attribute) ===
-                $this->castAttribute($key, $original);
```
alcaeus, line 400: Should casts be applied before we convert to BSON?

PR #3509, opened 2026-05-08
fix: Sync correct timestamps offset

zigzagdev:
### Summary                                                          
  Refactored `getNextAvailableJobAndReserve` to ensure that the `available_at` filter condition and the `reserved_at` update value share an identical timestamp. Previously, `Carbon::now()->getTimestamp()` was called twice separately, potentially causing slight discrepancies between the two values. 
  
### Changes
  `MongoQueue.php`
  - Captured a single timestamp into a `$now` variable.
  - Updated both the filter query and the $set operation to use $now, ensuring consistent timing within the atomic `findOneAndUpdate` operation. 
  
  
### Before

``` php
            [
                'queue' => $this->getQueue($queue),
                'reserved' => ['$ne' => 1],
                'available_at' => ['$lte' => Carbon::now()->getTimestamp()],
            ],
            [
                '$set' => [
                    'reserved' => 1,
                    'reserved_at' => Carbon::now()->getTimestamp(),
                ],
                '$inc' => ['attempts' => 1],
            ],
            
```

### After

```php
$now = Carbon::now()->getTimestamp();

            [
                'queue' => $this->getQueue($queue),
                'reserved' => ['$ne' => 1],
                'available_at' => ['$lte' => $now],
            ],
            [
                '$set' => [
                    'reserved' => 1,
                    'reserved_at' => $now,
                ],
               '$inc' => ['attempts' => 1],
            ],
```
  
### Checklist
スクリーンショット 2026-05-09 12 47 48

### Note
Closed to #3496

### DIFF
--- a/src/Queue/MongoQueue.php
+++ b/src/Queue/MongoQueue.php
@@ -78,16 +78,18 @@ class MongoQueue extends DatabaseQueue
      */
     protected function getNextAvailableJobAndReserve($queue)
     {
+        $now = Carbon::now()->getTimestamp();
+
         $job = $this->database->getCollection($this->table)->findOneAndUpdate(
             [
                 'queue' => $this->getQueue($queue),
                 'reserved' => ['$ne' => 1],
-                'available_at' => ['$lte' => Carbon::now()->getTimestamp()],
+                'available_at' => ['$lte' => $now],
             ],
             [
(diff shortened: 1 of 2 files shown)


GromNaN (approved):
Thank you @zigzagdev

Issue #3498, opened 2026-04-15
Bug: `increment()` and `decrement()` ignore `$amount` and `$extra` for embedded models


zigzagdev:
### Is your feature request related to a problem?

When calling `increment()` or `decrement()` on an embedded model, the `$amount` and `$extra` parameters are silently ignored. The current implementation retrieves the current value and saves it back unchanged, meaning the column value is never actually incremented or decremented.

**Affected Code:**
```php
public function increment($column, $amount = 1, array $extra = [])
{
    $relation = $this->model->getParentRelation();
    if ($relation) {
        $value = $this->model->{$column};
        $this->model->{$column} = null;
        $this->model->syncOriginalAttribute($column);
        return $this->model->update([$column => $value]); // $amount and $extra are never used
    }
    return parent::increment($column, $amount, $extra);
}
```

### Describe the solution you'd like

The `$amount` parameter should be added to the current value, and `$extra` fields should be included in the update.
```php
$value = $this->model->{$column} + $amount;
$this->model->{$column} = $value;
$this->model->syncOriginalAttribute($column);
return $this->model->update(array_merge([$column => $value], $extra));
```

### Describe alternatives you've considered

No alternatives considered. The fix is straightforward — apply `$amount` to the current value and pass `$extra` to the update call.

### Additional context

The same issue exists in the `decrement()` method. If this fix looks good, I'd be happy to open a pull request.

GromNaN:
We should ensure [`Illuminate\Database\Eloquent\Model::incrementOrDecrement()`](https://github.com/laravel/framework/blob/ff97ab481f3e2abf8b122e0294838649c8748ac2/src/Illuminate/Database/Eloquent/Model.php#L1108-L1109) is called.
Please open a PR. The tests should match Laravel tests on this feature (but I can't find them).

zigzagdev:
@GromNaN 

I've opened a PR to address this: #3510 

So, I'd appreciate it if you could review it when you get an opportunity.

zigzagdev:
Since the PR for #3510 has been merged, I'll go ahead and close this.

PR #3510, opened 2026-05-10
Fix: `increment` & `decrement` on embedded models to delegate to `Model::incrementOrDecrement()`

zigzagdev:
This PR is related to Issue(#3498)

### Why

The previous `Builder::increment()` and `Builder::decrement()` for embedded models manually set the column to `null`, called `syncOriginalAttribute()`, and ignored the `$extra` parameter. This bypassed `Model::incrementOrDecrement()`, which meant:

  - The `updating` / `updated` events were not fired
  - Extra attributes passed to `increment()`/`decrement()` were silently dropped
  - Custom amounts worked only by coincidence (computed by the caller, not the builder)

 So, this PR simplifies the embedded-model path to just persist the already-computed value via `$this->update()`, letting `Model::incrementOrDecrement()` handle value computation, event firing, and `syncOriginalAttribute()`.
 
 ### Changes

1.  **`Builder::increment()`/`decrement()`** — Remove manual attribute manipulation; forward `$extra` to `update()`
2. **Tests** — Add coverage for custom amounts, extra attributes, and `updating`/`updated` event assertions

### Notes
**`Builder`** — Move `use QueriesRelationships` before the constant declaration to follow PHP trait ordering conventions.

### Checklist(Test Results)

スクリーンショット 2026-05-10 19 01 31

Any feedback or suggestions are welcome — feel free to leave a comment.

### DIFF
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -22,6 +22,7 @@ use Override;
 
 use function array_key_exists;
 use function array_map;
+use function array_merge;
 use function array_replace;
 use function collect;
 use function is_array;
@@ -37,6 +38,7 @@ use function value;
 class Builder extends EloquentBuilder
 {
     use QueriesRelationships;
+
     private const DUPLICATE_KEY_ERROR = 11000;
 
     /**
--- a/src/Eloquent/Builder.php
+++ b/src/Eloquent/Builder.php
@@ -36,8 +36,8 @@ use function value;
  */
 class Builder extends EloquentBuilder
 {
-    private const DUPLICATE_KEY_ERROR = 11000;
     use QueriesRelationships;
+    private const DUPLICATE_KEY_ERROR = 11000;
 
     /**
      * The methods that should be returned from query builder.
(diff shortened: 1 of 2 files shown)


src/Eloquent/Builder.php
```diff
@@ -201,16 +203,7 @@ public function increment($column, $amount = 1, array $extra = [])
...
-            $value = $this->model->{$column};
-
-            // When doing increment and decrements, Eloquent will automatically
-            // sync the original attributes. We need to change the attribute
-            // temporary in order to trigger an update query.
-            $this->model->{$column} = null;
-
-            $this->model->syncOriginalAttribute($column);
-
-            return $this->model->update([$column => $value]);
```
GromNaN, line 213: This was introduced by 415ed78e3fbd1b621c30937410f2a28b7f116d50, the test that was added is still successful.

GromNaN (approved):
LGTM, thank you!

src/Eloquent/Builder.php
zigzagdev: @GromNaN 

Thanks for the comment ! 
I wasn't aware of the original intent from [#415ed78](https://github.com/mongodb/laravel-mongodb/commit/415ed78e3fbd1b621c30937410f2a28b7f116d50).

Should I revert this deletion to preserve the original behavior, or is removing it fine given that the test still passes?

zigzagdev:
Hi @paulinevos  , just checking in on this PR. 

Let me know if any changes are needed!

paulinevos:
@zigzagdev @GromNaN will have another look at your outstanding question today and see if we can include it in the upcoming release

src/Eloquent/Builder.php
GromNaN: Your change seems correct.

GromNaN (approved):

GromNaN:
Thank you @zigzagdev

zigzagdev:
@paulinevos 
I apologise if I came across as pushy — that wasn't my intention. I look forward to continuing to work with you.

@GromNaN 
Thank you for merging my PR! 🙌

paulinevos:
@zigzagdev Don't worry, it didn't ;) Thank you for contributing

PR #3514, opened 2026-05-25
Fix: prevent `getCastType` from mutating model-wide dateFormat

zigzagdev:
### Description
  - `getCastType()` called `setDateFormat()` as a side effect, which mutated the model-wide `$dateFormat` property. When a model had multiple datetime casts with different formats (e.g.`datetime:Y-m-d` and `datetime:U`), accessing one attribute would overwrite the format used by all other attributes.
  - Removed the `getCastType()` override and moved the per-attribute format resolution into `transformModelValue()`.

### Add New Test
 - Added a new `datetimeWithAnotherFormatField` (`datetime:Y-m-d\TH:i:s`) to the `Casting` test model.
 - Added `testMultipleDatetimeFormatsDoNotInterfere` to verify that accessing multiple datetime attributes with different formats does not cause interference in either access order.
 
 ### Test Cases Result
 
スクリーンショット 2026-05-25 11 56 53


GromNaN (approved):
Thank you for this fix!

The root cause is confirmed: `getCastType()` was mutating the model-wide `$dateFormat` as a side effect, but only for custom-format datetime casts. Accessing a plain `datetime` cast after a `datetime:format` cast would pick up the wrong format in `transformModelValue()`.

I added a regression test (`testCustomFormatDoesNotContaminateDefaultDatetimeFormat`) that specifically demonstrates this: it fails on the old code and passes with the fix. The existing test was not catching this because both attributes in it have custom formats, so each `getCastType()` call correctly resets the format before use.

PR #3517, opened 2026-06-08
PHPORM-494 Fix `MongoStore::unserialize()` for empty arrays and objects

### LINKED ISSUE #3515: Bug: MongoDB\Laravel\Cache\MongoStore::unserialize() doesn't unserialise cached empty array
- Laravel-mongodb Version: 5.7.0
- PHP Version: 8.5.4
- Database Driver & Version:

## Summary
The MongoDB cache store fails to properly deserialize cached empty arrays, returning the serialized string `"a:0:{}"` instead of an actual empty array `[]` when retrieved from a warm cache.

## Description
When caching an empty array using `Cache::remember()` or similar cache methods with the MongoDB cache driver, the cached value is returned as the serialized string `"a:0:{}"` on subsequent cache hits (within the TTL), rather than being properly deserialized back into an empty array.

### Root Cause
The `MongoStore::unserialize()` method (line 315–320) checks for the presence of a semicolon to determine whether a value needs unserialization:

```php
private function unserialize($value): mixed
{
    if (! is_string($value) || ! str_contains($value, ';')) {
        return $value;
    }

    return unserialize($value);
}
```

The serialized representation of an empty array (`a:0:{}`) does not contain a semicolon, so the check fails and the raw string is returned instead of being unserialized.

## Steps to Reproduce

```php
// In a service or controller using Cache::remember()
$result = Cache::remember('testEmptyArray', 10, function () {
    return []; // Return an empty array
});

// First call (cache miss): returns []
var_dump($result); // array(0) { }

// Second call within TTL (cache hit): returns serialized string
var_dump($result); // string(7) "a:0:{}"
```

## Expected Behavior
Both the first and subsequent cache retrievals should return an actual empty array `[]`, not a serialized string representation.

## Actual Behavior
- First call (cache miss): Returns `[]` as expected
- Subsequent calls (cache hit, within TTL): Returns `"a:0:{}"` as a string

GromNaN:
Fix `MongoStore::unserialize()` returning the raw serialized string instead of the value when the cache contains an empty array (`[]`) or an empty object.

## Root Cause

`MongoStore::serialize()` stores integers and floats as native numeric values; all other types go through PHP's `serialize()`. Any string value in the collection is therefore always a PHP-serialized payload.

The `str_contains($value, ';')` guard was meant to detect serialized strings, but PHP serializes empty arrays as `a:0:{}` and empty objects as `O:8:"stdClass":0:{}` — neither contains a semicolon, so unserialization was skipped.

The guard is removed. `insertToCacheTable()` in the test helper is also fixed to serialize values the same way `put()` does, which was the original cause of the workaround.

Fixes https://github.com/mongodb/laravel-mongodb/issues/3515
Jira: https://jira.mongodb.org/browse/PHPORM-494

### DIFF
--- a/src/Cache/MongoStore.php
+++ b/src/Cache/MongoStore.php
@@ -16,7 +16,6 @@ use function is_float;
 use function is_int;
 use function is_string;
 use function serialize;
-use function str_contains;
 use function unserialize;
 
 final class MongoStore implements LockProvider, Store
@@ -312,7 +311,7 @@ final class MongoStore implements LockProvider, Store
 
     private function unserialize($value): mixed
     {
-        if (! is_string($value) || ! str_contains($value, ';')) {
+        if (! is_string($value)) {
             return $value;
         }
 
--- a/tests/Cache/MongoCacheStoreTest.php
+++ b/tests/Cache/MongoCacheStoreTest.php
@@ -2,15 +2,22 @@
 
 namespace MongoDB\Laravel\Tests\Cache;
 
+use Generator;
 use Illuminate\Cache\Repository;
 use Illuminate\Support\Carbon;
 use Illuminate\Support\Facades\Cache;
 use Illuminate\Support\Facades\DB;
 use MongoDB\BSON\UTCDateTime;
 use MongoDB\Laravel\Tests\TestCase;
+use PHPUnit\Framework\Attributes\DataProvider;
+use stdClass;
 
 use function assert;
 use function config;
+use function get_debug_type;
+use function is_float;
+use function is_int;
+use function serialize;
 
 /** Tests imported from {@see \Illuminate\Tests\Integration\Database\DatabaseCacheStoreTest} */
 class MongoCacheStoreTest extends TestCase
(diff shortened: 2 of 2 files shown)


src/Cache/MongoStore.php
```diff
@@ -312,7 +311,7 @@ private function serialize($value): string|int|float
 
     private function unserialize($value): mixed
     {
-        if (! is_string($value) || ! str_contains($value, ';')) {
```
paulinevos, line 315: Wait so this guard was only here to be able to cover a test case? Am I understanding correctly?

GromNaN: Essentially yes. The test helper `insertToCacheTable()` was inserting raw (non-serialized) strings directly into the collection, bypassing `MongoStore::serialize()`. The guard was compensating for that.

In production, `MongoStore::put()` always passes non-numeric values through PHP's `serialize()`, so any string in the collection is always a PHP-serialized payload. The guard was never needed for production code — it just happened to work around the flawed test helper.

The fix updates `insertToCacheTable()` to mirror the same logic as `MongoStore::serialize()`: store ints and floats as-is, and serialize everything else.

paulinevos (approved):

PR #3525, opened 2026-06-11
Fix: Recurse into `stdClass` values in `prepareFieldsForQuery`

zigzagdev:
### Problem
In `Grammar.php`, `prepareFieldsForQuery()` method did not recurse into `stdClass` values. 
This meant that fields inside `stdClass` objects were not processed, causing `DateTimeInterface` values to remain unconverted and `id` fields to not be renamed to `_id`.

### What I did for fixing
Added handling for `stdClass` values in `prepareFieldsForQuery()`. 
When a `stdClass` value is encountered, it is converted to an array, recursively processed, and converted back to `stdClass`.

### Checklist

### Test Results
スクリーンショット 2026-06-11 20 16 13

**Notes**
`testPrepareFieldsForQueryProcessesStdClassWithConfigDisabled` test manually resets `renameEmbeddedIdField` to `true` at the end of the test instead of using `tearDown()`.
This is intentional to avoid modifying existing `tearDown()` logic in a separate PR.
If you prefer to consolidate the reset into `tearDown()`, please let me know !

### DIFF
--- a/src/Query/Grammar.php
+++ b/src/Query/Grammar.php
@@ -88,6 +88,13 @@ class Grammar extends BaseGrammar
                 $value = $this->prepareFieldsForQuery($value, false);
             } elseif ($value instanceof DateTimeInterface) {
                 $value = new UTCDateTime($value);
+            } elseif ($value instanceof stdClass) {
+                $value = $this->prepareFieldsForQuery(
+                    get_object_vars($value),
+                    false,
+                );
+
+                $value = (object) $value;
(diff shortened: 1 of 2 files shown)


GromNaN (approved):
Thank you for this contribution!

The fix is correct and symmetric with `prepareFieldsForResult()`, which already handles `stdClass` objects. I verified the regression: the new tests fail without the fix and pass with it.

PR #3533, opened 2026-06-20
fix: `MongoLock` lottery never pruned expired locks

zigzagdev:
### Content
The lottery condition in `MongoLock::acquire()` was logically inverted, because of expired lock pruning to never execute.

### Current the problems

**First Problem**
The original condition `$this->lottery[0] <= 0` is `true` only when the lottery is disabled (chance = 0). 
Combined with `random_int(1, total) <= 0`, which can never be true since `random_int` returns a minimum of 1, the pruning block was unreachable under all circumstances.

**Second Problem**
Because of  a `ValueError` from `random_int(1, 0)` due to PHP requiring min ≤ max. 
The fix short-circuits before calling `random_int` in this case.

### What I have changed
- Fix lottery condition from `<= 0` to `> 0` so pruning executes with the correct probability.
- Add unit tests to verify pruning runs when lottery hits (`[1, 1]`) and does not run when lottery is disabled (`[0, 0]`).

### Test Results
スクリーンショット 2026-06-20 14 32 40

### Checklist

### DIFF
--- a/src/Cache/MongoLock.php
+++ b/src/Cache/MongoLock.php
@@ -81,7 +81,7 @@ final class MongoLock extends Lock
             ],
         );
 
-        if ($this->lottery[0] <= 0 && random_int(1, $this->lottery[1]) <= $this->lottery[0]) {
+        if ($this->lottery[0] > 0 && random_int(1, $this->lottery[1]) <= $this->lottery[0]) {
             $this->collection->deleteMany(['expires_at' => ['$lte' => $this->getUTCDateTime()]]);
         }
 
--- a/tests/Cache/MongoLockTest.php
+++ b/tests/Cache/MongoLockTest.php
@@ -39,6 +39,32 @@ class MongoLockTest extends TestCase
         );
     }
 
+    public function testLotteryPrunesExpiredLocks(): void
+    {
+        $collection = $this->createMock(Collection::class);
+        $collection->expects($this->once())
+            ->method('findOneAndUpdate')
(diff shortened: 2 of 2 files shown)


GromNaN (approved):
Thank you for this fix!

The bug is real and the fix is correct. The original condition `$this->lottery[0] <= 0` was inverted: combined with `random_int(1, N) <= 0` (which can never be true since `random_int` returns at least 1), the pruning block was completely unreachable. The replacement `$this->lottery[0] > 0` is the right guard.

The fix also resolves a secondary issue: with a `[0, 0]` lottery (MongoLock's documented way to disable pruning), the original code would reach `random_int(1, 0)` and throw a `ValueError`. The `> 0` guard short-circuits before that call.

Note: Laravel's `DatabaseLock` disables pruning by passing `null` to the lottery parameter (guarded by `count($this->lottery ?? []) === 2`), which is a different design. MongoLock uses `[0, 0]` instead, so the `> 0` guard is the right equivalent here.

**Test coverage**: I verified both new tests regress against the unfixed code as expected:
- `testLotteryPrunesExpiredLocks` fails because `deleteMany` is never called (pruning was broken)
- `testDisabledLotteryDoesNotPruneExpiredLocks` throws a `ValueError` from `random_int(1, 0)` (the second bug)

All 12 tests pass with the fix applied.

zigzagdev:
@GromNaN 

Thank you for reviewing my PR and approving it !

PR #3534, opened 2026-06-20
feat(scout): support sorting by relevance score (`_score`) in ScoutEngine

zigzagdev:
## Summary

Adds support for `orderBy('_score', 'desc')` in the MongoDB Laravel Scout engine, mapping it to Atlas Search's relevance score sort (`$meta: searchScore`) instead of treating it as a regular field sort.

### Current Problem

Previously, `$sort[$order['column']] = $order['direction'] === 'asc' ? 1 : -1;` treated every order column the same way, including `_score`.
Atlas Search does not support sorting by a numeric `1`/`-1` direction on `_score` — relevance score sorting requires the dedicated `$meta: searchScore` syntax.
As a result, `orderBy('_score', 'desc')` would silently produce an invalid sort and not work correctly against Atlas Search.

### What I have changed

- In `performSearch()`, the sort-building loop now branches on the order column:
  - If the column is `_score`, it maps to `'score' => ['$meta' => 'searchScore']`.
  - Otherwise, it falls back to the existing `1`/`-1` direction sort.
- `orderBy('_score', 'asc')` throws an `InvalidArgumentException` because `$meta: searchScore` does not support ascending direction — only descending is valid in Atlas Search.
- Sorting by `_score` together with a field named `score` throws an `InvalidArgumentException` because both map to the same key in the Atlas Search sort stage.
- Added unit test coverages:
  - `ordered by score`: verifies `_score` alone maps to the relevance sort.
  - `ordered by score and field`: verifies `_score` and a regular field can be combined in the same sort.
  - `testSearchRejectsAscendingScoreSort`: verifies that `orderBy('_score', 'asc')` throws.
  - `testSearchRejectsScoreFieldWithScoreSort`: verifies the `score` field name collision throws.
- Added Atlas Search integration tests comparing the Scout result order against a reference raw aggregation pipeline using `$meta: searchScore` directly.

### Test Results
screenshot 2026-06-20 17 58 05

### Checklist

### DIFF
--- a/src/Scout/ScoutEngine.php
+++ b/src/Scout/ScoutEngine.php
@@ -260,10 +260,15 @@ final class ScoutEngine extends Engine
             $compound['mustNot'][] = ['in' => ['path' => $field, 'value' => $value]];
         }
 
-        // Sort by field value only if specified
+        // Sort by field value only if specified.
+        // '_score' maps to Atlas Search's relevance score via $meta: searchScore (always descending).
         $sort = [];
         foreach ($builder->orders as $order) {
-            $sort[$order['column']] = $order['direction'] === 'asc' ? 1 : -1;
+            if ($order['column'] === '_score') {
+                $sort['score'] = ['$meta' => 'searchScore'];
+            } else {
+                $sort[$order['column']] = $order['direction'] === 'asc' ? 1 : -1;
+            }
         }
 
         $pipeline = [
--- a/src/Scout/ScoutEngine.php
+++ b/src/Scout/ScoutEngine.php
@@ -270,6 +270,10 @@ final class ScoutEngine extends Engine
         $sort = [];
         foreach ($builder->orders as $order) {
             if ($order['column'] === '_score') {
+                if ($order['direction'] === 'asc') {
+                    throw new InvalidArgumentException("Cannot sort by '_score' in ascending order; Atlas Search relevance score always sorts descending.");
+                }
+
                 $sort['score'] = ['$meta' => 'searchScore'];
             } else {
                 $sort[$order['column']] = $order['direction'] === 'asc' ? 1 : -1;
--- a/src/Scout/ScoutEngine.php
+++ b/src/Scout/ScoutEngine.php
@@ -198,6 +198,18 @@ final class ScoutEngine extends Engine
      */
     private function performSearch(Builder $builder, ?int $offset = null): array
     {
+        // Validate sort before any I/O so unit tests can assert on these exceptions without a real collection.
+        $columns = array_column($builder->orders, 'column');
+        if (in_array('_score', $columns, true) && in_array('score', $columns, true)) {
+            throw new InvalidArgumentException("Cannot sort by a field named 'score' together with Atlas Search's '_score' relevance sort.");
+        }
+
+        foreach ($builder->orders as $order) {
+            if ($order['column'] === '_score' && $order['direction'] === 'asc') {
+                throw new InvalidArgumentException("Cannot sort by '_score' in ascending order; Atlas Search relevance score always sorts descending.");
+            }
+        }
+
         $collection = $this->getSearchableCollection($builder->model);
 
         if ($builder->callback) {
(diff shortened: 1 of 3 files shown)


src/Scout/ScoutEngine.php
```diff
@@ -260,10 +260,15 @@ private function performSearch(Builder $builder, ?int $offset = null): array
...
+        // Sort by field value only if specified.
+        // '_score' maps to Atlas Search's relevance score via $meta: searchScore (always descending).
         $sort = [];
         foreach ($builder->orders as $order) {
-            $sort[$order['column']] = $order['direction'] === 'asc' ? 1 : -1;
+            if ($order['column'] === '_score') {
+                $sort['score'] = ['$meta' => 'searchScore'];
+            } else {
+                $sort[$order['column']] = $order['direction'] === 'asc' ? 1 : -1;
+            }
```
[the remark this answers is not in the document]
zigzagdev: I fixed [it](https://github.com/mongodb/laravel-mongodb/pull/3534/commits/37967ce383713e774ad18e581e63f54b8a6a17d0).

tests/Scout/ScoutEngineTest.php
```diff
@@ -404,6 +404,44 @@ function () {
                 ],
             ]),
         ];
+
+        yield 'ordered by score' => [
+            function () {
+                $builder = new Builder(new SearchableModel(), 'lar');
+                $builder->orderBy('_score', 'desc');
```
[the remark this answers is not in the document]
zigzagdev: Thanks for the suggestion, but I'll leave this as-is.

Because your proposed changeset doesn't actually verify direction-independence — calling orderBy('_score', 'desc') followed by `orderBy('_score', 'asc')` on the same builder just adds two order entries for the same column rather than testing whether asc alone is ignored. 
Given the existing test already exercises the always-descending behavior implicitly (via desc), and the suggested code wouldn't isolate the regression it's meant to catch, I don't think a separate assertion is warranted here.

zigzagdev (commented):
I replied to all reviews.

src/Scout/ScoutEngine.php
```diff
@@ -260,10 +260,20 @@ private function performSearch(Builder $builder, ?int $offset = null): array
...
+        $columns = array_column($builder->orders, 'column');
+        if (in_array('_score', $columns, true) && in_array('score', $columns, true)) {
+            throw new InvalidArgumentException("Cannot sort by a field named 'score' together with Atlas Search's '_score' relevance sort.");
+        }
+
         $sort = [];
         foreach ($builder->orders as $order) {
-            $sort[$order['column']] = $order['direction'] === 'asc' ? 1 : -1;
+            if ($order['column'] === '_score') {
+                $sort['score'] = ['$meta' => 'searchScore'];
```
paulinevos, line 280: I think aside from the comment, we should explicitly validate that no ascending order is passed here. Otherwise a user may expect an ascending score and the behavior will be silently different.

zigzagdev: @paulinevos 

Good point ! 
I [added](https://github.com/mongodb/laravel-mongodb/pull/3534/commits/5a12809815a85c2a45519538f760423437a65cea) explicit validation that throws an InvalidArgumentException when _score is passed with ascending order.

paulinevos: You have two failing test cases on that now but other than that it looks good to me :)

GromNaN (approved):
The implementation looks correct. I've added two Atlas Search integration tests on the PR branch (`testItCanSortByRelevanceScore` and `testItCanSortByRelevanceScoreAndField`).

Both tests follow the same pattern: run the Scout query with `orderBy('_score', 'desc')` and compare the returned ID order against a reference raw aggregation pipeline that applies `{ score: { $meta: searchScore } }` sort directly. This makes the tests sensitive to the actual sort behavior — if ScoutEngine generates the wrong sort syntax (e.g. `{ _score: -1 }`), Atlas Search silently accepts it but returns a different order, causing the assertion to fail.

A few notes for the implementation:

- The `orderBy('_score', 'asc')` exception is a valid strict choice, but the PR description says it "still sorts descending" (implying silent fallback). Please update the description to match the actual behavior (throws).
- The `score` field name collision check leaks an internal mapping detail into the user-facing error. Worth noting in the docs or error message that `_score` is internally mapped to the key `score` in the Atlas Search sort stage.

PR #3540, opened 2026-07-06
PHPLARA-81 Add AsBsonDocument and AsBsonArray casts

GromNaN:
Introduces two Eloquent casts that store attributes as **native BSON** documents and arrays instead of JSON-encoded strings.

- `MongoDB\Laravel\Eloquent\Casts\AsBsonDocument` returns a `MongoDB\Model\BSONDocument` on read.
- `MongoDB\Laravel\Eloquent\Casts\AsBsonArray` returns a `MongoDB\Model\BSONArray` on read.

```php
use MongoDB\Laravel\Eloquent\Casts\AsBsonArray;
use MongoDB\Laravel\Eloquent\Casts\AsBsonDocument;

class Product extends Model
{
    protected \$casts = [
        'specs' => AsBsonDocument::class,
        'variants' => AsBsonArray::class,
    ];
}

\$product->specs['storage']['primary'] = '1TB SSD';   // nested mutation
\$product->variants[] = ['color' => 'Red'];           // append
\$product->save();
```

## Benefits

Same ergonomics as [Laravel's `AsArrayObject` / `AsCollection`](https://laravel-news.com/as-arrayobject-and-as-collection-custom-casts):
- Mutation of nested paths without full reassignment (`\$model->specs['a']['b'] = 'x'`), thanks to `ArrayObject` reference semantics.
- Automatic dirty tracking on nested mutations.
- Property-style access (`\$doc->foo`) in addition to array-style.

Plus **MongoDB-specific advantages that JSON string storage cannot deliver**:
- Fields are queryable and indexable by nested path (`specs.storage.primary`).
- Partial updates via `\$set`, `\$push`, `\$inc`, etc. work naturally.
- Aggregation pipelines can project, filter, and group on nested fields.
- Storage size is smaller than an escaped JSON string.
- No JSON encode/decode overhead on every read/write.

## Why not `Illuminate\Database\Eloquent\Casts\AsArrayObject` and `AsCollection`?

**Avoid these casts on MongoDB models.** Their `castUsing()` returns an anonymous class that calls `Json::encode(\$value)` in `set()` and `Json::decode(\$attributes[\$key])` in `get()`. Values are stored as JSON strings in MongoDB, not as BSON documents. This defeats every native MongoDB capability listed above: no indexing on nested fields, no `\$set` on `specs.storage.primary`, no aggregations on inner values, and larger on-disk size.

Because the encoding is hardcoded inside anonymous classes returned by `castUsing()`, there is no clean override point for a MongoDB-friendly variant. New casts are required.

The same issue affects `AsEnumArrayObject`, `AsEnumCollection`, `AsEncryptedArrayObject`, `AsEncryptedCollection`. Follow-ups may add BSON-native equivalents.

## Dirty tracking

`AsBsonDocument` and `AsBsonArray` implement `compare(\$model, \$key, \$original, \$value)` that normalizes both operands via `MongoDB\BSON\Document::fromPHP()` and compares the serialized BSON bytes. This means:
- Reassigning an equivalent-but-different container (e.g. array vs `BSONDocument` with the same content) is not dirty.
- Round-trip BSON types are compared correctly.

`DocumentModel::originalIsEquivalent()` now consults `isClassComparable()` before the loose-equality fallback, so casts opting in to `compare()` get honored.

## Related

Complementary to #3501, which generalizes BSON-aware comparison across all attribute types.

### DIFF
--- a/src/Eloquent/Casts/AsBsonArray.php
+++ b/src/Eloquent/Casts/AsBsonArray.php
@@ -1,0 +1,68 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Laravel\Eloquent\Casts;
+
+use Illuminate\Contracts\Database\Eloquent\Castable;
+use Illuminate\Contracts\Database\Eloquent\CastsAttributes;
+use MongoDB\BSON\Document;
+use MongoDB\Model\BSONArray;
+
+use function is_array;
+
+class AsBsonArray implements Castable, CastsAttributes
+{
+    public static function castUsing(array $arguments): static
+    {
+        return new static();
+    }
+
+    public function get($model, string $key, $value, array $attributes)
+    {
+        if ($value === null) {
+            return null;
+        }
+
+        if ($value instanceof BSONArray) {
+            return clone $value;
+        }
+
+        if (! is_array($value)) {
+            return null;
+        }
+
+        return new BSONArray($value);
+    }
+
+    public function set($model, string $key, $value, array $attributes)
+    {
+        if ($value === null) {
+            return [$key => null];
+        }
+
+        if ($value instanceof BSONArray) {
+            return [$key => clone $value];
+        }
+
+        return [$key => new BSONArray((array) $value)];
+    }
+
+    public function compare($model, string $key, $original, $value): bool
+    {
+        return self::normalize($original) === self::normalize($value);
+    }
+
+    private static function normalize($value): ?string
+    {
+        if ($value === null) {
+            return null;
+        }
+
+        if ($value instanceof BSONArray) {
+            $value = $value->getArrayCopy();
+        }
+
+        return (string) Document::fromPHP(['v' => (array) $value]);
+    }
+}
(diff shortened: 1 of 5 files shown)


src/Eloquent/Casts/AsBsonArray.php
```diff
@@ -0,0 +1,68 @@
...
+class AsBsonArray implements Castable, CastsAttributes
+{
+    public static function castUsing(array $arguments): static
+    {
+        return new static();
+    }
+
+    public function get($model, string $key, $value, array $attributes)
+    {
+        if ($value === null) {
```
paulinevos: This seems superfluous considering the `is_array` check on L31.

src/Eloquent/Casts/AsBsonArray.php
```diff
@@ -0,0 +1,68 @@
...
+            return [$key => null];
+        }
+
+        if ($value instanceof BSONArray) {
+            return [$key => clone $value];
+        }
+
+        return [$key => new BSONArray((array) $value)];
```
paulinevos, line 40: What if `$value` isn't castable to array?

paulinevos (commented):
Taking this PR over. Comments are for my own info! pls disregard.

src/Eloquent/Casts/AsBsonArray.php
paulinevos: A: apparently Laravel doesn't care in the Laravel equivalent, so we don't care either for consistency.

paulinevos (approved):

PR #3545, opened 2026-07-10
PHPLARA-247: Add plugin install suggestion to Boost guidelines

paulinevos:
Create minimal third-party guidelines for Boost users that point to the official MongoDB agent-skills plugin (only if not already installed and if the user requests MongoDB related things not already supported by Boost)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,11 @@
+## MongoDB-related coding assistance
+
+If the user needs MongoDB-related skills that aren't otherwise available through e.g. Boost, such as:
+
+- Inspecting Atlas system resources
+- Managing an Atlas cluster
+- Analysing MongoDB schema anti-patterns
+- Migrating an SQL schema to MongoDB
+- Setting up the official MongoDB MCP server
+
+Suggest installing the MongoDB agent skills plugin if not already installed. Instructions for installing it are in the [MongoDB agent skills plugin README](https://github.com/mongodb/agent-skills/blob/main/README.md).


paulinevos:
I'm actually not convinced this is Boost-specific. Maybe this should be a general guideline.

AGENTS.md
kevinAlbs, line 1: To check: should the extension `.md` rather than `.blade.php` to match contents?

https://laravel.com/docs/13.x/boost#adding-custom-ai-guidelines notes either extension.

AGENTS.md
```diff
@@ -0,0 +1,15 @@
...
+This application uses the [MongoDB Laravel](https://github.com/mongodb/laravel-mongodb) package, so it likely has one or more `mongodb` driver connections. Boost's database tools assume a SQL database and won't work against those.
+
+## Other database operations
+
+If the user needs MongoDB-related skills that Boost doesn't offer offer, such as:
+
+- Inspecting Atlas system resources,
+- Managing an Atlas cluster,
+- Analysing MongoDB schema anti-patterns,
+- Migrating an SQL schema to MongoDB
```
kevinAlbs, line 8: Comma for consistency with other bullets:

```suggestion
- Migrating an SQL schema to MongoDB,
```

kevinAlbs (approved):
LGTM with a possible file rename and a (likely unnecessarily) comma for consistency.

AGENTS.md
paulinevos: To match contents? What do you mean? Either way, both extensions work and will be picked up.

kevinAlbs: I assumed the contents were markdown rather than PHP.

> both extensions work and will be picked up.

Ack. Resolving.

AGENTS.md
GromNaN: This file targets LLM, I bet the ponctuation is unnecessary at the end of lines for a list.

PR #3546, opened 2026-07-15
PHPLARA-100: Create agent skills

paulinevos:
Adds agent skills to the Laravel integration. This is basically @GromNaN's PR to the mongodb agent-skills repo but we decided to have it live here instead.

Also copies over necessary skills review workflow.


AGENTS.md
GromNaN, line 1: Should you add a chapter in this file for skill maintenance. I want the agents to be reminded of updating the skill when there is a new feature.

paulinevos: That sounds like a good idea, but maybe OOS? Let me know if I should add a Jira issue for this, because I feel like we might want to decide how that should work. Maybe it could be another, separate skill that includes opening a PR. It's also a little confusing to me as to when an AGENTS.md file targets users of the library and when it targets development of the library.

paulinevos: Just looked it up. As far as I understand, the standard is to use a dual-file architecture. The root-level guidelines would be for internal development, and the nested level stuff is for end users of the library. Will change.

skills/laravel-mongodb/references/search-engine.md
```diff
@@ -0,0 +1,133 @@
...
+        'mappings' => [
+            'dynamic' => false,
+            'fields'  => [
+                'name'        => ['type' => 'string'],
+                'description' => ['type' => 'string'],
+                'tags'        => ['type' => 'string'],
+                'price'       => ['type' => 'number'],
+            ],
+        ],
+    ]);
```
kevinAlbs:
```suggestion
    ], 'scout');
```

Following a suggestion from Claude:

> `$c->searchIndex([...])` with no name argument creates an index named `default` [...] while the Scout engine only ever queries/manages an index named scout (`ScoutEngine.php:56`, `INDEX_NAME = 'scout'`). So the manually created index won't be used by `Product::search()`

skills/laravel-mongodb/references/vector-search.md
```diff
@@ -0,0 +1,207 @@
...
+                    ['type' => 'filter', 'path' => 'category'],
+                ],
+            ], 'products_vector');
+        });
+    }
+};
+```
+
+Available models: `voyage-4`, `voyage-4-large`, and others from MongoDB-supported providers.
+Optional fields: `numDimensions`, `quantization` (`float`|`scalar`|`binary`), `similarity` (`euclidean`|`cosine`|`dotProduct`), `indexingMethod` (`flat`|`hnsw`), `hnswOptions`.
```
kevinAlbs: Suggested by Claude since `binaryNoRescore` is also [documented](https://github.com/mongodb/laravel-mongodb/blob/f8ce003dfbbd39ca49ba29a54369b2538cc066bd/src/Schema/Blueprint.php#L117).

```suggestion
Optional fields: `numDimensions`, `quantization` (`float`|`scalar`|`binary`|`binaryNoRescore`), `similarity` (`euclidean`|`cosine`|`dotProduct`), `indexingMethod` (`flat`|`hnsw`), `hnswOptions`.
```

skills/laravel-mongodb/references/connection.md
```diff
@@ -0,0 +1,68 @@
...
+        ],
+    ],
+];
+```
+
+## `.env`
+
+```
+DB_CONNECTION=mongodb
+MONGODB_URI="mongodb+srv://user:pass@cluster0.mongodb.net/?retryWrites=true&w=majority"
```
kevinAlbs:
```suggestion
MONGODB_URI="mongodb+srv://user:pass@cluster0.mongodb.net/"
```

Minor: suggest removing `retryWrites=true&w=majority` as I expect these are unnecessary. IIRC these were previous recommendations for drivers. Quoting DOCSP-16255:

> Including "retryWrites=true&w=majority" in the example connection string complicates it and is unnecessary

`w=majority` is the [server default](https://www.mongodb.com/docs/manual/reference/mongodb-defaults/). `retryWrites=true` is a driver default as of PHPC-1324 (resolved in 2019).

skills/laravel-mongodb/references/connection.md
```diff
@@ -0,0 +1,68 @@
...
+// deprecated: getMongoDB() → getDatabase(), getMongoClient() → getClient()
+
+$collection = DB::connection('mongodb')->getCollection('logs');
+// note: ->collection() does not exist; use ->table() for the query builder
+```
+
+## Connection pooling and timeouts
+
+```
+mongodb+srv://.../db?maxPoolSize=50&minPoolSize=5&serverSelectionTimeoutMS=5000&socketTimeoutMS=30000
```
kevinAlbs:
```suggestion
mongodb+srv://.../db?maxPoolSize=50&serverSelectionTimeoutMS=5000&socketTimeoutMS=30000
```

The `minPoolSize` URI option was removed in C driver 2.0: CDRIVER-2390

skills/laravel-mongodb/references/eloquent-models.md
```diff
@@ -0,0 +1,92 @@
...
+        // Do NOT cast native MongoDB arrays to 'array' — stored natively;
+        // 'array' cast triggers deprecation and may serialize as JSON.
+    ];
+}
+```
+
+## `_id` vs `id`
+
+- BSON field is `_id` (ObjectId); PHP exposes both `$model->id` and `$model->_id` — both return a string.
+- In API resources, always cast to string so clients receive `"6708..."` not `{"$oid":"6708..."}`:
```
kevinAlbs: Suggest clarifying that ObjectId is the default, but not necessarily the type for `_id`:

```suggestion
- BSON field is `_id` (default ObjectId); PHP exposes both `$model->id` and `$model->_id` — both return ObjectId a string.
- In API resources, always cast ObjectId to string so clients receive `"6708..."` not `{"$oid":"6708..."}`:
```

Drivers add a default ObjectId `_id` on insert if there is no `_id`. But an explicit `_id` with other BSON types can be used. For example, with an integer:

```
$ mongosh --eval "db.coll.insertOne({'_id': 123})"
```

And I expect the Eloquent models do not require an ObjectId for `_id` either, since `getIdAttribute` [checks the BSON type](https://github.com/mongodb/laravel-mongodb/blob/f8ce003dfbbd39ca49ba29a54369b2538cc066bd/src/Eloquent/DocumentModel.php#L72-L95).

skills/laravel-mongodb/references/support.md
```diff
@@ -0,0 +1,23 @@
...
+Before opening an issue, search existing issues and pull requests in the relevant repository — the bug may already be reported or fixed.
+
+Report issues to the repository that owns the layer where the bug occurs:
+
+| Layer                                            | Repository |
+|--------------------------------------------------|---|
+| Eloquent integration, Laravel-specific behaviour | [mongodb/laravel-mongodb](https://github.com/mongodb/laravel-mongodb/issues) |
+| PHP query API, BSON, GridFS                      | [mongodb/mongo-php-library](https://github.com/mongodb/mongo-php-library/issues) |
+| C driver, connection, PHP extension              | [mongodb/mongo-php-driver](https://github.com/mongodb/mongo-php-driver/issues) |
+| Documentation                                    | [mongodb/docs](https://github.com/mongodb/docs/issues) |
```
kevinAlbs: mongodb/docs has issues disabled. Suggest replacing with description similar to [this one](https://www.mongodb.com/docs/php-library/current/get-started/#next-steps):

> submit feedback by using the **Rate this page** tab

skills/laravel-mongodb/references/support.md
```diff
@@ -0,0 +1,23 @@
+# Support and Issue Reporting
+
+Before opening an issue, search existing issues and pull requests in the relevant repository — the bug may already be reported or fixed.
```
kevinAlbs, line 3: Though Jira may be less useful to an agent (requires logging in), consider also referencing the PHPLARA, PHPLIB, PHPC Jira projects (similar to [README](https://github.com/mongodb/mongo-php-library#reporting-issues)) since they may be tracking an existing issue.

tools/review-skill/SKILL.md
```diff
@@ -0,0 +1,178 @@
...
+
+If `SKILL.md` does not exist at the given path, tell the user this is not a
+valid skill directory and ask them to provide the correct path.
+
+## Step 3: Run Structural Validation
+
+Run the full check suite:
+
+```bash
+skill-validator check <path>
```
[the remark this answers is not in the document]
paulinevos: Intentional

.github/workflows/validate-skills.yml
```diff
@@ -0,0 +1,43 @@
...
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: stable
+          cache: false
+
+      - name: Install skill-validator
+        run: go install github.com/agent-ecosystem/skill-validator/cmd/skill-validator@latest
+
```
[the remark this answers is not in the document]
paulinevos: Adding this. The test was originally just for local validation but could be useful for CI.

skills/laravel-mongodb/references/support.md
kevinAlbs: Poke: this might be still relevant.

kevinAlbs (approved):
LGTM with consideration to refer to Jira projects.

PR #3548, opened 2026-07-17
PHPLARA-251: Create database query tool for Boost

paulinevos:
This was originally a PR on `laravel/boost` (see: https://github.com/laravel/boost/pull/874), but was rejected because third-party packages can register their own tools in the service provider.

This adds a MongoDB-specific database query tool that Boost users can use instead of the SQL-specific Boost equivalent. Boost would otherwise attempt to run SQL on an MDB connection, which, well. Doesn't work.

### DIFF
--- a/src/Tools/DatabaseQuery.php
+++ b/src/Tools/DatabaseQuery.php
@@ -1,0 +1,219 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Laravel\Tools;
+
+use Illuminate\Contracts\JsonSchema\JsonSchema;
+use Illuminate\JsonSchema\Types\Type;
+use Illuminate\Support\Facades\DB;
+use InvalidArgumentException;
+use Laravel\Mcp\Request;
+use Laravel\Mcp\Response;
+use Laravel\Mcp\Server\Tool;
+use Laravel\Mcp\Server\Tools\Annotations\IsReadOnly;
+use MongoDB\Laravel\Connection;
+use Throwable;
+
+use function array_key_first;
+use function array_merge;
+use function array_walk;
+use function count;
+use function implode;
+use function in_array;
+use function is_array;
+use function sprintf;
+use function str_replace;
+
+/**
+ * MongoDB equivalent of Boost's DatabaseQuery tool.
+ *
+ * Runs read-only MQL commands against a MongoDB connection. Boost's own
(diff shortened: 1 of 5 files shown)


src/Tools/DatabaseQuery.php
```diff
@@ -0,0 +1,200 @@
...
+
+        if (count($pipeline) > self::MAX_STAGES_PER_PIPELINE) {
+            throw new InvalidArgumentException(sprintf('A pipeline exceeds the maximum of %d stages.', self::MAX_STAGES_PER_PIPELINE));
+        }
+
+        // These aggregation stages may contain nested aggregation pipelines
+        // and must be checked for write ops recursively.
+        $supportsNestedWrites = ['facet', 'lookup', 'unionWith'];
+
+        $allowList = array_merge($supportsNestedWrites, [
```
kevinAlbs, line 147: IIUC, `$allowList` is expected to be all known aggregation stages to avoid possibly missing write stages added in the future.

I expect this can be extended with other read-only stages in https://www.mongodb.com/docs/manual/reference/mql/aggregation-stages (`changeStream`, `collStats`, `search`, etc.).

src/Tools/DatabaseQuery.php
```diff
@@ -0,0 +1,200 @@
...
+        $operation = array_key_first($command);
+
+        if (! in_array($operation, $allowList, true)) {
+            throw new InvalidArgumentException(sprintf('Only read commands are allowed (%s).', implode(', ', $allowList)));
+        }
+
+        if ($operation === 'aggregate') {
+            // Check nested write ops recursively with conservative allow list
+            $this->ensureNoNestedWriteInAggregation($command['pipeline'] ?? []);
+        }
```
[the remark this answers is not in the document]
paulinevos: Not implementing these -- the agent calling the tool will catch and figure this out

src/Tools/DatabaseQuery.php
paulinevos: Yeah, Glean made a conservative allow list for me. I can add more to them. The only thing is at some point the allow list will be stale :/ Might want to keep an eye on that. I wanted to use an allow list instead of a deny list because it's safer... but maybe it's overly cautious? Idk. For now I'll add these. We can always switch strategies later, it's just an agent tool.

kevinAlbs (approved):
LGTM!

PR #3550, opened 2026-07-21
Move agent skill to Laravel Boost convention

pushpak1300:
Currently Laravel Boost cannot discover the package skill because it is stored outside the third-party package skill path. This moves the unchanged skill and references to `resources/boost/skills/laravel-mongodb`, following Laravel's [package skill convention](https://laravel.com/docs/13.x/boost#third-party-package-skills).

No tests, file relocation only.

Screenshot 2026-07-21 at 3 48 05 PM


paulinevos:
Heya @pushpak1300, thanks a lot for the PR.

We split the over-arching skill and Boost-specific guidelines (there is a Boost-specific guideline file present in `resources/boost/guidelines/core.blade.php`), because we wanted to avoid that users that don't use Boost wouldn't be able to discover the skill 🤔

Is it a valid thought that we could suggest using the root-level skill in the `resources/boost` guidelines file instead of moving the entire skill? Or would that not work?

GromNaN:
Thanks for the move. I looked into whether we should also keep a top-level `skills/` directory (via symlink, CI mirror, or a `.claude-plugin/plugin.json` manifest) so the skill stays discoverable by skills.sh in addition to Laravel Boost. Findings below in case they save the next reviewer the same detour.

**Ecosystem audit.** GitHub code search for `path:resources/boost/skills SKILL.md` returns ~17 Laravel packages that ship a Boost skill ([lorisleiva/laravel-actions](https://github.com/lorisleiva/laravel-actions), [serversideup/spin](https://github.com/serversideup/spin), [achyutkneupane/laravel-seo](https://github.com/achyutkneupane/laravel-seo), [wotzebra/laravel-boost-kit](https://github.com/wotzebra/laravel-boost-kit), and others). None of them keep a duplicate `skills/` directory at the repo root. All publish only to `resources/boost/skills/`, matching what this PR does.

GromNaN (approved):

GromNaN:
The skill is correctly listed on https://www.skills.sh/mongodb/laravel-mongodb/laravel-mongodb, but does not appear in the search results on https://skills.laravel.cloud/?search=mongodb.

@pushpak1300 Do you have any idea why our skill isn't showing up?

PR #3553, opened 2026-07-27
Fix groupBy() building an invalid $group stage when _id is selected

### LINKED ISSUE #3551: GroupBy + select that has _id throw an error
- Laravel-mongodb Version: 5.7.1
- PHP Version: 8.4

### Description:
```
 use Modules\Contacts\Contacts\Models\Contact;
      
  Contact::query()
      ->where('account_id', 7)
      ->groupBy('whatsapp_number')
      ->select(['_id', 'whatsapp_number'])
      ->get();
  ```

  This throws:
  MongoDB\Driver\Exception\CommandException: Executor error during aggregate
  command on namespace: datbase.contacts :: caused by :: $last's argument
  must be an array, but is objectId.
      
  Why: dump the compiled pipeline with ->toMql() and you see the actual $group stage:
  ```    
  [   
      ['$match' => ['account_id' => 7]],          
      ['$group' => [
          '_id' => ['$last' => '$_id'],           // should be ['whatsapp_number' => '$whatsapp_number']
          'whatsapp_number' => ['$last' => '$whatsapp_number'],
      ]
]
```

The issue on  [src/Query/Builder.php line ](https://github.com/mongodb/laravel-mongodb/blob/1bf7f53f474446fcda04c4bac51f38897af6fb89/src/Query/Builder.php#L319) that overrides the correct group by

### Steps to reproduce
just call a query with a groupBy and select that has an _id in it.

ZayanKhan-12:
Fixes #3551.

`Query\Builder` builds the `$group` stage with `_id` holding the grouping-key document. When the selected columns also contain `_id` (e.g. `->groupBy('foo')->select(['_id', 'foo'])`), the last-value loop overwrote that key document with `['$last' => '$_id']` — and inside `_id`, `$last` is parsed as the *array expression operator*, so the server rejects the pipeline with **"$last's argument must be an array"** (exactly the error in the issue).

Two commits:
1. Skip `_id` in the last-value column loop — the grouping key is already emitted as `_id`, so selecting it needs no accumulator.
2. Regression test in `tests/Query/BuilderTest.php` asserting the built pipeline for `select('_id', 'foo')->groupBy('foo')` keeps `_id` as the grouping-key document — **fails without commit 1** (verified).

**Verification (macOS arm64, PHP 8.5, ext-mongodb via pecl, local standalone mongod):** full `vendor/bin/phpunit` run: 826 tests, 3521 assertions — the only error is `ModelTest::testCreateOrFirst` ("did not remove its own error handlers"), which errors **identically on unmodified master** in this environment and is unrelated to this change; 50 skipped (feature-gated).

Prepared with AI assistance (Claude); the root cause and fix were verified against the built pipeline as described.

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -319,6 +319,12 @@ class Builder extends BaseBuilder
             // Add the last value of each column when there is no aggregate function.
             if ($this->groups && ! $this->aggregate) {
                 foreach ($columns as $column) {
+                    // The _id field holds the grouping keys and must not be
+                    // overwritten by an accumulator when it is selected.
+                    if ($column === '_id') {
+                        continue;
+                    }
+
                     $key = str_replace('.', '_', $column);
 
                     $group[$key] = ['$last' => '$' . $column];
--- a/tests/Query/BuilderTest.php
+++ b/tests/Query/BuilderTest.php
@@ -1241,6 +1241,16 @@ class BuilderTest extends TestCase
             fn (Builder $builder) => $builder->groupBy('foo'),
         ];
 
+        yield 'groupBy with _id in select' => [
+            [
+                'aggregate' => [
+                    [['$group' => ['_id' => ['foo' => '$foo'], 'foo' => ['$last' => '$foo']]]],
+                    [], // options
+                ],
+            ],
+            fn (Builder $builder) => $builder->select('_id', 'foo')->groupBy('foo'),
+        ];
+
         yield 'sub-query' => [
             [
                 'find' => [


GromNaN (approved):
Thanks @ZayanKhan-12 for the clean fix and regression test.

Verified locally:
* The new data provider case reproduces the bug on 5.x without the patch (actual pipeline has `_id => ['$last' => '$_id']`, which the server rejects).
* With the one-line `continue` in the last-value loop, `_id` correctly remains the grouping-key document and the test passes.
* Root cause analysis in the PR description matches the code path: the accumulator loop was clobbering the grouping key when `_id` appeared in the selected columns.

Scope of the change is appropriately narrow, and the fix aligns with the existing pattern (the grouping keys are already emitted as `_id` on line 311).

PR #3558, opened 2026-07-29
PHPLARA-256 Add Schema::ensureVectorExtensionExists() to check Atlas Search availability

GromNaN:
[PHPLARA-256](https://jira.mongodb.org/browse/PHPLARA-256)

Laravel 12.51 added `Schema::ensureVectorExtensionExists()` to verify that the vector extension is installed before creating vector columns and indexes. On PostgreSQL it creates the `pgvector` extension; on other connections the base implementation throws "Extensions are only supported by Postgres".

On MongoDB, the equivalent requirement for `Blueprint::vectorSearchIndex()` is Atlas Search. The method now lists the search indexes of a collection, which does not need to exist, and converts the "not supported" server error into an explicit `RuntimeException`:

```php
// Throws unless the deployment supports Atlas Search
Schema::ensureVectorExtensionExists();

Schema::create('books', function (Blueprint $collection) {
    $collection->vectorSearchIndex([
        'fields' => [['type' => 'vector', 'path' => 'embedding', 'numDimensions' => 1536, 'similarity' => 'cosine']],
    ], 'vector');
});
```

`MongoDB\Exception\SearchNotSupportedException` is caught when the installed version of mongodb/mongodb throws it, with a fallback on the error codes for older versions.

### DIFF
--- a/src/Schema/Builder.php
+++ b/src/Schema/Builder.php
@@ -7,10 +7,12 @@ namespace MongoDB\Laravel\Schema;
 use Closure;
 use MongoDB\Collection;
 use MongoDB\Driver\Exception\ServerException;
+use MongoDB\Exception\SearchNotSupportedException;
 use MongoDB\Laravel\Connection;
 use MongoDB\Model\CollectionInfo;
 use MongoDB\Model\IndexInfo;
 use Override;
+use RuntimeException;
 
 use function array_column;
 use function array_fill_keys;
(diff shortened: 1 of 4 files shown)


src/Schema/Builder.php
```diff
@@ -382,9 +384,45 @@ protected function getAllCollections()
...
+    public function ensureVectorExtensionExists($schema = null)
+    {
+        $collection = $this->connection->getDatabase($schema)->getCollection('any');
+        assert($collection instanceof Collection);
+
+        $message = 'Atlas Search is not available on this MongoDB deployment. Use MongoDB Atlas or a local Atlas deployment. ';
+
+        try {
+            // The collection doesn't need to exist to know whether Atlas Search is available
+            $collection->listSearchIndexes(['name' => 'any']);
```
GromNaN, line 405: We can bump the minimum version of the lib to 2.2, so this exception is always thrown. And not try to catch the exception.

This would be done with PHPLARA-255.

GromNaN (commented):
I mark this PR as "draft", waiting for [PHPLARA-255](https://jira.mongodb.org/browse/PHPLARA-255).

Issue #3003, opened 2024-06-10
withCount is not working 


Rahul-codoffer:
- Laravel-mongodb Version: 4.3
- PHP Version:8.3.7

### Description:
Getting **This method is not supported by MongoDB. Try \"toMql()\" instead** on withCount function

omerkoseoglu:
+1

Rahul-codoffer:
@omerkoseoglu ?

GromNaN:
Could you please provide a sample code to reproduce the issue. Some features are not supported with MongoDB or require some additional work.

hooman-mirghasemi:
I have same problem in laravel 10 php 8.3 and this package latest version 4.5

GromNaN:
@hooman-mirghasemi Same question. 

> Could you please provide a sample code to reproduce the issue. Some features are not supported with MongoDB or require some additional work.

hooman-mirghasemi:
make a model in mysql "post"
then make an other one in mongo db "comments"
make hybrid relationships

`Post::withCount('comments')->get();`

MiladAheshmeh:
same problem

MiladAheshmeh:
@jenssegers would you please fix this issue?

alcaeus:
@MiladAheshmeh please refrain from pinging people directly, especially as they no longer maintain the library. We are aware of the issue and will fix it in due time. If you'd like to help speed up the process, creating a pull request with a failing test case is a great way to do so!

arjundexbytes:
Same issue with me

Athul-Suresh:
Same issue with me also 
 "php": "^8.2",
 "laravel/framework": "^12.0",
 "mongodb/laravel-mongodb": "^5.4",

Sample Code

Post::where('status",1)->withCount('comments')->get()

Error:
 "message": "This method is not supported by MongoDB. Try \"toMql()\" instead.",
    "exception": "BadMethodCallException",

alcaeus:
We are tracking this internally as PHPORM-238 and will close out this issue once we've added support for `withCount`.

Athul-Suresh:
> We are tracking this internally as [PHPORM-238](https://jira.mongodb.org/browse/PHPORM-238) and will close out this issue once we've added support for `withCount`.

Thanks for the update!

Abdullahbutt3434:
Same issue with me

alcaeus:
[...]

Issue #2470, opened 2022-11-22
withCount and loadCount methods not working on hybrid hasMany relationships


MOstadi:
- Laravel-mongodb Version: #.#.#
- PHP Version: 8.1.6
- Database Driver & Version: MongoDB 6.0.2 Community

### Description:

It seems it's trying to get count from mysql instead of mongodb.

### Steps to reproduce

I have User and Post class as below:
```
use Illuminate\Database\Eloquent\Model;
use Jenssegers\Mongodb\Eloquent\HybridRelations;

class User extends Model
{
    use HybridRelations;

    public function posts()
    {
        return $this->hasMany('App\Models\Post');
    }
}
```

```
use Jenssegers\Mongodb\Eloquent\Model;

class Post extends Model
{

    protected $collection = 'posts';

    protected $connection = 'mongodb';

    public function user(){
        return $this->belongsTo('App\Models\User');
    }
}
```

now when I try to get count with this :
```

$user->posts()->count()
```
it's work
but if I try this :
```
User::withCount('posts')->first()
//or
$user->loadCount('posts');
```
it fails and gives this error:
```
SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '"testdb"."posts" where "user_id" exists ?) as `posts_count` f' at line 1 (SQL: select `users`.*, (select "user_id" from "testdb"."posts" where "user_id" exists 1) as `posts_count` from `users` where `users`.`id` = 462372 limit 1) 
```

GromNaN:
Issue confirmed. Tracked in PHPORM-69

hans-thomas:
@GromNaN I have been working on this one lately. I noticed there is no way to add an aggregation pipeline to the query instance.
I created this aggregation that can solve the problem.
```php
[
    [
        '$lookup' => [
            'from' => 'items',
            'localField' => '_id', 
            'foreignField' => 'user_id',
            'pipeline' => [['$project' => ['_id' => 1]]], 
            'as' => 'items_count'
        ]
    ], ['$addFields' => ['items_count' => ['$size' => '$items_count']]]
]
```
I can run this using [`MongoDB/Collection`](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBCollection-aggregate/) class, but it's not what we want in this case. There is a problem in the `select` methods that leads us to an error.

GromNaN:
[...]

Issue #2435, opened 2022-08-17
Adding withCount in mongodb request I got Illegal offset type error


sergeynilov:
In laravel 9 with MongoDB(jenssegers/mongodb 3.9) I added withCount in request :

    $subscription = Subscription
        ::getById($subscriptionId)
        ->withCount('userSubscriptions')
        ->firstOrFail();

and got error :
 

    [2022-08-16 14:02:18] local.ERROR: Illegal offset type {"userId":"62fb9ff9f74eb576f803a7b2","exception":"[object] (TypeError(code: 0): Illegal offset type at Project/vendor/jenssegers/mongodb/src/Query/Builder.php:373)
    [stacktrace]
    #0 Project/vendor/jenssegers/mongodb/src/Query/Builder.php(201): Jenssegers\\Mongodb\\Query\\Builder->getFresh()
    #1 Project/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php(698): Jenssegers\\Mongodb\\Query\\Builder->get()
    #2 Project/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php(682): Illuminate\\Database\\Eloquent\\Builder->getModels()
    #3 Project/vendor/laravel/framework/src/Illuminate/Database/Concerns/BuildsQueries.php(296): Illuminate\\Database\\Eloquent\\Builder->get()
    #4 Project/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php(582): Illuminate\\Database\\Eloquent\\Builder->first()
    #5 Project/app/Http/Controllers/Admin/SubscriptionController.php(162): Illuminate\\Database\\Eloquent\\Builder->firstOrFail()
    #6 Project/vendor/laravel/framework/src/Illuminate/Routing/Controller.php(54): App\\Http\\Controllers\\Admin\\SubscriptionController->edit()
    #7 Project/vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php(45): Illuminate\\Routing\\Controller->callAction()
    #8 Project/vendor/laravel/framework/src/Illuminate/Routing/Route.php(261): Illuminate\\Routing\\ControllerDispatcher->dispatch()
    #9 Project/vendor/laravel/framework/src/Illuminate/Routing/Route.php(204): Illuminate\\Routing\\Route->runController()
[...]

Issue #1801, opened 2019-08-06
How to count optimal data relations


trungdiep92:
How to count optimal data relations:

Collection: photos, photos_like, photos_comment
Result: Photo i, count like, count comment

Case 1:

$photos = Photo::with(['like', 'comment' => function ($query) {
            $query->where('is_approved', '1');}])
            ->select('photo_featured', 'title', 'view_count')->where('status', "1")->orderBy('created_at', 'desc')
            ->get();
			
view:
$photo->like->count()
$photo->comment->count()

But it selects all data in like and comment collection and then counts
->consuming memory, slow processor

Case 2: Use withCount

->return: Illegal offset type

https://github.com/jenssegers/laravel-mongodb/issues/1339
https://github.com/jenssegers/laravel-mongodb/issues/1147
https://github.com/jenssegers/laravel-mongodb/issues/1416

I did not find a way to fix the error

Case 3: Use query builder and raw

$photos = DB::table('photos')
            ->select(['photos.photo_featured', 'photos.title', 'photos.view_count', DB::raw("COUNT(photos_like._id) as like_count"), 
			DB::raw("COUNT(photos_comment._id) as comment_count")])
			->....
			
->return: Illegal offset type

error when using DB::raw

https://github.com/jenssegers/laravel-mongodb/issues/1123
https://github.com/jenssegers/laravel-mongodb/issues/1693

How to get count from table relation???

Smolevich:
@trungdiep92, let's go to slack, you can find link [here](https://github.com/jenssegers/laravel-mongodb/issues/1791)

trungdiep92:
The function withCount relation of this library is faulty. I research and found two solutions:
1. Relation: Use aggregate mongodb

2. Embed: Use embed array and count size of array or use like_count filed(increment)

Hope this helps!

GromNaN:
This feature is implemented by https://github.com/mongodb/laravel-mongodb/pull/3560

PR #3560, opened 2026-07-30
PHPLARA-225 Support relation aggregates on document models

GromNaN:
Adds support for `withCount()`, `withExists()`, `withSum()`, `withAvg()`, `withMin()` and `withMax()` on document models. `loadCount()`, `loadExists()` and `loadAggregate()` work as well, since Laravel funnels them through the same `withAggregate()` method.

Before this change, any of them failed with `BadMethodCallException: This method is not supported by MongoDB. Try "toMql()" instead.`, because `Illuminate\Database\Eloquent\Concerns\QueriesRelationships::withAggregate()` builds a correlated subselect and calls `toSql()`.

Resolves PHPLARA-225, PHPLARA-93 and PHPLARA-237. Reported as #3003, #2470, #2435, #1801 and #1339.

## How it works

MongoDB has no correlated subquery, so the values cannot be selected with the parent documents as Eloquent does. The new `MongoDB\Laravel\Helpers\QueriesRelationshipAggregates` trait records the requested aggregates in `withAggregate()`, then computes them in `eagerLoadRelations()`, after the parent documents are read, and sets them as attributes.

The cost is one additional query per aggregate, whatever the number of parent documents:

- `HasOneOrMany` and its morph variants are aggregated by the server with a `$group` on the foreign key.
- `BelongsTo`, `BelongsToMany` and `MorphToMany` cannot be grouped server side (a single related document per parent, or parent keys stored in an array field), so the eager loading `match()` logic is reused and the values are folded in PHP.
- `EmbedsOne` and `EmbedsMany` need no query at all, they are computed from the parent document.

`exists` is computed as a grouped count mapped to `> 0` and stored as a real boolean, so no `bool` cast is added on the model.

Aliases follow Illuminate exactly, so `withCount('books')` gives `books_count`, `withMax('items', 'amount')` gives `items_max_amount` and `withCount('books as total')` gives `total`. Defaults with no related document are `0` for `count`, `false` for `exists` and `null` for the others.

## Limitations

Cases that cannot be supported throw a `LogicException` rather than returning a wrong value:

- `orderBy()` on an aggregate alias: the value does not exist server side, so a `$sort` on it would silently return unsorted results. A `$lookup` pipeline is required to sort on an aggregate.
- `MorphTo` and other relation types outside the set supported by `has()`.
- Hybrid relations, where the related model is not stored in MongoDB.
- Constraint closures on embedded relations (PHPORM-292).

`cursor()` and `lazy()` do not call `eagerLoadRelations()`, so the aliases are absent on those paths, as with any eager loading.

The Boost skill and its query builder reference are updated, since they documented these methods as unsupported.

## Why not `$lookup`?

Computing the aggregates inside the parent query with a `$lookup` stage would avoid the extra queries, but it is not possible today, for five reasons.

**The query builder cannot emit a pipeline for a regular `get()`.** It only switches from `find()` to `aggregate()` when `$groups` or `$aggregate` is set (`src/Query/Builder.php:303`), and it has no `$lookup` support at all. Adding one would force every query carrying a `withCount()` through an aggregation pipeline, changing the execution plan of the parent query itself, not just of the aggregate.

**`$lookup` does no BSON type coercion.** Foreign keys are frequently stored as strings while `_id` is an `ObjectId`, to the point that the package documentation recommends casting them with `'author_id' => 'string'`. Eloquent normalizes both sides when matching in PHP, `$lookup` does not, so the counts would silently be `0`. That is precisely the failure mode this PR avoids by throwing on the cases it cannot compute correctly.

**Server compatibility.** A correlated `$lookup` (`let` + `pipeline`), which is what counting without materializing the related documents requires, needs MongoDB 5.0, while CI still covers 4.4. The `localField` / `foreignField` form works on 4.4 but builds the full array of related documents before `$size`, so an unbounded relation can hit the 16 MB document limit or the 100 MB per stage memory limit.

**One pipeline generator per relation type.** `belongsToMany` stores the keys in an array field, `morphToMany` in pivot subdocuments discriminated by a morph type, and `embedsMany` is already part of the parent document. That is four distinct pipeline builders to write and test, instead of reusing the eager loading `match()` logic that already handles those dictionaries.

**Scope.** PHPLARA-93 is the open spike about rewriting the query builder on top of aggregation pipelines. This PR delivers the feature now, at one extra query per aggregate whatever the number of parent documents, and `$lookup` remains the natural follow up: it is the only way to lift the `orderBy()` limitation on an aggregate alias.

### DIFF
--- a/resources/boost/skills/laravel-mongodb/SKILL.md
+++ b/resources/boost/skills/laravel-mongodb/SKILL.md
@@ -56,7 +56,7 @@ Suggest installing the MongoDB agent skills plugin if not already installed. Ins
 - Cast FK fields to `string` via `$casts` on the child model when FK values may come from outside model attributes (imports, raw ObjectIds) — prevents BSON type mismatches on direct `where('author_id', $id)` queries.
 - Eager-load with `::with()` — MongoDB does no server-side joins for Eloquent relations.
 - Use aggregation pipeline for grouping, counting per group, `$lookup`, and `$sample`.
-- Relation aggregates (`withCount()`, `withExists()`, `withSum()`, `withAvg()`, `withMin()`, `withMax()`) are supported, but each one runs an extra query after the parent documents are read. Use a `$lookup` pipeline when the aggregated value must be filtered, sorted or paginated on.
+- Relation aggregates (`withCount()`, `withExists()`, `withSum()`, `withAvg()`, `withMin()`, `withMax()`) are supported. Use a `$lookup` pipeline when the aggregated value must be filtered, sorted or paginated on.
 - Create indexes in migrations: `Schema::connection('mongodb')->create('posts', fn (Blueprint $c) => $c->index('user_id'))`.
 - Use `DB::connection('mongodb')->transaction(...)` only on replica set / sharded cluster.
 
--- a/resources/boost/skills/laravel-mongodb/SKILL.md
+++ b/resources/boost/skills/laravel-mongodb/SKILL.md
@@ -129,36 +129,7 @@ final class User extends Model
 }
 ```
 
-### 3. Relation aggregates, and when to use `$lookup` instead
-
-```php
-<?php
-
-use App\Models\Post;
-
-// One extra query per aggregate, values set as attributes on the models
-$posts = Post::withCount('comments')->withExists('author')->withMax('comments', 'score')->get();
-
-$posts[0]->comments_count;      // int
-$posts[0]->author_exists;       // bool
-$posts[0]->comments_max_score;  // null when there is no comment
-
-// Sorting or filtering on the aggregated value requires a pipeline,
-// Post::withCount('comments')->orderBy('comments_count') throws.
-$posts = Post::raw(fn ($collection) => $collection->aggregate([
-    ['$lookup' => [
-        'from'         => 'comments',
-        'localField'   => '_id',
-        'foreignField' => 'post_id',
-        'as'           => 'comments',
-    ]],
-    ['$addFields' => ['comments_count' => ['$size' => '$comments']]],
-    ['$project'   => ['comments' => 0]],
-    ['$sort'      => ['comments_count' => -1]],
-]));
-```
-
-### 4. Queue job
+### 3. Queue job
 
 ```php
 <?php
@@ -183,7 +154,7 @@ final class IndexPostJob implements ShouldQueue
 IndexPostJob::dispatch((string) $post->_id)->onConnection('mongodb');
 ```
 
-### 5. Feature test (Pest)
+### 4. Feature test (Pest)
 
 ```php
 <?php
--- a/resources/boost/skills/laravel-mongodb/SKILL.md
+++ b/resources/boost/skills/laravel-mongodb/SKILL.md
@@ -8,14 +8,14 @@ metadata:
 
 # Laravel MongoDB
 
-Implementation skill for `mongodb/laravel-mongodb`. Exists to prevent the common mistakes: auto-increment IDs, `withCount()`, `toSql()`, SQL `JOIN`, `distinct()->get()` expecting scalar arrays, and `belongsTo()` over native ObjectId FKs.
+Implementation skill for `mongodb/laravel-mongodb`. Exists to prevent the common mistakes: auto-increment IDs, `toSql()`, SQL `JOIN`, `distinct()->get()` expecting scalar arrays, and `belongsTo()` over native ObjectId FKs.
 
 ## Core Workflow
 
 1. Identify layer: model, query builder, relationship, schema/index, queue/cache/session, search, or transaction.
 2. Confirm model extends `MongoDB\Laravel\Eloquent\Model` (or uses `DocumentModel` trait).
 3. Map every FK and `_id`: ObjectId in DB, **string** in Eloquent — cast accordingly.
-4. Replace unsupported helpers (`withCount`, `toSql`, `groupByRaw`, `whereFulltext`, `union`, `inRandomOrder`, `whereColumn`) with MongoDB alternatives.
+4. Replace unsupported helpers (`toSql`, `groupByRaw`, `whereFulltext`, `union`, `inRandomOrder`, `whereColumn`) with MongoDB alternatives.
 5. Validate: `php artisan migrate`, Pest tests, `phpcs`/`phpstan`.
 
 ## Related MongoDB tooling
@@ -35,7 +35,7 @@ Suggest installing the MongoDB agent skills plugin if not already installed. Ins
 | Topic | Reference file | Load When |
 |---|---|---|
 | Models, casts, `_id` mapping | `references/eloquent-models.md` | Defining or modifying a model |
-| Query builder gotchas, aggregation | `references/query-builder.md` | Writing queries, `withCount`, `distinct`, grouping |
+| Query builder gotchas, aggregation | `references/query-builder.md` | Writing queries, relation aggregates, `distinct`, grouping |
 | Embedded, hybrid, cross-database relations | `references/relationships.md` | `belongsTo`, `hasMany`, `embedsMany`, `hasManyIn` |
 | Connection setup | `references/connection.md` | `config/database.php`, multiple connections |
 | Indexes, migrations | `references/schema.md` | Creating indexes, migrations, collections |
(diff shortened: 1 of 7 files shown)


resources/boost/skills/laravel-mongodb/references/query-builder.md
```diff
@@ -33,15 +32,43 @@ $genres = Movie::raw(fn ($c) => $c->aggregate([
 
 Do **not** use `Movie::raw(fn ($c) => $c->distinct('field'))` via Eloquent if you expect a Collection — use `->distinct()->pluck()` instead.
 
-## Replacing `withCount`
+## Relation aggregates
 
-`withCount()`, `withAvg()`, and `withSum()` are **not supported** on MongoDB models — they silently return wrong results or throw. Replace with a `$lookup` + `$size` / `$avg` / `$sum` aggregation pipeline.
+`withCount()`, `withExists()`, `withSum()`, `withAvg()`, `withMin()` and `withMax()` are supported, as well as
```
paulinevos, line 37: I am not sure about the gain here in replacing "xyz is not supporteed" with an explicit line about them being supported and what the implementation details are. The implementation details and tiny extra cost add needless context for usage. I think we should just remove the "not supported" message.

As the methods produce the same result as the base Eloquent builder (sets extra data as an attribute), that's all that matters for usage and a code sample should be sufficient.

resources/boost/skills/laravel-mongodb/references/query-builder.md
```diff
@@ -33,15 +32,43 @@ $genres = Movie::raw(fn ($c) => $c->aggregate([
...
+$posts[0]->comments_count;   // int, 0 when there is no related document
+$posts[0]->author_exists;    // bool
+$posts[0]->top_score;        // null when there is no related document
+
+// Constrained aggregate
+Post::withCount(['comments' => fn ($query) => $query->where('approved', true)])->get();
+```
+
+Supported relations: `hasOne`, `hasMany`, `morphOne`, `morphMany`, `belongsTo`, `belongsToMany`, `morphToMany`,
+`morphedByMany`, `embedsOne` and `embedsMany`. Embedded aggregates cost no extra query, they are computed from
```
paulinevos: Again not sure if the agent should care about the extra cost. The skill should guide it towards what it should use, so if we think it should use these aggregates then it should just say that imo

resources/boost/skills/laravel-mongodb/SKILL.md
```diff
@@ -128,14 +129,22 @@ final class User extends Model
 }
 ```
 
-### 3. Aggregation replacing `withCount`
+### 3. Relation aggregates, and when to use `$lookup` instead
```
paulinevos: This now seems to be duplicate context. We should pick which one we want to keep. To me this feels like it should go in the query builder skill instead so it can be loaded into context only when needed.

src/Helpers/QueriesRelationshipAggregates.php
```diff
@@ -0,0 +1,275 @@
...
+        if (! in_array($function, self::AGGREGATE_FUNCTIONS, true)) {
+            throw new InvalidArgumentException(sprintf(
+                'Aggregate function "%s" is not supported by MongoDB. Supported functions are: %s.',
+                $function ?? 'null',
+                implode(', ', self::AGGREGATE_FUNCTIONS),
+            ));
+        }
+
+        if (! is_string($column)) {
+            throw new InvalidArgumentException('Expressions are not supported as aggregate column by MongoDB.');
```
paulinevos: The exception message assumes that if it's not a string, it must be an expression. Instead of doing that, should we say that "Aggregate column name must be a string" or something similar?

src/Helpers/QueriesRelationshipAggregates.php
```diff
@@ -0,0 +1,275 @@
...
+                $subQuery = $relation->getRelated()->newQuery();
+                $constraints($subQuery);
+
+                if ($subQuery->getQuery()->wheres) {
+                    throw new LogicException(sprintf(
+                        'Constraints on the embedded relation "%s" are not supported. See https://jira.mongodb.org/browse/PHPORM-292',
+                        $name,
+                    ));
+                }
+            } elseif ($parentKey !== null && $this->getQuery()->columns !== null) {
```
paulinevos: I'm not a huge fan of this agentic way of writing code that has nested `if` and `elseif`s withing `foreach`es etc. and then comments explaining what things do. I find it a bit unreadable. Do you think you could apply private methods with good naming, early returns etc. to let the code communicate itself what it's doing?

GromNaN:
Discussion notes:
- Add a note in the docs that an optimisation is to store the aggregated value in the main document. That would solve orderBy compatibility issue and prevent from doing a 2nd query to retrieve the aggregation results.
- Not covering vector search features because they require aggregation pipelines.

resources/boost/skills/laravel-mongodb/references/query-builder.md
GromNaN: Removed. The section now shows the methods with a code sample, without implementation details or cost notes.

resources/boost/skills/laravel-mongodb/references/query-builder.md
GromNaN: Removed the cost note as well.

resources/boost/skills/laravel-mongodb/SKILL.md
GromNaN: Removed the section from SKILL.md. The example now lives only in the query-builder reference, which the reference table already points to for relation aggregates.

src/Helpers/QueriesRelationshipAggregates.php
GromNaN: Changed to "The aggregate column name must be a string."

src/Helpers/QueriesRelationshipAggregates.php
GromNaN: Refactored: the hydration is split into hydrateGroupedAggregate and hydrateMatchedAggregate, the alias resolution and the embedded constraint check are extracted, and the branches use early returns. The initial elseif was also removed, as the two branches were independent.

GromNaN:
I applied the review feedback and added two hardening checks after a security review:

- A column name that starts with "$" is rejected. The query builder prefixes the column with "$" to build a field path, so a value such as "$ROOT" would become the variable reference "$$ROOT" in the aggregation pipeline.
- A stored relation key that is not a scalar, a Binary, or a stringable object now causes a descriptive InvalidArgumentException instead of a PHP cast error during hydration.

Both checks have tests. Ready for another look.

resources/boost/skills/laravel-mongodb/SKILL.md
```diff
@@ -56,12 +56,13 @@ Suggest installing the MongoDB agent skills plugin if not already installed. Ins
 - Cast FK fields to `string` via `$casts` on the child model when FK values may come from outside model attributes (imports, raw ObjectIds) — prevents BSON type mismatches on direct `where('author_id', $id)` queries.
 - Eager-load with `::with()` — MongoDB does no server-side joins for Eloquent relations.
 - Use aggregation pipeline for grouping, counting per group, `$lookup`, and `$sample`.
+- Relation aggregates (`withCount()`, `withExists()`, `withSum()`, `withAvg()`, `withMin()`, `withMax()`) are supported, but each one runs an extra query after the parent documents are read. Use a `$lookup` pipeline when the aggregated value must be filtered, sorted or paginated on.
```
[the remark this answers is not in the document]
GromNaN: Removed the cost note from the bullet, in line with the earlier review feedback.

src/Helpers/QueriesRelationshipAggregates.php
```diff
@@ -0,0 +1,335 @@
...
+
+        $subQuery = $relation->getRelated()->newQuery();
+        $constraints($subQuery);
+
+        if ($subQuery->getQuery()->wheres) {
+            throw new LogicException(sprintf(
+                'Constraints on the embedded relation "%s" are not supported.',
+                $name,
+            ));
+        }
```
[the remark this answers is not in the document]
GromNaN: Good catch. The guard now also rejects limit, offset and distinct, with a test. orderBy and select stay allowed as they have no effect on the six aggregate functions, same as a SQL subselect.

src/Helpers/QueriesRelationshipAggregates.php
```diff
@@ -0,0 +1,336 @@
...
+        ));
+    }
+
+    private function assertAggregateRelationSupported(Relation $relation, string $name): void
+    {
+        if ($relation instanceof EmbedsOneOrMany) {
+            return;
+        }
+
+        if (! DocumentModel::isDocumentModel($relation->getRelated()) || $this->isAcrossConnections($relation)) {
```
GromNaN, line 295: Note on the `isAcrossConnections()` condition: it is conservative. Both hydration paths query the connection of the related model (server side $group for HasOneOrMany, eager match in PHP for the others), so a relation between two MongoDB connections could probably work. We keep the restriction for now because cross-connection MongoDB relations are rare and the path is untested. Relaxing this guard later is not a breaking change.

paulinevos (approved):

PR #3563, opened 2026-08-04
Fix invalid BSON regex flags in `DatabasePresenceVerifier::getCount()`

zigzagdev:
### Description
`DatabasePresenceVerifier::getCount()` was passing `'/i'` as the flags argument to `MongoDB\BSON\Regex`, instead of the plain PCRE flag character `'i'`. 
`Regex` flags must only contain flag characters (`i`, `m`, `x`, `s`, ...), not a delimiter-style string. 
The sibling method `getMultiCount()` already used the correct `'i'`.

This PR has no observable effect against MongoDB's current PHP driver/server: the BSON encoder silently strips invalid flag characters before a query ever reaches the server (verified by inspecting the actual command sent via `CommandSubscriber`). 
Still, constructing an invalid `Regex` value is incorrect and worth fixing for correctness / forward-compatibility.

### Test Plans
Added `testGetCountBuildsRegexWithValidFlags()` to `ValidationTest.php`.
Since the invalid flags can't be observed end-to-end (see above), the test asserts against the raw `Regex` object passed to `where()`, using a lightweight stub in place of the query builder/connection.

### Checklist

### Test Results
```bash
vendor/bin/phpunit tests/ValidationTest.php --testdox
PHPUnit 10.5.47 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.32
Configuration: /var/www/laravel-mongodb/phpunit.xml

...                                                                 3 / 3 (100%)

Time: 00:00.950, Memory: 26.00 MB

Validation (MongoDB\Laravel\Tests\Validation)
 ✔ Unique
 ✔ Exists
 ✔ Get count builds regex with valid flags

OK (3 tests, 19 assertions)
```

### DIFF
--- a/src/Validation/DatabasePresenceVerifier.php
+++ b/src/Validation/DatabasePresenceVerifier.php
@@ -17,7 +17,7 @@ class DatabasePresenceVerifier extends \Illuminate\Validation\DatabasePresenceVe
     #[Override]
     public function getCount($collection, $column, $value, $excludeId = null, $idColumn = null, array $extra = [])
     {
-        $query = $this->table($collection)->where($column, new Regex('^' . preg_quote($value) . '$', '/i'));
+        $query = $this->table($collection)->where($column, new Regex('^' . preg_quote($value) . '$', 'i'));
 
         if ($excludeId !== null && $excludeId !== 'NULL') {
             $query->where($idColumn ?: 'id', '<>', $excludeId);
--- a/tests/ValidationTest.php
+++ b/tests/ValidationTest.php
@@ -5,7 +5,9 @@ declare(strict_types=1);
 namespace MongoDB\Laravel\Tests;
 
 use Illuminate\Support\Facades\Validator;
+use MongoDB\BSON\Regex;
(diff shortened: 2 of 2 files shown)


GromNaN (approved):
The fix is correct. `MongoDB\BSON\Regex` expects flag characters only (`i`, `m`, `x`, ...) without PCRE delimiters, so replacing `'/i'` with `'i'` is the right change, consistent with what `getMultiCount()` already does.

**Tests:** The new `testGetCountBuildsRegexWithValidFlags` test correctly validates the flags using a lightweight stub, and the assertion would fail with the old code (`getFlags()` would return `'/i'`). The existing `testUnique` and `testExists` tests cover the end-to-end behavior.

**Security review:** No issues. `preg_quote()` properly escapes user input, anchors `^`/`$` prevent partial matches, and the regex has no quantifiers that could cause ReDoS.

**Performance:** The invalid `/` flag was stripped by the BSON encoder before reaching the server (as noted in the PR description), so no query behavior change is expected.

Thanks for spotting this inconsistency and for the thorough PR description explaining the observable impact!

PR #3564, opened 2026-08-09
fix: `morphOne()` using the wrong MorphOne relation class

zigzagdev:
## Summary

- Fix `morphOne()` instantiating the native `MorphOne` relation instead of a MongoDB-specific one, which threw `BadMethodCallException` when eager loading with an integer parent key type.

### what I have done

- Added `MongoDB\Laravel\Relations\MorphOne`, mirroring the existing `MorphMany` override, which always resolves `whereInMethod()` to `whereIn()`.
- Updated `HybridRelations::morphOne()` to use this MongoDB-specific class instead of `Illuminate\Database\Eloquent\Relations\MorphOne`.
- Added two regression tests to `tests/Ticket/GH2783Test.php`:
  - `testMorphOneUsesMongoRelationClass` — asserts `morphOne()` returns the MongoDB-specific relation class.
  - `testMorphOneEagerLoadWithIntegerKeyType` — reproduces the eager-load failure end-to-end with an integer `$keyType` parent.

### Motivation

`HybridRelations::morphOne()` already branched on `isDocumentModel()` like its sibling relation methods (`hasOne`, `hasMany`, `morphMany`, ...), but still instantiated the native `Illuminate\Database\Eloquent\Relations\MorphOne`. 
That class's `whereInMethod()` resolves to `whereIntegerInRaw()` when eager loading multiple parents with an integer `$keyType`, and `MongoDB\Laravel\Query\Builder` does not support that method (it's intentionally unsupported, since MongoDB has no bind-parameter concept to optimize around), causing a `BadMethodCallException`. 
String keys, the default for Mongo models, were unaffected either way, which is why this went unnoticed until now.

### Test Plans

- Run `tests/Ticket/GH2783Test.php` against a MongoDB instance and confirm both new regression tests pass, along with the existing tests in the file.
- Confirm the full test suite still passes with no regressions.

### Checklist

### Test Results
```bash
root@27ca77bae12d:/var/www/laravel-mongodb# vendor/bin/phpunit tests/ValidationTest.php tests/Ticket/GH2783Test.php --filter 'testGetCountBuildsRegexWithValidFlags|testMorphOneUsesMongoRelationClass|testMorphOneEagerLoadWithIntegerKeyType' --testdox
PHPUnit 11.5.56 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.32
Configuration: /var/www/laravel-mongodb/phpunit.xml

..                                                                  2 / 2 (100%)

Time: 00:03.765, Memory: 32.00 MB

GH2783 (MongoDB\Laravel\Tests\Ticket\GH2783)
 ✔ Morph one uses mongo relation class
 ✔ Morph one eager load with integer key type

OK, but there were issues!
Tests: 2, Assertions: 4, PHPUnit Deprecations: 1.
root@27ca77bae12d:/var/www/laravel-mongodb# 
```

### DIFF
--- a/src/Eloquent/HybridRelations.php
+++ b/src/Eloquent/HybridRelations.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace MongoDB\Laravel\Eloquent;
 
 use Illuminate\Database\Eloquent\Concerns\HasRelationships;
-use Illuminate\Database\Eloquent\Relations\MorphOne;
 use Illuminate\Database\Eloquent\Relations\Pivot;
 use Illuminate\Support\Str;
 use MongoDB\Laravel\Helpers\EloquentBuilder;
@@ -14,6 +13,7 @@ use MongoDB\Laravel\Relations\BelongsToMany;
 use MongoDB\Laravel\Relations\HasMany;
 use MongoDB\Laravel\Relations\HasOne;
 use MongoDB\Laravel\Relations\MorphMany;
+use MongoDB\Laravel\Relations\MorphOne;
 use MongoDB\Laravel\Relations\MorphTo;
 use MongoDB\Laravel\Relations\MorphToMany;
 
--- a/src/Relations/MorphOne.php
+++ b/src/Relations/MorphOne.php
@@ -1,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Laravel\Relations;
+
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\MorphOne as EloquentMorphOne;
+use Override;
+
+/**
+ * @template TRelatedModel of Model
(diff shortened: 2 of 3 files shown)


GromNaN (approved):
Thanks for the fix, this is exactly the same pattern as the existing MorphMany override, and the regression test reproduces the BadMethodCallException cleanly.

I checked out the branch locally, ran the two new tests against a local MongoDB instance (both pass), reverted the fix temporarily to confirm both tests fail without it (UnknownClassOrInterfaceException and BadMethodCallException as expected), then restored the fix and confirmed the full relations test suite still passes (63 tests). phpcs/phpcbf report no violations.

I also had two independent passes done on the diff for performance and security concerns: no issues found in either case, this is a pure inheritance fix with no new query patterns or algorithmic complexity.

I rebased the branch on the latest 5.x to keep it up to date and pushed it back to your fork, no content changes. Merging now.

PR #3565, opened 2026-08-13
PHPLARA-133: Add support for Laravel 13 attributes

paulinevos:
Ensures support for the new attributes added to Laravel in this PR: https://github.com/laravel/framework/pull/58578 by adding passing tests for them

### Checklist

### DIFF
--- a/src/Eloquent/DocumentModel.php
+++ b/src/Eloquent/DocumentModel.php
@@ -4,12 +4,14 @@ declare(strict_types=1);
 
 namespace MongoDB\Laravel\Eloquent;
 
+use BadMethodCallException;
 use Carbon\CarbonInterface;
 use DateTimeInterface;
 use DateTimeZone;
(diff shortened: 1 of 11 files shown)


GromNaN (approved):

src/Eloquent/DocumentModel.php
```diff
@@ -94,6 +97,26 @@ public function getIdAttribute($value = null)
         return $value;
     }
 
+    /** @inheritdoc */
+    public function initializeModelAttributes()
+    {
+        parent::initializeModelAttributes();
```
GromNaN, line 108: Should you check if `method_exists(parent::class, __METHOD__)`? And throw an exception if it does not exists?

You might say that, with or without it, it results in an error. And this method isn’t intended to be called directly by the developer.

paulinevos: added

PR #3568, opened 2026-08-22
PHPLARA-250 Keep numeric string relation keys as strings in has queries

### LINKED ISSUE #3541: has() / whereHas() fails when custom relationship keys are numeric strings
# `has()` / `whereHas()` fails when custom relationship keys are numeric strings

## Environment

* Laravel: 13.7.0
* mongodb/laravel-mongodb: 5.7.1
* mongodb/mongodb: 2.2.0

## Description

`has()` and `whereHas()` appear to fail when a relationship uses custom string keys whose values are **numeric strings** (e.g. `"1"`, `"2"`).

The relationship itself works correctly, but relationship existence queries (`has`, `whereHas`) return no results.

Using non-numeric strings (e.g. `"item-1"`) works as expected.

### Models

```php
class ParentModel extends Model
{
    public function children()
    {
        return $this->hasMany(ChildModel::class, 'parent_key', 'custom_key');
    }
}

class ChildModel extends Model
{
    public function parent()
    {
        return $this->belongsTo(ParentModel::class, 'parent_key', 'custom_key');
    }
}
```

### Documents

**parents**

```json
{
    "custom_key": "1"
}
```

**children**

```json
{
    "parent_key": "1"
}
```

### Working

```php
ParentModel::first()->children;
ParentModel::first()->children()->count();
ChildModel::where('parent_key', '1')->exists();
```

All of the above return the expected results.

### Failing

```php
ParentModel::has('children')->get();
ParentModel::whereHas('children')->get();
```

These return an empty collection.

## Query log

The generated query contains BSON integers instead of strings:

```json
{
    "custom_key": {
        "$in": [
            { "$numberInt": "1" }
        ]
    }
}
```

However, the stored values are BSON strings:

```json
{
    "custom_key": "1"
}
```

Since MongoDB performs strict type matching, the query returns no results.

## Additional observation

If the values are changed from numeric strings to non-numeric strings, for example:

```json
{
    "custom_key": "item-1"
}
```

and

```json
{
    "parent_key": "item-1"
}
```

then `has()` and `whereHas()` work correctly.

This suggests that somewhere during the relationship existence query, numeric strings are being coerced into integers before the `$in` query is constructed.

Hashim1999164:
Fixes #3541

has() and whereHas() collect related ids with array_count_values then array_keys. PHP promotes numeric string keys to integers, so the MongoDB $in query used integers against stored BSON strings and returned no documents.

This change stringifies the returned keys so values like "1" stay strings.

### Checklist

* [x] Add tests and ensure they pass

### DIFF
--- a/src/Helpers/QueriesRelationships.php
+++ b/src/Helpers/QueriesRelationships.php
@@ -16,7 +16,6 @@ use LogicException;
 use MongoDB\Laravel\Eloquent\Model;
 use MongoDB\Laravel\Relations\MorphToMany;
 
-use function array_count_values;
 use function array_filter;
 use function array_keys;
 use function array_map;
@@ -24,6 +23,7 @@ use function class_basename;
 use function collect;
 use function in_array;
 use function is_array;
+use function is_int;
 use function is_object;
 use function is_string;
 use function method_exists;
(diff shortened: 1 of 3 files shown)


src/Helpers/QueriesRelationships.php
```diff
@@ -200,9 +201,19 @@ protected function getHasCompareKey(Relation $relation)
      */
     protected function getConstrainedRelatedIds($relations, $operator, $count)
     {
+        $ids = is_array($relations) ? $relations : $relations->flatten()->toArray();
+
+        $originalIds = [];
```
paulinevos: Could you rename this `$idsByStringKey` for accuracy?

paulinevos (changes requested):
Thanks for contributing! I added some test cases to ensure no regressions, there's still an edge case that fails for integer IDs. 

Full disclosure: apart from that we're still on the fence if this would constitute a breaking change

src/Helpers/QueriesRelationships.php
Hashim1999164: Renamed, and switched the key itself. A string only map still merged int 1 with string "1", which is the integer edge case in the tests you added. idsByTypedKey keeps those two separate and still restores the original scalar for Mongo \$in.

paulinevos:
Hi @Hashim1999164, thank you for your PR and the changes. For now, we've decided to close this as this would be a breaking change and we'd like to plan a bit more for it.

PR #3571, opened 2026-08-26
PHPLARA-259 Apply maxTimeMS from connection options

GromNaN:
This change applies a default maxTimeMS from database connection config (`options.maxTimeMS`) to query operation options.

Precedence order remains:
- explicit query option `maxTimeMS`
- builder `timeout()`
- connection `options.maxTimeMS`

The implementation centralizes this behavior in inherited connection options, so it is consistently applied across supported query paths.

Unit tests now include a dedicated data provider that covers inherited connection options for `find`, `aggregate`, and `distinct`, plus precedence and empty-config cases.

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -447,10 +447,6 @@ class Builder extends BaseBuilder
         $options = [];
 
         // Apply order, offset, limit and projection
-        if ($this->timeout) {
-            $options['maxTimeMS'] = (int) ($this->timeout * 1000);
-        }
-
         if ($this->orders) {
             $options['sort'] = $this->grammar->prepareFieldsForQuery($this->orders);
         }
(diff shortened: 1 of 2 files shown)


src/Query/Builder.php
```diff
@@ -1778,6 +1774,18 @@ private static function vectorSearchSupportsAutoEmbedding(): bool
      */
     private function inheritConnectionOptions(array $options = []): array
     {
+        if (! isset($options['maxTimeMS'])) {
+            $queryTimeout = $this->timeout ? (int) ($this->timeout * 1000) : null;
```
paulinevos: Could you tell it to do something like:

```php
$options['maxTimeMs'] = $options['maxTimeMs'] ?? $this->someGoodMethodName()
```

with early returns instead of this nested condition?

paulinevos (approved):

PR #3573, opened 2026-08-28
PHPLARA-262 Support serializable classes allowlist in MongoStore cache

paulinevos:
Laravel added an opt-in allowlist of classes that cache stores may instantiate while unserializing, configured with cache.serializable_classes. MongoStore was the last store not honouring it, so applications hardening their cache had a gap they could not close.

The default stays unrestricted, so existing behaviour is unchanged.

### Checklist

### DIFF
--- a/resources/boost/skills/laravel-mongodb/references/cache-sessions.md
+++ b/resources/boost/skills/laravel-mongodb/references/cache-sessions.md
@@ -7,6 +7,10 @@
 return [
     'default' => env('CACHE_STORE', 'mongodb'),
 
+    // Optional: restrict which classes may be unserialized from the cache.
+    // null allows all, false (shown here) allows none, or list allowed classes.
+    'serializable_classes' => false,
+
     'stores' => [
         'mongodb' => [
             'driver'          => 'mongodb',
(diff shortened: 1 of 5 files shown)


resources/boost/skills/laravel-mongodb/references/cache-sessions.md
```diff
@@ -7,6 +7,10 @@
 return [
     'default' => env('CACHE_STORE', 'mongodb'),
 
+    // Optional: restrict which classes may be unserialized from the cache.
+    // null (default) allows all, false allows none, or list allowed classes.
+    'serializable_classes' => null,
```
GromNaN, line 12: Maybe set to `false`, to be secure by default when using this feature.
```suggestion
    'serializable_classes' => false,
```

tests/Cache/MongoCacheStoreTest.php
```diff
@@ -302,6 +342,14 @@ private function getStore(): Repository
         return $repository;
     }
 
+    private function getStoreAllowing(array|bool|null $serializableClasses): Repository
```
GromNaN, line 345: `null` is not used.
```suggestion
    private function getStoreAllowing(array|bool $serializableClasses): Repository
```

GromNaN (commented):
The PR will have to target the current minor (5.9)

GromNaN (approved):

PR #3577, opened 2026-09-07
PHPLARA-275 Add auto encryption: configuration and collection creation

### LINKED ISSUE #2699: [Feature Request]  Queryable Encryption or CSFLE
No but if possible it will have a smooth transition with mongodb nodejs. 

Im not sure if there is any way to make the connection to have automatic encryption

nodejs

GromNaN:
Adds first-class Queryable Encryption (QE) support to mongodb/laravel-mongodb, configuration-driven with no model metadata. The user declares the encrypted fields once in `driver_options.autoEncryption.encryptedFieldsMap`; the driver and the server handle encryption and decryption transparently.

What this adds:

- Connection: validation of `driver_options.autoEncryption` (keyVaultNamespace, kmsProviders, local key size, crypt_shared), plus `getClientEncryption()` and `getEncryptionOptions()`. The `encryptedFieldsMap` accepts a simplified keyed-by-path syntax and the list form, normalized to the driver format.
- Data keys: each field resolves by an alternate key name instead of a manual base64 keyId. The default name is `<database>.<collection>/<path>` and an explicit `keyAltName` per field overrides it. Resolution is delegated to libmongocrypt (>= 1.18). See DRIVERS-3637.
- Encrypted collection lifecycle: `Schema::createEncrypted()` creates the collection, its metadata collections and indexes, and generates missing data keys, idempotently. `dropAllTables()` documents that dropping the database also drops the key vault and every data key.
- Models: the server-managed `__safeContent__` field is hidden from serialization and protected from writes. Encrypted collections use single-document updates automatically.
- Artisan: `mongodb:encrypted:create` and `mongodb:encrypted:diagnose`, each with a `--no-server` config-only mode.
- Safety guard: a write to a collection mapped in `encryptedFieldsMap` fails fast when the collection is not created as an encrypted collection, so fields are never silently stored in plaintext. The check batches the server lookup. See DRIVERS-3647.

Internals: the encryption logic lives in an internal `MongoDB\Laravel\Encryption\AutoEncryption`, held weakly and created lazily only when `autoEncryption` is configured.

This requires `mongodb/mongodb` `^1.21.2|^2.1.1`.

The Queryable Encryption integration scenarios (encrypted collection creation, equality and range queries, key lifecycle) run against a MongoDB Enterprise or Atlas replica set with crypt_shared and are gated behind a skip guard; community lanes stay green.

Closes https://github.com/mongodb/laravel-mongodb/issues/2699


paulinevos:
Brother why 😭💔
image

resources/boost/skills/laravel-mongodb/references/queryable-encryption.md
```diff
@@ -0,0 +1,87 @@
+# Queryable Encryption
```
paulinevos, line 1: Did you benchmark this?

src/Commands/Encrypted/CreateEncryptedCommand.php
```diff
@@ -0,0 +1,57 @@
...
+    public function handle(): int
+    {
+        $collection = $this->argument('collection');
+
+        try {
+            $connection = $this->connection();
+            $config = $this->autoEncryptionConfig($connection);
+            $connection->validateAutoEncryptionConfig($config);
+
+            if (! is_array($config['encryptedFieldsMap'][$collection] ?? null)) {
```
paulinevos: 2 questions:
- Why validate this separately when there's a `validateAutoEncryptionConfig` call 2 lines earlier?
- Why validate this at all when the server will throw and many would catch this by phpstan array shape? I guess it saves a server trip but it also means we have to maintain parity with server validation rules.

GromNaN:
Going back and forth on the implementation. Can be squashed.

src/Commands/Encrypted/EncryptedCommand.php
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Get the validated driver_options.autoEncryption configuration.
+     *
+     * @return array<string, mixed>
+     */
+    protected function autoEncryptionConfig(Connection $connection): array
+    {
+        $config = $connection->getConfig('driver_options.autoEncryption');
+
+        if (! is_array($config)) {
```
paulinevos, line 49: If we do validate auto encryption config in multiple places it might be nice to encapsulate that validation in one place (maybe a VO?)

src/Encryption/AutoEncryption.php
```diff
@@ -0,0 +1,494 @@
...
+        }
+
+        $driverOptions['autoEncryption'] = $autoEncryption;
+
+        return $driverOptions;
+    }
+
+    /**
+     * Determine whether automatic encryption (Queryable Encryption or CSFLE)
+     * is requested on this connection.
```
paulinevos: I feel like we don't need 90% of the comments on this class. The fact that GH wouldn't render it because "large diff" should tell us something 😅

src/Encryption/AutoEncryption.php
```diff
@@ -0,0 +1,494 @@
...
+    /**
+     * Validate the shape of the driver_options.autoEncryption configuration and
+     * return a normalized copy. This only checks the static configuration; it
+     * never contacts the server.
+     *
+     * @param  array<string, mixed> $autoEncryption
+     *
+     * @return array<string, mixed>
+     */
+    public function validateAutoEncryptionConfig(array $autoEncryption): array
```
paulinevos, line 129: Yeah so see this I feel like could be extracted into a VO instead

src/Encryption/AutoEncryption.php
```diff
@@ -0,0 +1,494 @@
...
+
+            if (isset($config['keyId'])) {
+                $field['keyId'] = $config['keyId'];
+            }
+
+            if (isset($config['queries']) && is_array($config['queries'])) {
+                $field['queries'] = $config['queries'];
+            } elseif (isset($config['queryType'])) {
+                $queryType = $config['queryType'];
+                if (! in_array($queryType, ['equality', 'range'], true)) {
```
paulinevos: Is there no other way we can handle this?

src/Schema/Builder.php
```diff
@@ -129,6 +131,64 @@ public function create($table, ?Closure $callback = null, array $options = [])
...
+     */
+    public function createEncrypted(string $collection, ?Closure $callback = null, array $options = []): array
+    {
+        $config = $this->connection->getConfig('driver_options.autoEncryption');
+
+        if (! is_array($config) || ! $this->connection->isEncryptionEnabled($config)) {
+            throw new InvalidArgumentException('Queryable Encryption is not enabled on this connection. Configure "driver_options.autoEncryption" with a "keyVaultNamespace" and "kmsProviders" first.');
+        }
+
+        if (empty($config['encryptedFieldsMap']) || ! is_array($config['encryptedFieldsMap'])) {
```
paulinevos: Yet more repeated validation that could be centralized somewhere imo

src/Connection.php
```diff
@@ -261,9 +288,106 @@ protected function createConnection(string $dsn, array $config, array $options):
             $driverOptions += ['connectionName' => $config['name']];
         }
 
+        // Normalize the automatic encryption configuration early. A broken
+        // autoEncryption block fails fast at connection time. Capability
+        // gating is lazy and only checked when encryption is actually used, so
```
paulinevos: Remove comment

src/Connection.php
```diff
@@ -261,9 +288,106 @@ protected function createConnection(string $dsn, array $config, array $options):
...
+     *
+     * @return array<string, mixed>
+     */
+    public function resolveOrCreateEncryptionKeys(array $encryptedFieldsMap): array
+    {
+        return $this->encryption()?->resolveOrCreateEncryptionKeys($encryptedFieldsMap) ?? $encryptedFieldsMap;
+    }
+
+    /**
+     * Determine whether automatic encryption (Queryable Encryption or CSFLE)
```
paulinevos: Remove comment. All these method names already communicate all they need to

paulinevos (changes requested):

resources/boost/skills/laravel-mongodb/references/queryable-encryption.md
GromNaN: Good catch on the wording. That line is about architecture, not performance: the encryption work happens in libmongocrypt (native code reached through the driver) and on the server, so there is no encrypt/decrypt loop in PHP. QE still has overhead (encryption, the `__safeContent__` array, extra indexes), just not in PHP userland. I have reworded the doc so it no longer reads as a zero-overhead claim.

GromNaN:
Thanks for the review! I squashed the branch onto the latest 5.x. Summary of what addresses your comments:

- **Validation centralization** (comments on `CreateEncryptedCommand`, `EncryptedCommand`, `AutoEncryption::validateAutoEncryptionConfig`, `Schema\Builder`): the repeated "is encryption enabled / is the collection mapped" checks now live in one method, `AutoEncryption::encryptedFieldsFor()`, used by the create command and `Schema::createEncrypted()`. The redundant `validateAutoEncryptionConfig` calls in the commands and in `getClientEncryption()` are gone, since the config is already validated at connection construction. I kept the existing public `Connection` methods (so the tests and `Query\Builder` stay unchanged) rather than introducing a VO.
- **queryType validation** (`AutoEncryption`): dropped the manual `[equality, range]` allow-list. libmongocrypt validates the value when the client is built, so we no longer maintain parity with its rules.
- **Comments** (`AutoEncryption`, `Connection`): removed the non-high-level comments and verbose docblocks, keeping only type tags, the `@internal` marker, and the `DRIVERS-3637` pointer.
- **`encryptedFieldsMap` is optional** on the connection (the server can supply it for server-side QE), recommended for security, and required only to create an encrypted collection. The doc now states this.

PR #3582, opened 2026-09-22
Refactor AtlasSearchTest to use static methods

MaddipatlaChetan24:
Moves the expensive Atlas Search fixture setup in `AtlasSearchTest`
from `setUp()`/`tearDown()` (run before/after every test method) to
`setUpBeforeClass()`/`tearDownAfterClass()` (run once for the whole
class).

The fixture — dropping/recreating the `books` collection, inserting 20
documents, creating 3 Atlas Search indexes, and waiting for all of
them to become ready — is the most expensive part of this test file.
None of the 7 test methods write to the shared collection or mutate
`$vectors`; they're all read-only queries against the same fixture
data, so recreating it before every single test was unnecessary
repeated cost with no isolation benefit in practice.

`private array $vectors` becomes `private static array $vectors`,
`addVector()` becomes a static method, and the two vector-search tests
now reference `self::$vectors[7]` instead of `$this->vectors[7]`.
`testEloquentBuilderVectorSearchAutoEmbedding`'s own extra index setup
is left inside that test method, since it's gated behind an env var
and specific to that one test.

**Not yet verified against a real Atlas cluster or run through
PHPUnit** — I don't have that environment available. Please confirm
it compiles and the full suite still passes, and in particular confirm
none of these tests actually depend on per-test isolation that this
change removes.

### Checklist


GromNaN:
Thanks for this. The goal (running the expensive Atlas Search fixture once per class instead of before every test) is the right one, and the speedup is real. Unfortunately the implementation as written cannot run. I put together a corrected version and validated it against `mongodb/mongodb-atlas-local:8`.

## Why the current change fails

Two problems, both fatal:

1. **Static calls on instance methods.** `getConnection()`, `waitForSearchIndexesDropped()`, and `waitForSearchIndexesReady()` are instance methods, from Laravel's `InteractsWithDatabase` trait and the `AtlasSearchIndexManagement` trait respectively. Calling them as `self::...()` from a static `setUpBeforeClass()` emits a deprecation on PHP 8.2 and then a fatal `Using $this when not in object context`, because `getConnection()` reads `$this->app->make('db')`.

2. **The Laravel app is not bootstrapped in `setUpBeforeClass()`.** Orchestra Testbench builds `$this->app` (providers, config, DB manager) in `setUp()` per test, not in `setUpBeforeClass()`. At class-setup time there is no app, no booted `MongoDBServiceProvider`, and no Eloquent connection resolver, so `Book::insert()`, `Schema::getIndexes()`, and the facades would all fail. Orchestra's `BeforeAll` attribute does not help either: its `beforeAll()` hook receives no app context.

There is also a test-isolation issue the description overstates. `testEloquentBuilderVectorSearchAutoEmbedding` is not read-only: it creates an extra `auto_embed` search index on the shared `books` collection. With per-class setup that index leaks to later tests and would break `testGetIndexes` and its `assertCount(4)` if the test execution order ever changes.

## What I did instead

Kept `setUp()` as an instance method, so `$this->app` is available, but guarded the expensive fixture with a static flag so it runs only once per class. Cached the `MongoDB\Collection` handle in a static property so a static `tearDownAfterClass()` can drop the collection without a bootstrapped app (the `Collection` carries its own driver `Manager`, which survives Orchestra's per-test teardown). Cached the "Atlas Search not supported" skip decision so the remaining tests skip cheaply instead of retrying the expensive setup.

For the auto-embedding test, the cleanup needed a targeted wait. `waitForSearchIndexesDropped()` waits for all indexes to reach zero, which would time out here because the three fixture indexes stay in place. I added a `waitForSearchIndexDropped(Collection, string $name)` helper to the `AtlasSearchIndexManagement` trait that polls `listSearchIndexes(['name' => $name])` until just that index is gone, and used it in a `finally` block so `testGetIndexes` sees exactly four indexes regardless of order.

## Performance

Measured against `mongodb/mongodb-atlas-local:8`. Same outcome in both cases: 8 tests, 29 assertions, 1 skipped (auto-embedding without a registered model), and the same 2 pre-existing PHPUnit deprecations.

| Metric | Before (per-test setup) | After (per-class setup) |
| --- | --- | --- |
| Tests / assertions | 8 / 29 | 8 / 29 |
| Skipped | 1 | 1 |
| Duration | 7.28 s | 0.23 s (warm) |

Roughly 30x faster, because the fixture (drop collection, insert 20 documents, create 3 search indexes, wait for ready) now runs once instead of eight times.

Happy to push this to the branch (maintainer edits are enabled) or open a separate PR if you prefer. The two corrected files are below.

<details>
<summary>tests/AtlasSearchIndexManagement.php</summary>

```php
<?php

namespace MongoDB\Laravel\Tests;

use MongoDB\Collection;
use RuntimeException;

use function hrtime;
use function usleep;

/**
 * Helpers for managing Atlas Search indexes in tests with awaiting mechanism.
 */
trait AtlasSearchIndexManagement
{
    /**
     * Waits for the search index created in the previous test to be deleted
     */
    public function waitForSearchIndexesDropped(Collection $collection)
    {
        $timeout = hrtime()[0] + 120;
        // Waits for the search index created in the previous test to be deleted
        while ($collection->listSearchIndexes()->count()) {
            if (hrtime()[0] > $timeout) {
                throw new RuntimeException('Timed out waiting for search indexes to be dropped');
            }

            usleep(1000);
        }
    }

    /**
     * Waits for a single named search index to be deleted.
     *
     * Unlike waitForSearchIndexesDropped(), this does not require the collection
     * to be empty of all indexes, so it can be used to clean up one extra index
     * while the shared fixture indexes remain in place.
     */
    public function waitForSearchIndexDropped(Collection $collection, string $name, int $timeoutSeconds = 120)
    {
        $timeout = hrtime()[0] + $timeoutSeconds;
        while ($this->hasSearchIndex($collection, $name)) {
            if (hrtime()[0] > $timeout) {
                throw new RuntimeException('Timed out waiting for search index "' . $name . '" to be dropped');
            }

            usleep(1000);
        }
    }

    private function hasSearchIndex(Collection $collection, string $name): bool
    {
        foreach ($collection->listSearchIndexes(['name' => $name]) as $index) {
            if (($index['name'] ?? null) === $name) {
                return true;
            }
        }

        return false;
    }

    /**
     * Waits for all search indexes to be ready
     */
    public function waitForSearchIndexesReady(Collection $collection, int $timeoutSeconds = 120)
    {
        $timeout = hrtime()[0] + $timeoutSeconds;
        do {
            if (hrtime()[0] > $timeout) {
                throw new RuntimeException('Timed out waiting for search indexes to be ready');
            }

            usleep(1000);
            $ready = true;
            foreach ($collection->listSearchIndexes() as $index) {
                $ready = $ready && $index['queryable'];
            }
        } while (! $ready);
    }
}
```
</details>

<details>
<summary>tests/AtlasSearchTest.php</summary>

```php
<?php

namespace MongoDB\Laravel\Tests;

use Illuminate\Database\Eloquent\Collection as EloquentCollection;
use Illuminate\Support\Collection as LaravelCollection;
use Illuminate\Support\Facades\Schema;
use MongoDB\Builder\Query;
use MongoDB\Builder\Search;
use MongoDB\Collection as MongoDBCollection;
use MongoDB\Driver\Exception\ServerException;
use MongoDB\Laravel\Schema\Builder;
use MongoDB\Laravel\Tests\Models\Book;
use PHPUnit\Framework\Attributes\Group;

use function array_map;
use function assert;
use function getenv;
use function mt_getrandmax;
use function rand;
use function range;
use function srand;
use function str_contains;
use function usort;

#[Group('atlas-search')]
class AtlasSearchTest extends TestCase
{
    use AtlasSearchIndexManagement;

    /** Vectors generated for the fixture, shared across all tests in this class. */
    private static array $vectors = [];

    /** Whether the expensive Atlas Search fixture has already been loaded for this class. */
    private static bool $fixtureLoaded = false;

    /** Whether the server reported that Atlas Search is not supported. */
    private static bool $atlasSearchNotSupported = false;

    /** Cached collection handle so tearDownAfterClass() can drop it without a bootstrapped app. */
    private static ?MongoDBCollection $collection = null;

    protected function setUp(): void
    {
        parent::setUp();

        // The fixture is read-only for every test except the auto-embedding one, which
        // cleans up its own extra index. Load it once per class instead of per test:
        // recreating the collection, inserting 20 documents, and waiting on 3 search
        // indexes is the most expensive part of this file and brings no isolation gain.
        if (self::$atlasSearchNotSupported) {
            self::markTestSkipped('Atlas Search not supported.');
        }

        if (! self::$fixtureLoaded) {
            $this->loadAtlasSearchFixture();
            self::$fixtureLoaded = true;
        }
    }

    private function loadAtlasSearchFixture(): void
    {
        $collection = $this->getConnection('mongodb')->getCollection('books');
        assert($collection instanceof MongoDBCollection);
        self::$collection = $collection;
        $collection->drop();

        Book::insert(self::addVector([
            ['title' => 'Introduction to Algorithms'],
            ['title' => 'Clean Code: A Handbook of Agile Software Craftsmanship'],
            ['title' => 'Design Patterns: Elements of Reusable Object-Oriented Software'],
            ['title' => 'The Pragmatic Programmer: Your Journey to Mastery'],
            ['title' => 'Artificial Intelligence: A Modern Approach'],
            ['title' => 'Structure and Interpretation of Computer Programs'],
            ['title' => 'Code Complete: A Practical Handbook of Software Construction'],
            ['title' => 'The Art of Computer Programming'],
            ['title' => 'Computer Networks'],
            ['title' => 'Operating System Concepts'],
            ['title' => 'Database System Concepts'],
            ['title' => 'Compilers: Principles, Techniques, and Tools'],
            ['title' => 'Introduction to the Theory of Computation'],
            ['title' => 'Modern Operating Systems'],
            ['title' => 'Computer Organization and Design'],
            ['title' => 'The Mythical Man-Month: Essays on Software Engineering'],
            ['title' => 'Algorithms'],
            ['title' => 'Understanding Machine Learning: From Theory to Algorithms'],
            ['title' => 'Deep Learning'],
            ['title' => 'Pattern Recognition and Machine Learning'],
        ]));

        try {
            $this->waitForSearchIndexesDropped($collection);

            $collection->createSearchIndex([
                'mappings' => [
                    'fields' => [
                        'title' => [
                            ['type' => 'string', 'analyzer' => 'lucene.english'],
                            ['type' => 'autocomplete', 'analyzer' => 'lucene.english'],
                            ['type' => 'token'],
                        ],
                    ],
                ],
            ]);

            $collection->createSearchIndex([
                'mappings' => ['dynamic' => true],
            ], ['name' => 'dynamic_search']);

            $collection->createSearchIndex([
                'fields' => [
                    ['type' => 'vector', 'numDimensions' => 4, 'path' => 'vector4', 'similarity' => 'cosine'],
                    ['type' => 'vector', 'numDimensions' => 32, 'path' => 'vector32', 'similarity' => 'euclidean'],
                    ['type' => 'filter', 'path' => 'title'],
                ],
            ], ['name' => 'vector', 'type' => 'vectorSearch']);
        } catch (ServerException $e) {
            if (Builder::isAtlasSearchNotSupportedException($e)) {
                self::$atlasSearchNotSupported = true;
                self::markTestSkipped('Atlas Search not supported. ' . $e->getMessage());
            }

            throw $e;
        }

        $this->waitForSearchIndexesReady($collection);
    }

    public static function tearDownAfterClass(): void
    {
        // Drop the shared collection once the whole class is done. The cached Collection
        // handle carries its own driver Manager, so this works without a bootstrapped app.
        self::$collection?->drop();

        self::$collection = null;
        self::$fixtureLoaded = false;
        self::$vectors = [];
        self::$atlasSearchNotSupported = false;

        parent::tearDownAfterClass();
    }

    public function testGetIndexes()
    {
        $indexes = Schema::getIndexes('books');

        self::assertIsArray($indexes);
        self::assertCount(4, $indexes);

        // Order of indexes is not guaranteed
        usort($indexes, fn ($a, $b) => $a['name'] <=> $b['name']);

        $expected = [
            [
                'name' => '_id_',
                'columns' => ['_id'],
                'primary' => true,
                'type' => null,
                'unique' => false,
            ],
            [
                'name' => 'default',
                'columns' => ['title'],
                'type' => 'search',
                'primary' => false,
                'unique' => false,
            ],
            [
                'name' => 'dynamic_search',
                'columns' => ['dynamic'],
                'type' => 'search',
                'primary' => false,
                'unique' => false,
            ],
            [
                'name' => 'vector',
                'columns' => ['vector4', 'vector32', 'title'],
                'type' => 'vectorSearch',
                'primary' => false,
                'unique' => false,
            ],
        ];

        self::assertSame($expected, $indexes);
    }

    public function testEloquentBuilderSearch()
    {
        $results = Book::search(
            sort: ['title' => 1],
            operator: Search::text('title', 'systems'),
        );

        self::assertInstanceOf(EloquentCollection::class, $results);
        self::assertCount(3, $results);
        self::assertInstanceOf(Book::class, $results->first());
        self::assertSame([
            'Database System Concepts',
            'Modern Operating Systems',
            'Operating System Concepts',
        ], $results->pluck('title')->all());
    }

    public function testDatabaseBuilderSearch()
    {
        $results = $this->getConnection('mongodb')->table('books')
            ->search(Search::text('title', 'systems'), sort: ['title' => 1]);

        self::assertInstanceOf(LaravelCollection::class, $results);
        self::assertCount(3, $results);
        self::assertIsArray($results->first());
        self::assertSame([
            'Database System Concepts',
            'Modern Operating Systems',
            'Operating System Concepts',
        ], $results->pluck('title')->all());
    }

    public function testEloquentBuilderAutocomplete()
    {
        $results = Book::autocomplete('title', 'system');

        self::assertInstanceOf(LaravelCollection::class, $results);
        self::assertCount(3, $results);
        self::assertSame([
            'Database System Concepts',
            'Modern Operating Systems',
            'Operating System Concepts',
        ], $results->sort()->values()->all());
    }

    public function testDatabaseBuilderAutocomplete()
    {
        $results = $this->getConnection('mongodb')->table('books')
            ->autocomplete('title', 'system');

        self::assertInstanceOf(LaravelCollection::class, $results);
        self::assertCount(3, $results);
        self::assertSame([
            'Database System Concepts',
            'Modern Operating Systems',
            'Operating System Concepts',
        ], $results->sort()->values()->all());
    }

    public function testDatabaseBuilderVectorSearch()
    {
        $results = $this->getConnection('mongodb')->table('books')
            ->vectorSearch(
                index: 'vector',
                path: 'vector4',
                queryVector: self::$vectors[7], // This is an exact match of the vector
                limit: 4,
                exact: true,
            );

        self::assertInstanceOf(LaravelCollection::class, $results);
        self::assertCount(4, $results);
        self::assertSame('The Art of Computer Programming', $results->first()['title']);
        self::assertSame(1.0, $results->first()['vectorSearchScore']);
    }

    public function testEloquentBuilderVectorSearch()
    {
        $results = Book::vectorSearch(
            index: 'vector',
            path: 'vector4',
            queryVector: self::$vectors[7],
            limit: 5,
            numCandidates: 15,
            // excludes the exact match
            filter: Query::query(
                title: Query::ne('The Art of Computer Programming'),
            ),
        );

        self::assertInstanceOf(EloquentCollection::class, $results);
        self::assertCount(5, $results);
        self::assertInstanceOf(Book::class, $results->first());
        self::assertNotSame('The Art of Computer Programming', $results->first()->title);
        self::assertSame('The Mythical Man-Month: Essays on Software Engineering', $results->first()->title);
        self::assertThat(
            $results->first()->vectorSearchScore,
            self::logicalAnd(self::isType('float'), self::greaterThan(0.9), self::lessThan(1.0)),
        );
    }

    public function testEloquentBuilderVectorSearchAutoEmbedding()
    {
        if (getenv('VOYAGE_API_KEY') === false || getenv('VOYAGE_API_KEY') === '') {
            self::markTestSkipped('Auto embedding requires VOYAGE_API_KEY to be set on the Atlas cluster.');
        }

        $collection = $this->getConnection('mongodb')->getCollection('books');
        assert($collection instanceof MongoDBCollection);

        try {
            Schema::table('books', function ($table) {
                $table->vectorSearchIndex([
                    'fields' => [
                        ['type' => 'autoEmbed', 'modality' => 'text', 'path' => 'title', 'model' => 'voyage-4-large'],
                    ],
                ], 'auto_embed');
            });
        } catch (ServerException $e) {
            if (str_contains($e->getMessage(), 'not registered')) {
                self::markTestSkipped('Auto embedding requires an Atlas cluster with a registered embedding model. Set VOYAGE_API_KEY.');
            }

            throw $e;
        }

        try {
            $this->waitForSearchIndexesReady($collection);

            // Query with a lighter model (voyage-4-lite) than the indexing model (voyage-4-large);
            // all voyage-4 embeddings are compatible.
            $results = Book::vectorSearch(
                index: 'auto_embed',
                path: 'title',
                query: 'machine learning textbook',
                model: 'voyage-4-lite',
                limit: 3,
            );

            self::assertInstanceOf(EloquentCollection::class, $results);
            self::assertCount(3, $results);
            self::assertInstanceOf(Book::class, $results->first());
            self::assertIsFloat($results->first()->vectorSearchScore);
        } finally {
            // This test mutates the shared fixture by adding an index. Drop it and wait
            // for the deletion to complete so the shared state stays at the 4 indexes
            // the other tests expect, regardless of test execution order. We cannot use
            // waitForSearchIndexesDropped() here: it waits for ALL indexes to be gone,
            // which would time out because the 3 fixture indexes remain.
            $collection->dropSearchIndex('auto_embed');
            $this->waitForSearchIndexDropped($collection, 'auto_embed');
        }
    }

    /** Generate random vectors using fixed seed to make tests deterministic */
    private function addVector(array $items): array
    {
        srand(1);
        foreach ($items as &$item) {
            self::$vectors[] = $item['vector4'] = array_map(fn () => rand() / mt_getrandmax(), range(0, 3));
        }

        return $items;
    }
}
```
</details>

GromNaN (approved):
Thanks!
I'll rebase the PR on #3583 to fix the unrelated CI failure.

GromNaN:
Thank you @MaddipatlaChetan24

PR #3586, opened 2026-09-23
Fix count method to handle null case

MaddipatlaChetan24:
Fixes EmbedsOneOrMany::count() calling count() on a potentially-null return from getEmbedded(). Since PHP 8.0, count(null) is a TypeError rather than returning 0. getEmbedded() returns null whenever the parent's embedded attribute isn't set (e.g. a freshly-instantiated model, or one where the field was never populated) — so calling ->count() on such a relation would fatal-error instead of correctly reporting 0.

Fix: count($this->getEmbedded()) → count($this->getEmbedded() ?? []).

Not verified against a running PHP/MongoDB test environment — confirmed via the language-level behavior change (PHP 8 rejecting null for count()), which is unambiguous, not something needing execution to check.

### Checklist


MaddipatlaChetan24:
please check this !!

GromNaN:
> please check this !!

Sure! My intention was to rebase your PR on 5.11, since it's a bugfix. I'm on it.

GromNaN (approved):

GromNaN:
Thank you @MaddipatlaChetan24

